PR #26, opened 2019-08-17
Refactored Boltzmann example docs and scripts.

def-mycroft:
Made substantial edits to the Boltzmann wealth distribution example. The
general train of thought wasn't changed in the .md file, but many
paragraph were restructured significantly and a few paragraphs were
added.

Also, the corresponding files in `examples/` were edited as well. Since
the docs page gives two examples of the Boltzmann model (with a grid and
without), two separate `.jl` files where created so that either
variation can be easily executed by the user. The code referenced in
docs page is exactly the same code in the two `.jl` files.

### DIFF
--- a/docs/src/boltzmann_example01.md
+++ b/docs/src/boltzmann_example01.md
@@ -1,159 +1,222 @@
 # Boltzmann wealth distribution
 
-This is a simple agent-based model in economics. Despite its simplicity, it shows striking emergent wealth distribution. The first model that we will does not have spatial structure.
+The Boltzmann wealth distribution is a simple agent-based model (ABM) in economics. Despite its simplicity, the model shows striking emergent wealth distribution. This model is the first in this set of examples that does not necessarily have a spatial structure. At first, we will build a model that does not have a spatial structure, and then we will modify the model to include a spatial structure. 
 
-* We start with a number of agents all of which have one unit of wealth.
+The Boltzmann model is conceptually very simple:
+
+* We start with a number of agents, each of which have one unit of wealth.
 * At every step, agents give one unit of their money (if they have any) to a random agent.
 * We will see how wealth will be distributed after a few steps.
 
-The code of this tutorial is in the `examples/boltzmann_wealth_distribution.jl` file on the Github repository.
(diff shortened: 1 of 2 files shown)


kavir1698:
This is very good. Thanks for your contribution.
There is a to-do for the `agent_step!` function where you want the agents to exchange money only with other agents in the same cell. The original function already has implemented it. Specifically, the lines below:

```jl
    available_agents = get_node_contents(agent, model)
    agent2id = rand(available_agents)
    agent.wealth -= 1
    agent2 = [i for i in model.agents if i.id == agent2id][1]
    agent2.wealth += 1
    # now move
    neighboring_nodes = node_neighbors(agent, model)
    move_agent!(agent, rand(neighboring_nodes), model)
```

Could you add them to the function and remove the to-do before I merge this PR?

def-mycroft:
@kavir1698 thanks for sharing the updated `agent_step!` code for the second part of the Boltzmann example, I'll update this and remove the TODO (will try to do it tonight, about 8 hours from now). I don't see that code in the docs on the master branch, but no matter I'll just update it now. This will close issue #27.

def-mycroft:
Below is what I added to the Boltzmann example. 

```julia
function agent_step!(agent::AbstractAgent, model::AbstractModel)
  # If the agent's wealth is zero, then do nothing.
  if agent.wealth == 0
    return
  # Otherwise..
  else
    #...create a list of all agents on the same node and select a random agent.
    available_agents = get_node_contents(agent, model)
    random_neighbor_agent_id = rand(available_agents)
    random_neighbor_agent = [i for i in model.agents
                             if i.id == random_neighbor_agent_id][1]
    # Then decrement the current agent's wealth and increment the neighbor's wealth.
    agent.wealth -= 1
    random_neighbor_agent.wealth += 1
  end
end
```

def-mycroft:
@kavir1698 should the below lines be included in the `agent_step!` function?

```julia
   # now move
    neighboring_nodes = node_neighbors(agent, model)
    move_agent!(agent, rand(neighboring_nodes), model)
```

I didn't include them in my latest commit, I just didn't see where that was explained in the Boltzmann example. I can add the "move agent" lines back in if needed, I just wasn't quite sure why the agents where being moved in `agent_step!` and would like to explain it in the example.

kavir1698:
Yes, agents should actually move for the model to work. Otherwise, same agents would give and receive among themselves. See [this article](https://arxiv.org/abs/1005.5006).

def-mycroft:
@kavir1698 , ah, I see, that is intuitive in hindsight. I added those lines to the example and wrote a comment and paragraph in the docs page mentioning this.

PR #32, opened 2019-10-05
paper: copy editing

jedbrown:
Cc: https://github.com/openjournals/joss-reviews/issues/1611


kavir1698:
Thank you for your edits.

Issue #38, opened 2019-10-22
The methods of the step! function


Datseris:
Hi there,

in the process of redesigning and vastly simplifying the API, we have to face the current `step!` function. The problem from my eyes, is that this function is unwieldy. 

Not only it has 10 methods, but it even does multiple dispatch in 3 different argument positions. Something tells me hat things will be much simpler if we just separate many of its methods into different functions. What was especially confusing for me was the story with the aggregators, propagators, and all that jazz. To my understanding, `step!` has four (!!!) different functionalities:

1. Normal stepping.
2. Collecting agent properties
3. Appying aggregators
4. Doing whatever `propagg` does (I have not understood this at all.

(also, it seems that numbers 2, 3 and 4 **cannot** happen at the same time)

In my eyes, this is way too much stuff that are done by a single function name. A function should do one, and only one thing. This keeps things clear, and most importantly **easy to learn**.

@kavir1698 , can you please give some input here, on how we can simplify this approach?

My suggestion would be to have `step!`, `collect!`, `aggregate!`, `propaggate!` ( still don't know what this is).

Datseris:
Or, we can take a page out of the book of DifferentialEquations.jl and DynamicalSystems.jl and create a `Stepper` object. This object will be doing the stepping, and _it_ will be passed around to the 3 data collection functions.

I like this idea _a lot_, because it removes the unnecessary dispatch on `model_step!` as well!

kavir1698:
Hello,

I don’t think that a function that does many different things is problematics. It would be similar to the `plot` function in most plotting packages, where one can only pass x and y arrays, or add a ton of keyword args on top of that to modify the plot.

Similarly, separating the arguments of the `step!` function to separate functions does not change how much a user should learn. It just makes them to write more lines of code to do the same thing.

Perhaps, making most of the arguments as keyword arguments, and better explaining them in the docs is enough to make things easier.

About the `propagg`. It is a dictionary of whose keys are “properties” and values “aggregators”. So basically, one can apply different aggregators to different agent or model fields.

I do not know how the stepper object works in those packages. I should check them out.

From: George Datseris
Sent: Wednesday, October 23, 2019 12:58 AM
To: JuliaDynamics/Agents.jl
Cc: Ali R. Vahdati; Mention
Subject: [JuliaDynamics/Agents.jl] The methods of the step! function (#38)

Hi there,
[...]

Issue #37, opened 2019-10-21
[DOCS] examples of non-grid spaces 


Datseris:
I think it would be great, and useful, if examples of scientific research using non-grid (i.e. arbitrary graphs) as spaces for agent based modelling are presented/cited in the documentation.

After I am done with #24 #36 we could also consider implementing one of these examples in this library.

yuxiliu1995:
Some ideas:

Examples from Complexity Explorables: https://www.complexity-explorables.org/topics/agent-based-models/

* Ising model
* Herd immunity model
* Opinion in a network https://www.complexity-explorables.org/explorables/echo-chambers/
* [Epidemic Dynamics on an Adaptive Network](https://core.ac.uk/download/pdf/44186473.pdf)
* Economics and finance. See [To bail-out or to bail-in? Answers from an agent-based model](https://core.ac.uk/download/pdf/33971665.pdf), [Bubbles, crashes and intermittency in agent based market models](http://finance.martinsewell.com/stylized-facts/scaling/GiardinaBouchaud2003.pdf)

Datseris:
Oh wow, besides the Ising model (which is traditionally associated with spatial structure), these examples are awesome!

kavir1698:
We have now examples of graph spaces (SIR model) and continuous spaces (Boid flocking). Can this issue be closed?

Issue #36, opened 2019-10-15
questions regarding the API


Datseris:
Hello there,

I've got some questions regarding the current API. I am looking specifically at the forest fire example, but I think the same thing applies to all examples.

1. Why use the lengthy syntax, like `for node in 1:gridsize(forest.space.dimensions)` instead of using `for node in verteces(space)` ? 
2. In this loop:
```
      tree = Tree(node, (1,1), true)
      add_agent!(tree, node, forest) # TODO: why is the agent getting (1,1)?
```
why do the agents get position `(1,1)`? Shouldn't they be getting the position of the `node`? This confuses me a lot...

3. Later on, there is `shuffle(1:gridsize(forest.space.dimensions))` which could instead become `shuffle(verteces(forest.space))`

---

In general, while I am working on #20 , I am trying to simplify the API. I am asking these questions to see if there is a fundamental design decision behind them. If not, then I think the simplest, clearest, and shortest version should be used.

I don't think it is beneficial to define a second API on top of the LightGraphs API, which I think is what happens at the moment with `gridsize`.

Datseris:
One more Q:

the documentation states that a field of a `Space` object must be:

> Second, `dimensions` of the grid or network in a Tuple.

What is this supposed to be for a social network graph?

Datseris:
The more I think about it, the more I wonder: can we completely remove the type `Space`, and use a graph directly as a subfield of the `Model`, and make `agent_positions` a subfield of the `Model`?

EricForgy:
That's what I was wondering too. Seems like a clean / elegant way to go.

Datseris:
Yeah, this would simplify things a *lot*, and I think LightGraphs provides quite a robust API, making it not necessary to make our own on top of it.

Datseris:
If @kavir1698 gives the okay, I'll start working on a new branch that removes the `Space` types and subtypes and instead links graphs directly.

EricForgy:
If you have energy, I say "Go for it" while its hot :)

kavir1698:
Hi,
Thank you for your suggestions.
Answers to the questions:

1. Yes, we could use the `vertices` function here, which would be used as `for node in vertices(forest.space.space)`. I do not have a strong opinion for either implementation.

[...]

Issue #22, opened 2019-08-01
[DOC] Statement of the APIs


Datseris:
From my understanding reading the docs, Agents seems to define some APIs that the user-defined types must follow. For example, it seems to me that `Model` and `scheduler` functions have a specific form, or should have a specific form.

is this true?

If yes, I think this has to be stated somewhere explicitly and not only via examples. E.g. the expected fields that subtypes of `AbstractModel` should have.

I am now looking at the Schelling's segregation model doc page. I realize that what I ask here is partly described there. This is (in my personal opinion) not an intuitive mix. API and Package Workflow should have their own dedicated page, and examples should follow after.

My suggestion would be to move the first section of that page to a dedicated page, and enrich it with API instructions for `AbstractModel`, `scheduler`, etc...

Datseris:
Hm, the more I read the Schelling's model tutorial, the more I see that what I ask for is already existing. This could be solved with minimal effort, by making Schelling's tutorial a first-level page by itself (instead of being grouped in the examples) and changing titles and making everything a bit more visible.

Some other title is necessary, so that it is obvious from a glance that this page does serve as an "introduction" to Agents.jl. I suggest to make the First level title `Agent's.jl Architecture` and add a second title later on `Schelling's segregation model Example`.

(what I said above, to state the expected API **independently** of an example, is still very useful)

Datseris:
The [Table of Contents](https://kavir1698.github.io/Agents.jl/dev/#Table-of-contents-1) also starts with the forest fire model. Some re-organization there could help as well, so that it starts with basic API introduction instead.

kavir1698:
Hi,
Thanks for the suggestions. I was on vacation. I will try to address your comments as soon as possible.

kavir1698:
The docs are now refactored as you suggested.

Datseris:
Cool, thanks, I'll have a look soon!

Datseris:
I am reading the new documentation currently. I have some comments that might sound nit-picky but I think can really help improve the docs. Recall that I am in no way a user already familiar with Agents.jl and thus my comments (I believe) do give you the impression of what a newcomer would think.

---

[...]

PR #41, opened 2019-10-26
API overhaul of Agents.jl

### LINKED ISSUE #24: Defining a space - Why?
This comes after reading the documentation tutorial, specifically the part "Defining a space".

I have not understood at all why a user has to define a space by creating a struct. At least not at all for the simple grid cases.

Wouldn't it be much better to avoid all these complications and simply provide simple constructors like:
```
space = Space(x, y, z; periodic = true, Moore = false)
```
where internally `Space` does something like
```
Space(args...; kwargs...) = Space((args...,,), grid(args...; kwags...), instantiate_agents(grid))
````
and returns the resulting `struct`? Here `Space` is the standard `Space` type of Agents.jl (currently not existing), which would replace the `MyGrid` of the documentation.

From what I have understood, the only information the user _really_ cares about is simply the dimensionality of the space, and whether it is Moore/periodic. Everything else is so that Agents.jl operates internally. Thus this information should stay *internal* and the user should have to care about it / create a `Space` struct.

---

Of course it should be stated that this `Space` struct contains the actual agent locations, but still this is advanced usage I feel.

Datseris:
This PR is the very first step of bringing Agents.jl into version 2.0, with an API that is not only overhauled, but also simplified and easier to learn and understand. In this process the amount of written source code has also been reduced quite dramatically. It should go without saying that this PR is **majorly breaking**.

The changes introduced here are inspired by a large number of discussions, #38 , #37 , #36, #24, #22, etc.

Closes #24

While working on this I have also cleaned up the source code a lot, and tried to make the docstrings more idiomatic to Julia's syntax. Now we also adhere much more to Julia's style guide.

---

@kavir1698 , @EricForgy, @def-mycroft please consider reviewing this PR. I would advise you to _not_ review based on the git diff :P Instead, I have updated the Tutorial.md file to reflect the changes. It makes sense to build this documentation locally using Documenter. IF not, then just read the crucial docstrings of the crucial functions. These are in the Tutorial.md file *just before* the data collection part.

The data collection is not yet updated. I cannot do this update unfortunately, @kavir1698 will have to do this part when he has time. I looked at the source but it was just too unfamiliar for me... :( 

(same goes for the data-collection part of `step!`)


Datseris:
Closing this in favor of #42 which merges to a different branch (`v2.0`)

PR #42, opened 2019-10-26
API overhaul of Agents.jl

### LINKED ISSUE #24: Defining a space - Why?

Datseris:
This PR is the very first step of bringing Agents.jl into version 2.0, with an API that is not only overhauled, but also simplified and easier to learn and understand. In this process the amount of written source code has also been reduced quite dramatically. It should go without saying that this PR is **majorly breaking**.

The changes introduced here are inspired by a large number of discussions, #38 , #37 , #36, #24, #22, etc.

Closes #24

While working on this I have also cleaned up the source code a lot, and tried to make the docstrings more idiomatic to Julia's syntax. Now we also adhere much more to Julia's style guide.

---

@kavir1698 , @EricForgy, @def-mycroft please consider reviewing this PR. I would advise you to _not_ review based on the git diff :P Instead, I have updated the Tutorial.md file to reflect the changes. It makes sense to build this documentation locally using Documenter. IF not, then just read the crucial docstrings of the crucial functions. These are in the Tutorial.md file *just before* the data collection part.

The data collection is not yet updated. I cannot do this update unfortunately, @kavir1698 will have to do this part when he has time. I looked at the source but it was just too unfamiliar for me... :( 

(same goes for the data-collection part of `step!`)


Datseris:
Of course, tests will fail here. But should you agree with the changes I'd still say to merge this and continue working from here. There is still some work to do, and if we merge this I'll make up a list of things to-do for 2.0 release.

With 2.0 we should also announce this on discourse.

kavir1698:
I will post my comment within the next few days.

Datseris:
Thanks.

@EricForgy , you participated in https://github.com/JuliaDynamics/Agents.jl/issues/24  and in https://github.com/JuliaDynamics/Agents.jl/issues/36 do you mind giving this PR a look? It addresses both of these issues (and more).

EricForgy:
Hi @Datseris 👋 

I would love to, but I am totally snowed. I will do my best to have a look when / if things settle a bit over here. I am watching with interest though 💪

kavir1698:
I have just carefully reviewed the changes. This is great work. The documentation looks better, and the `Space` function is just so easy to use. I think these changes are only beneficial, I do not see any drawbacks.

One small comment regarding the `Space` function is to make the `periodic` and `moore` arguments keyword arguments for easier usage.

I have more free time now, and I can start making changes to the data collection functions.

Thank you.

Datseris:
Hi, very glad that you like the changes. Okay, then I'll do the update of making these two arguments keywords, and then we should merge this into the branch `v2.0`. Starting to work from this point will make it easier!

Issue #40, opened 2019-10-26
doubled information: agent.pos and space.agent_positions


Datseris:
Seems to me that for any agent, the information contained in `agent.pos` and `space.agent_positions` is fully equivalent and convertible from one form to another.

Is this true? If it is true, then **why** should both informations exist? 

Can't one get away by completely removing the `pos` field of the agent struct?

Datseris:
I think I understand why: it would be terribly inneficient to do such conversion every time you search for a specific position.

kavir1698:
Yes. But here is actually something I have been thinking about a lot. Now that you mention it, it maybe a good time to discuss it.

I am looking for a more efficient way of storing agent positions in the `space.agent_positions` vector. Right now, agent IDs are saved in the `space.agent_positions` vector. The problem is that whenever one wants to retrieve all the agents in a node, we have to find agents that correspond to the IDs saved in a node. The `id_to_agent` function does that with the following implementation:

```jl
function id_to_agent(id::Integer, model::AbstractModel)
  agent_index = findfirst(a-> a.id==id, model.agents)
  agent = model.agents[agent_index]
  return agent
end
```

The first line of this function is a major bottleneck. I once tried to save `Ref`s to the agents instead of their IDs in the `space.agent_positions` vector. But it turned out to be even slower. Do you have any suggestion for this situation?

The "forest fire" example is a good model to test the performance of different the performance of our ideas regarding this issue.

Datseris:
I think a smart way to resolve this is to **not allow users to provide the Agent ID**. Each agent gets an id when added to a model, and the agent's id is just the number it was added at, starting from 1 and incremeting by 1 for each agent.

Then the agent's ID is the same as the agent's index in the `model.agent` vector.

I don't see a reason why the user would have to specify the agent ID. Once again this seems to me a 100% internal information the user should never have to care about, similarly with making the `agent_positions` when initializing a space. If there is any model-related information about the ID, the user should just create a new field for their agent type.

---

**Importantly** such a change would eliminate the ID as a field of the Agent type. Since the agent's id is identical with its index in `model.agents`, there is no reason to doubly store the information.

kavir1698:
A problem with such approach arises when some agents die during a model.
Then the agent ID will be different from it's index.

Datseris:
Hm, good point that I have not though.

But seems to me that creating a custom Index type would probably solve our problems. I'll contemplate on it and let you know.

[...]

PR #51, opened 2019-11-10
New storage of agents in a model

Datseris:
This PR addresses #40 and also addresses the fact that users have to make a `struct` for their model instead of calling a function. 

The main idea in this implementation is already described in #40 where we replace killed agents with `missing`.

---

**TODO before merging:**

- [ ] Add a "renormalization" function that removes all missings
- [ ] Update the space creation
- [ ] Update agent_space_interactions
- [ ] Update documentation / tutorials for this
- [ ] Check if the data collection works properly with this (sorry @kavir1698 but only you can do this)
- [ ] Benchmark this with the current `v2.0` branch: it may be that this change introduces regressions. (super step!)

For the benchmarking thing, we have to come up with a smart way to benchmark it, which not only creates and kills agents very frequently, but also cares very much about getting the agents in a specified position (because this is exactly the problem that #40 discusses) . Maybe a modified version of the forest fire will do the trick?


kavir1698:
A large forest fire model should be good because agent IDs keep increasing.

PR #52, opened 2019-11-10
New storage of agents in the model (and new model type)

Datseris:
This PR addresses #40 and also addresses the fact that users have to make a `struct` for their model instead of calling a function. 

The main idea in this implementation is already described in #40 where we replace killed agents with `missing`.

---

**TODO before merging:**

- [ ] Update agent_space_interactions
- [x] Update the space creation
- [ ] Add a "renormalization" function that removes all missings
- [ ] Update documentation / tutorials for this
- [ ] Check if the data collection works properly with this (sorry @kavir1698 but only you can do this)
- [ ] Benchmark this with the current `v2.0` branch: it may be that this change introduces regressions. (super step!)

For the benchmarking thing, we have to come up with a smart way to benchmark it, which not only creates and kills agents very frequently, but also cares very much about getting the agents in a specified position (because this is exactly the problem that #40 discusses) . Maybe a modified version of the forest fire will do the trick?

### DIFF
--- a/src/simulations/step.jl
+++ b/src/simulations/step.jl
@@ -39,9 +39,9 @@ function step! end
 dummystep(model) = nothing
 dummystep(agent, model) = nothing
 
-step!(model::AbstractModel, agent_step!, n::Int = 1) = step!(model, agent_step!, dummystep, n)
+step!(model::ABM, agent_step!, n::Int = 1) = step!(model, agent_step!, dummystep, n)
 
-function step!(model::AbstractModel, agent_step!, model_step!, n::Int = 1)
+function step!(model::ABM, agent_step!, model_step!, n::Int = 1)
   for i in 1:n
     activation_order = model.scheduler(model)
     for index in activation_order
@@ -55,9 +55,9 @@ end
 # data collection
 #######################################################################################
 
-step!(model::AbstractModel, agent_step!, n::Int, properties; when::AbstractArray{Int}, nreplicates::Int=0) = step!(model, agent_step!, dummystep, n::Int, properties, when=when, nreplicates=nreplicates, parallel=false)
+step!(model::ABM, agent_step!, n::Int, properties; when::AbstractArray{Int}, nreplicates::Int=0) = step!(model, agent_step!, dummystep, n::Int, properties, when=when, nreplicates=nreplicates, parallel=false)
 
-function step!(model::AbstractModel, agent_step!, model_step!, n::Int, properties; when::AbstractArray{Int}, nreplicates::Int=0, parallel::Bool=false)
+function step!(model::ABM, agent_step!, model_step!, n::Int, properties; when::AbstractArray{Int}, nreplicates::Int=0, parallel::Bool=false)
 
   if nreplicates > 0
     if parallel
--- a/src/simulations/step.jl
+++ b/src/simulations/step.jl
@@ -43,9 +43,9 @@ step!(model::ABM, agent_step!, n::Int = 1) = step!(model, agent_step!, dummystep
 
 function step!(model::ABM, agent_step!, model_step!, n::Int = 1)
   for i in 1:n
-    activation_order = model.scheduler(model)
-    for index in activation_order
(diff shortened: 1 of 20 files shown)


src/simulations/step.jl
```diff
@@ -43,9 +43,9 @@ step!(model::ABM, agent_step!, n::Int = 1) = step!(model, agent_step!, dummystep
 
 function step!(model::ABM, agent_step!, model_step!, n::Int = 1)
   for i in 1:n
-    activation_order = model.scheduler(model)
-    for index in activation_order
-      agent_step!(model.agents[index], model)
+    actived_agents = model.scheduler(model)
```
kavir1698: typo: "activated_agents"

Datseris:
Hm, I now realize that some thought must be put into how `kill_agent!` works.

we have to find the agent in the list of agents.
but now it is possible that two agents are identical (with same properties in their fields).
thus one cannot type `findfirst(a -> a == agent, model.agents)` because it could find a different agent.

I think it is best to re-work the agent iteration protocol that the scheduler function does.
The scheduler should return agent indices directly and the user should use these indices directly as well. So in a sense, the user defined stepping function would do:

```
for agentid in model.scheduler(model)
agent = model.agents[i] # or even define model[i] directly as per overloading getindex
# do stuff with your agent.... 
# or kill it by doing:
kill_agent!(agentid, model)
```

Thoughts?

kavir1698:
The agents should still have the `id` field. That was the whole point for these changes, so that we can quickly retrieve an agent from its ID. It is a common practice to go randomly through the nodes, retrieves agents in those nodes and updates them. We cannot always go through all the agents.

Datseris:
> That was the whole point for these changes, so that we can quickly retrieve an agent from its ID.

I don't think so. This started so that finding an agent is a performant operation, and this is only possible when the ID is equivalent with the index in the agent's array. Because the bottleneck is exactly `findfirst(a -> a == agent, model.agents)` ...

Datseris:
I think we should spend more time thinking about this and close the PR...

(Regardless of how the Agents are stored and found, should we go through the change that introduces `AgentBasedModel`, and makes it so that the users do not define their own struct? I can separate this in a new PR)

kavir1698:
> I don't think so. This started so that finding an agent is a performant operation, and this is only possible when the ID is equivalent with the index in the agent's array. Because the bottleneck is exactly findfirst(a -> a == agent, model.agents) ...

Exactly. When the ID is equivalent with the index in the agents' array, we won't need to use `findfirst(a -> a == agent, model.agents)`. We could simply say `agent = model.agents[some_id]`.

> should we go through the change that introduces AgentBasedModel, and makes it so that the users do not define their own struct?

Yes, that's fine. Users only define an agent struct now.

Datseris:
So you are saying that what I discussed in #40 should still happen, with the difference of keeping `id` as the agent's field? Then `kill_agent!` would become:
```julia
function kill_agent!(agent, model)
    model.agents[agent.id] = missing
    # update space accordingly
end
```
Now, how will `# update space accordingly` happen? Because one still has to find the agent in the space.

Maybe yet one more field of the `Agent`, called `node`, which stores the node number that the agent is currently located at, and this field is changed whenever `add_agent!` or `move_agent!` is called?

kavir1698:
> So you are saying that what I discussed in #40 should still happen, with the difference of keeping id as the agent's field?

Yes, this is what I had in mind.

> Now, how will # update space accordingly happen? Because one still has to find the agent in the space.

> Maybe yet one more field of the Agent, called node, which stores the node number that the agent is currently located at, and this field is changed whenever add_agent! or move_agent! is called?

Well, agents already have a `pos` field. So, one should basically find the agent among the agents in a node. The current implementation is not that bad:

```jl
splice!(agent_positions(model)[agentnode],
          findfirst(a->a==agent.id, agent_positions(model)[agentnode]))
```

Datseris:
Okay, better to progress step by step though, to make everything easier.

Issue #20, opened 2019-08-01
Proposal: Refactor documentation to use the examples of `examples`, Literate.jl and Documenter.jl


Datseris:
I have a proposal to make, which will take some effort but it won't bring new features to Agents.jl. Nevertheless, I think it is worth it. 

The proposal is to refactor the examples part of the documentation, so that instead of being static code, to be real Julia code that runs and whose output is shown in the documentation. This has the following benefits:
1. It ensures that the code is runnable. If it isn't the docs won't build properly.
2. It ensures that everything is up to date.
3. It also makes it easier to update the documentation after an update on the code, or when changing a parameter.
4. The `examples` folder is used directly in the documentation.

Here is how this works. We do it in TimeseriesPrediction.jl. We have some examples here:
https://github.com/JuliaDynamics/TimeseriesPrediction.jl/tree/master/examples 
These examples are valid Julia code, which is runnable. They follow the Literate.jl syntax, which means Literate can process then and make them markdown files for the docs. Then in the [`make.jl`](https://github.com/JuliaDynamics/TimeseriesPrediction.jl/blob/master/docs/make.jl) we first convert them to markdown, and then documenter makes the doc page. 

Here is how it looks: https://juliadynamics.github.io/TimeseriesPrediction.jl/latest/stexamples/

*In the TimeseriesPrediction example we do one  extra step to include all examples in a single file; I think here this should be avoided and each example should stay a separate file. So don't mind the function `replace_includes`*

I can contribute here once the JOSS review is over.

def-mycroft:
> refactor the examples part of the documentation, so that instead of being static code, to be real Julia code that runs and whose output is shown in the documentation. 

I think this is really important and I love the way it looks in the examples that you provided. 

I've been in the middle of working on some documentation edits and have been working to create a correspondence between the examples and the documentation page. 

I had been working on the Boltzmann example before you made this post and will try to finish soon (was sidetracked on another project last week). 

So I think what I'll do (unless there is a better idea) is continue to refactor the examples like I have been, and then after there is a tight correspondence between the docs pages the scripts in `examples/` it will be easier to upgrade to using Literate.jl.

Datseris:
Cool, happy to see you liking the idea! Don't forget that this is something I can easily contribute to after the JOSS review is done, since I have already done it a couple of times for various doc pages (e.g. also JuliaMusic).

kavir1698:
[...]

PR #62, opened 2019-11-15
Runnable documentation

Datseris:
With this PR we should try to ensure that we can use the **latest** version of documenter and successfully build the documentation locally. 

We should then view it and ensure that all is a-ok, in preparation of the 2.0 release.


Datseris:
@kavir1698 the docs finally run locally for me, excluding AgentPlots. I don't use plots, or any other plotting stuff beside PyPlot, so please do me a favor and check out this branch and also make the docs once locally, to ensure that plotting works as expected.

Then we merge. Ps I am already working on #20 on the forest fire model, see `forest` branch.

Ps2: once you update the readme of AgentsPlots.jl I'll try to register it. The Julia guys want at least a proper readme to accept registration.

Datseris:
This is in a very good state, so I'll merge. @kavir1698 please do not forget to check that plotting works. Not necessary for the 2.0 release though. We can always add the plotting docs after 2.0.

PR #77, opened 2019-11-27
First implementation of grid search function

kavir1698:
This is the first try to implement a function (`gridsearch`) that automatically runs simulations with different parameter value combinations. Here is an example using the Schelling model:

```jl
using Agents

mutable struct SchellingAgent <: AbstractAgent
  id::Int # The identifier number of the agent
  pos::Tuple{Int,Int} # The x, y location of the agent
  mood::Bool # whether the agent is happy in its node. (true = happy)
  group::Int # The group of the agent,
             # determines mood as it interacts with neighbors
end

# NB for the gridsearch function to work, we need a function that accepts a single argument
# and outputs a model object
function instantiate(properties::Dict)
    space = Space(properties[:griddims], moore = true) # make a Moore grid
    model = ABM(SchellingAgent, space, scheduler=random_activation, properties=properties)
    for n in 1:properties[:numagents]
        agent = SchellingAgent(n, (1,1), false, n < properties[:numagents]/2 ? 1 : 2)
        add_agent_single!(agent, model)
    end
    return model
end

    agent.mood == true && return # do nothing if already happy
    minhappy = model.properties[:min_to_be_happy]
    neighbor_cells = node_neighbors(agent, model)
    count_neighbors_same_group = 0
    # For each neighbor, get group and compare to current agent's group
    # and increment count_neighbors_same_group as appropriately.
    for neighbor_cell in neighbor_cells
        node_contents = get_node_contents(neighbor_cell, model)
        # Skip iteration if the node is empty.
        length(node_contents) == 0 && continue
        # Otherwise, get the first agent in the node...
        agent_id = node_contents[1]
        # ...and increment count_neighbors_same_group if the neighbor's group is
        # the same.
        neighbor_agent_group = model.agents[agent_id].group
        if neighbor_agent_group == agent.group
            count_neighbors_same_group += 1
        end
    end
    # After counting the neighbors, decide whether or not to move the agent.
    # If count_neighbors_same_group is at least the min_to_be_happy, set the
    # mood to true. Otherwise, move the agent to a random node.
    if count_neighbors_same_group ≥ minhappy
        agent.mood = true
    else
        move_agent_single!(agent, model)
    end
    return
end

happyperc(moods) = count(x -> x == true, moods)/length(moods)


# parameters for which we want to test different values
param_ranges = Dict(:min_to_be_happy=>2:5, :numagents=>[200,300])
# data to collect. Should be a dictionary
collect_fields = Dict(:mood=>[happyperc])
# input of the instantiate function
model_properties = Dict(:numagents=>320, :griddims=>(20, 20), :min_to_be_happy=>3)


gridsearch(param_ranges=param_ranges, model_properties=model_properties, n=3,
collect_fields=collect_fields, when=1:3, model_initiation=instantiate,
agent_step=agent_step!, model_step=dummystep)

32×4 DataFrames.DataFrame
│ Row │ happyperc(mood) │ step  │ min_to_be_happy │ numagents │        
│     │ Float64         │ Int64 │ Int64           │ Int64     │
├─────┼─────────────────┼───────┼─────────────────┼───────────┤
│ 1   │ 0.0             │ 0     │ 2               │ 200       │
│ 2   │ 0.67            │ 1     │ 2               │ 200       │
│ 3   │ 0.865           │ 2     │ 2               │ 200       │
│ 4   │ 0.93            │ 3     │ 2               │ 200       │
│ 5   │ 0.0             │ 0     │ 2               │ 300       │
│ 6   │ 0.833333        │ 1     │ 2               │ 300       │
│ 7   │ 0.963333        │ 2     │ 2               │ 300       │
│ 8   │ 0.983333        │ 3     │ 2               │ 300       │        
│ 9   │ 0.0             │ 0     │ 3               │ 200       │
│ 10  │ 0.27            │ 1     │ 3               │ 200       │
│ 11  │ 0.46            │ 2     │ 3               │ 200       │        
│ 12  │ 0.63            │ 3     │ 3               │ 200       │        
│ 13  │ 0.0             │ 0     │ 3               │ 300       │        
│ 14  │ 0.593333        │ 1     │ 3               │ 300       │        
│ 15  │ 0.83            │ 2     │ 3               │ 300       │        
│ 16  │ 0.916667        │ 3     │ 3               │ 300       │        
│ 17  │ 0.0             │ 0     │ 4               │ 200       │        
│ 18  │ 0.155           │ 1     │ 4               │ 200       │        
│ 19  │ 0.305           │ 2     │ 4               │ 200       │        
│ 20  │ 0.41            │ 3     │ 4               │ 200       │        
│ 21  │ 0.0             │ 0     │ 4               │ 300       │        
│ 22  │ 0.433333        │ 1     │ 4               │ 300       │        
│ 23  │ 0.663333        │ 2     │ 4               │ 300       │        
│ 24  │ 0.743333        │ 3     │ 4               │ 300       │        
│ 25  │ 0.0             │ 0     │ 5               │ 200       │        
│ 26  │ 0.025           │ 1     │ 5               │ 200       │        
│ 27  │ 0.085           │ 2     │ 5               │ 200       │        
│ 28  │ 0.12            │ 3     │ 5               │ 200       │        
│ 29  │ 0.0             │ 0     │ 5               │ 300       │        
│ 30  │ 0.15            │ 1     │ 5               │ 300       │        
│ 31  │ 0.29            │ 2     │ 5               │ 300       │        
│ 32  │ 0.413333        │ 3     │ 5               │ 300       │  
```


Datseris:
(I'll review this once tests are passing)

kavir1698:
Tests pass for Julia v1.2. The problem of Julia v1.0 is not with this PR.

kavir1698:
1. `dict_list` is a cool function. It would be even better if it supported ranges (e.g. `1:3`). Do you think you can modify it to support ranges in addiction to arrays?
2. Having all the parameters and ranges in a single file is a good idea, so that we can ditch `model_properties`.
3. About the function's name, I was also looking for something better. `gridsearch` is the name of a similar function in Scikit-Learn. But in that context searching makes sense. `paramscan` is a good alternative.

Datseris:
Re:

1. Unfortunately nope, the decision to not support ranges was made on purpose `*`, to make the API as simple as possible. Unfortunately extending this now to also work with ranges would be a breaking change, so I am not willing to do that. But you could just do `collect(1:50)`.
2. great
3. I am reserved to use the `grid` word for the reasons I mentioned above... (Besides, we can break the cycle and make a difference with better names!)

`*` The idea was simple: we wanted `dict_list` to be able to work with parameters that are user-defined types, Tuples, NamedTuples, and anything else. Now, the thing is, many of these things are iterable by themselves. But simply because a structure is iterable, it doesn't mean that the user would want it to be iterable over. In the very example of this PR we have `gridsize = (20,30)`. This is iterable. Does it mean that `gridsize` is one parameter that should be used in one simulation, or is two parameters that we want to iterate over, 20 and then 30? Thus resulting in two simulations?

A line had to be draw, and using Julia's multiple dispatch was the best way to do this. And the line is simple: If it is subtype of `Vector`, it is something that contains many parameters. if it is _not_ a subtype of `Vector` it is one parameter. (Notice that `Array`s are _not_ subtypes of vectors)

kavir1698:
> In the very example of this PR we have gridsize = (20,30). This is iterable. 

That's a good point.

kavir1698:
Please see the updated function.

src/simulations/data_collector.jl
```diff
@@ -180,4 +180,213 @@ function series_replicates(model, agent_step!, model_step!, properties, when, n,
...
+It should have arguments for every key in the `parameters` dict, even if it does not use them..
+
+`properties` is the same dictionary used in the `step!` function that determines
+what information should be collected.
+
+Running replicates is not implemented, yet.
+
+# Example
+
+Here is an example of using `paramscan` with the Schelling model:
```
Datseris: please put this in the `tutorial.md` file as runnable Julia code. Add one more `@example` block.

This is just way too big for a documentation string. Let's reuse the fact that Tutorial.md already defines the schelling model...

Datseris:
Hi @kavir1698 , I think this is almost done. I've just done some updates that improve the documentation, and also make calling `paramscan` a bit easier, see the docs.

From my end, there are two things missing before we can merge this:

1. `paramscan` needs to be tested.
2. We should decide whether we want this line: `  changing_params = [k for (k, v) in parameters if typeof(v)<:Vector]` . Isn't there a benefit in actually listing all parameters in the dataframe, regardless of if they change or not? At the moment only parameters that change are listed. I think both approaches are valid. Therefore, maybe it is worth it to add a keyword argument `show_constants` that also includes constant parameters in the final dataframe?

kavir1698:
>  Isn't there a benefit in actually listing all parameters in the dataframe, regardless of if they change or not?

I vote against always including the constant variable because it can increase the dataframe size considerably, especially in long simulations with many parameters. It will be a limiting factor while adding little value.
Of course I do not disagree with adding an argument to include constant variables.

Datseris:
> because it can increase the dataframe size considerably

Good point. Then the default value of the extra argument should be `false`.

kavir1698:
> paramscan needs to be tested.

I will add test after merging the other branches because they have breaking changes.

PR #78, opened 2019-11-29
Restructure output

kavir1698:
This PR changes the output structure when collecting raw data.
Currently, when we collect raw data from a model at different steps, the output data frame creates a new column for each step. But a more standard structure is to add a `step` column. This way the output is easier to parse and understand.

Current output for the Schelling model looks like this:

```
370×19 DataFrames.DataFrame. Omitted printing of 11 columns
│ Row │ id     │ pos_0  │ mood_0 │ group_0 │ pos_1  │ mood_1 │ group_1 │ pos_2  │
│     │ Int64⍰ │ Int64⍰ │ Bool⍰  │ Int64⍰  │ Int64⍰ │ Bool⍰  │ Int64⍰  │ Int64⍰ │
├─────┼────────┼────────┼────────┼─────────┼────────┼────────┼─────────┼────────┤
│ 1   │ 1      │ 5      │ 0      │ 2       │ 393    │ 0      │ 2       │ 140    │
│ 2   │ 2      │ 368    │ 0      │ 2       │ 368    │ 1      │ 2       │ 368    │
│ 3   │ 3      │ 367    │ 0      │ 1       │ 367    │ 1      │ 1       │ 367    │
│ 4   │ 4      │ 360    │ 0      │ 1       │ 361    │ 0      │ 1       │ 361    │
│ 5   │ 5      │ 333    │ 0      │ 1       │ 333    │ 1      │ 1       │ 333    │
│ 6   │ 6      │ 192    │ 0      │ 1       │ 192    │ 1      │ 1       │ 192    │
│ 7   │ 7      │ 391    │ 0      │ 2       │ 391    │ 1      │ 2       │ 391    │
│ 8   │ 8      │ 232    │ 0      │ 1       │ 232    │ 1      │ 1       │ 232    │
```

And the output from this PR looks like this:

```
2220×5 DataFrames.DataFrame
│ Row  │ id    │ pos   │ mood │ group │ step  │
│      │ Int64 │ Int64 │ Bool │ Int64 │ Int64 │
├──────┼───────┼───────┼──────┼───────┼───────┤
│ 1    │ 1     │ 292   │ 0    │ 2     │ 0     │
│ 2    │ 2     │ 385   │ 0    │ 2     │ 0     │
│ 3    │ 3     │ 118   │ 0    │ 1     │ 0     │
⋮
│ 2217 │ 367   │ 4     │ 1    │ 2     │ 5     │
│ 2218 │ 368   │ 169   │ 1    │ 1     │ 5     │
│ 2219 │ 369   │ 393   │ 1    │ 2     │ 5     │
│ 2220 │ 370   │ 45    │ 1    │ 1     │ 5     │
```

### DIFF
--- a/src/simulations/data_collector.jl
+++ b/src/simulations/data_collector.jl
@@ -80,11 +80,12 @@ function data_collecter_raw(model::ABM, properties::Array{Symbol}; step=1)
     begin
       dd[!, :id] = sort(collect(keys(model.agents)))
     end
-    fieldname = Symbol(join([string(fn), step], "_"))
+    fieldname = fn
     begin
       dd[!, fieldname] = temparray
     end
   end
+  dd[!, :step] = [step for i in 1:size(dd, 1)]
   return dd
 end
 
@@ -123,7 +124,7 @@ end
 
 function data_collector(model::ABM, properties::Array{Symbol}, when::AbstractArray{T}, step::Integer, df::DataFrame) where T<:Integer
   d = data_collecter_raw(model, properties, step=step)
-  df = join(df, d, on=:id, kind=:outer)
+  df = vcat(df, d) #join(df, d, on=:id, kind=:outer)
   return df
 end
 
--- a/src/simulations/data_collector.jl
+++ b/src/simulations/data_collector.jl
@@ -1,4 +1,4 @@
-export combine_columns!, gridsearch
+export combine_columns!, paramscan
 
 """
     data_collecter_aggregate(model::ABM, field_aggregator::Dict; step=1)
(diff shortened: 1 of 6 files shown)


kavir1698:
What remains to be fixed is to re-structure the output when of multiple replicates. Similarly, for that we should avoid adding columns or data frames for each replicate, and instead, add a column named "replicate" and specify the number of replicate there.

Datseris:
Hi @kavir1698 , I like this a lot. I do have the same comment as #79 though. Can you please start a new branch from `2.0` instead of continuing the `paramscan` ? It is more hygienic this way. (Also, you can combine the 0-step with this change into one PR, they do target the same function `step!` after all)

Datseris:
(oh, by the way @kavir1698 : what I am asking is not hard at all. You just have to "cherry-pick" your latest 2 commits in this branch. This is a standard git command that will immediatelly apply these commits to your chosen new branch, that has started of `2.0`)

kavir1698:
#80 has these commits from branch v2.0.

Issue #74, opened 2019-11-24
Numbering of columns. Shouldn't it start from 0 ?


Datseris:
```
julia> data
2000×6 DataFrames.DataFrame
│ Row  │ id     │ wealth_1 │ wealth_3 │ wealth_5 │ wealth_7 │ wealth_9 │
│      │ Int64⍰ │ Int64⍰   │ Int64⍰   │ Int64⍰   │ Int64⍰   │ Int64⍰   │
├──────┼────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
│ 1    │ 1      │ 1        │ 0        │ 3        │ 4        │ 2        │
│ 2    │ 2      │ 1        │ 1        │ 2        │ 0        │ 1        │
│ 3    │ 3      │ 1        │ 2        │ 0        │ 1        │ 1        │
│ 4    │ 4      │ 1        │ 0        │ 1        │ 0        │ 1        │
│ 5    │ 5      │ 1        │ 1        │ 0        │ 1        │ 0        │
│ 6    │ 6      │ 1        │ 0        │ 0        │ 1        │ 2        │
│ 7    │ 7      │ 1        │ 0        │ 0        │ 0        │ 0        │
```
this is what I see when I play around with the wealth distribution example. The problem is, the column `wealth_1` is clearly the distribution _before_ running the model. I think this `1` is misleading, as it denotes the "step number". The step number in this case is 0 not one. If we save the initial state of the model before stepping, then we should be consistent and use this number to denote "how many steps are taken".

kavir1698:
https://github.com/JuliaDynamics/Agents.jl/commit/6d21ba41b18746a3cf8829f275b23172fc45de53 fixes it.

Datseris:
Wait, with your change does `when = 1:5` include collecting data for the initial condition?

@kavir1698 similar with my suggestion on the max characters per line of code, can you please start adding things using Pull Requests, instead of just "pushing" to the `2.0` branch? Not only this improves collaboration experience (as I always have to pull your changes to my development branches), but most importantly makes it possible to review things _before_ they are merged into the main branch. I think that you can agree to the fact that it is likely that any change will have mistakes, and being seen by a second pair of eyes is helpful.

kavir1698:
> with your change does when = 1:5 include collecting data for the initial condition?
Yes, it does.

> can you please start adding things using Pull Requests, instead of just "pushing" to the 2.0 branch?
[...]

PR #79, opened 2019-11-29
Step0

kavir1698:
Adds the option not to collect data at step 0. Answers #74.


Datseris:
I am confused @kavir1698 , because this PR seems to include `paramscan` related functionality. I think this will lead to several merge conflicts if we leave it as is. Could you please start a new branch from the `2.0` branch, and only add the 0-step related functionality? 

I am sorry if this sounds tedious, but I can already see in the commit sequence that this will lead to conflicts otherwise.

kavir1698:
#80 has these commits from branch v2.0.

PR #88, opened 2019-11-30
Update comparison with Mesa

### LINKED ISSUE #75: A fairer comparison with Mesa
Mesa seems to have very advanced functionality that supports browser-based plotting and interaction.

In our comparison with Mesa we should be fair: we should say that our performance is much higher, and our API is much simpler. But we should also say that they have much stronger visualization capabilities: https://github.com/projectmesa/mesa/blob/master/docs/images/Mesa_Screenshot.png

kavir1698:
Updates the comparison between Agents.jl and Mesa. This time, the figure shows grid sizes of maximum 40k because the Python code is too slow on my computer. Now, Agents.jl is more than 13 times faster than Mesa, as opposed to 8 times in version 1.

It will close #75


Datseris:
> Now, Agents.jl is more than 13 times faster than Mesa

Oh yeah baby

PR #92, opened 2019-12-01
CA examples

kavir1698:


examples/game_of_life_2D_CA.jl
```diff
@@ -0,0 +1,30 @@
+using Agents
+using Agents.CA2D
+using AgentsPlots
+
+# Agents.jl provides a module to create and plot 2D cellular automata.
+
+# Define the rule
```
Datseris: This should be made `# ## Define the rule` to be displayed as markdown titles.

Datseris:
This PR can be only merged _after_ `AgentsPlots` is registered, which is _after_ 2.0 is through. I'll tag 2.0 now!

kavir1698:
Can we merge this now?

Datseris:
Ops, good point! As long as the codeblocks return a plot type from Plots, it will be displayed in the docs (so, does `plot_CA2D` return the plot type?). If the last thing of the code blocks is not a plot, the plot won't be displayed, even if it was plotted.

Maybe you want to also put some plots for the forest fire model here?

kavir1698:
>  does plot_CA2D return the plot type?

Actually not. The last line of code is `savefig` function. Returning a plot without saving it does not make much sense. Does Plots.jl allow returning a plot after it is saved? Is it a good approach?

> Maybe you want to also put some plots for the forest fire model here?

Yeah, I'll do that.

Datseris:
> Actually not. The last line of code is savefig function. Returning a plot without saving it does not make much sense. Does Plots.jl allow returning a plot after it is saved? Is it a good approach?

It is not a good approach to save anything actually. `plot_CA2D` should only plot and return the plot object, and the user should decide whether to save or not. `plot_CA2D` should just create a plot, just like `plot([1, 2], [1, 2])` plots a line.

kavir1698:
Where should `AgentsPlots` be added to be available to tests?

Datseris:
Hi,

> Where should AgentsPlots be added to be available to tests?

I don't agree with this. `AgentsPlots` should be tested in its own repo and not in Agents.jl. We only use it to build the docs, but other than that it should be as detached from the main repo as possible.

I mean, lines such as these:
```
  @test typeof(plot_CA1D(data, nodesize=2)) <: AgentsPlots.AbstractPlot
```
make sense to be written only in `AgentsPlots`.

kavir1698:
>  @test typeof(plot_CA1D(data, nodesize=2)) <: AgentsPlots.AbstractPlot

This should have been removed. I agree with you on this.

But `CA2D` is different. I decided to add frames to the `Animation` object at every time step of the model. That is because with a 2D CA, if we run all simulation steps and then create animation frames from its output, the dataframe size soon become too large.

I have tried to make the `CA2D.ca_run` function as decoupled from AgentsPlots as possible. To that end, this function takes as argument a plotting function (`plot_CA2Dgif`) from AgentsPlots.
But for running tests we still need to call AgentsPlots.

kavir1698:
A workaround is to create a dummy function. So that `CA2D.ca_run` runs without an error, but does not actually plot anything.

kavir1698:
Now, only the documentation fails because it cannot find `AgentsPlots`. I don't see where is the problem.

Datseris:
> Now, only the documentation fails because it cannot find AgentsPlots. I don't see where is the problem.

I'll take care of this don't worry. But I need a bit more time to go through this, please be patient just a bit more. I'll have to do this over the weekend...

kavir1698:
Of course. I have no hurry and I am not trying to push you. All the best.

Datseris:
Okay, I fixed it, now it works and it properly interlaces the plots with the documentation.

@kavir1698 I want a favor from you. How difficult do you think it is for you to modify the wealth distribution example so that:

At the point where I calculate the 2D martrices, we instead have a grid plot (just like forest fire model).
The color of its node is its wealth, by mapping to a pre-determined colormap.

is this possible to be done? If yes could you please do it?

kavir1698:
> Okay, I fixed it, now it works and it properly interlaces the plots with the documentation.

Great!

> At the point where I calculate the 2D martrices, we instead have a grid plot (just like forest fire model).
> The color of its node is its wealth, by mapping to a pre-determined colormap.

That's straightforward. See the latest commit.

Datseris:
> That's straightforward. See the latest commit

But that is not what I was asking... Because you still compute the Matrices.

Let's phrase the question differently. Is there a simple way to plot an ABM, by passing a function `f` that gets as an input all agents of a node and returns a number as an output?

kavir1698:
I am still not clear on what you are asking. Do you mean to create a plotting function that calculates the matrices and returns the heat-maps? So basically hiding the `wealth_distr` function?

Datseris:
Here is the most reasonable scenario I would expect one would want to plot an ABM, which I think is also the most general case.

You plot a graph (either as a grid or just as a graph), but the color of the nodes is what you care to specify. The most general way to do this is to provide a function `f`. This function takes as an input the agents that are on this node and returns a number (or a color) as an output. This is what is used to plot. 

How did you set up the forest fire model? It was not done like this? Because it certainly _can_ be done like this. But the proposal I have here allows you to straight forwardly do what I am asking for the wealth model, by passing the function `f(agents) = mean(a.wealth for a in agents)`.

Datseris:
Okay, I see that you do: `:status, t=2, cc=Dict(true=>"green", false=>"red"),`. This is limiting, as it can only accomodate plotting with 1 agent per node. In addition it detaches the color output from the argument type `:status`, which dissallows further processing.

Wouldn't it better to combine these 2 arguments to a single function that does what I describe above?

kavir1698:
I see what you mean. It's a good suggestions. I would have to do it at some point in the future because right now I am getting a bit busy.

Datseris:
ok then. let's leave it as is for now, ill merge using the heatmap. Can you just ensure that tests pass?

Issue #98, opened 2019-12-10
Scheduler function & number of steps


NV2019:
I have two issues specific to my model. 

1. Right now, among the scheduler functions there is one that picks agents for update uniformly at random. How can I instead choose them according to a predetermined distribution?
2. What if I want the step function to run as long as a certain condition is satisfied? Without mentioning the number of steps beforehand?

kavir1698:
> Right now, among the scheduler functions there is one that picks agents for update uniformly at random. How can I instead choose them according to a predetermined distribution?

You can define your own scheduler function. A scheduler takes the model object as input and returns an list of agent IDs to be activated in turn. For example, the `random_activation` function is just this:

```jl
function random_activation(model::ABM)
  order = shuffle(collect(keys(model.agents)))
end
```

> What if I want the step function to run as long as a certain condition is satisfied? Without mentioning the number of steps beforehand?

This is not implemented. But I think it should be easy to add.

Datseris:
@NV2019 to asnswer 1, something like this (but needs to be tested!):

```julia
function by_weight(model)
w = [(a.id, a.weight) for a in model.agents]
s = sortperm(w, by = x -> x[2])
return [x[1] for x in w[s]]
end
```

Datseris:
Seems useful enough to be added to the existing schedulers, if you are up for it, and want to obtain a contributor status to Agents.jl, you can do a Pull Request @NV2019 .

Datseris:
@NV2019 I actually just implemented the point 1 that you ask in PR #101 . Once this is merged and a new version is tagged you can do:

```
model(AgentType; scheduler = property_activation(:property)
```
(with `:property` the actual field of the agents you want to compare, e.g. `:weight`)

NV2019:
Can I know if this modification is available and if the documentation has been updated regarding this and this issue[issue](https://github.com/JuliaDynamics/Agents.jl/issues/102#issue-535967690)

Datseris:
Huh? I don't understand the question. Everything is implemented, you can find it at the documentation of `step!` or at the API page: https://juliadynamics.github.io/Agents.jl/stable/api/#Schedulers-1   or    https://juliadynamics.github.io/Agents.jl/stable/examples/schelling/#Creating-an-ABM-1

PR #101, opened 2019-12-10
Add a scheduler API and improve scheduler documentation

Datseris:
I have changed `as_added`, to `by_id`, because "as added" is a misleading name. Dictionary keys have no sequence and also no memory of which is added first.

I've added scheduler `fastest` and made it the default, which is just the key iterator of the dictionary.

Added basic API tests for schedulers.

### DIFF
--- a/test/api_tests.jl
+++ b/test/api_tests.jl
@@ -58,11 +58,11 @@ a = model.scheduler(model)
 @test a[1] == 74 # reproducibility test
 
 # by property
-mutable struct Agent2 <: AbstractAgent
+mutable struct Agentw <: AbstractAgent
   id::Int
   w::Float64
 end
-model = ABM(Agent2; scheduler = property_activation(:w))
+model = ABM(Agentw; scheduler = property_activation(:w))
 for i in 1:N; add_agent!(model, rand()/rand()); end
 
 Random.seed!(12)
--- a/test/api_tests.jl
+++ b/test/api_tests.jl
@@ -20,3 +20,39 @@ pos1 = model1.space.agent_positions[coord2vertex((1,1), model1)]
 @test length(pos1) == 0
 pos2 = model1.space.agent_positions[coord2vertex((2,2), model1)]
 @test pos2[1] == 1
+
+# %% Scheduler tests
+N = 1000
+mutable struct Agent0 <: AbstractAgent
+  id::Int
+end
+# fastest
+model = ABM(Agent0)
+for i in 1:N; add_agent!(model); end
+@test sort!(collect(keys(model.agents))) == 1:N
+
+# by_id
+model = ABM(Agent0; scheduler = by_id)
+for i in 1:N; add_agent!(model); end
+
+@test sort!(collect(keys(model.agents))) == 1:N
+
+@test model.scheduler(model) == 1:N
+
+using Random
+Random.seed!(12)
+
+# random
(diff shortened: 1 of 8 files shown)


Datseris:
Ah, and I have re-worked `partial_activation` which was first of all not updated to Agents 2.0 (since it tried to do `model.partial_activation`. 

I have added the possibility to just use it as a functor (a function tied with variables), or so that it can access a model property (I personally never liked tying the model properties with activation parameters, but I can see why one would like to have the activation parameter as part of the `model.properties`).

Datseris:
Also added `property_activation` scheduler to address #98

Datseris:
@kavir1698 the more I think about `partial_activation` the less I like the possibility that it searches a model parameter. I mean, it makes everything messier. If the user wants they can add this probability as a model parameter but we don't have to search for it. In the end of the day it is advised that a model is created via a function anyway, so this function can take the activation probability and make it both a model parameter and also call `partial_activation(p)`.

I vote to remove the functionality of `partial_activation` that looks into the model's parameters.

kavir1698:
> "as added" is a misleading name. Dictionary keys have no sequence and also no memory of which is added first.

Well, agent ids show the addition sequence of agents. When we add a new agent, its ID is automatically 1+the largest current ID.

kavir1698:
> I've added scheduler fastest and made it the default, which is just the key iterator of the dictionary.

Great idea.

kavir1698:
> In the end of the day it is advised that a model is created via a function anyway, so this function can take the activation probability and make it both a model parameter and also call partial_activation(p).

I see. So it is automatically added to the model properties, instead of the user adding it, correct? It is a cleaner approach.

test/api_tests.jl
```diff
@@ -20,3 +20,55 @@ pos1 = model1.space.agent_positions[coord2vertex((1,1), model1)]
 @test length(pos1) == 0
 pos2 = model1.space.agent_positions[coord2vertex((2,2), model1)]
 @test pos2[1] == 1
+
```
kavir1698: It is worth putting tests inside a `@testset` block.

Datseris:
> Well, agent ids show the addition sequence of agents. When we add a new agent, its ID is automatically 1+the largest current ID.

That is true if the addition of agents is clean and done from scracth. If for some reason someone plays with the model a bit, and adds and kills agents, some intermediate IDs are lost. In addition, since in principle one could add agents with arbitrary id, this could be not safe. But I think the syntax of `by_id` cannot raise ambiguity, don't you think?

> I see. So it is automatically added to the model properties, instead of the user adding it, correct? It is a cleaner approach.

No, this is not what happens (and it is also not possible). The function `partial_activation` itself becomes a functor i.e. it stores internal variables. See "Fuctors" in the Julia docs for more. The `model.parameters` still remains `nothing` if it was set to `nothing`. I just think that trying to search for model parameters inside the scheduler can be too restrictive and if a user wants to do it they should do it on their own end, instead of being forced via our API.

test/api_tests.jl
Datseris: I am on it!

kavir1698:
> But I think the syntax of by_id cannot raise ambiguity, don't you think?

Agree.

> I just think that trying to search for model parameters inside the scheduler can be too restrictive and if a user wants to do it they should do it on their own end, instead of being forced via our API.

Alright, let's do it.

Issue #102, opened 2019-12-10
Stepping until condition is met


Datseris:
In #98 @NV2019 requested the following:

> What if I want the step function to run as long as a certain condition is satisfied? Without mentioning the number of steps beforehand?

which I think is a great and non-breaking addition. The way I propose to implement this is to use Multiple Dispatch. If `n` (the number of steps) is integer, nothing changes. If that argument is a function `f` instead then `step!` runs until `f(model)` gives `true`.

kavir1698:
I was thinking along the same lines. I will do it.

NV2019:
Thank you!

PR #103, opened 2019-12-11
Conditionally stop simulations

kavir1698:
This answers #102.
I have not updated the version, so that we can merge merge this with #101 into v2.1.0

### DIFF
--- a/src/simulations/data_collector.jl
+++ b/src/simulations/data_collector.jl
@@ -86,20 +86,20 @@ function data_collecter_raw(model::ABM, properties::Array{Symbol}; step=1)
 end
 
 """
-    data_collector(model::ABM, field_aggregator::Dict, step::Integer [, df::DataFrame]) where T<: Integer
+    data_collector(model::ABM, field_aggregator::Dict, step::Integer [, df::DataFrame])
 
 Used in the `step!` function.
 
 Returns a DataFrame of collected data. If `df` is supplied, appends to collected data to it.
 """
-function data_collector(model::ABM, field_aggregator::Dict, step::Integer) where T<: Integer
+function data_collector(model::ABM, field_aggregator::Dict, step::Integer)
   d, colnames = data_collecter_aggregate(model, field_aggregator, step=step)
   dict = Dict(Symbol(colnames[i]) => d[i] for i in 1:length(d))
   df = DataFrame(dict)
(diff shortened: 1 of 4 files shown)


kavir1698:
I see the following warning, but I don't see where the problem is. Any idea?

```jl
WARNING: Method definition step!(Agents.AgentBasedModel{A, S, F, P} where P where F where S<:Union{Nothing, Agents.AbstractSpace} where A<:Agents.AbstractAgent, Any, Any) in module Agents at /home/travis/build/JuliaDynamics/Agents.jl/src/simulations/step.jl:56 overwritten at /home/travis/build/JuliaDynamics/Agents.jl/src/simulations/step.jl:59.

  ** incremental compilation may be fatally broken for this module **
```

kavir1698:
There are some warnings that I should fix before it is merged.

src/simulations/step.jl
```diff
@@ -7,6 +7,9 @@ Update agents `n` steps. Agents will be updated as specified by the `model.sched
 If given the optional function `model_step!`, it is triggered _after_ every scheduled
 agent has acted.
 
+`n` can be a function instead of a number of step. If so, the model runs as long as `n`
```
Datseris: > the model runs as long as `n` returns `false`.

A better way to phrase this would be "as long as `n(model)` returns `false`.

Datseris:
Very good! I only have minor syntactic comments before merging!

src/simulations/data_collector.jl
```diff
@@ -133,11 +133,23 @@ function _step(model, agent_step!, model_step!, properties, when, n, step0)
...
+    end
+  else
+    ss = 1
+    while !n(model)
+      step!(model, agent_step!, model_step!, 1)
+      # collect data
+      if ss in when
+        df = data_collector(model, properties, ss, df)
+      end
+      ss += 1
```
Datseris: I'd personally recommend to make these 2 loops a simple function that dispatches on `n`. The syntax `typeof(n) .... ` is typically not used in Julia to dispatch. Having the core loop a function will also make the code more readable and also easier to debug and extend.

Datseris:
LGTM, merging when tests pass!

Datseris:
@JuliaRegistrator register()

PR #104, opened 2019-12-12
Add sample! function

kavir1698:
A common task during agent-based modeling, at least in biology and ecology, is replacing the agents with a sample of them. Since doing that with a dictionary is more difficult than an array, I have written this `sample!` function.

### DIFF
--- a/examples/wright-fisher.jl
+++ b/examples/wright-fisher.jl
@@ -1,59 +1,68 @@
 # # Wright-Fisher model of evolution
 
-# This is one of the simplest models of population genetics.
-# We implement a simple case of the model with a single locus in a haploid 
-# population of constant size.
+# This is one of the simplest models of population genetics that demonstrates the
+# use of [`sample!`](@ref).
+# We implement a simple case of the model where we study haploids (cells with a single set
+# of chromosomes) while for simplicity focusing only on one locus (a specific gene).
+# In this example we will be dealing with a population of constant size.
 
 # ## A neutral model
 
-# * Imagine a population of 100 haploid individuals.
-# * At each generation, 100 offsprings replace the parents.
+# * Imagine a population of `n` haploid individuals.
+# * At each generation, `n` offsprings replace the parents.
 # * Each offspring chooses a parent at random and inherits its genetic material.
 
 
 using Agents
+n = 100
 
 # Let's define an agent. The genetic value of an agent is a number (`trait` field).
-mutable struct Agent <: AbstractAgent
+mutable struct Haploid <: AbstractAgent
     id::Int
     trait::Float64
 end
 
-m = ABM(Agent)
+# And make a model without any spatial structure:
+m = ABM(Haploid)
 
-# Start 100 random individuals.
-for i in 1:100
-    add_agent!(m, rand()/rand())
+# Create `n` random individuals:
+for i in 1:n
+    add_agent!(m, rand())
 end
 
-# To create a new generation, we can use the `sample!` function. It chooses 100
+# To create a new generation, we can use the `sample!` function. It chooses
 # random individuals with replacement from the current individuals and updates
-# the model.
+# the model. For example:
 sample!(m, nagents(m))
 
 # The model can be run for many generations and we can collect the average trait
-# value of the population.
+# value of the population. To do this we will use a model-step function (see [`step!`](@ref))
+# that utilizes [`sample!`](@ref):
 
-# First, put the `sample!` in a step function that accepts a single argument:
-# model object `m`.
+modelstep_neutral!(m) = sample!(m, nagents(m))
 
-modelstep_neutral!(m::ABM) = sample!(m, nagents(m))
-
-# We can now run the model and collect data:
-using StatsBase: mean
+# We can now run the model and collect data. We use `dummystep` for the agent-step
+# function (as the agents perform no actions).
+using Statistics: mean
 
 data = step!(m, dummystep, modelstep_neutral!, 20, Dict(:trait => [mean]))
+
+# As expected, the average value of the "trait" remains around 0.5.
 
 # ## A model with selection
 
 # We can sample individuals according to their trait values, supposing that their
 # fitness is correlated with their trait values.
 
-m = ABM(Agent)
+m = ABM(Haploid)
 for i in 1:100
-    add_agent!(m, rand()/rand())
+    add_agent!(m, rand())
 end
 
 modelstep_selection!(m::ABM) = sample!(m, nagents(m), :trait)
 
 data = step!(m, dummystep, modelstep_selection!, 20, Dict(:trait => [mean]))
+
+# Here we see that as time progresses the trait comes closer and closer to 1,
+# which is expected as agents with higher traits have higher probability of being
+# sampled for the next "generation".
--- a/examples/wright-fisher.jl
+++ b/examples/wright-fisher.jl
@@ -1,0 +1,59 @@
+# # Wright-Fisher model of evolution
+
+# This is one of the simplest models of population genetics.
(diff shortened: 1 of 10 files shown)


src/core/agent_space_interaction.jl
```diff
@@ -185,3 +185,49 @@ function add_agent_single!(model::ABM, properties...)
...
+Replace agents of the `model` with a random sample of the current agents with
+size `n`. Optionally, choose an agent property `weight`(Symbol) to weight the sampling.
+
+# Keywords
+
+* `n`: the number of new agents.
+* `replace=true` whether sampling is performed with replacement.
+* `rng`::AbstractRNG=Random.GLOBAL_RNG: Optionally, specify a random number generator.
+"""
+function sample!(model::ABM, weight=nothing; n::Int, replace=true, ordered=false, rng::AbstractRNG=Random.GLOBAL_RNG)
```
Datseris: Keyword `ordered` exists in the function but is not documented. What does it achieve?

src/core/agent_space_interaction.jl
```diff
@@ -185,3 +185,49 @@ function add_agent_single!(model::ABM, properties...)
...
+        newids = sample(rng, collect(keys(model.agents)), weights, n, replace=replace, ordered=ordered)
+    else
+        newids = sample(rng, collect(keys(model.agents)), n, replace=replace, ordered=ordered)
+    end
+
+    newagents = [deepcopy(model.agents[i]) for i in newids]
+
+    for k in keys(model.agents)
+        delete!(model.agents, k)
+    end
```
Datseris: I wonder if there is some more efficient way to do this...

src/core/agent_space_interaction.jl
```diff
@@ -185,3 +185,49 @@ function add_agent_single!(model::ABM, properties...)
...
+"""
+function sample!(model::ABM, weight=nothing; n::Int, replace=true, ordered=false, rng::AbstractRNG=Random.GLOBAL_RNG)
+    if weight != nothing
+        weights = Weights([getproperty(i, weight) for i in values(model.agents)])
+        newids = sample(rng, collect(keys(model.agents)), weights, n, replace=replace, ordered=ordered)
+    else
+        newids = sample(rng, collect(keys(model.agents)), n, replace=replace, ordered=ordered)
+    end
+
+    newagents = [deepcopy(model.agents[i]) for i in newids]
```
Datseris: Why are the agents deepcopied? In general, what does this line aim to achieve?

kavir1698:
> @kavir1698 I must admit, I have not understood what this function aims to do. What does Replace mean? Do some agents stay in the model and others leave? I really want to understand what the function does without looking at the code (because I think this code is not the most performant to do what the function wants to do).

This is simply sampling (with replacement). It is a standard model for natural selection. If `model.agents` was an `Array` instead of a `Dict`, then what this function does could be done like this:

```jl
n = length(model.agents)
newids = sample(1:n, WeightsVector, n)
model.agents = model.agents[newids] 
```

> How? How is weight used to sample? because for StatsBase the weights must be either frequencies or formal probabilities, adding to 1. How is this used here where any arbitrary property can be used?

`weight` parameter is an agent field that determines the replacement likelihood. I think it does not have to sum to 1. The sample function normalizes them internally.

src/core/agent_space_interaction.jl
```diff
@@ -185,3 +185,49 @@ function add_agent_single!(model::ABM, properties...)
...
+
+####################################
+### sample
+####################################
+
+"""
+        sample!(model::ABM [, weight]; kwargs...)
+
+Replace agents of the `model` with a random sample of the current agents with
+size `n`. Optionally, choose an agent property `weight`(Symbol) to weight the sampling.
```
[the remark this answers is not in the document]
kavir1698: I wanted to keep the order of parameters same as the `sample` function. But yeah, I'll change it.

src/core/agent_space_interaction.jl
kavir1698: It should be removed. I overlooked it.

src/core/agent_space_interaction.jl
kavir1698: Me too. Although, it seems to be not too bad.

src/core/agent_space_interaction.jl
kavir1698: Because I noticed that without deepcopying, when an agent is replaced more than once, all the resulting new agents are connected: mutating one of them would mutate them all.

Datseris:
I still need more explanation. For a moment imagine that I do not know what `sample` means. Please confirm or disprove:

This function `sample!` "normalizes" `model.agents`, so that only a subest of `n` agents remains alive, randomly choosing this subset, while all others die. If weights are provided then this random choice is in fact not equibprobable but weighted.

I am phrasing the question in this way because I think it is important for algorithm performance design.

kavir1698:
> This function sample! "modifies" model.agents, so that only a subest of n agents remains alive, randomly choosing this subset, while all others die. If weights are provided then this random choice is in fact not equibprobable but weighted.

Basically correct. But the agents that remain alive can be as many as or more than the current agents. That would be equivalent to "reproduction". It is achieved by sampling with replacement, meaning that one "fitter" agent (defined by the weight parameter) can represent multiple agents in the next generation while some agents would not have any children.

Datseris:
I'll think a bit about this more over the weekend and I'll let you know. I somehow feel convinced we can make this more performant.

Datseris:
Sorry, I was changing apartments over the weekend, I'll do this soon!

kavir1698:
No worries! All the best From: George DatserisSent: Monday, December 16, 2019 4:21 PMTo: JuliaDynamics/Agents.jlCc: Ali R. Vahdati; MentionSubject: Re: [JuliaDynamics/Agents.jl] Add sample! function (#104) Sorry, I was changing apartments over the weekend, I'll do this soon!—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

Datseris:
Alright, so here is the thing: I want to spend a bit of time here to eradicate the `haspos` loop. I want to define a function that "adds an agent to the position specified by its field `pos` (if it has one).

Datseris:
Okay @kavir1698 , can you have a look at my changes? I simplified `sample!` significantly, and did some documentation improvements. 

In addition, since you use these kind of functionalities in your work, maybe you are familiar with a simple example of natural selection on ABMs you could write for our documentation, that highlights the use of `sample!` ?

kavir1698:
Thank you. I will add an example to the docstring.

Datseris:
I meant more like an example like the forest fire, or the wealth distribution or so. The more examples we have in this column the better!

kavir1698:
> I meant more like an example like the forest fire, or the wealth distribution or so. The more examples we have in this column the better!

Yeah sure.

Datseris:
Awesome, this is good to go!

examples/wright-fisher.jl
```diff
@@ -1,59 +1,68 @@
...
+
 # ## A model with selection
 
 # We can sample individuals according to their trait values, supposing that their
 # fitness is correlated with their trait values.
 
-m = ABM(Agent)
+m = ABM(Haploid)
 for i in 1:100
-    add_agent!(m, rand()/rand())
```
kavir1698: Let's keep `rand()/rand()` so that the range of values is larger.

Datseris:
@kavir1698  with `rand()` even a user not really familiar with statistics and distributions can understand that the mean value should be 0.5 in the first example and 1.0 in the second (that is why I changed it). With `rand()/rand()` it is not so clear as it depends so much on the initial distribution.

PR #110, opened 2019-12-18
Bug fix: update space in add_agent!

kavir1698:


Datseris:
Hm, maybe in this branch its a good idea to add a `@test` that uses `add_agent!` and then really looks into `model.space.agent_positions` to ensure that the id is added?

PR #111, opened 2019-12-18
Add agent bug

kavir1698:
I have added more tests

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,7 @@
 using Test, Agents, Random
 
+@testset "all tests" begin
+
 include("api_tests.jl")
 include("forest_fire_defs.jl")
 include("space_test.jl")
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -1,9 +1,13 @@
 using Test, Agents, Random
 
+@testset "all tests" begin
```
Datseris: This choice could lead to problems in CI Travis if the entirety of the test set takes more than 10 minutes. Because with this choice there are messages printed only when all tests finish (or there is an aerror). Therefore if all tests take more than 10 minutes, Travis will error.

What is the motivation behind it?

kavir1698: The motivation is to see all test failures at once. If a test fails, other tests are still run.
Currently, tests don't take more than a minute.

Datseris: Okay then, but let's keep this in mind as we add more tests.

Issue #72, opened 2019-11-24
Increase test coverage


Datseris:
At the moment there are functions in the API of Agents.jl that are not tested _directly_. Direct tests are useful, because they make bugs harder to hide, and during updates or enhancements they point out directly to the source of the problem.

Besides, increasing test coverage is an excellent way to contribute for a newcomer!

Datseris:
Another thing we constantly do is defining agent types all over the test suite. We should just define three dummy types
```
Agent1 
id
end

Agent2
id
pos
end

Agent3
id
pos
a
```

and just use them throughout. We can define them in either `Agents` module or the top of the `runtests.jl`.

PR #121, opened 2019-12-24
Add unit test for nodes function

suzil:
- Additionally aggregates dummy agents into `runtests.jl`

Refs: #72


test/space_test.jl
```diff
@@ -152,4 +152,23 @@ end
...
+  end
+  space = Space((3,3))
+  model = ABM(MyAgent, space)
+  for node in nodes(model)
+    if rand() > 0.7
+      add_agent!(node, model)
+    end
+  end
+  @test nodes(model) == 1:9
+  @test nodes(model, by=:random) == [8, 9, 1, 7, 6, 5, 3, 2, 4]
```
Datseris: Shouldn't you set the seed here, to ensure that this random sequence will be the same all the time?

suzil: There's a seed set at the beginning of the file.

test/runtests.jl
```diff
@@ -1,5 +1,37 @@
...
+  weight::Float64
+end
+
+mutable struct Agent3 <: AbstractAgent
+  id::Int
+  pos::Tuple{Int,Int}
+  weight::Float64
+end
+
+mutable struct Agent4 <: AbstractAgent
```
suzil: Maybe some of these could be further combined? I wasn't sure

Datseris: Yeah, in total we need only 4 types of agents for the tests, the multiplicity of with and without position, and with and without some extra parameter `weight` or `a` or whatever.

Datseris: Sorry, I was mistaken. You are correct we need 6 agents, because no position, integer position and tuple position are 3.

PR #122, opened 2019-12-24
Add unit test for model show()

suzil:
Refs:  #72


Project.toml
```diff
@@ -9,6 +9,7 @@ Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
```
Datseris: This should not be a dependency of the package, but a test dependency instead. An example where this happens is here: https://github.com/JuliaDynamics/DrWatson.jl/blob/master/Project.toml#L20-L30

suzil: Thank you! I'll update it

suzil: I'm a bit stuck on how to fix this :see_no_evil: It seems to work fine for Test, but for some reason Suppressor doesn't want to get installed. I have added the latest version of Suppressor.
```
ERROR: Unsatisfiable requirements detected for package Suppressor [fd094767]:

 Suppressor [fd094767] log:

 ├─possible versions are: 0.1.1 or uninstalled

 ├─restricted to versions * by an explicit requirement, leaving only versions 0.1.1

 └─restricted by compatibility requirements with Compat [34da2185] to versions: uninstalled — no versions left

   └─Compat [34da2185] log:

     ├─possible versions are: [1.0.0-1.0.1, 1.1.0, 1.2.0, 1.3.0, 1.4.0, 1.5.0-1.5.1, 2.0.0, 2.1.0, 2.2.0, 3.0.0] or uninstalled

     └─restricted to versions 3.0.0 by an explicit requirement, leaving only versions 3.0.0
```

Datseris: I don't know either. Probably suppressor and compat have compatibility problems. There doesn't seem to be something we can do on our side.

Datseris: If there is a way to do this without Supressor then its better. If not, then its okay if we can't solve this, in the end of the day `show` is not part of the API, its okay if it is not tested.

suzil: Aha OK. I'll close this PR as it's not worth the trouble for this.

PR #128, opened 2020-01-23
Fix ca2d

### LINKED ISSUE #117: Is the game of life animation correct?
@kavir1698 , when I look here: https://juliadynamics.github.io/Agents.jl/dev/examples/game_of_life_2D_CA/

you see that in the animation all pixels "fall down", i.e. move downwards. Even a group of 2 pixels side-by-side move downwards, even though this clearly contradicts the game's of life rules.

kavir1698:
Do not merge yet. Closes #117 
I found bugs in the old CA2D.jl but now all the cells die out. Can you spot the problem here?


src/CA2D.jl
```diff
@@ -24,86 +24,33 @@ function build_model(;rules::Tuple, dims=(100,100), Moore=true)
...
-    bottomright = model.agents[coord2vertex((after[1], before[2]), model)].status
-    bottomleft = model.agents[coord2vertex((before[1], before[2]), model)].status
-
-    if model.properties[:Moore]
-      nstatus = [topleft, top, topright, left, right, bottomleft, bottom, bottomright]
-      nlive = length(findall(a->a=="1", nstatus))
-    else
-      nstatus = [top, left, right, bottom]
-      nlive = length(findall(a->a=="1", nstatus))
+  new_status = Array{String}(undef, nagents(model))
```
Datseris: shouldn't this be `Bool` instead of `String`...? It takes up much less memory.

kavir1698: How would you write a Bool array for the rules?

Datseris: By looking the code further down I see that `new_status` takes only the values `"0"` or `"1"`, so true or false?

Datseris:
@kavir1698  bump here: I am not that familiar with game of life rules and how they work, I am not sure I can be of much help for the actual physics. For performance coding I gave some ideas already :)

kavir1698:
I have updated the game of life example. It behaves as expected now. I also changed it and CA1D from a special module to a normal example. 

Only remaining point is visualizing 1D CA because the old plotting functions from AgentsPlots are removed now.

examples/rule22_1D_CA.jl
```diff
@@ -1,11 +1,60 @@
...
+end
+
+function ca_step!(model)
+  agentnum = nagents(model)
+  new_status = Array{String}(undef, agentnum)
+  for agid in 1:agentnum
+    agent = model.agents[agid]
+    center = agent.status
+    if agent.id == agentnum
+      # right = "0"  # this is for when there space is not toroidal
```
Datseris, line 38: there -> the

examples/rule22_1D_CA.jl
```diff
@@ -1,11 +1,60 @@
...
+      right =  model.agents[1].status
+      left = model.agents[agent.id-1].status
+    elseif agent.id == 1
+      # left = "0"  # this is for when there space is not toroidal
```
Datseris, line 42: When using Literate.jl, comments inside code must start with `##` instead of just 1. OTherwise they are interpreted as markdown, and this ruins all code.

Datseris:
My opinion is that having 2D CA is enough, no reason for 1D as well.

Datseris:
> I also changed it and CA1D from a special module to a normal example.

Great! Having them as modules was always offputting for me.

kavir1698:
All is clear then.

Datseris:
Here is the doc preview: https://juliadynamics.github.io/Agents.jl/previews/PR128/  I don't see the game of life. I think you have to re-enable it in `make.jl` (i've removed it from both literate as well as make)

Datseris:
The animation is very fast, but more importantly, it is very difficult to understand what is going on with your eyes: https://juliadynamics.github.io/Agents.jl/previews/PR128/examples/game_of_life_2D_CA/

maybe make a smaller system and much smaller FPS?

Libbum:
Just found out that John Conway died from COVID-19 two days ago. :disappointed:

kavir1698:
Found and fixed another bug. This is important to remember. In Plots.jl, `(1, 2)` means second row, first column (why?!). In Agents.jl, it means first row, second column. This was the reason game of life animations did not look right.

kavir1698:
> Just found out that John Conway died from COVID-19 two days ago. 

☹

Datseris:
@kavir1698 there are some old stuff in the example that need to be updated. It starts by saying "agents provide a module..."

Also, it is better if the docstring of `build_model` is made normal text in the markdown code blocks.

Also: "Now we define a stepping function for the model to apply the rules to agents This function creates a model " missing `.` before "This"

Lastly: please set a `Random.seed!` that generates a nice animation, so that the doc shows the same thing in every commit

kavir1698:
It's not easy to find an initialization that makes one of those guns or self replicating shapes. We are initializing it randomly. I have added a seed, though.

Datseris:
Maybe change `model.properties[:rules]` to `model.rules` throughout? will eliminate scrolling in the largest code block (big plus)

PR #131, opened 2020-03-12
use JuliaDynamics theme

Datseris:
PR is visible here: https://juliadynamics.github.io/Agents.jl/previews/PR131 

Ali it would be cool to also have a logo for the sidebar, it is displayed above the name on the left side. I wonder what we could use for that. Maybe we can just crop the capital A of the full logo?


kavir1698:
The code names inside backticks are a bit difficult to read. It's good to improve their contrast.

Datseris:
> The code names inside backticks are a bit difficult to read. It's good to improve their contrast.

Is this for the light or dark theme? Also, does this refer to inline code, like `code`, entire codeblocks like 
```
code
```
or inline code _with_ links, like [`code`](link)?

kavir1698:
This is for light theme (how can I see the dark theme?). And for the inline `code`. See for example, the inline code on the Tutorial page, numbered list.

Datseris:
You can see the dark theme by clicking the gear on the top right corner. I just saw the docs, and I realize that it used incorrect fonts. I believe the new fonts are much more readable, please have a look once the docs build and if you still find it unreadable let me know. The code text color is pure black, so the only thing we can do better is make the code background even lighter.

kavir1698:
The dark theme has everything well visible.

The font color is green, though.

Datseris:
Okay, then you are talking about the inline code _with_ links. The links and these things get green color. 

Let me know how this is for the new font. You see examples here: https://juliadynamics.github.io/DrWatson.jl/dev/name/#  and let me know whether you would like it more dark, more light, or even a different color. I find it perfectly readable as is (with the correct fonts)

kavir1698:
The new font solves the problem. Perfect.

PR #133, opened 2020-03-13
Add SIR model

kavir1698:
We can make this model better by constructing inner nodes for each city.


Datseris:
Awesome, I'll contribute once I return home!

Datseris:
@kavir1698 by the way, should we maybe simplify this, by making the graph not rectangular but some arbitrary abstract graph? I mean, maybe it is more intuitive to represent connected cities this way? Or you think the rectangle is better for visualizations?

kavir1698:
The graph connections don't matter. Connections between cities are encoded in `migration_rates`. The graph structure is used just to record agent positions.

Datseris:
@kavir1698 is it okay if we name the agent type `PoorSoul`? :D

Datseris:
I am currentlyu making a plotting function for ABMs on arbitrary grids. It takes a huge ass amount of time, but I bet many dollars it will be worth it in the end.

Datseris:
@kavir1698 I've updated the initial part of the example with a sketch of how it will be with cities on a plottable graph. Is it okay if we make the parameters `Bu` and `Bd` numbers, i.e. all cities get the same number ? It will simplify things drastically I think.

kavir1698:
> is it okay if we name the agent type PoorSoul? :D

I don't mind :)

> I am currentlyu making a plotting function for ABMs on arbitrary grids. It takes a huge ass amount of time, but I bet many dollars it will be worth it in the end.

Great!

> Is it okay if we make the parameters Bu and Bd numbers, i.e. all cities get the same number ? It will simplify things drastically I think.

Allowing heterogeneity is a selling point for using ABM to study disease spread. I encourage keeping Bu and Bd as arrays.

kavir1698:
> I've updated the initial part of the example with a sketch of how it will be with cities on a plottable graph.

I don't quite see the advantage of this addition. Our cities can be arbitrarily connected with the `migration_rates` matrix. What would those random edges between cities mean?

Datseris:
> I don't quite see the advantage of this addition.

> What would those random edges between cities mean?

They are not random, they are supposed to be the "roads" between cities :D 

I hope that we are able to animate the percentage of infected per city using the new plotting functionality I PR'd to AgentsPlots.jl. If we are able to visualize it, I think it can become a much more seen example.

(in this case the transmission probability is just proportional to the city size)

kavir1698:
> They are not random, they are supposed to be the "roads" between cities :D

> (in this case the transmission probability is just proportional to the city size)

I see. But transmission probabilities among cities, and connections between them (be it by land or air) are already encoded in `migration_rates`. I suggest we build the edges between nodes from the `migration_rates` matrix, where edge size corresponds to amount of travel/migration.

Datseris:
I actually just learned some basics of networks overnight, and choose this network type as it appropriatelly create hubworlds, which are probably relevant for reality.... I don't know how to make a network from migration rates and vice versa, if you can contribute this it would be great!

kavir1698:
Ok, in this commit, I plot the network of cities with migration rates as edge thickness.

kavir1698:
Thanks. I think we need to import LightGraphs too.

> # Notice this example is mostly for education purposes, not
> # research purposes. In a research scenario one would need more complex agents, with
> # age information, and all infection parameters would not be constants but instead
> # depend on the agent's properties.

This is not necessarily true. The level of model realism depends on research questions. This model can be used for research.

Datseris:
I count this as a huge step of progress for Agents.jl :)

@kavir1698 one thing I want to ask: is it possible you think to modify the model so that there are more cities, or larger network? The one we have is nice, but not so "real looking" visually, I'd like to make a larger one. How hard do you think it would be for you to create the network and transmission rates programmatically given `Ns` ?

(so that our version is a bit closer visually to the Pathogen.jl one)

kavir1698:
> How hard do you think it would be for you to create the network and transmission rates programmatically given Ns ?

Not hard at all. I'll change the example with more cities.

PR #136, opened 2020-03-15
Social distancing model

kavir1698:
This is a disease spread model inspired by [this article](https://www.washingtonpost.com/graphics/2020/world/corona-simulator/) and our SIR model. What I show here is agents moving in a continuous space. I used a SQL data structure for quickly finding agents in a given area. 

To do:
[] Add some explanations.

### DIFF
--- a/examples/social_distancing.jl
+++ b/examples/social_distancing.jl
@@ -1,10 +1,18 @@
+# # A model of social distancing for spread of disease.
+
+# This is a model similar to our SIR model of disease spread. But instead of having different cities, we let agents move in one continuous space and transfer the disease if they come into contact with one another. This model is partly inspired by [this article](https://www.washingtonpost.com/graphics/2020/world/corona-simulator/).
+
+# For a detailed description of the basics of the model, see the SIR example.
+
+# We note that Agents.jl does not have an API for modeling continuous space. As we implement it, this page will be updated as well.
+
 using Agents, Random, DataFrames, SQLite, Plots
 using DrWatson: @dict
 
 mutable struct Agent <: AbstractAgent
   id::Int
   pos::Tuple{Float64, Float64}
-  dir::Float64
+  dir::Float64  # direction of movement
   days_infected::Int  # number of days since is infected
   status::Symbol  # 1: S, 2: I, 3:R
 end
@@ -16,6 +24,7 @@ function model_initiation(;infection_period = 30, moveprob = 0.4,
   Random.seed!(seed)
   # movedist = Normal(0, movement)
   # Create a database of agent positions
+  # A database allows fast access to agents within any area.
   db = SQLite.DB()
   stmt = "CREATE TABLE tab (
     x REAL,
--- a/examples/social_distancing.jl
+++ b/examples/social_distancing.jl
@@ -60,7 +60,7 @@ function move!(agent, model)
   newx > 1.0 && (newx = newx - 1.0)
   newy < 0.0 && (newy = newy + 1.0)
   newy > 1.0 && (newy = newy - 1.0)
-  agent.pos = Tuple([newx, newy])
+  agent.pos = (newx, newy)
 end
 
 function transmit!(agent, model)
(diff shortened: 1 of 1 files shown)


examples/social_distancing.jl
```diff
@@ -0,0 +1,130 @@
...
+
+function move!(agent, model)
+  rand() > model.properties[:moveprob] && return
+  newx = agent.pos[1] + (model.properties[:movement] * cos(agent.dir))
+  newy = agent.pos[2] + (model.properties[:movement] * sin(agent.dir))
+  newx < 0.0 && (newx = newx + 1.0)
+  newx > 1.0 && (newx = newx - 1.0)
+  newy < 0.0 && (newy = newy + 1.0)
+  newy > 1.0 && (newy = newy - 1.0)
+  agent.pos = Tuple([newx, newy])
```
Datseris: Why not just `agent.pos = (newx, newy)`...?

Datseris:
I must be honest here, the way continuous space exists seems like a "hack" to me... Agents.jl does not support continuous space yet, but somehow this example manages to do it. I understrand how of course: the space is in fact represented by the agents themselves, there is no need for `Space`. But I still feel like there is something "off" here... Can you talk me more into it?

examples/social_distancing.jl
```diff
@@ -0,0 +1,130 @@
...
+    if contact.status == :S || (contact.status == :R && rand() ≤ prop[:reinfection_probability])
+      contact.status = :I
+    end
+  end
+  # change direction
+  firstcontact = id2agent(r[1,:id], model)
+  dir1 = deepcopy(agent.dir)
+  dir2 = deepcopy(firstcontact.dir)
+  agent.dir = dir2
+  firstcontact.dir = dir1
```
Datseris:
```suggestion
  agent.dir, firstcontact.dir = firstcontact.dir, agent.dir
```

this is enough to exchange directions, `Float64` is immutable.

examples/social_distancing.jl
kavir1698: Yeah, that works. I don't know what error I got with that.

kavir1698:
> I must be honest here, the way continuous space exists seems like a "hack" to me... Agents.jl does not support continuous space yet, but somehow this example manages to do it, using something I am entirely unfamiliar with (this database thingy). I understand how continuous space exists here: the space is in fact represented by the agents themselves, there is no need for Space. But I still feel like there is something "off" here... Can you talk me more into it?

Agents.jl does not support continuous space, yet. What I am showing here is a general and efficient way to implement it in Agents.jl. This approach can incorporate 1, 2, or 3D continuous space in an efficient way. You can see the code runs pretty fast. At its core, the basic functionality of a continuous space is two fold: first, being able to quickly retrieve agents in any defined area. Second, moving agents in this space by providing a direction and step size. The SQLite database makes it possible to quickly retrieve agents with a given position. There may also be other data structures to achieve this objective, like JuliaDB (?) or some sorted lists. We can think about that. But with SQLite it is already clean and efficient, IMO.

I am not sure how to best put this continuous space in Agents.jl's user interface. I hope we can brainstorm about that.

Datseris:
Hm, this is something that needs a bunch of thinking. I'll need some days before I can formulate an opinion. In general I think this should be supported properly as you suggest as well. On the other hand, this is a nice example, perfectly runnable, and not that slow either. So I think we should publish it, but its very first line being something like "this is an example for continuous space, which does not yet have a core API in Agents.jl. We are thinking of the best way to implement continuous space in Agents.jl, and then this example will be updated accordingly."  (or something less verbose).

Can you go ahead and add some more text, referring to the previous example to not replicate that much, so that it is a "properly documented example", and we can merge this?

---

For discussing the continuous space, I think we should have a different PR with the simplest possible ABM: just agents exchanging their position, so that we can focus on the API. The thing with continuous space that adds even more complication is that one shoould think about "radius of interaction".

kavir1698:
> this is a nice example, perfectly runnable, and not that slow either.

Yeah, the slow part is plotting. Running 500 generations of simulation and data collection takes about 2 seconds for me.

> Can you go ahead and add some more text, referring to the previous example to not replicate that much, so that it is a "properly documented example", and we can merge this?

Sure, I'll do that tomorrow.

> For discussing the continuous space, I think we should have a different PR with the simplest possible ABM: just agents exchanging their position, so that we can focus on the API. The thing with continuous space that adds even more complication is that one shoould think about "radius of interaction".

I have some idea for making that decision automatic. We can talk about it in the new PR.

Issue #137, opened 2020-03-16
Continuous space brainstorming


Datseris:
So, let's get this party started!!! The end goal of this discussion is a new space type, `ContinuousSpace` and robust support for this in Agents.jl.

First, let's make a wishlist of all the things that are at least in imagination possible, and then as the ideas mature we will see which ones are possible:

1. A `ContinuousSpace` struct.
1. from 1 to 3 dimensions.
1. Support for periodic or not boundary conditions.
1. Concrete way for the agents to have velocity (does it also have to be a specfic field of the agent?)
1. Interaction radius: in principle could take any real number as value.
1. Interaction radius could be on a CityBlock metric, or Euclidean metric (square shape versus circle shape)
1. Interaction a' la billiards: agents have a "size", and interaction is literally collision between the balls of the agents. We can use HardSphereDynamics.jl for this.
1. Can all agents move in parallel, or is it only possible to move sequentially...? (For HardSphereDynamics, all spheres move in parallel)
1. Avoiding double interactions: the framework should be able to understand that two agents have _just interacted_, and thius, even if they are in the interaction radius of each other, they shouldn't interact again.
1. Allow generic equations of motion, i.e. besides linear motion, acceleration, deceleration, repulsion

*I will be updating this list as we talk about this more*

When comparing _different_ space implementations, these are the things that should be compared performance-wise:

* Initializing spatial structure
* moving 1 agent in spatial structure
* adding 1 agent in spatial structure
* finding neighbors of 1 agent in spatial structure, for several different radii

Datseris:
@kavir1698, please describe here your idea of using a database like you used in the social distancing model. 

---

Another idea comes from a colleague of mine @lhupe , and goes like this: 

* There is an underlying graph, and it is the same as the graph of `GridSpace` (i.e. there is a grid).
* One keeps track of the Agent's continuous position (it is their field `pos`), and also their position on the `grid`.
* The interaction units are multiples of the grid spacing.
* Finding all agents our agent interacts with is the same as finding the neighboring nodes.
* An internal "move" function is used at each agent, and then interaction is checked.

> one keeps track both of a continuous position and a position on a grid space, with one grid unit corresponding to the transmission range. Then, you can efficiently find all particles within the transmission range by iterating over the particles on the neighbouring grid cells – without any SQL.

kavir1698:
[...]

PR #138, opened 2020-03-16
Continuous space

kavir1698:
A suggestion for #137 . Minimum working example of continuous space.


examples/continuous_space.jl
```diff
@@ -0,0 +1,90 @@
...
+# This is a proof of concept for continuous space.
+# The final api can use ideas in this example.
+
+using Agents, Random, DataFrames, SQLite, Plots
+using DrWatson: @dict
+
+mutable struct Agent <: AbstractAgent
+  id::Int
+  pos::Tuple{Float64, Float64}
+  dir::Float64  # direction of movement
```
Datseris: I think it is best if `dir` is the same type as `pos`, i.e. it explicitly provides all components of the velocity. This will allow us changing velocities of agents as well as direction. Thus, I vote for a field `vel::Tuple{2}`. You can update the position of the agent by simply doing `a.pos .+= δt .* a.vel`

kavir1698: Good point!

Datseris:
@kavir1698 , while I think about the implementation of the grid version (so that we can compare), do you think you can create a `ContinuousSpace` struct, which stores the database entry? And when initializing the continuous space, the database is also initialized.

What you have set up already looks very promising so I guess we should think the user API now.

Once I have a rough implementation with the graph version I mentioned, I will benchmark the two. If the version here is much more performant, I won't even bother to continue.

kavir1698:
I'll give it a try.

kavir1698:
@Datseris I have created a new space type. The example is not working yet. `Base.show` and `add_agent!` need to be updated to support the new space object.

examples/continuous_space.jl
```diff
@@ -0,0 +1,140 @@
...
+  q2 = DBInterface.prepare(db, searchq)
+  ContinuousSpace(db, q, q2)
+end
+
+"Add many agents to the database"
+function fill_db!(agents, model::ABM{A, S}) where {A, S<:ContinuousSpace}
+  db = model.space.db
+  for agent in agents
+    p1, p2 = agent.pos
+    DBInterface.execute(model.space.insertq, [p1, p2, agent.id])
```
Datseris: Do all these execute commands require the arguments to be passed in as vectors, or `Tuple` is possible as well? Because this results to a lot of allocation.s

kavir1698: Tuple is possible too. It does improve performance a lot!

Datseris: Look out for these things! `[stuff...]` allocates memory! I've seen you use it quite often, but be careful! Most of the times it is not really necessary!

Datseris:
@kavir1698 I think we have something really solid here. What I'd recommend is that we actually merge this PR, even though it is not entirely complete. Then we individually start working on some of the TODOs as we find more spare time. What do you think?

kavir1698:
Alright, let's do it.

PR #139, opened 2020-03-17
Implement and test genocide!(model::ABM, f::Function)

### LINKED ISSUE #134: add genocide!(model, f::Function)
`genocide!(model, f::Function)` kills all agents where the function `f(agent)` returns `true`.

Libbum:
This adds unit tests for the `genocide!` functions in the API, and implements the requested additional `genocide!(model::ABM, f::Function)`, which closes #134.

I've attempted to keep to the style of similar methods and tests as much as practicable.

For the moment we just throw a `TypeError` if `f` doesn't return `Bool`. Fairly certain that's OK based on other unit tests in the project, but could be easily changed.

### DIFF
--- a/src/simulations/sample.jl
+++ b/src/simulations/sample.jl
@@ -26,7 +26,7 @@ end
 
 """
     genocide!(model::ABM, f::Function)
-Kills all agents where the function `f(agent)` returns `true`.
+Kill all agents where the function `f(agent)` returns `true`.
 """
 function genocide!(model::ABM, f::Function)
     for (k, v) in model.agents
--- a/src/simulations/sample.jl
+++ b/src/simulations/sample.jl
@@ -25,6 +25,18 @@ function genocide!(model::ABM, n::Int)
 end
 
 """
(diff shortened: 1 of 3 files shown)


src/simulations/sample.jl
```diff
@@ -24,6 +24,18 @@ function genocide!(model::ABM, n::Int)
     end
 end
 
+"""
+    genocide!(model::ABM, f::Function)
+Kills all agents where the function `f(agent)` returns `true`.
```
Datseris:
```suggestion
Kill all agents where the function `f(agent)` returns `true`.
```

Datseris:
Thanks a lot, great contribution. THe only change I did was `kills` -> `kill`, because it is advised by the Julia style guide to write docstrings in imperative tone "do this, do that".

Datseris:
Ah one last thing @Libbum : please increment minor version in `Project.toml`.

Libbum:
Sure thing&mdash;done.

PR #143, opened 2020-03-18
periodicity in continuous space

Datseris:
Adds periodicity for continuous space

### DIFF
--- a/examples/continuous_space.jl
+++ b/examples/continuous_space.jl
@@ -15,7 +15,7 @@ end
 
 function model_initiation(;N=100, speed=0.005, space_resolution=0.001, seed=0)
   Random.seed!(seed)
-  space = Space(2)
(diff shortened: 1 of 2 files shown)


Datseris:
@kavir1698 I am a bit confused with this database. at the moment `move_agent!` doesn't do anything to the database. How does the database knows that the agent has been moved?

kavir1698:
@Datseris Yeah, let me try to understand what is happening.

kavir1698:
Ok. The database doesn't get updated, but agent objects do. And the plotting function in the example uses agent objects, not the database.

Datseris:
Okay but then we have to fix this right? Because when calling `space_neighbors`, this checks the database. So after each agent is moved the database has to be updated for it to be correct?

kavir1698:
Yes, correct. I'll take care of that. We will need another statement for updating the database.

PR #144, opened 2020-03-18
Extend kill_agent! function to continuous space

kavir1698:


Datseris:
kill_agents! seems very familiar with `genocide!`. Do we need two names or you think `genocide!` is enough?

src/core/continuous_space.jl
```diff
@@ -146,3 +149,35 @@ function move_agent!(agent::A, model::ABM{A, S, F, P}) where {A<:AbstractAgent,
...
+  "DELETE FROM tab WHERE id IN $(Tuple(agentIDs))")
+  for id in agentIDs
+    delete!(model.agents, id)
+  end
+  return model
+end
+
+function genocide!(agents::AbstractArray, model::ABM{A, S}) where {A, S<:ContinuousSpace}
+  DBInterface.execute(model.space.db,
+  "DELETE FROM tab WHERE id IN $(Tuple([a.id for a in agents]))"
```
Datseris: Once again here `[]` is not necessary. `Tuple(rand() for i in 1:100)` is valid syntax. But making large tuples is bad for performance. What is the goal here? Is it to just interpolate into a string all ids? `"1, 23, 5, 6"` with commas?

src/core/continuous_space.jl
```diff
@@ -146,3 +149,35 @@ function move_agent!(agent::A, model::ABM{A, S, F, P}) where {A<:AbstractAgent,
...
   return agent.pos
 end
+
+function kill_agent!(agent::AbstractAgent, model::ABM{A, S}) where {A, S<:ContinuousSpace}
+  DBInterface.execute(model.space.deleteq, (agent.id,))
+  delete!(model.agents, agent.id)
+  return model
+end
+
+function genocide!(agentIDs::AbstractArray{Int}, model::ABM{A, S}) where {A, S<:ContinuousSpace}
```
Datseris: Maybe it makes sense to use the already existing and intuitive API of `genocide!`:  https://juliadynamics.github.io/Agents.jl/stable/api/#Agents.genocide!  and have the input arguments be the exact same 3 combinations?

src/core/continuous_space.jl
kavir1698: we need `(1,2,3...)` in the string.

Datseris: Then please use the same approach I have done for the other queries using `join(...)`. String operation and manipulation is what we need at this occassion, and will also bring performance benefits.

src/core/continuous_space.jl
```diff
@@ -125,7 +124,24 @@ function add_agent!(model::ABM{A, <:ContinuousSpace}, properties...) where {A}
   # the column "id" of the database? There _has_ to be a way for it.
```
kavir1698: I don't think it is inefficient. It's like saying `maximum(somelist)` in julia.

PR #150, opened 2020-03-18
Some tests on continuous space

kavir1698:
For #146

PR #152, opened 2020-03-19
space_neighbors functionality

Datseris:
After we merge #150, I'll update this and add tests for space_neighbors for both discrete and continuous space.

Good news: this is not even breaking, because `node_neighbors` does a different thing and we can keep it!

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -7,7 +7,7 @@ Agents.jl is part of [JuliaDynamics](https://juliadynamics.github.io/JuliaDynami
 To get started, please read the [Tutorial](@ref) page.
 
 !!! info "Latest news"
-    Welcome to Agents.jl 3.0! In this major release we have full support for agent based models on _continuous_ space. We also simplified some interfaces, see the [CHANGELOG](https://github.com/JuliaDynamics/Agents.jl/blob/master/CHANGELOG.md) for a full list of changes.
+    Welcome to Agents.jl 3.0! In this major release we have full support for agent based models on [`ContinuousSpace`](@ref)! We also simplified some interfaces, see the [CHANGELOG](https://github.com/JuliaDynamics/Agents.jl/blob/master/CHANGELOG.md) for a full list of changes.
 
     There are also two new examples about the spread of COVID-19 using ABMs!
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -7,9 +7,7 @@ Agents.jl is part of [JuliaDynamics](https://juliadynamics.github.io/JuliaDynami
 To get started, please read the [Tutorial](@ref) page.
 
 !!! info "Latest news"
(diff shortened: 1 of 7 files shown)


src/core/space.jl
```diff
@@ -374,27 +374,52 @@ Return an agent given its ID.
...
+
+Return the ids of the agents neighboring the given `position` (which must match type
+with the spatial structure of the `model`). If `r` is given, it is the radius to search
+for agents.
+
+For `DiscreteSpace`s `r` must be integer and defines higher degree neighbors.
+For example, for `r=2` include first and second degree neighbors,
+that is, neighbors and neighbors of neighbors.
+
+For `ContinuousSpace`, `r` is real number and finds all neighbors within distance `r`
```
kavir1698: I think we should add another column to the database: diameter. When searching for neighbors, we should check whether diameter of neighbors is within r. Currently, it checks the center coordinates.

docs/src/index.md
```diff
@@ -7,7 +7,9 @@ Agents.jl is part of [JuliaDynamics](https://juliadynamics.github.io/JuliaDynami
 To get started, please read the [Tutorial](@ref) page.
 
 !!! info "Latest news"
-    We now have an example about the spread of COVID-19, and a way to plot any agent based model using [`plotabm`](@ref)!
+    Welcome to Agents.jl 3.0! In this major release we have full support for agent based models on [`ContinuousSpace`](@ref)! We also simplified some interfaces, see the [CHANGELOG](https://github.com/JuliaDynamics/Agents.jl/blob/master/CHANGELOG.md) for a full list of changes.
+
+    There are also two new examples about the spread of COVID-19 using ABMs!
```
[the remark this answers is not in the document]
Datseris: you are right. This will only be visible in the master version of the docs. Buyt some people read the master anyway...!

src/core/space.jl
Datseris: this is a good point, but I'd prefer to discuss it in a separate issue / pr. It seems to be a strict improvement of the existing functionality, but its good to have the basic in first.

src/core/agent_space_interaction.jl
```diff
@@ -113,20 +113,40 @@ end
 #######################################################################################
 # Adding agents
 #######################################################################################
+# TODO: the Source code of `add_agent!` for discrete space is overly complex.
```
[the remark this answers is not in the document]
Datseris: no, this is still valid. The source code wasn't changed at all, only the docstrings were. the source still uses the complicated format where `add_agent_pos!` calls `add_agent!` instead of the other way around.

src/core/space.jl
Datseris: @kavir1698 can you please open an _issue_ for this idea so we won]'t forget it and have a place to think about it?

Issue #148, opened 2020-03-18
Rethink gridspace: don't expose network to user


Datseris:
I think that we unecessarily complicate things by exponsing the underlying network structure of `GridSpace` to user, and exporting functions such as `vertex2coord` and `coord2vertex` or allowing the position to be specified either as a node or a tuple.

Instead, we should simply ensure / assert that for grid spaces the "position" is always of type `NTuple{Int}` and this is the only type used. 

it will also simplify the source code and simplify the entire discussion.

---

while technically a breaking change, it is an extremely mild one, basically removing obsolete functions from the API.

Libbum:
Sure, I can take this on!

Issue #147, opened 2020-03-18
Agent field (+ types) assertion errors


Datseris:
At some point in the code we should add assertion checks that check whether the agent fields are appropriate. e.g. for grid space that a field `pos` exists _and_ that it is of type `Tuple`.

Datseris:
```
julia> fieldtype(Rational, :num)
Integer
julia> fieldtype(Rational{Int}, :num)
Int64
```
can be used, so that this check happens in `AgentBasedModel`.

PR #156, opened 2020-03-20
[WIP] Verbose spaces and Agent constructor guards

Libbum:
Starting the move to a more verbose `Space` layout, I've changed the three current space constructors to have the underlying struct's name, so `GraphSpace`, `GridSpace` & `ContinuousSpace`.

- `Space(model::ABM)` has been left in, since it may be a useful helper (although it doesn't seem to be used currently)
- Test cases are functional
- benchmarks and examples have not completely been updated
- most likely some more documentation to fix
- deprecations are added, although I'm having problems with those macros&mdash;they seem to want to spit a type error at you along with the warning. 

Then, there's a start on the work for #148, for which we now have guards and tests for user implemented `AbstractAgent` subtypes. 

- This is a bit more abstract than I'd like, but it needs to deal with the possibility of composite types. We therefore can't use nice helpers like `hasfield(A, :id)` or `A.mutable`, since they only work on `DataTypes`s. With that being said, the current implementation works across our requirements.

I need a bit of clarification from this point on though.
`vertex2coords` and `coords2vertex` are exposed in `core/space`, but the functions are actually without the pluralisation (`vertex2coord` and `coord2vertex`), so there's a bit of general confusion there.

> underlying network structure

Do you mean the ability to do this?
```julia
space = GridSpace((10,10))
space.graph
```
if that's the case, `node_neighbors(node, model::ABM{A, <:DiscreteSpace} [, r]) → nodes` should probably be changed to something like `agent_neighbors(agent, model::ABM{A, <:DiscreteSpace} [, r]) → agents`?

What other things do you think would make sense for this PR?

### DIFF
--- a/examples/continuous_space.jl
+++ b/examples/continuous_space.jl
@@ -15,7 +15,7 @@ end
 
 function model_initiation(;N=100, speed=0.005, space_resolution=0.001, seed=0)
   Random.seed!(seed)
-  space = Space(2; periodic = true, extend = (1, 1))
+  space = ContinuousSpace(2; periodic = true, extend = (1, 1))
   model = ABM(Agent, space);
 
   ## Add initial individuals
--- a/examples/forest_fire.jl
+++ b/examples/forest_fire.jl
@@ -39,7 +39,7 @@ end
 # We then make a setup function that initializes the model
 function model_initiation(; f, d, p, griddims, seed = 111)
     Random.seed!(seed)
-    space = Space(griddims, moore = true)
+    space = GridSpace(griddims, moore = true)
     properties = Dict(:f => f, :d => d, :p => p)
     forest = AgentBasedModel(Tree, space; properties=properties)
 
--- a/examples/schelling.jl
+++ b/examples/schelling.jl
@@ -25,7 +25,7 @@ mutable struct SchellingAgent <: AbstractAgent
 end
 
 # Notice that the position of this Agent type is a `Tuple{Int,Int}` because
-# we will use a grid `Space`.
+# we will use a `GridSpace`.
 
 # We added two more fields for this model, namely a `mood` field which will
 # store `true` for a happy agent and `false` for an unhappy one, and an `group`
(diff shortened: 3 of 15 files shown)


Datseris:
I am very sorry @Libbum , I have done the change `Space` -> OtherSpaces already in master.... I am really sorry I should have tagged you. Can you please see current master and start working there for #148 and the guards for the agent types...?

I am really sorry!

Libbum:
Ah, no worries. Bed time for me now, will rebase any useful changes tomorrow.

Datseris:
I haven't added any deprecations. I've only added an error to `Space`. deprecations are much better though!

---

Addressing your comments:

* For the agent guard I think the approach of #147 will "just work": In the end we only have to check that the type has field `pos` and that this field is of specific type, which Julia provides functions to check that.

* For `vertex/coord` whatever: these functions will _not_ be part of the public API in 3.0, and in addition we aim to _minimize_ their usage in the source code. At the moment they are used very frequently, because we allowed someone to use GridSpace with pos type `Int` (referencing the node number directly). This won't be the case anymore, and thus we can safely assume that all GridSpace models will give and take their position as Tuple. This will remove some usage of `vertex2coord`, but not all.

* `space.graph` will still be possible, but the user won't know abouit it and thus won't use it. Even now, before we fix #148 , nowhere in the public API it is stated that you can access the underlying graph. Although this gives me the idea that specifically in the docstring of `GraphSpace`, we should say that the graph is accessible via `space.graph`.

* Hm, your comment on `node_neighbors` is a valid one. `node_neighbors` still doesn't expose the network structure, as it returns tuples for GridSpace. But I understand that the name is not so super-duper fitting for it... `agent_neighbors` isn't much better, because it doesn't return agents, and it is also incorrect to say that "it returns the neigghbors of the agent" because it doesn't do that: it returns the neighboring "nodes". I think we can live with the name staying `node_neighbors`, although if you have better alternatives go ahead.

Libbum:
Closing in favour of #158 and #159. Fixes for #148 in a subsequent PR.

PR #159, opened 2020-03-23
Add agent guards to ABM constructor

### LINKED ISSUE #147: Agent field (+ types) assertion errors

Libbum:
Closes #147. Full test of requirements for `AbstractAgent` inside the constructor of `AgentBasedModel`.

### DIFF
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -109,7 +109,7 @@ function agent_validator(::Type{A}, space::S) where {A<:AbstractAgent, S<:SpaceT
     isbitstype(A) && throw(ArgumentError("Agent struct must be mutable"))
     (any(isequal(:id), fieldnames(A)) && fieldnames(A)[1] == :id) || throw(ArgumentError("First field of Agent struct must be `id`"))
     if space != nothing
-        (any(isequal(:pos), fieldnames(A)) && fieldnames(A)[2] == :pos) || throw(ArgumentError("Agent struct must have a `pos` field when using a space"))
+        (any(isequal(:pos), fieldnames(A)) && fieldnames(A)[2] == :pos) || throw(ArgumentError("Second field of Agent struct be `pos` when using a space"))
         # Check `pos` field in A has the correct type
         pos_type = fieldtype(A, :pos)
         space_type = typeof(space)
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -106,9 +106,9 @@ Checks for mutability and existance and correct types for fields depending on `S
 function agent_validator(::Type{A}, space::S) where {A<:AbstractAgent, S<:SpaceType}
     # Check A for required properties & fields
     isbitstype(A) && throw(ArgumentError("Agent struct must be mutable"))
-    any(isequal(:id), fieldnames(A)) || throw(ArgumentError("Agent struct must have an `id` field"))
+    (any(isequal(:id), fieldnames(A)) && fieldnames(A)[1] == :id) || throw(ArgumentError("Agent struct must have an `id` field"))
     if space != nothing
-        any(isequal(:pos), fieldnames(A)) || throw(ArgumentError("Agent struct must have a `pos` field when using a space"))
+        (any(isequal(:pos), fieldnames(A)) && fieldnames(A)[2] == :pos) || throw(ArgumentError("Agent struct must have a `pos` field when using a space"))
         # Check `pos` field in A has the correct type
         pos_type = fieldtype(A, :pos)
         space_type = typeof(space)
@@ -120,11 +120,11 @@ function agent_validator(::Type{A}, space::S) where {A<:AbstractAgent, S<:SpaceT
             # `vel` field in A must be checked alongside `pos`
             any(isequal(:vel), fieldnames(A)) || throw(ArgumentError("Agent struct must have a `vel` field when using ContinuousSpace"))
             vel_type = fieldtype(A, :vel)
-            if !(pos_type <: NTuple{D, Float64} where {D}) && !(vel_type <: NTuple{D, Float64} where {D})
+            if !(pos_type <: NTuple{D, <:AbstractFloat} where {D}) && !(vel_type <: NTuple{D, <:AbstractFloat} where {D})
                 throw(ArgumentError("`pos` and `vel` fields in Agent struct must be of type NTuple{Float64} when using ContinuousSpace."))
-            elseif !(pos_type <: NTuple{D, Float64} where {D})
+            elseif !(pos_type <: NTuple{D, <:AbstractFloat} where {D})
                 throw(ArgumentError("`pos` field in Agent struct must be of type NTuple{Float64} when using ContinuousSpace."))
-            elseif !(vel_type <: NTuple{D, Float64} where {D})
+            elseif !(vel_type <: NTuple{D, <:AbstractFloat} where {D})
                 throw(ArgumentError("`vel` field in Agent struct must be of type NTuple{Float64} when using ContinuousSpace."))
             end
         end
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -107,7 +107,7 @@ function agent_validator(::Type{A}, space::S) where {A<:AbstractAgent, S<:SpaceT
     # Check A for required properties & fields
     isconcretetype(A) || throw(ArgumentError("Agent struct must be a concrete type"))
     isbitstype(A) && throw(ArgumentError("Agent struct must be mutable"))
-    (any(isequal(:id), fieldnames(A)) && fieldnames(A)[1] == :id) || throw(ArgumentError("Agent struct must have an `id` field"))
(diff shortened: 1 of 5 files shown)


Datseris:
Thanks a lot! could you please separate the checks into a function that is called inside the code of `AgentBasedModel`, to keep its code clear?

src/core/model.jl
```diff
@@ -96,6 +98,39 @@ function Base.show(io::IO, abm::ABM{A}) where {A}
...
 end
 
+"""
+    agent_validator(agent, space)
+Validate the user supplied agent (subtype of `AbstractAgent`).
+Checks for mutability and existance and correct types for fields depending on `SpaceType`.
+"""
+function agent_validator(::Type{A}, space::S) where {A<:AbstractAgent, S<:SpaceType}
+    # Check A for required properties & fields
+    isbitstype(A) && throw(ArgumentError("Agent struct must be mutable"))
```
Datseris: Although technically correct, it is better to use `isimmutable(A)` here.

src/core/model.jl
```diff
@@ -96,6 +98,39 @@ function Base.show(io::IO, abm::ABM{A}) where {A}
...
+    any(isequal(:id), fieldnames(A)) || throw(ArgumentError("Agent struct must have an `id` field"))
```
Datseris: I am not sure whether this is entirely true (I don't remember the entire codebase). Let's be safe here and also assert that `:id` not only exists, but it is the first field. (i.e. `fieldnames(A)[1] == :id`.

src/core/model.jl
```diff
@@ -96,6 +98,39 @@ function Base.show(io::IO, abm::ABM{A}) where {A}
...
+    if space != nothing
+        any(isequal(:pos), fieldnames(A)) || throw(ArgumentError("Agent struct must have a `pos` field when using a space"))
```
Datseris: Similar with above, please ensure that `:pos` is not only an existing field, but also the *second* field.

src/core/model.jl
```diff
@@ -96,6 +98,39 @@ function Base.show(io::IO, abm::ABM{A}) where {A}
...
+        space_type = typeof(space)
+        if space_type <: GraphSpace && !(pos_type <: Integer)
+            throw(ArgumentError("`pos` field in Agent struct must be of type Int when using GraphSpace."))
+        elseif space_type <: GridSpace && !(pos_type <: NTuple{D, Integer} where {D})
+            throw(ArgumentError("`pos` field in Agent struct must be of type NTuple{Int} when using GridSpace."))
+        elseif space_type <: ContinuousSpace
+            # `vel` field in A must be checked alongside `pos`
+            any(isequal(:vel), fieldnames(A)) || throw(ArgumentError("Agent struct must have a `vel` field when using ContinuousSpace"))
+            vel_type = fieldtype(A, :vel)
+            if !(pos_type <: NTuple{D, Float64} where {D}) && !(vel_type <: NTuple{D, Float64} where {D})
```
Datseris: `NTuple{D, Float64}` -> `NTuple{D, <:AbstractFloat}` . Nothing stops the user from using other float types.

src/core/model.jl
Libbum: I've done it this way since `isimmutable` only takes `::DataType`. When users provide composite types (for example `Cell` in CA1D.jl), isimmutable ends up testing against `UnionAll` and returns true on a mutable struct.

Is there a nicer way to test this without having to make `Cell` a concrete type like `Cell{Integer, String}`?

Datseris: Providing a Union type is incorrect and should not be accepted because it reduces performane by orders of magnitude.. I am fine with throwing an extra error on union types, saying "please give concrete type" or something.

The code in CA1D.jl should be changed as well. I didn't realize it before.

Datseris: This is something important and not discussed so far in tutorial.md! Thanks for pointing it out! Althought by throwing the error here we won't have to discuss it explicitly

Libbum: OK, great. I'll add a concrete type check and swap this out.

src/core/model.jl
Libbum: I've added this as an additional check, since we'll get a `BoundsError` if we try this directly with `GridSpace` and `Agent0` for example.

Datseris:
@Libbum this discussion made me think: we should add two constructors for `AgentBasedModel`. One that takes an Agent _instance_ and thus dispatches on `a::AbstractAgent` and simply propagates `typeof(a)` , and one that takes in directly the data type, `a::Type{A}` and does what the current constructor does anyway.

src/core/model.jl
Datseris: @kavir1698 please have a look at this discussion. I often see you using the same code, but its type unstable. If you define `mutable struct Agent{D, F<:AbstractFloat}` and then you give `Agent` to `ABM`, this will create a dictioanry with value type `Agent`. But `Agent` is not a concrete type, this is type instability, and therefore bad for performance.

Libbum:
(sorry, got busy)

A second method with `a::AbstractAgent` is certainly something I can get running easily. We can use `isimmutable` there, but still hitting walls with using it in the context of our Type. From the docs: [if you give it a type, it will tell you that a value of DataType is mutable](https://docs.julialang.org/en/v1/base/base/#Base.isimmutable).

The question I'd raise there though, is what is the utility of the two separate constructors? Does it allow users two specific use cases? Personally, I'd see it as cumbersome to make essentially a 'dummy' agent first, only to pass it in to `ABM` then forget about it.

Datseris:
Yeah, I thought about this more. We shouldn't add a second constructor, following the principle "there should be 1 way to make the thing, the simplest way".

Libbum:
Agreed.

I think therefore there's nothing currently outstanding on this one, so long as @kavir1698 has no fundamental issues with us confining implementations of `AbstractAgent` to concrete types.

The `isimmutable` vs `isbitstype` issue isn't really resolvable as far as I can see if we stick to the raw type. That being said, `isbitstype` does function correctly for our purposes.

src/core/model.jl
```diff
@@ -96,6 +98,40 @@ function Base.show(io::IO, abm::ABM{A}) where {A}
...
+"""
+    agent_validator(agent, space)
+Validate the user supplied agent (subtype of `AbstractAgent`).
+Checks for mutability and existance and correct types for fields depending on `SpaceType`.
+"""
+function agent_validator(::Type{A}, space::S) where {A<:AbstractAgent, S<:SpaceType}
+    # Check A for required properties & fields
+    isconcretetype(A) || throw(ArgumentError("Agent struct must be a concrete type"))
+    isbitstype(A) && throw(ArgumentError("Agent struct must be mutable"))
+    (any(isequal(:id), fieldnames(A)) && fieldnames(A)[1] == :id) || throw(ArgumentError("Agent struct must have an `id` field"))
```
Datseris:
```suggestion
    (any(isequal(:id), fieldnames(A)) && fieldnames(A)[1] == :id) || throw(ArgumentError("First field of Agent struct must be `id`"))
```

Libbum:
Should we change the `pos` error to "Second field of ..." as well?

Datseris:
Where in the source code is the problem...? I didn't understand...

Libbum:
I think d4eeca1 will fix it. Creating types inside of `@testset` [isn't a good idea](https://github.com/JuliaLang/julia/issues/23794).

src/core/model.jl
kavir1698: @Datseris I thought this is the recommended way to define types in [the Julia docs](https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Avoid-fields-with-abstract-type-1).

Libbum: Interesting discussion. Perhaps we should run some benchmarks on both of these implementations.

Datseris:
I was sloppy to merge this: the examples still use the old syntax. I am updating some of the now, but we should check all of them.

Libbum:
Ah, sorry. Yeah&mdash;I ran the tests but not the examples.

src/core/model.jl
kavir1698: @Datseris could you clarify a bit here? the Julia manual clearly states contrary to what you suggest.

Libbum: Let's continue in #171, I think I know what happened here.

Datseris: Guys, no, there is no reason to run any benchmarks, this is basic Julia knowledge.

Yes, `struct Agent{D, F<:AbstractFloat}` is the correct way to define a parameteric type, but `Agent` is not a concrete type. `Agent{Int, Float64}` is.

Therefore `Dict{Int, Agent}` is NOT a concrete type either. `Dict{Int, Agent{Int, FLoat64}}` is. This is not an opinion, this is the truth of the type system.

PR #161, opened 2020-03-23
Update the continuous space example

kavir1698:


Datseris:
@kavir1698 this example is without the SIR model right? I remember that at some point you had made a continuous version of this SIR, and you have even produced an animation of the agents moving around and becoming red and stuf....

Do you still have this around ? Do you think it is sensible to merge this with the current example? I think it would be great to show that you can do these things with Agents.jl. It is a quite hot topic now and these animations about COVID and social distancing attract a lot of attention.

kavir1698:
Yes, I am preparing that example too. This one is just basic movements. In the animation, I see balls moving over one another without changing directions. Do you spot any mistakes?

Datseris:
I checked this out locally and spent quite some time on it.... I really cannot see what is the problem... P.s.: use random color for the agents, e.g. do 
```
colors = rand(200)
```
and give argument ` marker_z=colors,` to `scatter`.

Datseris:
@kavir1698 I've been working on it the last hour and I've encountered 2 problems: (a) sometimes some agents have the same position (maybe due to randomness?) and more importantly:

(b): `space_neighbors` sometimes returns a list of IDs that contains the agent's own ID.... `space_neighbors` needs to be tested more I think. I wrote a simple loop that called `space_neighbors` 
```juia

function collide!(agent, model)
  agent.moved && return
  r = space_neighbors(agent.pos, model, agent.diameter)

  length(r) == 0 && return
  # change direction
  for contactid in 1:length(r)
    contactid == agent.id && error("same ide???")
  end
end
```
and the error actually triggered. 

Can you please have a look at space neighbors and do a separate PR that fixes this, and maybe even tests more `space_neigghbors` ?

Datseris:
@kavir1698  check the `step!` at the end of the animation loop. You don't use the model step!!! All agents have `moved==false` all the simulation!!! Does it work for you if you correct that part?

kavir1698:
> check the step! at the end of the animation loop. You don't use the model step!!!

Yeah!!!! What an overlook. Great!

Datseris:
But please if possible still have a look on the database side of `space_agents` nevertheless, because the error I encountered is unrelated with the absence of `model_step!`.

kavir1698:
> space_neighbors sometimes returns a list of IDs that contains the agent's own ID

Yes, because we removed the condition to exclude agent's own ID from the list. Before, `searchq` had `AND id != agent.id`, but it is removed now. Below is a possible fix after reverting `searchq`:

```julia
"""
    space_neighbors(pos::Tuple, model::ABM, r::Real)
Return IDs of all agents within radius `r` from a particular position `pos` for any `SpaceType`.
"""
function space_neighbors(pos::Tuple, model, r::Real)
  left = pos .- r
  right = pos .+ r
  res = interlace(left, right)
  collect_ids(DBInterface.execute(model.space.searchq, (res...,-1)))
end

"""
    space_neighbors(agent::AbstractAgent, model::ABM, r::Real)
Return neighbours of a particular agent, within radius `r` for any `SpaceType`.
"""
function space_neighbors(agent::AbstractAgent, model, r::Real)
  left = agent.pos .- r
  right = agent.pos .+ r
  res = interlace(left, right)
  collect_ids(DBInterface.execute(model.space.searchq, (res...,agent.id)))
end
```

If you think this is consistent with the API, i'll create a PR.

Datseris:
Yes I Think this is fine!

PR #162, opened 2020-03-23
social distancing example

kavir1698:


Datseris:
I think we should merge the continuous space example and the social distancing example. They are too much similar, and so far the examples in the docs do a good job of being diverse and each example showcasing a unique scenario so to speak.

Datseris:
We can have one nice example that is "introduction" to continuous space. Starts with balls just jumping off each other, then adds virus transmission, then adds social distancing (i.e. makes some balls fixed).

kavir1698:
Good idea. I will merge the two examples.

Datseris:
Note to self: Once this is updated to use #170 , I want to add a note for people to understand that Agents.jl is _not_ a billiards simulator, and all this ball scattering is highly inaccurate, and should use DynamicalBilliards.jl or HardSphereDynamics.jl for more rigorous stuff.

Datseris:
HELL YEAH!!!

socialdist5

Datseris:
https://juliadynamics.github.io/Agents.jl/previews/PR162

Datseris:
@Libbum have a look at `plotabm` function. This is the way to plot abms in continous space in my opinion, (from the API perspective). It seems natural to use this function to plot this nice model you have shown us with the houses etc. 

(`plotabm` is in AgentsPLots.jl)

Libbum:
Indeed. Frame by frame this works great. 

A discussion for another thread we'll need to have sometime in the near future is how to animate and save data simultaneously.

kavir1698:
Great job, @Datseris !

PR #164, opened 2020-03-24
Update space_neighbors

kavir1698:


src/core/continuous_space.jl
```diff
@@ -215,13 +215,24 @@ end
 # filter!(...) where the filtering function checks distances w.r.t. `r`.
 """
     space_neighbors(pos::Tuple, model::ABM, r::Real)
-Return neighbours of a particular position `pos`, within radius `r` for any `SpaceType`.
+Return IDs of all agents within radius `r` from a particular position `pos` for any `SpaceType`.
```
Datseris: `SpaceType` can be missleading: this is not a type defined in Agents.jl. You should replace it with just the english word "space", without back ticks. 

In general if something is in backticks, it is implied that it exists in code, either as given argument, or as exported name by the module.

Datseris:
There is another thing:
```
collect_ids(DBInterface.execute(model.space.searchq, (res...,-1)))
```
I don't think anywhere in Agents.jl we pose the restriction that agent IDs need to be positive. In fact, I very much imagine a scenario where a user would want to distinguish between agents with positive or negative ID. 

I think the best way to resovle this is to simply make two query strings. One with the `id != id` clause, and one without.

src/core/continuous_space.jl
kavir1698: Ok. `SpaceType` was there before I edit this function.

kavir1698:
> I think the best way to resovle this is to simply make two query strings. One with the id != id clause, and one without.

Yes, it's cleaner. I was hesitant to add one more query to the space.

Issue #163, opened 2020-03-23
iterate till convergence


IgorDouven:
I recently discovered this package and am quite excited about it. I'm particularly interested in the implementation of the Hegselmann-Krause model. In the example, the number of iterations (how often the agents update their opinions) is set by hand. However, often in this model one would like the updating to go on until convergence is reached (for all agents, their new opinions are equal to -- or within a very close range of -- their old opinions). I can't figure out how to do that with this package. Is it possible at all? If not, it would be a nice feature to have.

Datseris:
Hi, thanks a lot for your kind words.

The `step!` function can take as an input an arbitrary function on when to terminate the evolution: https://juliadynamics.github.io/Agents.jl/dev/tutorial/#Agents.step! .

If this doesn't work for you please re-open this, but I think it will be fine.

Datseris:
By the way, if you feel you can _enhance_ the existing example file so that indeed the model stops running automatically once convergence is reached, the please consider doing a Pull Request to contribute this here!

Datseris:
*(or just paste here the solution code and I'll take care of enhancing the docs)*

Datseris:
Here is a sketch of the function you have to give as `n`:
```julia
function terminate(model)
    agents = values(model.agents)
    if any(!isapprox(a.old_opinion, a.new_opinion) for a in agents)
        return false
    else 
         return true
    end
end

IgorDouven:
Many thanks for your response, George -- much appreciated! My understanding
of the package is not at a level where I'm able to figure out how to
implement this. For instance, just adding the `terminate` function to
`model_step!` doesn't work. More generally, it is not clear to me how this
is going to replace `iterations` in `model_run`. Where I now have `when =
0:iterations`, I'd obviously like to say that I'd want to record the output
from the zeroth step still the point of convergence. But how would I define
that point using the `terminate` function?

Op di 24 mrt. 2020 om 02:41 schreef George Datseris <
notifications@github.com>:

Datseris:
Alright, I will prepare an example sometime soon!

IgorDouven:
This looks fantastic! Many thanks for the work you put into this. I'm going
to have a closer look tomorrow.

Op ma 6 apr. 2020 om 22:29 schreef George Datseris <notifications@github.com

PR #167, opened 2020-03-24
Better functional termination of model run

Datseris:
This PR addresses #163 , by improving the API of `step!`.

It allows `when` to be just `true` (i.e. collect at every step). 

As I was coding this, I realized it is extremely important to have a fail safe mechanism. Some way to terminate evolution even if `n` doesn't give true.

This is done by the keyword `Nmax`.

But now @kavir1698 you really have to help me here. The source code inside `data_collector.jl`is a nightmare to deal with:

* No keyword propagation technique was used whatsoever, making the code base unmaintainable (I just wanted to add a new keyword, and now all hell broke loose... :D ) . To fix this I started using keyword propagation in step!, but we need much more.
* arguments are in entirely arbitrary order, completely incinsistent with `step!`, which has `n` before `properties`
* the `_step!` functions have crazy syntax. All keyword arguments are given as arguments, and then `df` is given.... Time to put all keyword arguments of `step!` to also be keyword arguments of `_step!`.

This is all so confusing. Can you please re-write this part of the code?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## Breaking changes
 * Deprecated `Space` in favor of the individual spaces: `Nothing, GridSpace, GraphSpace, ContinuousSpace`.
-* The `n::Function` argument of `step!` now takes two arguments.
 * Reworked the public API of `GridSpace` to be simpler: position must be `NTuple{Int}`. As a result `vertex2coord` and stuff no longer exported, since they are obsolete.
 * It is now mandatory to provide a concrete type when initializing an `ABM` (for type stability and performance)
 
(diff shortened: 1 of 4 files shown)


kavir1698:
> Can you please re-write this part of the code?

It could take a bit for me to get to focus on this. But yeah.

Libbum:
@kavir1698 I'd be up for assisting on this if need be.

Datseris:
Sure, go ahead, but I think it is a hard task. I don't do it because I don't have too much familiarity with the data collector, and don't want to get a head-ache getting all methods correct, but if you want to put the effort, sure, go ahead!

kavir1698:
Hi @Libbum , yes that would great. It might be a few days before I can start working on this.

Datseris:
Closing in favor of #191

PR #168, opened 2020-03-24
Remove coord/vertex from API

Libbum:
Here's some work connected to #148. There was a lot of different discussions there, and some have already been fixed via the `Space` depreciation and #159.

This PR drops the public facing `coord2vertex` and `vertex2coord` function and replaces their use in example cases where needed. 

One place where this hasn't been done explicitly is `src/CA2D.jl`, which is going to be dropped in #128 anyhow.

Doesn't seem like this simplifies all that much, so would be good to get some feedback on what additional things could be done here.

### DIFF
--- a/src/core/agent_space_interaction.jl
+++ b/src/core/agent_space_interaction.jl
@@ -65,21 +65,26 @@ in the model and remove it from the old position
 """
 function move_agent!(agent::AbstractAgent, pos::Tuple, model::ABM)
   nodenumber = coord2vertex(pos, model)
-  _move_agent!(agent, nodenumber, model)
+  move_agent!(agent, nodenumber, model)
 end
 
-function _move_agent!(agent::AbstractAgent, pos::Integer, model::ABM)
+function move_agent!(agent::AbstractAgent, pos::Integer, model::ABM)
   # remove agent from old position
-  oldnode = coord2vertex(agent.pos, model)
-  splice!(model.space.agent_positions[oldnode], findfirst(a->a==agent.id, model.space.agent_positions[oldnode]))
-  agent.pos = vertex2coord(pos, model)  # update agent position
+  if typeof(agent.pos) <: Tuple
+    oldnode = coord2vertex(agent.pos, model)
+    splice!(model.space.agent_positions[oldnode], findfirst(a->a==agent.id, model.space.agent_positions[oldnode]))
+    agent.pos = vertex2coord(pos, model)  # update agent position
+  else
+    splice!(model.space.agent_positions[agent.pos], findfirst(a->a==agent.id, model.space.agent_positions[agent.pos]))
+    agent.pos = pos
+  end
   push!(model.space.agent_positions[pos], agent.id)
   return agent
 end
 
 function move_agent!(agent::AbstractAgent, model::ABM)
   nodenumber = rand(1:nv(model.space))
-  _move_agent!(agent, nodenumber, model)
+  move_agent!(agent, nodenumber, model)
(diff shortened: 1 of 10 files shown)


Libbum:
Additional:

Seems like `src/simulations/sample.jl` has a function thats never used: `clean_space!`

src/core/agent_space_interaction.jl
```diff
@@ -13,11 +13,7 @@ move_agent_single!, kill_agent!, coord2vertex, vertex2coord, genocide!
 Remove an agent from model, and from the space if the model has a space.
 """
 function kill_agent!(agent::AbstractAgent, model::ABM{A, S}) where {A, S<:AbstractSpace}
-  if typeof(agent.pos) <: Tuple
-    agentnode = coord2vertex(agent.pos, model)
-  else
-    agentnode = agent.pos
-  end
+  agentnode = coord2vertex(agent.pos, model)
```
kavir1698: Agents on `GraphSpace`  still have integers for their positions, don't they? If so, this change will be wrong.

src/core/agent_space_interaction.jl
```diff
@@ -69,26 +65,21 @@ in the model and remove it from the old position
...
   # remove agent from old position
-  if typeof(agent.pos) <: Tuple
-    oldnode = coord2vertex(agent.pos, model)
-    splice!(model.space.agent_positions[oldnode], findfirst(a->a==agent.id, model.space.agent_positions[oldnode]))
-    agent.pos = vertex2coord(pos, model)  # update agent position
-  else
-    splice!(model.space.agent_positions[agent.pos], findfirst(a->a==agent.id, model.space.agent_positions[agent.pos]))
-    agent.pos = pos
-  end
+  oldnode = coord2vertex(agent.pos, model)
```
kavir1698: Same here.

Datseris:
Yes you can remove that. It is a remnant of `genocide!`

test/api_tests.jl
```diff
@@ -38,17 +38,9 @@ model1 = ABM(Agent1, GridSpace((3,3)))
...
-@test length(pos1) == 1
-@test pos1[1] == 1
 
 move_agent!(agent, (2,2), model1)
-
 @test agent.pos == (2,2)
-pos1 = model1.space.agent_positions[coord2vertex((1,1), model1)]
-@test length(pos1) == 0
-pos2 = model1.space.agent_positions[coord2vertex((2,2), model1)]
-@test pos2[1] == 1
```
kavir1698: Shouldn't the tests be updated instead of removed?

Libbum: Yep, I killed that too early. Will post a fix soon.

src/core/agent_space_interaction.jl
Libbum: This *may* be the case and if so, this was some overzealous cutting. I'm writing a test suite for these functions & will update the result whatever the outcome.

Libbum: This wasn't a problem here, but was indeed the case for `move_agent!`. Fixed with 
0cbc01f.

Libbum:
- `clean_space!` is in fact used in `sample!`, so it should remain.
- The latest commit adds a number of checks for `kill_agent!` and `move_agent!` to make sure they function as expected.

Datseris:
I am really sorry, did I sleep on the wheel on this? Is it just good to go?

Libbum:
Was just waiting on a sign off from Ali. Otherwise it should be good, yeah,

PR #172, opened 2020-03-24
Pair-wise interactions in continuous space

### LINKED ISSUE #170: Iterator over agent pairs
See https://github.com/JuliaDynamics/Agents.jl/pull/169 for motivation.

Often wants to iterate of pair of nearest neighbors of agents, and do some kind of interaction. Almost always, this interaction happens once between `agent1, agent2` and need not be repeated for `agent2, agent1`. I.e. almost all interactions are **commutative**.

We should create a function / iterator that iterates over pairs of nearest neighbors agents (i.e. the iterator returns two agents). This function can be then used in a `model_step!` to make everything massively more performant.

@kavir1698 with the database approach, is there an easy way to find nearest neighbors?

Datseris:
Closes #170 

This further advances continuous space interactions.

Once this is merged, PR #162 should be updated to use this new system. See the file `test/collision_tests.jl`. There there is also cool animation (commented out)

There is one little problem though. if you look at the test file you will see it. `@test x == model.properties[:c] # test that at least 10 agents have collided
` fails, even though at least conceptually it should be correct.

any ideas why?


Datseris:
animation:

movement

Datseris:
movement


And with this, we are finally able to replicate all those cool animations can see online.

PR #176, opened 2020-03-26
Type Stability & better constructors

### LINKED ISSUE #174: Ensure all examples are type stable, see #159
Ensure all examples are type stable, see #159

Libbum:
Fixes to the `ABM` constructor following from #159 and subsequent discussions.
- New `AgentBasedModel(agent::A ...) where {A<:AbstractAgent}` constructor
- Updated documentation
- More verbose and helpful errors when `agent` is not instantiated according to our requirements
- Examples and benchmarks are now up to date and functioning in accordance with these changes. Closes #174.

### DIFF
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -66,8 +66,8 @@ agent IDs, while the values are the agents themselves.
 It is recommended however to use [`id2agent`](@ref) to get an agent.
 
 `space` can be omitted, in which it will equal to `nothing`.
-This means that all agents are virtualy in one node and have no spatial structure.
-If space is omitted, some functions that fascilitate agent-space interactions will not work.
+This means that all agents are virtually in one node and have no spatial structure.
+If space is omitted, some functions that facilitate agent-space interactions will not work.
 
 Optionally provide a `scheduler` that creates the order with which agents
 are activated in the model, and `properties`
@@ -76,6 +76,33 @@ This is accessed as `model.properties` for later use.
 """
 function AgentBasedModel(
         ::Type{A}, space::S = nothing;
+        scheduler::F = fastest, properties::P = nothing
+        ) where {A<:AbstractAgent, S<:SpaceType, F, P}
+    agent_validator(A, space)
+
+    agents = Dict{Int, A}()
+    return ABM{A, S, F, P}(agents, space, scheduler, properties)
+end
+
+"""
+    AgentBasedModel(agent [, space]; scheduler, properties)
+Similar to the constructor above, but created via an instance of your agent.
+
+```julia
+mutable struct YourAgent <: AbstractAgent
+    id::Int
+    pos::Tuple{Int, Int}
+end
+agent = YourAgent(1, (1,1))
+model = AgentBasedModel(agent, GridSpace((10,10))
+```
+
+Note that the specific values you use when creating `agent` are not important and wont be used here.
+This process just helps Agents.jl do some performance checks for you, if you're not that familiar with
+Julia's type system.
+"""
+function AgentBasedModel(
+        agent::A, space::S = nothing;
         scheduler::F = fastest, properties::P = nothing
         ) where {A<:AbstractAgent, S<:SpaceType, F, P}
     agent_validator(A, space)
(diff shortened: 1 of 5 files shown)


Libbum:
Hold on reviewing this for a second. Something went wrong here. `core/model.jl` is not what it should be.

Libbum:
OK 'fixed'.
Unsure if 6582893 is necessary or not though&mdash;let me know if that's needed there.

Datseris:
All tests fail in Julia 1.0 https://travis-ci.com/github/JuliaDynamics/Agents.jl/jobs/304872032 

(I personally I am not too much resticted on staying on Julia 1.0, if it is impossible to test this in 1.0)

Libbum:
Looks like the same typeset scope issues I had before. Should be fixable. 

-----Original Message-----

Libbum:
So the issue is `fieldtypes`, which requires v1.1.
I've added this in due to this case:

```julia
julia> mutable struct AmbiguousAgent <: AbstractAgent    
           id::Int    
           pos::NTuple{2, Int}    
           # The following types are abstract    
           weight::Real    
           info::AbstractString    
       end  

julia> isconcretetype(AmbiguousAgent)
true
```

4115d35 adds a compatibility shim.

Libbum:
Bah. It's fixed 1.0 and broken 1.4.0. Hang on..

src/core/model.jl
```diff
@@ -84,6 +84,33 @@ function AgentBasedModel(
     return ABM{A, S, F, P}(agents, space, scheduler, properties)
 end
 
+"""
+    AgentBasedModel(agent [, space]; scheduler, properties)
```
Datseris:
```suggestion
    AgentBasedModel(agent_instance [, space]; scheduler, properties)
```

src/core/model.jl
```diff
@@ -84,6 +84,33 @@ function AgentBasedModel(
...
+This process just helps Agents.jl do some performance checks for you, if you're not that familiar with
+Julia's type system.
+"""
+function AgentBasedModel(
+        agent::A, space::S = nothing;
+        scheduler::F = fastest, properties::P = nothing
+        ) where {A<:AbstractAgent, S<:SpaceType, F, P}
+    agent_validator(typeof(agent), space)
+
+    agents = Dict{Int, typeof(agent)}()
```
Datseris:
```suggestion
    agents = Dict{Int, A}()
```

Here `A` and `typeof(agent)` are textually equivalent.

src/core/model.jl
```diff
@@ -101,32 +128,43 @@ end
...
 Validate the user supplied agent (subtype of `AbstractAgent`).
-Checks for mutability and existance and correct types for fields depending on `SpaceType`.
+Checks for mutability and existence and correct types for fields depending on `SpaceType`.
 """
 function agent_validator(::Type{A}, space::S) where {A<:AbstractAgent, S<:SpaceType}
     # Check A for required properties & fields
-    isconcretetype(A) || throw(ArgumentError("Agent struct must be a concrete type"))
-    isbitstype(A) && throw(ArgumentError("Agent struct must be mutable"))
-    (any(isequal(:id), fieldnames(A)) && fieldnames(A)[1] == :id) || throw(ArgumentError("First field of Agent struct must be `id`"))
+    if VERSION < v"1.1"
```
Datseris: Add `@static` before the if.

src/core/model.jl
```diff
@@ -101,32 +128,43 @@ end
...
+        for f in 1:fieldcount(A)
+            if !isconcretetype(fieldtype(A, f))
+                isconcrete = false
+                break
+            end
+        end
+    else
+        iscontrete = all(isconcretetype, fieldtypes(A))
+    end
+    isconcrete || throw(ArgumentError("Agent struct must be a concrete type. If your agent is parametrically typed, this probably happened because you gave `Agent` instead of `Agent{T}` to this function. You can also create and instance of your agent and pass it to this function."))
```
Datseris: Here instead of `Agent{T}` write `Agent{Float64}`. `Agent{T}` is ambiguous depending on `T`

Datseris: Typo: `and instance` -> `an instance`

src/core/model.jl
```diff
@@ -101,32 +128,43 @@ end
...
+    isbitstype(A) && throw(ArgumentError("Agent struct must be mutable. Try adding the `mutable` keyword infront of `struct` in your agent definition."))
+    (any(isequal(:id), fieldnames(A)) && fieldnames(A)[1] == :id) || throw(ArgumentError("First field of Agent struct must be `id` (it should be of type `Integer`)."))
     if space != nothing
-        (any(isequal(:pos), fieldnames(A)) && fieldnames(A)[2] == :pos) || throw(ArgumentError("Second field of Agent struct be `pos` when using a space"))
+        (any(isequal(:pos), fieldnames(A)) && fieldnames(A)[2] == :pos) || throw(ArgumentError("Second field of Agent struct be `pos` when using a space (it should be of type `NTuple{Integer}`)."))
```
Datseris: This is incorrect. If `GraphSpace` is used, type `Int` is used for the position. You don't need the added sentence `(it shoud...)` here, since you already take care of it below.

src/core/model.jl
```diff
@@ -101,32 +128,43 @@ end
...
         # Check `pos` field in A has the correct type
         pos_type = fieldtype(A, :pos)
         space_type = typeof(space)
         if space_type <: GraphSpace && !(pos_type <: Integer)
-            throw(ArgumentError("`pos` field in Agent struct must be of type Int when using GraphSpace."))
+            throw(ArgumentError("`pos` field in Agent struct must be of type `Integer` when using GraphSpace."))
```
Datseris: `Integer` is abstract type, and this might invoke bad behavior. Replace all instances of `Integer` with `Int` to ensure that even newbies get the error message.

src/core/model.jl
```diff
@@ -101,32 +128,43 @@ end
...
         elseif space_type <: GridSpace && !(pos_type <: NTuple{D, Integer} where {D})
-            throw(ArgumentError("`pos` field in Agent struct must be of type NTuple{Int} when using GridSpace."))
+            throw(ArgumentError("`pos` field in Agent struct must be of type `NTuple{Integer}` when using GridSpace."))
         elseif space_type <: ContinuousSpace
             # `vel` field in A must be checked alongside `pos`
-            any(isequal(:vel), fieldnames(A)) || throw(ArgumentError("Agent struct must have a `vel` field when using ContinuousSpace"))
+            any(isequal(:vel), fieldnames(A)) || throw(ArgumentError("Agent struct must have a `vel` field when using ContinuousSpace (it should be of type `NTuple{AbstractFloat}`)."))
```
Datseris: `AbstractFloat` is also abstract. Here you can type `NTuple{ <: AbstractFloat}` instead, to really clarify that it should use a _subtype_ of `AbstractFloat`, and not the type itself.

src/core/model.jl
```diff
@@ -101,32 +128,43 @@ end
...
         pos_type = fieldtype(A, :pos)
         space_type = typeof(space)
         if space_type <: GraphSpace && !(pos_type <: Integer)
-            throw(ArgumentError("`pos` field in Agent struct must be of type Int when using GraphSpace."))
+            throw(ArgumentError("`pos` field in Agent struct must be of type `Integer` when using GraphSpace."))
         elseif space_type <: GridSpace && !(pos_type <: NTuple{D, Integer} where {D})
-            throw(ArgumentError("`pos` field in Agent struct must be of type NTuple{Int} when using GridSpace."))
+            throw(ArgumentError("`pos` field in Agent struct must be of type `NTuple{Integer}` when using GridSpace."))
         elseif space_type <: ContinuousSpace
             # `vel` field in A must be checked alongside `pos`
```
Datseris: Notice that `vel` is _not_ a mandatory field for continuous space. Because you are not absolutely required to use `move_agent!` if you use continuous space. You might want the benefits of continuous space for other reasons. Therefore this check must be removed.

test/api_tests.jl
```diff
@@ -13,24 +13,49 @@ mutable struct DiscreteVelocity <: AbstractAgent
...
+    id::Int
+    pos::NTuple{2, Int}
+    # The following types are abstract
+    weight::Real
+    info::AbstractString
+end
 
 @testset "Model construction" begin
     # Cannot use ImmutableAgent since it cannot be edited
     @test_throws ArgumentError ABM(ImmutableAgent)
```
Datseris: This is probably too much: we don't want to babysit the user so much as to check every single field of the agent. We should only check the following:

* The given agent type to ABM is mutable and concrete
* The given agent type's first field is `id`
* The second field `pos`'s type matches the space.

That is it, we shouldn't check for anything beyond that.

Libbum: Oh, OK. Well, if that's the case I'll remove this compatibility shim for 1.0 too.

src/core/model.jl
Libbum: Good to know. Thought that would be the case.

src/core/model.jl
Libbum: Yep, an oversight.

test/api_tests.jl
Libbum: Have changed to a parametric type test. This also gives an example of how users who are not familiar with the type system can use inference to get started.

Libbum:
Thanks for the suggestions. All are updated. Have tested on 1.0.5 and 1.4.0 locally. I assume the CI will pass now.

test/api_tests.jl
```diff
@@ -13,12 +13,11 @@ mutable struct DiscreteVelocity <: AbstractAgent
     vel::NTuple{2, Int}
     diameter::Float64
 end
-mutable struct AmbiguousAgent <: AbstractAgent
-    id::Int
-    pos::NTuple{2, Int}
-    # The following types are abstract
-    weight::Real
-    info::AbstractString
+mutable struct ParametricAgent{T} <: AbstractAgent where {T <:Integer}
```
Datseris:
```suggestion
mutable struct ParametricAgent{T <: Integer} <: AbstractAgent
```
Is the correct syntax. `where` is reserved for method definitions :)

Libbum: Oh, yes&mdash;thanks. Updated.

Datseris:
Darn, again! Could you please add a sentence about this in CHANGELOG in your next PR ? (even if it is an unrelated PR).

Something like: `AgentBasedModel` now checks and errors if the given agent type is not concrete.

Libbum:
Sure, will do!

PR #181, opened 2020-03-27
Relax some errors to warnings.

### LINKED ISSUE #178: Regarding #176 : warn, not error
Regarding the PR #176 , we should warn, and not error, if any of these checks fail, e.g. if agent type is not concrete, etc. 

(besides the `id` field being first and an integer, and the `pos` field being appropriate. There we should always error)

There is a simple reason: some users may want to provide `Union{Tree, Woodcutter}` as their agent type, i.e. create a **mixed agent-based-model**. I guess @Libbum you would be one of those people, given the figure in #141 already uses 2 types of stuff.

Of course, you could do this by creating two agent-based models, but maybe you want to use the API more directly, including data collection, etc.

---

@Libbum once again I am sorry for merging stuff before thorough thinking about them.

Libbum:
Closes #178 
- Drops severity of discussed checks from errors to warnings
- Add test for `id::Int`, which I had previously missed
- Add `warn` keyword to `ABM` constructors to allow users to suppress warnings
- Tests and documentations for all
- CHANGELOG :P


Datseris:
ABM is a central point of the Tutorial, so it should have as concise docstring as possible. 

I've reduced it and remove a lot of stuff.

The woodcutter docs is good, but it is much better to demonstrate this via a new Example than to have it in the documentation string.

PR #182, opened 2020-03-27
Two convenience syntaxes for getting model propertries and agents

### LINKED ISSUE #166: Feature request: propagate model.symbol to model.properties[:symbol]
For `:symbol` not already in model fields. 

(I have done this before in DynamicalBilliards.jl, I will try it here in a PR as well)

Datseris:
This PR introduces two syntaxes that I find much beetter than what we have now:

1. `model[id]` gives the agent with `id`, i.e. it is `model.agents[id]`.
2. `model.property` gives the entry of `model.properties`. For example, `model.β` gives `model.properties[:β]`.

Motivation: I was extremely displeased about how super verbose it is to get model properties. I was also displeased that getting the agent with `id` is so complicated, given that the dictionary is central and should be more easy to do that. 

This PR resolves everything. It deprecates the function `id2agent` which is useless now.

Closes #166 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,14 @@
 * Agents can be accessed from the model directly. `model[id]` is equivalent with `model.agents[id]` and replaces `id2agent`.
 * If `model.properties` is a dictionary with key type Symbol, then the
   convenience syntax `model.prop` returns `model.properties[:prop]`.
+* Version of `add_agent!` now has keyword propagation as well (in case you make your types with `@kwdef` or Parameters.jl)
 
 ## Breaking Changes
 * Deprecated `Space` in favor of the individual spaces: `Nothing, GridSpace, GraphSpace, ContinuousSpace`.
 * Reworked the public API of `GridSpace` to be simpler: position must be `NTuple{Int}`. As a result `vertex2coord` and stuff no longer exported, since they are obsolete.
 - `AgentBasedModel` checks the construction of your agent and will return errors when it is malformed (no `id` or `pos` when required, incorrect types). Warnings when possible problems may occur (immutable agents, types which are not concrete, `vel` not of the correct type when using `ContinuousSpace`).
 - `id2agent` is deprecated in favor of `getindex(model, id) == model[id]`
-* Version of `add_agent!` now has keyword propagation as well (in case you make your types with `@kwdef` or Parameters.jl)
+
 
 # v2.1
 * Renamed the old scheduler `as_added` to `by_id`, to reflect reality.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * New convenience function `allagents`
 * New continuous space functions `nearest_neighbor` and `elastic_collision!`
 * New iterator `interacting_pairs`
-* Agents can be accessed from the model directly. `model[id]` is equivalent with `model.agents[id]` and replaces `id2agent`.
+* Agents can be accessed from the model directly. `model[id]` is equivalent with `model.agents[id]` and replaces `id2agent`. `model[id] = agent` also works.
 * If `model.properties` is a dictionary with key type Symbol, then the
   convenience syntax `model.prop` returns `model.properties[:prop]`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * New convenience function `allagents`
 * New continuous space functions `nearest_neighbor` and `elastic_collision!`
 * New iterator `interacting_pairs`
-* Agents can be accessed from the model directly. `model[id]` is equivalent with `model.agents[id]` and replaces `id2agent`. `model[id] = agent` also works.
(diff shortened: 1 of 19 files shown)


Datseris:
TODO: replace all `model.properties[:whatever]` with `model.whatever` in the docs.

Datseris:
@Libbum I altered some text, but the tests fail. I've tried to correct them. What did I do wrong...?

Datseris:
@Libbum I have done your hotfixes here already.

Libbum:
Test is failing because the warning text is not updated on line 63 in `api_tests.jl`.

Don't have time to do a proper review tonight, but the changes look great. This really makes life easier. I'll take a deeper look tomorrow.

itsdfish:
Will the conditional return type introduce additional type instability? 
```
function Base.getproperty(m::AgentBasedModel, s::Symbol)
    if s ∈ (:agents, :space, :scheduler, :properties)
        return Base.getfield(m, s)
    else
        return Base.getindex(m.properties, s)
    end
end
```


```
julia> @code_warntype model.x
Variables
  #self#::Core.Compiler.Const(getproperty, false)
  m::AgentBasedModel{Shopper,Agents.GraphSpace{SimpleDiGraph{Int64}},typeof(fastest),Dict{Symbol,Any}}
  s::Symbol

Body::Any
1 ─ %1 = (:agents, :space, :scheduler, :properties)::Core.Compiler.Const((:agents, :space, :scheduler, :properties), false)
│   %2 = (s ∈ %1)::Bool
└──      goto #3 if not %2
2 ─ %4 = Base.getfield::Core.Compiler.Const(getfield, false)
│   %5 = (%4)(m, s)::Union{typeof(fastest), Agents.GraphSpace{SimpleDiGraph{Int64}}, Dict}
└──      return %5
3 ─ %7 = Base.getindex::Core.Compiler.Const(getindex, false)
│   %8 = Base.getproperty(m, :properties)::Any
│   %9 = (%7)(%8, s)::Any
└──      return %9
```

```
@code_warntype model.properties[:x]
Variables
  #self#::Core.Compiler.Const(getindex, false)
  h::Dict{Symbol,Any}
  key::Symbol
  val::Union{}
  index::Int64

Body::Any
1 ─       Core.NewvarNode(:(val))
│         (index = Base.ht_keyindex(h, key))
│         $(Expr(:inbounds, true))
│   %4  = (index < 0)::Bool
└──       goto #3 if not %4
2 ─ %6  = Base.KeyError(key)::KeyError
│         Base.throw(%6)
└──       Core.Compiler.Const(:(return %7), false)
3 ┄ %9  = Base.getproperty(h, :vals)::Array{Any,1}
│   %10 = Base.getindex(%9, index)::Any
│   %11 = Core.typeassert(%10, $(Expr(:static_parameter, 2)))::Any
└──       return %11
4 ─       $(Expr(:inbounds, :pop))
└──       Core.Compiler.Const(:(return val), false)
```

As a point of reference, here is a case in which properties is a struct rather than a dictionary:

```
@code_warntype model.properties.stock
Variables
  #self#::Core.Compiler.Const(getproperty, false)
  x::Store
  f::Symbol

Body::Int64
1 ─ %1 = Base.getfield(x, f)::Int64
└──      return %1
```

Datseris:
Thanks for pointing this out @itsdfish ! I've opened a question about this: https://discourse.julialang.org/t/overloading-getproperty-in-a-type-stable-manner/36638

The way I have done this before, was to use only `getproperty` and `getfield`. Now I am mixing things and type instability occurs. 

If it is not possible to get rid of type instability, I will obviously remove this change (but change `model[id]` remains, and is type stable as well).

Datseris:
Thanks for pointing this out @itsdfish ! I've opened a question about this: https://discourse.julialang.org/t/overloading-getproperty-in-a-type-stable-manner/36638

The way I have done this before, was to use only `getproperty` and `getfield`. Now I am mixing things and type instability occurs. 

If it is not possible to get rid of type instability, I will obviously remove this change (but change `model[id]` remains, and is type stable as well).

itsdfish:
No problem. I'm glad that I can help a bit.

I've encountered a similar problem before. In many cases, multiple dispatch can be used as an alternative to an if statement, but that looks unlikely in this case. Parameters.jl has a macro that unpacks all fields of a struct (maybe a dictionary could work too), but it is described as "somewhat dangerous" (last paragraph https://mauro3.github.io/Parameters.jl/stable/manual/). It would be nice if there is a better way to access nested content. The best workaround I know of to limit verbosity is to use an alias at the beginning of a function , such as p = model.properties.

Datseris:
Alright, I've resolved this. The new method _is indeed type stable_. Notice that this type stability can only exist within functions not in global scope (otherwise no constant propagation can happen).

But, by definition, nothing is type stable in global scope so we can't even care about the instability in global scope.

Datseris:
> but that looks unlikely in this case

Not really, I have done this before for another package, but here it was only a hair more tricky. Anyway, it works now exactly as originally intended :)

itsdfish:
I'm glad you figured out something. The solution seems a little weird to me. I'm still trying wrap my head around it. The changes you made also have the benefit of working on structs as well as dictionaries. Nice!

PR #183, opened 2020-03-27
Testing hotfixes

Libbum:
98d081e87 introduced some test failures and recent updates concerning periodic bounds in `ContinuousSpace` broke a test written in #168.


Datseris:
I include it already in #182 , so let's save the complexity of rebasing and go there instead.

PR #189, opened 2020-03-29
Additional documentation

Libbum:
Great work on #182 - it all looks amazing! Sorry I've been AFK all weekend.

Since there's been a lot of churn and I didn't get the change to do a proper review, I've gone through the final version and think it's solid.

I went through the docs and made sure all exported functions are now represented, as well as if we have any dangling exports. The only method we don't account for in the API docs currently is `dummystep`.
Also, the `Base` extensions could be considered as documentation, although I'm guessing it's not that important to have them explicitly discussed if we use the `model.x` and `model[id]` forms throughout.

If however, it makes sense to be a little verbose in the docs, I've added some words to that effect here. An exception to that is `setindex!` in accordance with 5f5b244.

Expecting these were considered omissions, so feel free to close this if such is the case.

**Concerning `id2agent`**:
Is the plan to completely depreciate that in v4, but keep it alive in v3 with a warning? We didn't give `Space` the same treatment. Should `id2agent` just be another deprecated, breaking change?

### DIFF
--- a/src/core/space.jl
+++ b/src/core/space.jl
@@ -376,7 +376,7 @@ get_node_agents(x, model) = [model[id] for id in get_node_contents(x, model)]
 Return an agent given its ID.
 """
 function id2agent(id::Integer, model::ABM)
-  @warn "`id2agent(id, model)` is deprecated. Use `model[id]` instead!"
+    @warn "`id2agent(id, model)` is deprecated. Use `model[id]` or `getindex(model, id)` instead!"
   model[id]
 end
 
--- a/src/core/space.jl
+++ b/src/core/space.jl
@@ -1,5 +1,5 @@
 export node_neighbors, find_empty_nodes, pick_empty, has_empty_nodes, get_node_contents,
-id2agent, NodeIterator, space_neighbors, nodes, get_node_agents
+NodeIterator, space_neighbors, nodes, get_node_agents
 export nv, ne
 export GraphSpace, GridSpace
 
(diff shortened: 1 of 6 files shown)


src/simulations/step.jl
```diff
@@ -52,7 +52,18 @@ function step! end
 #######################################################################################
 # basic stepping
 #######################################################################################
+
+"""
+    dummystep(model)
+
+Ignore the model dynamics on this `step!`. Use in-place of `model_step!`.
```
Datseris:
```suggestion
Ignore the model dynamics on this `step!`. Use instead of `model_step!`.
```

src/simulations/step.jl
```diff
@@ -52,7 +52,18 @@ function step! end
...
+"""
+    dummystep(model)
+
+Ignore the model dynamics on this `step!`. Use instead of `model_step!`.
+"""
 dummystep(model) = nothing
+"""
+    dummystep(agent, model)
+
+Ignore the agent dynamics on this `step!`. Use in-place of `agent_step!`.
```
Datseris:
```suggestion
Ignore the agent dynamics on this `step!`. Use instead of `agent_step!`.
```

Datseris:
Yes `id2agetn` is another deprecated breaking change. If I didn't do this in #189 I probably forgot about it. Let's do it here.

src/core/space.jl
```diff
@@ -1,5 +1,5 @@
 export node_neighbors, find_empty_nodes, pick_empty, has_empty_nodes, get_node_contents,
-id2agent, NodeIterator, space_neighbors, nodes, get_node_agents
+NodeIterator, space_neighbors, nodes, get_node_agents
```
Datseris: It doesn't make sense to deprecate something without exporting it :D otherwise users will get "Undef" error!

Datseris: Op sorry, I take this back! `@deprecate` exports automatically!

Libbum: Yep! That's the point&mdash;keeps things clean so we can just drop one line later if we want.

Datseris:
@Libbum please from now on respect 92 characters per line of code, even for documentation strings. 

GitHub wraps lines so I missed this here, but Julia recommends not having more than 92 characters per line.

Libbum:
Will do. Are you using something like [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl) to keep things sane?

Datseris:
I just use Atom and make it display a vertical line after 92 characters. I just never cross this line :)

Libbum:
Fair enough. I'll just set my vimrc to do the same @ line 92 for Julia files.

Issue #188, opened 2020-03-29
Collect data from model environment and agent simulatenously.


itsdfish:
Currently, the data collection feature is limited to information about the agent or the model. It would be convenient to have the ability to collect information from both. In addition, it might be useful to collect variables that are computed from other variables (although this might not be easy to do in a general way).

In the meantime, I have found that it is possible to pass a dataframe and collect the information manually like so:

`step!(model, (x,y)->agent_step!(x, y, df), model_step!, 150)`

Libbum:
Yep, this is certainly a limitation of our currently implementation. (See #151 also). I'm currently looking at scenarios that better satisfy all user needs in this matter and we'll be reworking the data collection methods for v3.

Will begin discussions over at #187.

Datseris:
When data are aggregated, you can still get model variables, as well as "variables from variables" (since you can pass any arbitrary function). E.g. use `Dict(:model =>[...], :wealth => [...])`

I assume that what you really try to say @itsdfish is that you want to collect model data at every step using the `properties::Vector{Symbol}` case. In this case, what is the `id` field of the model? (The only thing that distinguishes agents in this part of the data collection is the ID. What is it for the model?)

---

> In addition, it might be useful to collect variables that are computed from other variables (although this might not be easy to do in a general way).

Please explain this in more detail.

itsdfish:
Perhaps the capabilities for these features already exist. I'll try to give more concrete examples of each use case just to clarify.

Suppose that there is a food resource in the environment that replenishes at some rate as long as at least one unit exists. In such a model, it would be useful to track the quantity of food available for each agent's interaction. The resulting data for two agents on step 1 might look like this:

```
2×3 DataFrame
│ Row │ id    │ food_units │ step  │
│     │ Int64 │ Int64      │ Int64 │
├─────┼───────┼────────────┼───────┤
│ 1   │ 1     │ 100        │ 1     │
│ 2   │ 2     │ 92         │ 1     │

```

As you noted, in some cases it is possible to compute variables from other variables in post processing. However, there are cases where that is difficult. Suppose I have a model in which agents interact at each step. How would I collect the id of the agents in each interaction and the result of that interaction? Perhaps based on some combination of variables, the interaction could be classified as good or bad. 

```
2×4 DataFrame
│ Row │ id    │ other_id │ step  │ interaction │
│     │ Int64 │ Int64    │ Int64 │ Symbol      │
[...]

Issue #187, opened 2020-03-29
Split `step!` into three functions


Datseris:
As far as I can tell, `step!` has three functionaities, all different with each other.

1. Evolve the model.
2. Evolve the model _and_ collect data _per step_
3. Evolve the model and _aggregate data_ across time (steps).

I think this is too much conceptual load put into one function. If we simply _rename_ these functionalities into three functions, I think it will make this package easier to use. 

Thoughts?

Libbum:
Totally agreed. These three aspects of the function muddy the waters and limit the users ability to efficiently collect data. It also limits our ability to both collect and plot 'live', which is something we should be working towards seeing.

I'm still in the process of getting together a set of use cases that with conflict the current method and solutions. I'll consider how they would change if we had four separate methods (evolution, data collection, data aggregation, plotting) and report back soon if there are any edge cases I identify.

Datseris:
data collection and data aggregation is the only case that can't be separated. You need the data first to aggregate them, and I am not sure whether splitting this into two pipelines (first collect, _then_ aggregate) is useful in the end. 

The way I see it the aggregating function should also do the data collection. Thankfully, plotting is already detached from anything else.

I've thought about names for the processes as well: `step!, collect!, aggregate!`

Libbum:
`step!` to me is singular, perhaps `evolve!` could be used.

An implementation problem we need to consider here relates to the collectors understanding what has happened since the last call to it.

```julia
step!(model, agent_step!, model_step!)
results = collect!(model, properties) #New dataframe
step!(model, agent_step!, model_step!, 30)
collect!(results, model, properties) #data appended to existing dataframe
```
Here, `properties` are *values on explicit steps*. This would be equivalent to
```julia
results = step!(model, agent_step!, model_step!, 31, properties; when = 1:30:31)
```
The separation means that the data collection method has no way of tracking the actual step number (the issue I solved with `model.properties[:tick]` in #151). 

This would need be solved by adding a `step` (or `tick`) to the `model` rather than having the evolution method handle this.

kavir1698:
Hi guys,
Sorry that I have not been able to contribute in the past few days. But I am following the discussions. About this PR, I am clear what will be the final shape of the API. After splitting `step!`, does the user have to write a loop to collect data? Specifically, when we want to collect aggregate data from each step, we do not want to first collect all possible data first, and then aggregate them. This is inefficient.

Datseris:
This:
```julia
[...]

Issue #151, opened 2020-03-19
A problematic data collection case


Libbum:
Here's an example where the current implementation of data collection is causing me some pain.

**Scenario:**
Dynamical system running daily (each `step!`) in the model, agents acting on properties of the model on a yearly basis.

**An attempt at an implementation with data collection:**
```julia
function bream(m)
   m.properties[:lake].bream 
end

function pike(m)
   m.properties[:lake].pike
end

dummy_step(agent, model) = nothing

function run(model, years_to_run=20; 
             agent_properties=[:pos, :oss, :implementation_lag], 
             model_properties=Dict(:model => [bream, pike, m -> m.properties[:lake].nutrients, m -> m.properties[:affectors], m -> m.properties[:tick]]))
    yearly_agent_data = DataFrame()
    daily_model_data = DataFrame()
    step!(model, agent_step!, model_step!, 1)
    for year in 1:years_to_run
        append!(daily_model_data, step!(model, dummy_step, model_step!, 364, model_properties))
        append!(yearly_agent_data, step!(model, agent_step!, model_step!, 1, agent_properties))
    end
    yearly_agent_data, daily_model_data
end
```
I've done a few different things here to highlight some issues
1. No access to current step information (as far as I can tell) for user provided `agent_step!` and `model_step!` functions.
   
Current solution: increment an internal counter in `model_step!`, assigned to `model.properties[:tick]`.

2. There's either an underlying assumption that a model will be run using `n` steps and that's it (which really means `step!` is equivalent to running a complete simulation), or that data is only needed from one call of `step!`.
   
Here, I'm stepping the model through a set number of years, taking yearly data and daily data separately. Since appending to a `DataFrame` is only an option in the internal `_step!` method, I need to use DataFrames `append!` to capture all the data, but now my step values are reset. Of course, the `:tick` values I have are correct, the model and agents have advanced correctly, but I have to now do a bunch of alignment and post processing to get my steps in order inside the resultant `DataFrame`. 

I can of course dump yearly values using a 'complete' simulation
```julia
step!(model, agent_step!, model_step!, 365*years_to_run, agent_properties, when = 365:365:365*years_to_run)
```
but then I have no option to dump daily information from the model. It's not the end of the world if I dropped the `when` clause and saved all of my agent information and model information in the one dataframe for this specific model (100 agents), it's just highly redundant. Scaling this to something larger though..

[...]

PR #191, opened 2020-03-31
Datacollection

kavir1698:
Addresses and #151 #187 #188

This is a work in progress and not ready to be merged. New functions are added to `collect.jl` which will replace `data_collector.jl` and `parallel.jl`.

Two main functions collect data: `collect_agent_data` and `collect_model_data`. A single function `aggregate_data` aggregates collected data. `_step!` function may remain as a convenience.

To be discussed and solved: What format should be the data collected from model. In case of agents, it is simple. Each field of an agent will be a single value. Data collection from a model is not that simple. It can be a single value, e.g. from `model.properties`, or an array, e.g. from each node in space. Such mixed data cannot be put in a data frame. An option is to dump all returned data in a named tuple. The downsize is that the user has to do post processing.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,268 @@
+# This file is machine-generated - editing it directly is not advised
+
+[[ArnoldiMethod]]
+deps = ["DelimitedFiles", "LinearAlgebra", "Random", "SparseArrays", "StaticArrays", "Test"]
+git-tree-sha1 = "2b6845cea546604fb4dca4e31414a6a59d39ddcd"
+uuid = "ec485272-7323-5ecc-a04f-4719b315124d"
+version = "0.0.4"
+
+[[Base64]]
+uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+
+[[BenchmarkTools]]
+deps = ["JSON", "Logging", "Printf", "Statistics", "UUIDs"]
+git-tree-sha1 = "9e62e66db34540a0c919d72172cc2f642ac71260"
+uuid = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+version = "0.5.0"
+
+[[BinaryProvider]]
+deps = ["Libdl", "SHA"]
+git-tree-sha1 = "5b08ed6036d9d3f0ee6369410b830f8873d4024c"
+uuid = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
+version = "0.5.8"
+
+[[CategoricalArrays]]
+deps = ["Compat", "DataAPI", "Future", "JSON", "Missings", "Printf", "Reexport", "Statistics", "Unicode"]
+git-tree-sha1 = "23d7324164c89638c18f6d7f90d972fa9c4fa9fb"
+uuid = "324d7699-5711-5eae-9e2f-1d82baa6b597"
+version = "0.7.7"
+
+[[Compat]]
+deps = ["Base64", "Dates", "DelimitedFiles", "Distributed", "InteractiveUtils", "LibGit2", "Libdl", "LinearAlgebra", "Markdown", "Mmap", "Pkg", "Printf", "REPL", "Random", "SHA", "Serialization", "SharedArrays", "Sockets", "SparseArrays", "Statistics", "Test", "UUIDs", "Unicode"]
+git-tree-sha1 = "0198d18b28c093bef39872a22f1a897218a925f5"
+uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
+version = "3.8.0"
+
+[[DBInterface]]
+git-tree-sha1 = "8687a1cc640f088ea2b9c4573b16739ba3917175"
+uuid = "a10d1c49-ce27-4219-8d33-6db1a4562965"
+version = "2.0.0"
+
+[[DataAPI]]
+git-tree-sha1 = "674b67f344687a88310213ddfa8a2b3c76cc4252"
+uuid = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
+version = "1.1.0"
+
+[[DataFrames]]
+deps = ["CategoricalArrays", "Compat", "DataAPI", "Future", "InvertedIndices", "IteratorInterfaceExtensions", "Missings", "PooledArrays", "Printf", "REPL", "Reexport", "SortingAlgorithms", "Statistics", "TableTraits", "Tables", "Unicode"]
+git-tree-sha1 = "7d5bf815cc0b30253e3486e8ce2b93bf9d0faff6"
+uuid = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
+version = "0.20.2"
+
+[[DataStructures]]
+deps = ["InteractiveUtils", "OrderedCollections"]
+git-tree-sha1 = "5a431d46abf2ef2a4d5d00bd0ae61f651cf854c8"
+uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
+version = "0.17.10"
+
+[[DataValueInterfaces]]
+git-tree-sha1 = "bfc1187b79289637fa0ef6d4436ebdfe6905cbd6"
+uuid = "e2d170a0-9d28-54be-80f0-106bbe20a464"
+version = "1.0.0"
+
+[[Dates]]
+deps = ["Printf"]
+uuid = "ade2ca70-3891-5945-98fb-dc099432e06a"
+
+[[DelimitedFiles]]
+deps = ["Mmap"]
+uuid = "8bb1440f-4735-579b-a4ab-409b98df4dab"
+
+[[Distributed]]
+deps = ["Random", "Serialization", "Sockets"]
+uuid = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+
+[[Future]]
+deps = ["Random"]
+uuid = "9fa8497b-333b-5362-9e8d-4d0656e87820"
+
+[[Inflate]]
+git-tree-sha1 = "f5fc07d4e706b84f72d54eedcc1c13d92fb0871c"
+uuid = "d25df0c9-e2be-5dd7-82c8-3ad0b3e990b9"
+version = "0.1.2"
+
+[[InteractiveUtils]]
+deps = ["Markdown"]
+uuid = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
+
+[[InvertedIndices]]
+deps = ["Test"]
+git-tree-sha1 = "15732c475062348b0165684ffe28e85ea8396afc"
+uuid = "41ab1584-1d38-5bbf-9106-f11c6c58b48f"
+version = "1.0.0"
+
+[[IteratorInterfaceExtensions]]
+git-tree-sha1 = "a3f24677c21f5bbe9d2a714f95dcd58337fb2856"
+uuid = "82899510-4779-5014-852e-03e436cf321d"
+version = "1.0.0"
+
+[[JSON]]
+deps = ["Dates", "Mmap", "Parsers", "Unicode"]
+git-tree-sha1 = "b34d7cef7b337321e97d22242c3c2b91f476748e"
+uuid = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
+version = "0.21.0"
+
+[[LibGit2]]
+deps = ["Printf"]
+uuid = "76f85450-5226-5b5a-8eaa-529ad045b433"
+
+[[Libdl]]
+uuid = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+
+[[LightGraphs]]
+deps = ["ArnoldiMethod", "DataStructures", "Distributed", "Inflate", "LinearAlgebra", "Random", "SharedArrays", "SimpleTraits", "SparseArrays", "Statistics"]
+git-tree-sha1 = "f40c4dbcd957cc3afc8cca0ff26c9f8304def00d"
+uuid = "093fc24a-ae57-5d10-9952-331d41423f4d"
+version = "1.3.1"
+
+[[LinearAlgebra]]
+deps = ["Libdl"]
+uuid = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+
+[[Logging]]
+uuid = "56ddb016-857b-54e1-b83d-db4d58db5568"
+
+[[MacroTools]]
+deps = ["Markdown", "Random"]
+git-tree-sha1 = "f7d2e3f654af75f01ec49be82c231c382214223a"
+uuid = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
+version = "0.5.5"
+
+[[Markdown]]
+deps = ["Base64"]
+uuid = "d6f4376e-aef5-505a-96c1-9c027394607a"
+
+[[Missings]]
+deps = ["DataAPI"]
+git-tree-sha1 = "de0a5ce9e5289f27df672ffabef4d1e5861247d5"
+uuid = "e1d29d7a-bbdc-5cf2-9ac0-f12de2c33e28"
+version = "0.4.3"
+
+[[Mmap]]
+uuid = "a63ad114-7e13-5084-954f-fe012c677804"
+
+[[OrderedCollections]]
+deps = ["Random", "Serialization", "Test"]
+git-tree-sha1 = "c4c13474d23c60d20a67b217f1d7f22a40edf8f1"
+uuid = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
+version = "1.1.0"
+
+[[Parsers]]
+deps = ["Dates", "Test"]
+git-tree-sha1 = "d6d82d5bdbb75048e574cd2d2c89dfbf2c74250c"
+uuid = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
+version = "1.0.0"
+
+[[Pkg]]
+deps = ["Dates", "LibGit2", "Libdl", "Logging", "Markdown", "Printf", "REPL", "Random", "SHA", "UUIDs"]
+uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+
+[[PooledArrays]]
+deps = ["DataAPI"]
+git-tree-sha1 = "b1333d4eced1826e15adbdf01a4ecaccca9d353c"
+uuid = "2dfb63ee-cc39-5dd5-95bd-886bf059d720"
+version = "0.5.3"
+
+[[Printf]]
+deps = ["Unicode"]
+uuid = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+
+[[REPL]]
+deps = ["InteractiveUtils", "Markdown", "Sockets"]
+uuid = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
+
+[[Random]]
+deps = ["Serialization"]
+uuid = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+
+[[Reexport]]
+deps = ["Pkg"]
+git-tree-sha1 = "7b1d07f411bc8ddb7977ec7f377b97b158514fe0"
+uuid = "189a3867-3050-52da-a836-e630ba90ab69"
+version = "0.2.0"
+
+[[SHA]]
+uuid = "ea8e919c-243c-51af-8825-aaa63cd721ce"
+
+[[SQLite]]
+deps = ["BinaryProvider", "DBInterface", "Dates", "Libdl", "Random", "SQLite_jll", "Serialization", "Tables", "Test", "WeakRefStrings"]
+git-tree-sha1 = "25d88a45ce804e04a13ca00f1d67a35ac0e2ede6"
+uuid = "0aa819cd-b072-5ff4-a722-6bc24af294d9"
+version = "1.0.3"
+
+[[SQLite_jll]]
+deps = ["Libdl", "Pkg"]
+git-tree-sha1 = "25675b6872c3fcbe2034510db3a467405b6dc79e"
+uuid = "76ed43ae-9a5d-5a62-8c75-30186b810ce8"
+version = "3.31.1+0"
+
+[[Serialization]]
+uuid = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
+
+[[SharedArrays]]
+deps = ["Distributed", "Mmap", "Random", "Serialization"]
+uuid = "1a1011a3-84de-559e-8e89-a11a2f7dc383"
+
+[[SimpleTraits]]
+deps = ["InteractiveUtils", "MacroTools"]
+git-tree-sha1 = "2bdf3b6300a9d66fe29ee8bb51ba100c4df9ecbc"
+uuid = "699a6c99-e7fa-54fc-8d76-47d257e15c1d"
+version = "0.9.1"
+
+[[Sockets]]
+uuid = "6462fe0b-24de-5631-8697-dd941f90decc"
+
+[[SortingAlgorithms]]
+deps = ["DataStructures", "Random", "Test"]
+git-tree-sha1 = "03f5898c9959f8115e30bc7226ada7d0df554ddd"
+uuid = "a2af1166-a08f-5f64-846c-94a0d3cef48c"
+version = "0.3.1"
+
+[[SparseArrays]]
+deps = ["LinearAlgebra", "Random"]
+uuid = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+
+[[StaticArrays]]
+deps = ["LinearAlgebra", "Random", "Statistics"]
+git-tree-sha1 = "5a3bcb6233adabde68ebc97be66e95dcb787424c"
+uuid = "90137ffa-7385-5640-81b9-e52037218182"
+version = "0.12.1"
+
+[[Statistics]]
+deps = ["LinearAlgebra", "SparseArrays"]
+uuid = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
+[[StatsBase]]
+deps = ["DataAPI", "DataStructures", "LinearAlgebra", "Missings", "Printf", "Random", "SortingAlgorithms", "SparseArrays", "Statistics"]
+git-tree-sha1 = "a6102b1f364befdb05746f386b67c6b7e3262c45"
+uuid = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+version = "0.33.0"
+
+[[TableTraits]]
+deps = ["IteratorInterfaceExtensions"]
+git-tree-sha1 = "b1ad568ba658d8cbb3b892ed5380a6f3e781a81e"
+uuid = "3783bdb8-4a98-5b6b-af9a-565f29a5fe9c"
+version = "1.0.0"
+
+[[Tables]]
+deps = ["DataAPI", "DataValueInterfaces", "IteratorInterfaceExtensions", "LinearAlgebra", "TableTraits", "Test"]
+git-tree-sha1 = "242b7fde70b8bc6a30d6476adf17ca3cf1ced6ee"
+uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
+version = "1.0.3"
+
+[[Test]]
+deps = ["Distributed", "InteractiveUtils", "Logging", "Random"]
+uuid = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[[UUIDs]]
+deps = ["Random", "SHA"]
+uuid = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
+
+[[Unicode]]
+uuid = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
+
+[[WeakRefStrings]]
+deps = ["DataAPI", "Random", "Test"]
+git-tree-sha1 = "28807f85197eaad3cbd2330386fac1dcb9e7e11d"
+uuid = "ea10d353-3f73-51f8-a26c-33c1cb351aa5"
+version = "0.6.2"
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -1,0 +1,80 @@
+"""
+Collects agent properties (fields of the agent object) into a dataframe.
+"""
+function collect_agent_data(model::ABM, properties::Array{Symbol}; step=1)
+  dd = DataFrame()
+  dd[!, :id] = collect(keys(model.agents))
+  for fn in properties
+    dd[!, fn] = getproperty.(values(model.agents), fn)
+  end
+  dd[!, :step] = repeat([step], size(dd, 1))
+  return dd
+end
+
+"""
+Collects agent properties (fields of the agent object) into a dataframe
+and appends them to the supplied `df`.
+"""
+function collect_agent_data!(df::DataFrame, model::ABM, properties::Array{Symbol}; step::Integer)
+  d = collect_agent_data(model, properties, step=step)
+  df = vcat(df, d)
+  return df
+end
+
+"""
+Collects model properties from functions provided in `properties`.
+"""
+function collect_model_data(model::ABM, properties::AbstractArray; step=1)
+  dd = DataFrame()
+  for fn in properties
+    r = fn(model)
+    if typeof(r) <: AbstractArray
+      d[!, Symbol(fn)] = r
+    else 
+      dd[!, Symbol(fn)] = [r]
+    end
+  end
+  dd[!, :step] = repeat([step], size(dd, 1))
+  return dd
+end
+
+# TODO: decide on the shape of model data. what if the output for each function has a different length?
+"""
+    aggregate_data(df::AbstractDataFrame, aggregation_dict::Dict)
+  
+Aggregate `df` columns  with some function(s) specified in `aggregation_dict`.
+Each key in `aggregation_dict` is a column name (Symbol), and each value is
+an array of function to aggregate that column.
+
+Aggregation occurs per step.
+"""
+function aggregate_data(df::AbstractDataFrame, aggregation_dict::Dict)
+  all_keys = collect(keys(aggregation_dict))
+  v1 = aggregation_dict[all_keys[1]]
+  final_df = by(df, :step,  all_keys[1] => v1[1])
+  for v2 in v1[2:end]
+    dd = by(df, :step,  k => v2)
+    final_df = join(final_df, dd, on=:step)
+  end
+  for k in all_keys[2:end]
+    v = aggregation_dict[k]
+    for v2 in v
+      dd = by(df, :step,  k => v2)
+      final_df = join(final_df, all_df[di], on=:step)
+    end
+  end
+
+  # rename columns
+  colnames = Array{Symbol}(undef, length(final_df, 2)) 
+  colnames[1] = :step
+  counter = 2
+  for (k,v) in aggregation_dict
+    for vv in v
+      colnames[counter] = Symbol(join([vv,"(", string(k), ")"], ""))
+      counter += 1
+    end
+  end
+  rename!(final_df, colnames)
+
+  return final_df
+end
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -1,7 +1,7 @@
 """
 Collects agent properties (fields of the agent object) into a dataframe.
 """
-function collect_agent_data(model::ABM, properties::Array{Symbol}; step=1)
+function collect_agent_data(model::ABM, properties::Array{Symbol}; step=0)
   dd = DataFrame()
   dd[!, :id] = collect(keys(model.agents))
   for fn in properties
@@ -23,13 +23,14 @@ end
 
 """
 Collects model properties from functions provided in `properties`.
+# TODO: decide the shape of model data. what if the output for each function has a different length?
 """
-function collect_model_data(model::ABM, properties::AbstractArray; step=1)
+function collect_model_data(model::ABM, properties::AbstractArray; step=0)
   dd = DataFrame()
   for fn in properties
     r = fn(model)
     if typeof(r) <: AbstractArray
-      d[!, Symbol(fn)] = r
+      dd[!, Symbol(fn)] = r
     else 
       dd[!, Symbol(fn)] = [r]
     end
@@ -38,7 +39,6 @@ function collect_model_data(model::ABM, properties::AbstractArray; step=1)
   return dd
 end
 
-# TODO: decide on the shape of model data. what if the output for each function has a different length?
 """
     aggregate_data(df::AbstractDataFrame, aggregation_dict::Dict)
   
@@ -53,19 +53,19 @@ function aggregate_data(df::AbstractDataFrame, aggregation_dict::Dict)
   v1 = aggregation_dict[all_keys[1]]
   final_df = by(df, :step,  all_keys[1] => v1[1])
   for v2 in v1[2:end]
-    dd = by(df, :step,  k => v2)
+    dd = by(df, :step,  all_keys[1] => v2)
     final_df = join(final_df, dd, on=:step)
   end
   for k in all_keys[2:end]
     v = aggregation_dict[k]
     for v2 in v
       dd = by(df, :step,  k => v2)
-      final_df = join(final_df, all_df[di], on=:step)
+      final_df = join(final_df, dd, on=:step)
     end
   end
 
   # rename columns
-  colnames = Array{Symbol}(undef, length(final_df, 2)) 
+  colnames = Array{Symbol}(undef, size(final_df, 2)) 
   colnames[1] = :step
   counter = 2
   for (k,v) in aggregation_dict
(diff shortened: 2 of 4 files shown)


Manifest.toml
```diff
@@ -0,0 +1,262 @@
+# This file is machine-generated - editing it directly is not advised
```
Datseris: Please add `Manifest.toml` to `.gitignore` and delete this manifest

src/simulations/collect.jl
```diff
@@ -0,0 +1,246 @@
...
+  end
+  dd[!, :step] = repeat([step], size(dd, 1))
+  return dd
+end
+
+"""
+Collects agent properties (fields of the agent object) into a dataframe
+and appends them to the supplied `df`.
+"""
+function collect_agent_data!(df::DataFrame, model::ABM, properties::Array{Symbol}; step::Integer)
```
Datseris: If `step` is mandatory argument, why is it a keyword? It probably makes sense to change `step` from a keyword to argument. The same for the above method, `; step = 0` -> `, step = 0`.

Datseris:
> It can be a single value, e.g. from model.properties, or an array, e.g. from each node in space. 

I didn't get this part... How can the model data be dependent on the "node", i.e. the location in space? There is one model, which has the same properties irrespectively at where you look at space. What do you mean by that?

I thought hte only model data collection can be done on `model.properties`.

kavir1698:
> I thought hte only model data collection can be done on model.properties.

One can write a function that returns the population size at each node. This will be an array. We can restrict users to write functions that return single values. Another option is to return a dataframe, and convert array entries to tuples.

Datseris:
But this is taken care of by the data aggregation, no? `:agents => length` is valid. 

I think we should not care too much about extreme cases. The main goal is to provide a simple and clean low-level interface so anyone with more advanced needs can write their own loops for data collection. 

> We can restrict users to write functions that return single values. 

Yes, I think this is so much simpler and more fitting for a simple API. We can define a low level call `get_data(model, s::Symbol) = getproperty(model, symbol)` and `get_data(model, f::Function) = f(model)` to allow using both symbols as well as functions.

Once again, anything more advanced can be done in between the loop and appended to a master dataframe.

kavir1698:
> EDIT: Why do we have to enforce that functions return signel values? In a dataframe it is not possible for a column's value to be vectors? I am pretty sure it is.

No, but they _can_ be a tuples.

Datseris:
I don't think you are correct, vectors of vectors can be columns of dataframes:
```
julia> using DataFrames

julia> df = DataFrame()
0×0 DataFrame


julia> df[!, :x] = [rand(5) for i in 1:5]
5-element Array{Array{Float64,1},1}:
 [0.14468217632365898, 0.9324895361972538, 0.985426627191043, 0.5695308059981801, 0.4693909045018321]
 [0.08625087385686125, 0.013455473650316074, 0.3421566969571488, 0.9355785975869249, 0.20811622241347982]
 [0.6406336155898598, 0.4958374930140659, 0.776680893056402, 0.8019539161285418, 0.6010993128801057]
 [0.9118357065420952, 0.9969826727317972, 0.52270059439928, 0.6990532308503792, 0.2641174807035138]
 [0.7202702762467486, 0.3539123459572411, 0.6097693172549279, 0.24555080628724002, 0.023878247634558036]

julia> df
5×1 DataFrame
│ Row │ x                                                    │
│     │ Array{Float64,1}                                     │
├─────┼──────────────────────────────────────────────────────┤
│ 1   │ [0.144682, 0.93249, 0.985427, 0.569531, 0.469391]    │
│ 2   │ [0.0862509, 0.0134555, 0.342157, 0.935579, 0.208116] │
│ 3   │ [0.640634, 0.495837, 0.776681, 0.801954, 0.601099]   │
│ 4   │ [0.911836, 0.996983, 0.522701, 0.699053, 0.264117]   │
│ 5   │ [0.72027, 0.353912, 0.609769, 0.245551, 0.0238782]   │
```

kavir1698:
Yes, I'd corrected the response. The issue is solved then.

kavir1698:
Yes, that's a very good idea.

src/simulations/collect.jl
```diff
@@ -1,18 +1,21 @@
+get_data(a, s::Symbol) = getproperty(a, s)
+get_data(a, f::Function) = f(a)
+
 """
-Collects agent properties (fields of the agent object) into a dataframe.
+Collect agent properties into a dataframe. `properties` can have symbols (agent fields) or functions that take an agent as input.
```
Datseris: The documentation strings need to have a call signature statement, since in Julia a function name doesn't equate a call signature due to multiple dispatch

```
    collect_agent_data(model, properties, step = 0)
Collect...
```

src/simulations/collect.jl
```diff
@@ -1,18 +1,21 @@
...
 """
-function collect_agent_data(model::ABM, properties::Array{Symbol}, step)
+function collect_agent_data(model::ABM, properties::AbstractArray, step)
   dd = DataFrame()
   dd[!, :id] = collect(keys(model.agents))
   for fn in properties
-    dd[!, fn] = getproperty.(values(model.agents), fn)
+    dd[!, fn] = get_data.(values(model.agents), fn)
   end
   dd[!, :step] = repeat([step], size(dd, 1))
```
Datseris: Instead of `repeat` here you should do `fill(step, size(dd, 1))`

src/simulations/collect.jl
```diff
@@ -0,0 +1,285 @@
...
+function collect_model_data(model::ABM, properties::AbstractArray, step::Int)
+  dd = DataFrame()
+  for fn in properties
+    r =  get_data(model, fn)
+    dd[!, Symbol(fn)] = [r]
+  end
+  dd[!, :step] = fill(step, size(dd, 1))
+  return dd
+end
+
```
Datseris, line 35: there used to be in place functions `collect_.._data!(df, ...)` , what happened to them?

Datseris: They are supposed to be part of the public API in this new design.

src/simulations/collect.jl
```diff
@@ -0,0 +1,285 @@
...
+      if !isnothing(model_properties)
+        df_model = _step_model_collect!(df_model, model, model_properties,  aggregation_dict, ss)
+      end
+      if !isnothing(agent_properties)
+        df_agent = _step_agent_collect!(df_agent, model, agent_properties,  aggregation_dict, ss)
+      end
+    end
+  end
+  return df_agent, df_model
+end
```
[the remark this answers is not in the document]
Datseris: same story can be done in the `n` condition, where we change 
```julia
until(ss, n::Int, model) = ss in 1:n
until(ss, n::AbstractVector{Int}, model) = ss in n
until(ss, n::Function, model) = !n(model)
```
and we change all `for ss in 1:n` to `while until(ss, n, model) ... ss += 1; end` 

This will lead to having only **one** function, in which every condition is satisfied. See also #167, and the issue that led to #167

src/simulations/collect.jl
```diff
@@ -0,0 +1,285 @@
...
+  rename!(final_df, colnames)
+
+  return final_df
+end
+
+## step functions
+###################
+
+"used in _step!"
+function _step_agent_collect!(df::DataFrame, model::ABM, agent_properties,  aggregation_dict, step::Int)
```
Datseris, line 84: I am reading this function again, but I can't understand why `_step!` is part of the name. Isn't this function just `collect_agent_data!(df ...) ` ?

src/simulations/collect.jl
```diff
@@ -0,0 +1,285 @@
...
+  if isnothing(aggregation_dict)
+    dft = collect_agent_data(model, agent_properties, step)
+    df = vcat(df, dft)
+  else
+    dfall = collect_agent_data(model, agent_properties, step)
```
Datseris, line 89: Line `dft = ...` and `dfall = ...` is identical in both clauses. Please put it outside the `if` clause to reduce code duplication.

Datseris:
There is still a lot of code duplication happening here, that we should actively try to avoid.

Libbum:
Joining the party now. Will take a look and see if I can identify some simplifications, tests and cases we need to take care of.

src/simulations/collect.jl
kavir1698: They were redundant. We have three basic functions now: collect agent data, collect model data, and aggregate. Updating an existing `df` is just as easy as `vcat(df, new_df)`.

kavir1698:
> What is _step! supposed to mean?

It is the backend to the `step!` function.

kavir1698:
What remains to be done:

- [ ] Test new code for bugs.
- [ ] Update `step.jl`.
- [ ] Update `paramscan`.
- [ ] Remove `data_collector.jl` and `parallel.jl`.
- [ ] Update tests.
- [ ] Update documentations.
- [ ] Update examples.

src/simulations/collect.jl
```diff
@@ -0,0 +1,268 @@
...
+
+function _step!(model, agent_step!, model_step!, n, when; model_properties=nothing, aggregation_dict=nothing, agent_properties=nothing)
+  df_agent = DataFrame()
+  df_model = DataFrame()
+  ss = 0
+  while until(ss, n, model)
+    step!(model, agent_step!, model_step!, 1)
+    if ss in when
+      if !isnothing(model_properties)
+        df_model = collect_model_data!!(df_model, model, model_properties,  aggregation_dict, ss)
```
Libbum, line 119: `!!`

Libbum: Oh, sorry. There is a function for that.

Libbum:
I have (at least) two problems with the current implementation.
First, how the column names are handled in the aggregate function:
```julia
colnames[counter] = Symbol(join([vv,"(", string(k), ")"], ""))
```
This is so abstract and ends up giving names like `#347` if an anonymous function is used. I think we should accept named tuples as responses and fall back to the above convention if there's no other choice.

Second, there's a lot of work done to export model properties such that we can use `model.x` now. This hasn't been considered yet in the implementation.

```julia
mutable struct Orchard    
    apple_trees::Int    
    peach_trees::Int    
    land::Float64 #m^2    
    irrigated::Bool    
end    
    
mutable struct Farm    
    cows::Int    
    feed::Float64 #tonne    
    tractors::Int     
end    
    
properties = Dict(:orchard => Orchard(560,324,1507.5,true),    
                  :farm => Farm(242, 26.8, 3),    
                  :year => 0,    
                  :bank_balance => 81051.50,    
                  :loans => 12050.82)    
model = ABM(Agent1, GridSpace((50,50)); properties = properties)
```

For now, my workers (`Agent1`) are not so important, but I want to know how many assets I have.
```julia
julia> Agents.collect_model_data(model, [m->m.orchard.apple_trees, m->m.orchard.peach_trees, m->m.farm.cows, m->m.farm.tractors, m->m.bank_balance-m.loans], 1)
1×6 DataFrames.DataFrame
│ Row │ #13   │ #14   │ #15   │ #16   │ #17     │ step  │
│     │ Int64 │ Int64 │ Int64 │ Int64 │ Float64 │ Int64 │
├─────┼───────┼───────┼───────┼───────┼─────────┼───────┤
│ 1   │ 560   │ 324   │ 242   │ 3     │ 69000.7 │ 1     │
```

Surely we can think of a better way to export these properties. I'm going to try a few things, but let me know your thoughts on this scenario.

Datseris:
The names suggestion provided as an optional tuple of symbols is great. I concur.

@kavir1698  the name `step!` should no longer be used to mean both stepping the model and data collection. A different name should be used for that. `step!` should exclusively mean "evolve the model for `n` steps". And we can also consider the name `evolve!` instead of `step!` as well. 

In addition, function names starting `_` are intended to be internal and not part of a public API.

src/simulations/collect.jl
```diff
@@ -0,0 +1,268 @@
...
+get_data(a, s::Symbol) = getproperty(a, s)
+get_data(a, f::Function) = f(a)
+
+"""
+    collect_agent_data(model::ABM, properties::AbstractArray, step::Int)
+
+Collect agent properties into a dataframe. `properties` can have symbols (agent fields) or functions that take an agent as input.
+"""
+function collect_agent_data(model::ABM, properties::AbstractArray, step::Int)
+  dd = DataFrame()
```
Libbum, line 12: Since these functions are always for one `step`, why not return a [`DataFrameRow`](https://juliadata.github.io/DataFrames.jl/stable/lib/types/#DataFrames.DataFrameRow)?

Datseris:
I'll start adding commits here as well on the weekend. This will take a lot of effort to update all docs and examples.

src/simulations/collect.jl
Libbum: See [constructing row by row](https://juliadata.github.io/DataFrames.jl/stable/man/getting_started/#Constructing-Row-by-Row-1).

kavir1698: Because they do not output a single row when you are collecting data from all agents.

Libbum: True. I'll think about that.

kavir1698:
> @kavir1698 the name step! should no longer be used to mean both stepping the model and data collection. A different name should be used for that. step! should exclusively mean "evolve the model for n steps". And we can also consider the name evolve! instead of step! as well.
> 
> In addition, function names starting _ are intended to be internal and not part of a public API.

`evolve!` is  not general enough and does not imply data collection. I suggest something explicit, like `step_and_collect!` or `run!`.

Also, `_step!` is intended to be internal. The API has `step!`.

src/simulations/collect.jl
```diff
@@ -0,0 +1,268 @@
...
+###################
+
+until(ss, n::Int, model) = ss in 1:n
+until(ss, n::AbstractVector{Int}, model) = ss in n
+until(ss, n::Function, model) = !n(model)
+
+function _step!(model, agent_step!, model_step!, n, when; model_properties=nothing, aggregation_dict=nothing, agent_properties=nothing)
+  df_agent = DataFrame()
+  df_model = DataFrame()
+  ss = 0
```
Libbum, line 114: `until` needs `ss>0`, otherwise the loop will never run.

src/simulations/collect.jl
kavir1698: `collect_model_data` does always outputs a single row. But its easier that both of these functions return the same object type.

kavir1698:
> Second, there's a lot of work done to export model properties such that we can use model.x now. This hasn't been considered yet in the implementation.

Well, it is considered in `collect_data`, so that the user can have a mix of symbols and functions in `agent_properties` array.

Datseris:
> I suggest something explicit, like `step_and_collect!` or `run!`.

I agree with `run!`.

Libbum:
Some explorations with my `Farm`/`Orchard` above. 
Since we already require users to write functions like `model_step!` and get them to write a `properties` dictionary for the model, why not continue this capability and allow data collection to be a function which returns a named dictionary ready for `DataFrame` insertion.

I'm using the exact model as above here, with some added methods:

```julia
function model_step!(model)
    model.loans /= 1.05
    model.bank_balance *= 1.2
    model.farm.cows += 2
    model.orchard.apple_trees -= 10
end

assets(model) =                                                                                                                                                            
    Dict(:trees => model.orchard.apple_trees + model.orchard.peach_trees,                                                                                                  
         :cows => model.farm.cows,                                                                                                                                         
         :tractors => model.farm.tractors,                                                                                                                                 
         :cash => model.bank_balance - model.loans)
```

Now, `assets` is a function that returns everything I require from the model, with arbitrary processing (#188) and a solution to the column naming issue.

With mockups of an implementation:
```julia
function collect_model(model::ABM, properties::Function, step::Int)                                                                                                        
    data = properties(model)                                                                                                                                               
    if typeof(data) <: Dict{Symbol,}                                                                                                                                       
        data[:step] = step                                                                                                                                                 
    else                                                                                                                                                                   
        # Array of properties, whatever. Or make this a guard and throw an ArgumentError                                                                                   
    end                                                                                                                                                                    
    data                                                                                                                                                                   
end                                                                                                                                                                        
                                                                                                                                                                           
function collect_model!(df::DataFrame, model::ABM, model_properties::Function, aggregation_dict, step::Int)                                                                
  new_row = collect_model(model, model_properties, step)                                                                                                                   
  if isnothing(aggregation_dict)                                                                                                                                           
    push!(df, new_row)                                                                                                                                                     
  else                                                                                                                                                                     
    #TODO: unimplemented                                                                                                                                                   
  end                                                                                                                                                                      
end

function _step!(model, agent_step!, model_step!, n, when; model_properties::Function, aggregation_dict=nothing, agent_properties=nothing)
  df_agent = DataFrame()

  init_model_output = model_properties(model)
  df_model = DataFrame([typeof(val) for val in values(init_model_output)], collect(keys(init_model_output)))
  df_model[!, :step] = Int[]

  #if step0
  #  init_model_output[:step] = 0
  #  push!(df_model, init_model_output)
  #end

  ss = 1
  while until(ss, n, model)
    step!(model, agent_step!, model_step!, 1)
    if ss in when
      if !isnothing(model_properties)
        collect_model!(df_model, model, model_properties,  aggregation_dict, ss)
      end
      if !isnothing(agent_properties)
        df_agent = collect_agent_data!(df_agent, model, agent_properties,  aggregation_dict, ss)
      end
    end
    ss += 1
  end
  return df_agent, df_model
end
```
We get an appropriate dataframe:
```julia
julia> _step_test!(model, dummystep, model_step!, 5, 1:5; model_properties = assets)
(0×0 DataFrame
, 5×5 DataFrame
│ Row │ tractors │ trees │ cash     │ cows  │ step  │
│     │ Int64    │ Int64 │ Float64  │ Int64 │ Int64 │
├─────┼──────────┼───────┼──────────┼───────┼───────┤
│ 1   │ 3        │ 874   │ 85784.8  │ 244   │ 1     │
│ 2   │ 3        │ 864   │ 105784.0 │ 246   │ 2     │
│ 3   │ 3        │ 854   │ 129647.0 │ 248   │ 3     │
│ 4   │ 3        │ 844   │ 158154.0 │ 250   │ 4     │
│ 5   │ 3        │ 834   │ 192240.0 │ 252   │ 5     │)
```

src/simulations/collect.jl
kavir1698: There is `ss += 1` in `_step!`. Do you mean something else?

Libbum: No, I mean the initialisation. `ss=0`, then the call to `until(ss, n, model)`.  0 in 1:n will always return false.

kavir1698: Oh yeah!

kavir1698:
It seems like there is no single approach that works best for all different scenarios. What you suggest works well for the example you provide. But if you are just collecting model fields, then this will be the more verbose approach than the current approach.

Column names problem can be solved by adding an option for the user to provide them.

Datseris:
I think we need to take a step back and write down an overview of what we want, and what we have, and what the incoming suggestions do. I will write the overview of the **low level api functions**, because this is what I've spent time thinking. If we have these functions, we checmark them.

`THING` can be `agent` or `model`.

- [ ] `step!(..., n)`, which evolves the model for `n` steps.
- [ ] `collect_THING_data(...)`, which initializes the data container and collects data at the current step.
- [ ] `collect_THING_data!(df, ....)` which inplace appends data to existing container
- [ ] `aggregate!(df, ...)` which does data aggregation over agents at current step

If we have these, everything else is simple higher level for loops. I think at the moment we have all of the functions I write above.

I am a bit lost now for _how_ the data collection is done. We have `properties`, which is a vector of symbols and/or functions. Then, for each intry in teh vector we get properties and/or functions of the THING. This to my understanding makes _every_ scenario possible.

@Libbum  you propose to change `properties` to be a dictionary. What is the benefit of that? I can see that one benefit is that it explicitly names the columns of the dataframe, but is there any other.?

Libbum:
- It solves the naming issue, yes. 
- A dictionary is the suggested method of appending to a `DataFrame`, so there's no conversion or performance loss.
- There's no additional user overhead to learn. If you can construct a model, you can construct a data extractor. 
- I still struggle to see how `assets(model)` is any more verbose than `[m->m.orchard.apple_trees+m.orchard.peach_trees, m->m.farm.cows, :loan ...]`, which also mixes types (confusing), and would require some other way of adding column names (are these separate vectors? What's the cognitive overhead to match my names and values?)
- Functions could provide greater control over output. If a user only wants aggregate agent data and some model properties, they can clearly use just one dataframe with this implementation.
- Aggregate functions are also possible through `:avg_x_pos => mean([first(a.pos) for a in values(model.agents)])` or somesuch.
- Big drop in code complexity.

Not sure if it is a better solution or not, but for the moment I still hit the wall with what's happening under the hood here. I perhaps need some more complex use cases than the current examples have to offer and how the current approach will answer these questions.

kavir1698:
> A dictionary is the suggested method of appending to a DataFrame, so there's no conversion or performance loss.

This wouldn't work for agent data collection, where the output is not a row but a dataframe, right?

> I still struggle to see how assets(model) is any more verbose than...

I meant it is more verbose than `[model.x, model.y, etc]`, i.e. when you directly collect model fields.

> There's no additional user overhead to learn.

Does the current method has additional overhead to learn?

> Functions could provide greater control over output. If a user only wants aggregate agent data and some model properties, they can clearly use just one dataframe with this implementation.

The current implementation supports functions too. The problem with having dataframe is for special cases that we have issues: collecting full agent data and model data. Otherwise, you could already collect aggregate agent data and model data before renewing data collection.

Libbum:
> I meant it is more verbose than [model.x, model.y, etc], i.e. when you directly collect model fields.

Agreed! But we can't do that (currently).

```julia
julia> model.properties
Dict{Symbol,Any} with 5 entries:
  :loans        => 12050.8
  :year         => 0
  :farm         => Farm(242, 26.8, 3)
  :orchard      => Orchard(560, 324, 1507.5, true)
  :bank_balance => 81051.5

julia> model.farm.cows
242

julia> Agents.collect_model_data(model, [:year], 1)
1×2 DataFrame
│ Row │ year  │ step  │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 0     │ 1     │

julia> Agents.collect_model_data(model, [model.year], 1)
ERROR: MethodError: no method matching get_data(::AgentBasedModel{Agent1,GridSpace{LightGraphs.SimpleGraphs.SimpleGraph{Int64},2,Int64},typeof(fastest),Dict{Symbol,Any}}, ::Int64)
Closest candidates are:
  get_data(::Any, ::Symbol) at /mnt/turtle/repos/Agents.jl/src/simulations/collect.jl:3
  get_data(::Any, ::Function) at /mnt/turtle/repos/Agents.jl/src/simulations/collect.jl:4
Stacktrace:
 [1] collect_model_data(::AgentBasedModel{Agent1,GridSpace{LightGraphs.SimpleGraphs.SimpleGraph{Int64},2,Int64},typeof(fastest),Dict{Symbol,Any}}, ::Array{Int64,1}, ::Int64) at /mnt/turtle/repos/Agents.jl/src/simulations/collect.jl:29
 [2] top-level scope at REPL[14]:1
```

kavir1698:
yes, my mistake. I meant this `Agents.collect_model_data(model, [:year], 1)`

Libbum:
Indeed. So my point here is that the current implementation makes a bunch of the work done in #182 inaccessible. You have to resort to the function approach to access nested data. Then, you need to be either very verbose:
```julia
cows(model) = model.farm.cows
Agents.collect_model_data(model, [cows], 1)
```
or resort to methods I've shown already `Agents.collect_model_data(model, [m->m.farm.cows], 1)` that give us naming issues.

Datseris:
I still don't see a problem. @Libbum you could write a method that dispatches on whether `properties` is `::Dict` and do this collection.

I like the current version, because then `[:property1, :property2]` is undeniably the simpler, and shorter syntax if one has direct properties without processing them, while I also like the functional approach, 
```julia
cows(model) = model.farm.cows
Agents.collect_model_data(model, [cows], 1)
```
because I think it leads to clearer scientific code base. 

---

Thankfully, Julia is by far the best language in existence and allows us to use dictionary for `properties` as well as the vector we use now, because of Multiple Dispatch. We only have to create two helper functions:

```julia
get_names(propertes::Vector) = ... # what we do now, where symbols are the names and functions make this crazy process with parenthesis
get_names(prop::Dict) = keys(prop) # here it is probably wise to sort keys by name
```
and the second helper function 
```
actually_get_data(data, prop::Vector, key::int) = what we do now
actually_get_data(data, prop::Dict, key) = what you have to do for dict
```
importantly, the loop in the datacollection doesn't change, beceause we loop over fields as:
```
for key in eachindex(prop)
```
and this works irrespectively if `prop` is a Dict or a vector, since:
```julia
julia> a = Dict(:a => .5, :b=> 0.2)
Dict{Symbol,Float64} with 2 entries:
  :a => 0.5
  :b => 0.2

julia> eachindex(a)
Base.KeySet for a Dict{Symbol,Float64} with 2 entries. Keys:
  :a
  :b
```

---

My problem is: **I am not sure whether we should do that**. I still don't see something that is _only possible_ with the dict approach. As far as I can tell, teverything that is possible with the dict approach is possible with the Vector approach. Thus, is it worth it complicating the API ? So far the only "benefit" is that you don't hwave to define the function `cow`, yet I say that if you use this dictionary at least more than once, then you already "defined cow" twice....?

kavir1698:
And, one would have to write separate functions anyway, if they are more complicated that this simple example.
________________________________

Datseris:
@kavir1698 for the checklist I wrote, everything is implemented, correct?

kavir1698:
Correct. Except, we still have to add option to supply column names.
________________________________

Datseris:
Alright, I've moved the TODOs in the first-most comment of this thread, so that it is easy to do. Whenever we have time we do some of them. I'll have a look now at the source in more detail.

Libbum:
OK, fair arguments! I concede this point. You're right, when it comes down to brass tacks I'm arguing over `cow(m) = m.farm.cow` vs `:cow => m.farm.cow`, and that's literally just semantics.

My pain point was just to see a good example of this.

I can even do this if I really want:
```julia
function assets(model)
    trees(model) = model.orchard.apple_trees + model.orchard.peach_trees                                                                                                  
    cows(model) = model.farm.cows                                                                                                                                         
    tractors(model) = model.farm.tractors                                                                                                                                 
    cash(model) = model.bank_balance - model.loans
    [trees, cows, tractors, cash]
end

julia> Agents.collect_model_data(model, assets2(model), 1)
1×5 DataFrame
│ Row │ trees │ cows  │ tractors │ cash    │ step  │
│     │ Int64 │ Int64 │ Int64    │ Float64 │ Int64 │
├─────┼───────┼───────┼──────────┼─────────┼───────┤
│ 1   │ 884   │ 242   │ 3        │ 69000.7 │ 1     │
```

Moving on!

Datseris:
oh, cool syntax!

Libbum:
This also scopes the helper functions such that you don't get conflicts when you want to name some value in the table the same as some higher order method.

Datseris:
I still don't get why `_step!` is internal, and why it has that name. To my understanding, `_step!` is the function `run!` we were talking about...?

Libbum:
Just for clarity: is `step!` still going to be our exposed endpoint? The internal mechanism `_step!` will be renamed to `run!`?

If so, no issues. If it's the other way round, we'll also need to re-document `model_step!` and `agent_step!`.

Datseris:
yes of course `step!(model, agent_step, model_step, n)` stays as is and is of course exported.

Libbum:
`step.jl` is updated to the point it now gets the HK example running. Haven't closely integrated the `aggregate_dict` value just yet.

Datseris:
So the current `step!` at line 96 of step.jl is the `run!` function right? anyone working on that now or should I do some quick changes? p.s: if you use the Julia slack, we have a channel for JuliaDynamics, are you in there?

Libbum:
I'm off to bed, so go right ahead. Don't use the slack often but I'm there. I'll open it up so we can chat easier about this over the coming days.

But no, line 96 is the 'complete' version of the API facing `step!`, which is a wrapper for the parallel and series replicates functions first, then does the direct work calling `run!` on line 107. That function is now line 121 in collect.jl

Datseris:
The function `step!` will NOT be doing any data collection. This is the reason issue #188 was opened in the first place. So, 

> the 'complete' version of the API 

is just `run!`. 

* `step!` evolves the model.
* `collect_whatever....` collects data for one step.
* `run!` is just a loop over `step!` and `collect!`. `run!` can of course also include parallel / replicates options.

Datseris:
The function `step!` will NOT be doing any data collection. This is the reason issue #188 was opened in the first place. So,

>    the 'complete' version of the API

is just `run!.

*    step! evolves the model.
 *   collect_whatever.... collects data for one step.
  *  run! is just a loop over step! and collect!. run! can of course also include parallel / replicates options.

Seems to me that `step!` originally in line 96 of `step.jl`, is just `run!` but deciding whether to do parallel / replicates.

---

kind of offtopic, @kavir1698 please respect 92 characters per line of code, even for docstrings. Currently the function call signatures are very hard to read on my computer and Julia style guide suggests limiting to 92 chars.

Datseris:
@Libbum why does the line 
```julia
    collect_model_data!(df_model, model, model_properties, aggregation_dict, 0)
```
exist in `run!` (originally in line 129 of `collect.jl`? Aggregation doesn't make sense for model data.

Datseris:
`isnothing` is not valid in Julia 1.0, in case we want to support that version.

Datseris:
Arguably the most useful thing I've learned while developing JuliaDynamics, and what I always teach when I give workshops on writing good scientific code, is the following:

* **Write the documentation string first, and _then_ the code**.

This helps you reflect on what you want the function to do, and how to design an API that you are happy with _before_ you start coding and get lost in all the functions. 

So, in the same spirit, let's first agree and finalize the docstring of `run!` that I just defined in the file `collect.jl`, and agree on that. Then we can do other things. I've written some TODOs, points to be discussed in the docstring. Please read them and report back here.

Datseris:
I am talking with @Libbum on slack, and we are both confused about many things, so please @kavir1698 . you wrote a lot of code but now take some time to explain things.

kavir1698:
@Datseris I've never used slack. how can I join you?

src/simulations/collect.jl
```diff
@@ -115,14 +115,25 @@ add_data(s, when::Bool) = when
...
 function _step!(
-    model, agent_step!, model_step!, n, when = true;
-    model_properties=nothing, aggregation_dict=nothing, agent_properties=nothing
+    model, agent_step!, model_step!, n;
+    collect0 = true, when = true,
+    model_properties=nothing, aggregation_dict=nothing, agent_properties=nothing,
   )
-  df_agent = DataFrame()
-  df_model = DataFrame()
+  if collect0
```
kavir1698, line 195: I find `collect0` unnecessary. The user can just choose to include 0 in when or not.

Datseris: That is a good point, but it doesn't work if `when` is a function instead of numbers.

Datseris:
@kavir1698 https://slackinvite.julialang.org/  and then find the channel `#dynamics-bridged`. Keep in mind that you can also use Gitter, there is a channel for Julia dynamics there as well. https://gitter.im/JuliaDynamics/Lobby

kavir1698:
> ? Aggregates the model data? What is the aggregation done over? For agents, the aggreation is done over agents. For model, there exists only one model. What is aggregated?

Good point. For a single step, there is nothing to aggregate. I think this function is useless, except that there are cases where collect model can return multiple rows.

src/simulations/collect.jl
```diff
@@ -0,0 +1,361 @@
...
+`aggregation_dict` decides whether the agent data should be aggregated over agents.
+
+#TODO: Describe what `aggregation_dict` is.
+
+# TODO: Well, how do you use aggregation on entries of `agent_properties` that are
+**functions**? Maybe it is better to make `aggregation_dict` keys to be integers,
+that are the entries of `agent_properties`, instead of `symbols` that are only limited
+to agent fields and not functions.
+
+# TODO: isn't it simpler to make `aggregation_dict` a VECTOR just like
```
kavir1698, line 35: A dict is just easier to write and read, especially when you want to apply multiple functions to each field.

src/simulations/collect.jl
```diff
@@ -0,0 +1,361 @@
...
+then the data for this entry is agent's field with that `name`. If an entry is a `f::Function`,
+then the data for this entry is just `f(a)` for each agent `a`.
+
+`model_properties::Vector` works exactly like `agent_properties` but for model level data.
+
+`aggregation_dict` decides whether the agent data should be aggregated over agents.
+
+#TODO: Describe what `aggregation_dict` is.
+
+# TODO: Well, how do you use aggregation on entries of `agent_properties` that are
```
kavir1698, line 30: I am not sure what you mean here. Aggregation happens on dataframe columns.

Datseris: Where? Where are you referencing.I don't know if you mean the first sentence or the second. You can reference multiple lines in the Github code commentiong functionality (just select many of them)

src/simulations/collect.jl
Datseris: good point. So how do we handle cases where the collected data are functions of agents instead of just fields?

Datseris:
> Good point. For a single step, there is nothing to aggregate. I think this function is useless, except that there are cases where collect model can return multiple rows.

I am still not satisfied by this answer and it confuses things even more. How can "model data collection" return "multiple rows" since **by definition** data collection happens ONCE per step? 

Please, explain this thing as explicitly as possible. In addition, why would you say that we can "aggregate over steps"? This was never part of the offered functionality, and I don't think we should try to do this either. Let the user do at least something! We should collect data *per step*, and that is it. Further aggregation _over_ steps should be done on the user side I feel. It just keeps everything clean and with a singular purpose.

src/simulations/collect.jl
Datseris: Wait wait. Aggregation happens on column _names_ so they can use the name of `f` ? like `:f` ?

Datseris: That's why I am asking you to *explain* what `aggregation_dict` is. Is it exactly the same thing as before? Just write it out explicitly, and then I don't have to assume any more...

src/simulations/collect.jl
kavir1698: I was talking about the second point. The `aggregate_data` function accepts only a dataframe and a dictionary. So, I don't see what you mean by agent_properties that are function.

Datseris: A user defines `f(a) = agent.something ^ 2`. And collects data with `agent_properties = [f]`. What is `aggregate_data` in this case if you want to aggregate over the mean of the function `f`?

kavir1698:
> I am still not satisfied by this answer and it confuses things even more. How can "model data collection" return "multiple rows" since by definition data collection happens ONCE per step?

It is run once per step. I was wondering if with some user functions, `collect_model_data` could possibly returns multiple rows per step. Probably not. Then this function is just redundant.

> In addition, why would you say that we can "aggregate over steps"?

Where did I say that?

Datseris:
You said "For a single step, there is nothing to aggregate", which made me _assume_ that you implied that for "multiple steps, it is possible to aggregate". But this doesn't make sense to me, since aggregation doesn't happen over steps, but over agent properties.

Datseris:
> Then this function is just redundant.

Alright, then we should remove it.

src/simulations/collect.jl
kavir1698: In this case, it will be `aggregate(df, Dict(Symbol("f(something)")=> mean))`.
Aggregation occurs over the column name that holds values of `f`.

kavir1698:
>  (in fact @Datseris votes to not even bother to do this, but instead show in the Tutorial (the Schelling example) how to use rename!.

I vote the same.

src/simulations/collect.jl
Datseris: perfect.

src/simulations/collect.jl
```diff
@@ -85,7 +77,8 @@ function collect_agent_data(model::ABM, properties::AbstractArray, step::Int = 0
   dd = DataFrame()
   dd[!, :id] = collect(keys(model.agents))
   for fn in properties
-    dd[!, fn] = get_data.(values(model.agents), fn)
+    colname = typeof(fn) <: Function ? Symbol(fn) : fn
```
Datseris, line 83: this `colname` should be made a function that uses multiple dispatch on `::Function` to avoid code duplication.

src/simulations/collect.jl
```diff
@@ -0,0 +1,340 @@
...
+"""
+    aggregate_data(df::AbstractDataFrame, aggregation_dict::Dict)
+
+Aggregate `df` columns  with some function(s) specified in `aggregation_dict`.
+Each key in `aggregation_dict` is a column name (Symbol), and each value is
+an array of function to aggregate that column.
+
+Aggregation occurs per step.
+"""
+function aggregate_data(df::AbstractDataFrame, aggregation_dict::Dict)
```
Libbum, line 119: Scenario: `Agent5`, `agent_properties=nothing`, `aggregation_dict=Dict(:weight => [mean])`.

`collect_agent_data!` is called, but since there is no `agent_properties`, `dft = DataFrame()`.

Thus, `dfnames` here will return `0-element Array{Symbol,1}`, thus  `aggregate_data` will return nothing.

kavir1698: An assertion that `aggregation_dict` must be `nothing` when `agent_properties` is nothing should solve such scenario.

Libbum: But what if all I want is `mean(weight)` an an output?

kavir1698: Then `agent_properties=[:weight]`, `aggregation_dict=Dict(:weight =>[mean])`

Datseris: then you _must_ give `agent_properties=[:weight]` ?

Libbum: I'll write up an example to test that workflow.

Datseris: Aha. Now I see a good point. `agent_properties` is _reduntant_ if `aggregation_dict` is given **EXCEPT** in the scenario where `agent_properties` contains functions. `aggregation_dict` doesn't know if a column of the dataframe came from applying a function or collecting a field.

Datseris: Let's not mess with all these complexities and expect that `agent_properties` must not be nothing if `aggregation_dict` is not nothing...?

kavir1698: A little redundancy here and there is a fair price for a simple source code.

Libbum: That may indeed be the outcome. I'll take that as the assumption and work on some tests surrounding this.

src/simulations/collect.jl
```diff
@@ -0,0 +1,338 @@
+###################################################
+# Definition of the data collection API
+###################################################
+get_data(a, s::Symbol) = getproperty(a, s)
+get_data(a, f::Function) = f(a)
+
+# TODO: `add_data` is a misleading name, we need a better one.
+add_data(s, model, when::AbstractVector) = s ∈ when
```
kavir1698, line 8: `inWhen`?

Datseris: `inwhen` is good, and I was also thinking something crazy verbose like `should_we_collect` :D If julia allowed `?` at the end of function names iot would be super cool :D

kavir1698: yeah. Or `time_to_collect`

src/simulations/collect.jl
```diff
@@ -0,0 +1,356 @@
...
+      counter += 1
+    end
+  end
+  rename!(final_df, colnames)
+
+  return final_df
+end
+
+"used in _run!"
+function collect_agent_data!(df::DataFrame, model::ABM, agent_properties,  aggregation_dict, step::Int)
```
Libbum, line 153: This is not at in-place function, so shouldn't have an `!` in its name.

Datseris: this would confilct with the version that initializes the data container.

Datseris:
At the moment we have:
```julia
function collect_agent_data(model::ABM, properties::AbstractArray, step::Int = 0)
  dd = DataFrame()
  dd[!, :id] = collect(keys(model.agents))
  for fn in properties
    dd[!, Symbol(fn)] = get_data.(values(model.agents), fn)
  end
  dd[!, :step] = fill(step, size(dd, 1))
  return dd
end

function collect_agent_data!(df::DataFrame, model::ABM, agent_properties,  aggregation_dict, step::Int)
  dft = collect_agent_data(model, agent_properties, step)
  if aggregation_dict == nothing
    df = vcat(df, dft)
  else
    dfa = aggregate_data(dft, aggregation_dict)
    df = vcat(df, dfa)
  end
  return df
end
```
and in the core `_run!` loop we do:
```julia
    if should_we_collect(s, model, when)
      dfm = collect_model_data(model, model_properties, s)
      df_model = vcat(df_model, dfm)
      df_agent = collect_agent_data!(df_agent, model, agent_properties, aggregation_dict, s)
    end
```
so, `collect_agent_data` without `!` is not used anywhere?

Datseris:
I very much dislike the current state that `aggregation_dict` needs `agent_properties` to be specified. I propose a simpler, and cleaner solution. 

1. If `aggregation_dict` is given, `agent_properties` is ignored if also given.
2. The user input aggregation_dict is a dictionary with keys either functions or symbols (for agent fields or functions to extract data). 
1. The data collected are the `keys` of this dictionary (thus, `agent_properties = keys(agdict)`).
3. The only minute change we need to do is change the line `available_keys = [k for k in all_keys if in(k, dfnames)]` of the function `aggregate_data` to `available_keys = [Symbol(k) for k in all_keys if in(k, dfnames)]`
4. ???
5. Profit

Datseris:
Let's not forget the datacollection API mentioned in https://github.com/JuliaDynamics/Agents.jl/pull/191#issuecomment-607233788  . Does this api exist? Is it still satisfied?

Libbum:
Yes, that's the point I was trying to make before. That can also strike a bunch of extra testing that I've been writing (uncommitted). 

I'll implement that now and re-align these tests.

src/simulations/collect.jl
kavir1698: It is updating `df`.

Libbum: Not currently, no. It returns a new dataframe, but the input `df` remains as it was before the call.

Anyhow, this will change with the new implementations we've discussed in slack.

kavir1698:
> so, collect_agent_data without ! is not used anywhere?

it is used in `collect_agent_data!`!

kavir1698:
> 1. If aggregation_dict is given, agent_properties is ignored if also given.

This is the separation of data collection and aggregation that we talked about. First data is collected and then aggregated. It doesn't make sense to both collect all data and aggregate them. If all data is collected, the user can aggregate them on their own.

> The user input aggregation_dict is a dictionary with keys either functions or symbols (for agent fields or functions to extract data).

No, keys are always symbols. They the column names of the df.

> The only minute change we need to do is change the line available_keys = [k for k in all_keys if in(k, dfnames)] of the function aggregate_data to available_keys = [Symbol(k) for k in all_keys if in(k, dfnames)]

Unnecessary. The keys are already Symbol.

kavir1698:
> The only minute change we need to do is change the line available_keys = [k for k in all_keys if in(k, dfnames)] of the function aggregate_data to available_keys = [Symbol(k) for k in all_keys if in(k, dfnames)]

I see now. This allows the user to input functions in aggregation_dict too. Good!

kavir1698:
Ok, I wasn’t following slack. From: Tim DuBoisSent: Thursday, April 2, 2020 6:34 PMTo: JuliaDynamics/Agents.jlCc: Ali R. Vahdati; MentionSubject: Re: [JuliaDynamics/Agents.jl] New data collection (#191) @Libbum commented on this pull request.In src/simulations/collect.jl:> +  colnames[1] = :step+  counter = 2+  for k in available_keys+    v = aggregation_dict[k]+    for vv in v+      colnames[counter] = Symbol(join([vv,"(", string(k), ")"], ""))+      counter += 1+    end+  end+  rename!(final_df, colnames)++  return final_df+end++"used in _run!"+function collect_agent_data!(df::DataFrame, model::ABM, agent_properties,  aggregation_dict, step::Int)Not currently, no. It returns a new dataframe, but the input df remains as it was before the call.Anyhow, this will change with the new implementations we've discussed in slack.—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

Datseris:
@kavir1698 I can't read your latest post :D

kavir1698:
Oops! what happened there? It's ok. it was nothing important.

src/simulations/collect.jl
kavir1698: `df = vcat(df, dfa)` this updates the `df`.

Datseris:
please try to have a look at the changes of the latest commit I did, https://github.com/JuliaDynamics/Agents.jl/pull/191/commits/699d62ca8691ef062410e0159153712175234476 . It uses the keyword propagation technique. The previous code was unsustainable, because no propagation was done anywhere, leading to the halt of #167 . 

When you wrote down the new methods this time, the same thing happened: no keyword propagation was used. If at some point in the future we wanted to add some new functionality to `run!` via a new keyword, we would have to alter around 5 to 6 functions. With the keyword propagation technique, there is only one central place that keywords are _actually defined_, `_run!`. So please have a look at the changes I did. I believe using this technique will help you not only here for developing Agents.jl, but also in your own code base. In fact, it will help you even more in your own codebase, since scientific codebases actually change much faster than packages!

src/simulations/collect.jl
Libbum: But it doesn't. `vcat` is not an in-place function and the `df` result is now locally scoped. To update in place there would be no need to return a dataframe and we'd expect the method to alter the input.

```julia
julia> df_agent = DataFrame();
julia> df = Agents.collect_agent_data!(df_agent, model, [:weight], nothing, 1);
julia> df_agent
0×0 DataFrame
julia> df
3×3 DataFrame
│ Row │ id    │ weight   │ step  │
│     │ Int64 │ Float64  │ Int64 │
├─────┼───────┼──────────┼───────┤
│ 1   │ 2     │ 0.300242 │ 1     │
│ 2   │ 3     │ 0.729594 │ 1     │
│ 3   │ 1     │ 0.145448 │ 1     │
```

Libbum:
8191189 drops `collect_agent_data!` and simplifies the aggregation interface by dispatching on `collect_agent_data::Dict`. Tests all passing for this aspect of the upgrade.

kavir1698:
Thanks, @Datseris . I now get what you meant by using keyword arguments. Indeed it is a very useful practice.

src/simulations/collect.jl
kavir1698: Right. I hadn't tested the method and I overlooked the local scope.

Libbum:
10abe9c updates testing around the CA modules. I've set one test to *broken*, since the size I'm seeing now is different. Perhaps I've refactored incorrectly. @kavir1698, perhaps you can check that?
In general, these files need to be updated regardless, and #128 is also on the way, so didn't want to spend too much time there.

572ff6c exports our `collect_*_data` function and implements a few initial tests for the low and high level collection API. Let's discuss some more as we go along, but for the moment these seem to fit our requirements.

Datseris:
Alright, @kavir1698 and @Libbum , this is finished (from API part). Everything works and I've updated the Schelling example to also work. The only thing left is to update the tests and the rest of the examples, which I'd appreciate if you could do it instead of me. Gotta take some time for other manners.

All in all, the lesson to be learned here is: if you take a step back and think about how the implementation should be, and only then code it, you'll get much better results! I think the data collection we have right now is truly beautiful:

* very small code
* keyword propagation
* all keywords are defined in one, and only one point (`_run!`)
* Huge performance benefits by not making dataframes a thousand times over

Another lesson to be learned is: **study the Julia performance tips** : https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-tips-1

Libbum:
Something has gone pretty wrong with these changes though.
`test/benchmark/collection.jl` benchmarking `run!`:

At 3904952
```
BenchmarkTools.Trial:·
  memory estimate:  15.33 MiB
  allocs estimate:  181639
  --------------
  minimum time:     216.462 ms (0.00% GC)
  median time:      221.095 ms (0.00% GC)
  mean time:        221.368 ms (0.50% GC)
  maximum time:     232.810 ms (0.00% GC)
  --------------
  samples:          23
  evals/sample:     1
```

At 7961853 
```
BenchmarkTools.Trial: 
  memory estimate:  1.01 GiB
  allocs estimate:  46029885
  --------------
  minimum time:     3.019 s (1.90% GC)
  median time:      3.027 s (1.86% GC)
  mean time:        3.027 s (1.86% GC)
  maximum time:     3.034 s (1.81% GC)
  --------------
  samples:          2
  evals/sample:     1% 
```

I'll start rolling back and see where the bottleneck is.

Datseris:
How can you benchmark data aggregation, since data aggregation was not implemented yet? :S 

Damn its really sad to see those numbers, something must be very off. It makes little sense that we have _more_ allocations. Probably there is a type instability somewhere.

Libbum:
Oh, sorry, didn't realise we hadn't got that far. 

Regardless, there's a big jump for the non-aggregate too. I'll dig a little.

Datseris:
I guess you benchmarked the old aggregation function with the dict stuff. Okay then we have to look into it in detail to see what is going on. I just removed discrete space from the benchmark. Let's try isolate what we benchmark as much as possible.

test/collect_tests.jl
```diff
@@ -94,25 +102,37 @@ end
     daily_agent_aggregate = init_agent_dataframe(model, agent_agg)
     yearly_agent_data = init_agent_dataframe(model, agent_props)
 
-    for year in 1:5
-        for day in 1:365
+    for year = 1:5
```
Datseris, line 105: why change `in` to `=`?

Datseris: if you don't like `in` use the `\in<TAB>`, but not equals. Equals should be reserved for assignement.

Datseris:
I see you use the formatter. great! I only dislike changing `in` to `=`.

Libbum:
I think I can turn that off. Or I can swap it back manually. `in` is nicer mostly, I agree.

test/collect_tests.jl
Libbum: That was the formatter, I've changed the defaults and reverted this already. Will push it soon.

src/simulations/collect.jl
```diff
@@ -140,10 +140,10 @@ function init_agent_dataframe(model::ABM, properties::AbstractArray)
 end
 
 function collect_agent_data!(df, model, properties::Vector, step::Int=0)
-    alla = allagents(model)
+    alla = sort(collect(values(model.agents)), by=a->a.id)
```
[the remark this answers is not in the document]
Libbum: It seemed to be faster in benchmarks. But I didn't try `sort!`, will take a look there.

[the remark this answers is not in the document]
Datseris: Okay, if it is faster then we could/should collect, but `sort!` is definitely better here.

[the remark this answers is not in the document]
Datseris: ( you can only `sort!` the collect part)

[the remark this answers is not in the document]
Libbum:
```
@btime alla = sort(collect(values(model.agents)), by=a->a.id)
  59.685 μs (5 allocations: 19.97 KiB)
  
@btime alla = sort!(collect(values(model.agents)), by=a->a.id)
  58.511 μs (4 allocations: 12.03 KiB)
  
@btime alla = values(sort(model.agents))
  88.709 μs (35 allocations: 94.83 KiB)
```
Without collecting `sort` fails on `values(...)`. So yep, `sort!` is mildly quicker, but keeping collect is worth it too.

Datseris:
@Libbum this is good to go, correct? (after a final review from me)

Libbum:
Essentially. I'm adding a few extra tests just to be sure. After that it's good from my end.

PR #195, opened 2020-04-05
Change biggest_id to next_id

itsdfish:
This PR changes `biggest_id` to `next_id` because all uses of `biggest_id` add 1 to the largest id to obtain the next id. This is also useful when creating models with heterogeneous agents because the ids need to be managed by the user when new agents are added dynamically throughout the model run.


Datseris:
Can you do the following please:

1. rename `next_id` to `nextid` 
2. Add a simple documentation string to it
3. export it
4. add it to the API.md file in the `docs/src` folder. Under the seciton agent information and retrieval

itsdfish:
I also updated the changelog. It wasn't clear to me where to add nextid to the API.md file. It didn't fit exactly in any section, but I didn't want to create a new section with only one entry.

PR #197, opened 2020-04-07
Fix a bug in sample!

### LINKED ISSUE #196: Problems with the `sample!` function
The sample function does not do what it claims.

1. `sample!` with `replace` is not tested.
2. `sample!` with `replace` does not work because:
```julia
    for (index, id) in enumerate(newids) # add new agents while adjusting id
        model.agents[index] = deepcopy(model[id])
        model.agents[index].id = index
    end
```
thus an agent that has been selected twice is only added once in the model.

3. The design of the `sample!` function has bad performance. It removes and re-adds everything from scratch, but I don't understand why. Why not just kill the agents that were not selected, and add any additional extra ones (e.g. add the extra duplicates if `resample` , or just add the extra created agents if `n > nagents(model)` ?

kavir1698:
Closes #196

### DIFF
--- a/src/simulations/sample.jl
+++ b/src/simulations/sample.jl
@@ -18,8 +18,8 @@ be chosen more than once.
 
 See the Wright-Fisher example in the documentation for an application of `sample!`.
 """
-function sample!(model::ABM, n::Int, weight=nothing; replace=true,
-  rng::AbstractRNG=Random.GLOBAL_RNG)
+function sample!(model::ABM{A, S}, n::Int, weight=nothing; replace=true,
+  rng::AbstractRNG=Random.GLOBAL_RNG) where{A, S}
   
   org_ids = collect(keys(model.agents))
   if weight != nothing
@@ -29,6 +29,7 @@ function sample!(model::ABM, n::Int, weight=nothing; replace=true,
     newids = sample(rng, org_ids, n, replace=replace)
   end
 
+  add_function = hasfield(A, :pos) ? add_agent_pos! : add_agent!
   nextid = maximum(keys(model.agents)) + 1
   for id in org_ids
     if !in(id, newids)
@@ -38,7 +39,7 @@ function sample!(model::ABM, n::Int, weight=nothing; replace=true,
       for t in 2:noccurances
         newagent = deepcopy(model.agents[id])
         newagent.id = nextid
-        add_agent_pos!(newagent, model)
+        add_function(newagent, model)
         nextid += 1
       end
     end
--- a/src/simulations/sample.jl
+++ b/src/simulations/sample.jl
@@ -7,7 +7,8 @@ using StatsBase: sample, Weights
 Replace the agents of the `model` with a random sample of the current agents with
 size `n`.
 
-Optionally, choose an agent property `weight` (Symbol) to weight the sampling.
+Optionally, provide a `weight`: Symbol (agent field) or function (input agent
+out put number) to weight the sampling.
 This means that the higher the `weight` of the agent, the higher the probability that
 this agent will be chosen in the new sampling.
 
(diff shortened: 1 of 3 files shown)


src/simulations/sample.jl
```diff
@@ -13,51 +13,34 @@ this agent will be chosen in the new sampling.
...
-        model.agents[index].id = index
-    end
-    # kill extra agents
-    if n < nagents(model)
-        genocide!(model, n)
+      noccurances = count(x->x==id, newids)
+      for t in 2:noccurances
+        newagent = deepcopy(model.agents[id])
+        newagent.id = nextid
+        add_agent_pos!(newagent, model)
```
Libbum, line 41: Tests are failing at this point. The test `Agent2` has no position.

Libbum: Could add a gate here with something like `hasfield(typeof(newagent),:pos)`?

Libbum:
Whilst we're fixing bugs in `sample!`, we should also add tests and logic to cover the case when a user attempts to sample from a pool of zero agents:

```julia
julia> model2 = ABM(Agent3, GridSpace((10, 10)))
AgentBasedModel with 0 agents of type Agent3
 space: GridSpace with 100 nodes and 180 edges
 scheduler: fastest

julia> sample!(model2, 10, :weight)
ERROR: MethodError: no method matching zero(::Type{Any})
```

Datseris:
We should add tests for `sample!` with replace in this pr as well.

kavir1698:
> nagents(model) > 0 || throw(ArgumentError("Model must have at least 1 agent to sample from."))
I think it's better to return nothing

Libbum:
Huh. We have uncovered another bug. There is no ability to add an instantiated agent without a space.

kavir1698:
> We should add tests for sample! with replace in this pr as well.

sampling with replacement is the default behavior. So current tests cover it.

kavir1698:
> Huh. We have uncovered another bug. There is no ability to add an instantiated agent without a space.

Are you talking about a model with `Agent2`? I don't receive an error. I am not on master, though.

Datseris:
`hasfield(A, :pos)` can't be used for mixed agent models.

Libbum:
> I don't receive an error.

That's odd. You should! I do when I check out this branch locally, [so does Travis](https://travis-ci.com/github/JuliaDynamics/Agents.jl/jobs/315941841#L359-L361).

kavir1698:
> That's odd. You should! I do when I check out this branch locally, so does Travis.

Maybe it's just because I am using the stable version. I will push a commit with a new `add_agent!` defined.

Datseris:
I've fixed the missing method. But @kavir1698 the tests still fail both locally and oline for other reason, I think you need to have a second look at the new `sample!` (some Keys are wrong).

Libbum:
I think this might actually be a test design flaw, more than the `sample!` implementation.

line 153 in api_tests.jl
```julia
julia> model.agents[1].weight = 2
ERROR: KeyError: key 1 not found

julia> model.agents
Dict{Int64,Agent2} with 12 entries:
  2  => Agent2(2, 3.0)
  11 => Agent2(11, 0.573621)
  21 => Agent2(21, 2.13182)
  7  => Agent2(7, 1.62655)
  26 => Agent2(26, 3.955)
  25 => Agent2(25, 3.955)
  10 => Agent2(10, 0.691542)
  22 => Agent2(22, 2.13182)
  24 => Agent2(24, 3.955)
  4  => Agent2(4, 0.987061)
  23 => Agent2(23, 3.955)
  15 => Agent2(15, 3.955)
```

Since sample! is not expected to re-align ids, I'll change these tests to take random agents instead of specific ones.

kavir1698:
> Don't even make sense? Should we just drop these?

These were to catch a bug that was caused by not using `deepcopy`. In their current form and the new code that changes all agent ids, they don't make sense anymore.

PR #202, opened 2020-04-08
allow both metrics

### LINKED ISSUE #201: Properly use metric in continuous space
(now we use a mix of cityblock and euclidean. E.g. nearest_neighbor uses Euclidean, while everything else uses cityblock)

Datseris:
Closes #201

TODO: Add tests

### DIFF
--- a/src/core/continuous_space.jl
+++ b/src/core/continuous_space.jl
@@ -21,35 +21,31 @@ end
 const COORDS = 'a':'z' # letters representing coordinates in database
 
 """
-    ContinuousSpace(D::Int [, update_vel!]; periodic::Bool = true, extend = nothing, metric = "cityblock")
+    ContinuousSpace(D::Int [, update_vel!]; kwargs...)
 Create a `ContinuousSpace` of dimensionality `D`.
 In this case, your agent positions (field `pos`) should be of type `NTuple{D, F}`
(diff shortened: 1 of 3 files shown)


src/core/continuous_space.jl
```diff
@@ -246,15 +246,19 @@ export nearest_neighbor, elastic_collision!, interacting_pairs
...
+space's metric. Valid only in continuous space.
+Return `nothing` if no agent is within distance `r`.
 """
 function nearest_neighbor(agent, model, r)
   n = space_neighbors(agent, model, r)
   length(n) == 0 && return nothing
   d, j = Inf, 0
   for i in 1:length(n)
-    @inbounds dnew = sqrt(sum(abs2.(agent.pos .- model[n[i]].pos)))
+    if model.space.metric == "euclidean"
```
Libbum, line 257: Any reason `metric` is a string? Would it not be better as a `Symbol` or abstract type?

Datseris: Yes symbol is better. I am debating whether it is worth it for us to add `Distances` as a dependency and have the metric be `<: Metric`...

Datseris: But unfortunately `Distances` doesn't support Tuples, which is what we use here, so this idea is out of the window.

PR #205, opened 2020-04-09
Keyword changes and CHANGELOG

Libbum:
Housecleaning aspects for the alterations #191.

### DIFF
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -1,6 +1,6 @@
 export run!, collect_agent_data!, collect_model_data!,
        init_agent_dataframe, init_model_dataframe, aggname
-       
+
 ###################################################
 # Definition of the data collection API
 ###################################################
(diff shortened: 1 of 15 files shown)


Datseris:
Maybe we should also mention the new keyword naming in the change log. And add an entry in the "additions" that model data collection is now possible.

src/simulations/collect.jl
```diff
@@ -19,21 +19,21 @@ data specified by the keywords, explained one by one below. Return the data as
...
   then the data for this entry is agent's field `weight`. If an entry is a `Function`, e.g.
   `f`, then the data for this entry is just `f(a)` for each agent `a`.
   The resulting dataframe columns are named with the input symbol (here `:weight, :f`).
 
-* `agent_properties::Vector{<:Tuple}`: if `agent_properties` is a vector of 2-tuples instead,
+* `acollect::Vector{<:Tuple}`: if `acollect` is a vector of 2-tuples instead,
   data aggregation is done over the agent properties. For each 2-tuple, the first entry
   is the "key" (any entry like the ones mentioned above, e.g. `:weight, f`). The second
   entry is an aggregating function that aggregates the key, e.g. `mean, maximum`. So,
   Continuing from the above example, we would have
```
Datseris, line 31: Continuing -> continuing

Datseris:
Can we do even better and use `adata, mdata`, or am I getting too greedy?

Libbum:
> Can we do even better and use adata, mdata, or am I getting too greedy?

I'd be fine with this. It's something users will use continuously & its around the docs, examples and tests everywhere.

Datseris:
@kavir1698 what do you think?

Datseris:
I added a small safety / depwarning so that `agent_properties` still works.

kavir1698:
@Datseris I agree with this change. Maybe explain what `adata` and `mdata` stand for on their first use in the documentation?

PR #207, opened 2020-04-11
Add kwargs to add_agent! for continuous space

itsdfish:
I didn't realize there were separate methods for continuous space models. This PR adds kwargs to those methods and adds a test.


Datseris:
Yeap there are different methods! The reason one needs to use `add_agent!` instead of just adding an agent to the dictionary of `model`, is because one also needs to add the agent into the space. And for continuous space this process is different!

PR #213, opened 2020-04-12
Doc update

itsdfish:
This PR notes in the doc string to use StaticArrays when tuples are not supported for linear algebra operations.


src/core/continuous_space.jl
```diff
@@ -41,6 +41,9 @@ By default no update is done this way.
   periodicity occurs. If `periodic ≠ true`, `extend` is only used at plotting.
 * `metric = :cityblock` : metric that configures distances for finding nearest neighbors
   in the space. The other option is `:euclidean` but cityblock is faster (due to internals).
+
+Note: if your model requires linear algebra operations for which tuples are not supported, a performant solution is to convert
+to convert the tuple to a SVector defined in StaticArrays.jl.
```
Datseris, line 46: "to convert" is repeated twice :) . It is also useful to show that one can convert directly: `s = SVector(x)` and `x = Tuple(s)`.

itsdfish:
Of course I failed at something simple =). I fixed the typo and added the example.

PR #214, opened 2020-04-12
Flock model

itsdfish:
Here is the flocking model based on Mesa. Please review to ensure it is correct and feel free to add changes as you see fit.

I'm not sure whether the documentation will render correctly. I am not aware of an easy way to test it locally.


examples/Flock.jl
```diff
@@ -0,0 +1,130 @@
...
+# The contribution of each rule defined above recieves an importance weight: `cohere_factor` is the importance of maintaining the average position of neighbors,
+# `match_factor` is the importance of matching the average trajectory of neighboring birds, and `seperate_factor` is the importance of maining the minimum
+# distance from neighboring birds.
+#
+# The following defines a Bird constructor that uses keyword arguments and default values.
+function Bird(id, pos; vel=(0.0,0.0), speed=1.0, cohere_factor=.25, separation=4.0,
+        seperate_factor=.25, match_factor=.01, visual_distance=5.0)
+    return Bird(id, pos, vel, speed, cohere_factor, separation, seperate_factor,
+        match_factor, visual_distance)
+end
```
Datseris, line 40: Why do this? Is there any benefit? It makes the code bigger. In the following `initialize_model` function, the fields are known.

examples/Flock.jl
```diff
@@ -0,0 +1,130 @@
...
+# can be overwritten by passing keyword arguments.
+function initialize_model(;n_birds=100, speed=1.0, cohere_factor=.25, separation=4.0,
+    seperate_factor=.25, match_factor=.01, visual_distance=5.0, dims=(100,100))
+    space2d = ContinuousSpace(2; periodic=true, extend=dims)
+    model = ABM(Bird, space2d, scheduler=random_activation)
+    for _ in 1:n_birds
+        vel = Tuple(rand(2)*2 .- 1)
+        add_agent!(model; vel=vel, speed=speed, cohere_factor=cohere_factor,
+        separation=separation, seperate_factor=seperate_factor, match_factor=match_factor,
+         visual_distance=visual_distance)
```
Datseris, line 53: Just have a look here. Every word is literaly repeated twice, yet there is no benefit in it.

examples/Flock.jl
itsdfish: kwargs are useful if you want to generate objects with default values and modify default values without remembering the positions. I removed that in the latest commit.

Datseris:
Ali here is the flock example. Let us know if it is sufficiently similar with Mesa so that you can use it as a comparison.

kavir1698:
I'll check it ASAP.

Datseris:
@itsdfish please change all comments that are inside code blocks (e.g. comments inside functions) to start with `##` instead of just `#`. Otherwise we can't use the examples in the documentation.

itsdfish:
Sorry. It should be fixed now.

PR #222, opened 2020-04-14
Daisyworld

### LINKED ISSUE #221: Reducing over an empty collection is not allowed
We've got a problem in the data collector when we try to save data from a set of 0 agents.

Suppose we have

```julia
julia> model.agents
Dict{Int64,Agent} with 5 entries
...
```
and `run!` with an agent aggregation function, say `(:weight, sum)`.

If `agent_step!` ends up using `kill_agent!()` or `genocide!()` such that
```julia
julia> model.agents
Dict{Int64,Agent} with 0 entries
```
at some point, we'll hit the `[1] _empty_reduce_error() at ./reduce.jl:295`.

Libbum:
Reasons:
1. It's pretty awesome
2. Closes #221 via discussion in `run!`

### DIFF
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -33,19 +33,9 @@ two `DataFrame`s, one for agent-level data and one for model-level data.
   use the function [`aggname`](@ref), and create something like `mean(weight)` or
   `maximum(f)`. This name doesn't play well with anonymous functions!
 
-  **A note on aggregation with models containing zero agents**
-
-  If a model uses [`kill_agent!`](@ref) or [`genocide!`](@ref), care must be taken when
-  attempting to collect aggregated data. When there are zero agents in the model
-  it is very hard (if not impossible) to know what the return type should be for a
-  general function. Therefore, it is best practice to provide a default return type
-  in this scenario.
-
-  For example, the `sum` over a variable of type `Float64`:
-
-  ```julia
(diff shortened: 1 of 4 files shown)


Datseris:
@Libbum you have write acccess here, the next PR you do please do it from the `origin` upstream. This we can we preview documentation on PRs!

Libbum:
Ah, that's why I couldn't see it! No worries.

I've built the docs locally and they run fine.

src/simulations/collect.jl
```diff
@@ -33,6 +33,20 @@ two `DataFrame`s, one for agent-level data and one for model-level data.
...
+  it is very hard (if not impossible) to know what the return type should be for a
+  general function. Therefore, it is best practice to provide a default return type
+  in this scenario.
+
+  For example, the `sum` over a variable of type `Float64`:
+
+  ```julia
+  total(v) = length(v) == 0 ? 0.0 : sum(v)
+  ```
+
```
Datseris, line 49: I think this is too much text and every word in a docstring should be counted as if it weights gold! I propose:

> **Notice:** Aggregating only works if there are agents to be aggregated over. If you remove agents during model run, you should modify the aggregating functions. E.g. instead of passing `mean`, pass `mymean(a) = isempty(a) ? 0.0 : mean(a)`.

Issue #220, opened 2020-04-14
Function to get all pairs within interaction radius


lhupe:
@Datseris asked me to implement a simple model for growing dense bacterial colonies to test the usability of the new continuous space API.

In a dense bacterial colony (or indeed any dense granular system), a single agent can interact mechanically with several other agents at the same time. However, `interacting_pairs` only returns pairs of nearest neighbours. It would be nice to have a function that actually returns **all** pairs of agents within one interaction radius of each other.

Datseris:
To be clear, `space_neighbors` is not good for your use case because for each interacting pair you only want to do the interaction **once**, correct? Having a look at the source code of `interacting_pairs`, it seems possible, to create a similar function that doesn't use `nearest_neighbor` but `space_neighbors` directly. In that source I have to keep track the pair IDs manually anyway, so it should be that much harder.

Datseris:
@lhupe how do you imagine the API of the function that you ask for here? I am having trouble to imagine what it should do and what it should return. Actually, I am not yet sure what are the input arguments.

lhupe:
@Datseris I think there was some confusion on my part about the exact function of `interacting_pairs!`. 
Does it return an iterator over *all*  pairs of particles that are within one `r` neighborhood of each other or just over pairs of nearest neighbours? 

I am getting some strange behaviour, which originally caused me to assume that the latter case is true. However, on reexamining my results, I find that the behaviour is inconsistent wih either hypothesis.

In the picture below, you can see a configuration of four of my `Agents` (plotted using a custom function). I have visualized the output of `interacting_pairs!`  (with a sufficiently large interaction radius) on this configuration with the connecting lines. As you can see, the connection from the beige to the cyan agent is missing. 

![strangeness](https://user-images.githubusercontent.com/34064555/79311369-9e8ba900-7efd-11ea-8add-979e67e95191.png)

Libbum:
Can you send us an MWE of this four agent cluster? We need to build some tests around this scenario I think. Then we can identify if `interacting_pairs!` needs to be modified (/fixed?) or we require a new method.

Datseris:
Hi Lukas, thanks for showing this, it means that `interacting_pairs` is bugged. You should not have a connection between orange and magenta. Let's see the current source:
```
function interacting_pairs(model, r)
  pairs = Tuple{Int, Int}[]
  for id in keys(model.agents)
    # Skip already checked agents
[...]

PR #224, opened 2020-04-16
fix interact based on #220

Datseris:
Still need to add docs.

### DIFF
--- a/src/core/continuous_space.jl
+++ b/src/core/continuous_space.jl
@@ -315,19 +315,27 @@ function elastic_collision!(a, b, f = nothing)
   return true
 end
 
+# TODO: Add "true_interacting_pairs" where it is guaranteed that each agent
+# is only paired with its true nearest neighbor
+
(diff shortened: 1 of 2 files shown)


src/core/continuous_space.jl
```diff
@@ -315,19 +315,27 @@ function elastic_collision!(a, b, f = nothing)
   return true
 end
 
+# TODO: Add "true_interacting_pairs" where it is guaranteed that each agent
```
Libbum, line 318: We could add a type keyword with [:unique, :single, :all] as options instead of three functions perhaps.

Datseris: This sounds like a good idea to me but I am not sure what `:unique` means. Does it mean the "true" ones?

Libbum: Well, we can deal with semantics later. I just added that since you put "that yields *unique* pairs" in the new doc string.

Datseris:
I don't have time to do the other 2 methods so I merge this.

PR #225, opened 2020-04-17
Extend interacting pairs

Libbum:
WIP for discussions around #220.

This implements the *all neighbors* scenario at the moment, with a currently unimplemented *true neighbors* to come soon. 

Current benchmarks using test case:
```julia
julia> @btime interacting_pairs(model, 2.0).pairs
  5.489 μs (73 allocations: 3.41 KiB)
2-element Array{Tuple{Int64,Int64},1}:
 (6, 5)
 (4, 3)

julia> @btime interacting_pairs(model, 2.0; method = :all).pairs
  10.959 μs (137 allocations: 6.47 KiB)
5-element Array{Tuple{Int64,Int64},1}:
 (3, 4)
 (4, 5)
 (4, 6)
 (3, 5)
 (5, 6)
```

### DIFF
--- a/src/core/continuous_space.jl
+++ b/src/core/continuous_space.jl
@@ -371,17 +371,47 @@ function scheduler_pairs!(pairs::Vector{Tuple{Int, Int}}, model::ABM, r, schedul
 end
 
 function all_pairs!(pairs::Vector{Tuple{Int, Int}}, model::ABM, r)
-  for id in keys(model.agents)
-    for nid in space_neighbors(model[id], model, r)
+  for a in allagents(model)
+    for nid in space_neighbors(a, model, r)
       # Sort the pair to overcome any uniqueness issues
-      new_pair = isless(id, nid) ? (id, nid) : (nid, id)
+      new_pair = isless(a.id, nid) ? (a.id, nid) : (nid, a.id)
       !(new_pair in pairs) && push!(pairs, new_pair)
     end
   end
 end
 
 function true_pairs!(pairs::Vector{Tuple{Int, Int}}, model::ABM, r)
-    #Unimplemented
+  distances = Vector{Float64}(undef, 0)
+  for a in allagents(model)
+    nid = nearest_neighbor(a, model, r).id
+    nid == nothing && break
+    # Sort the pair to overcome any uniqueness issues
+    new_pair = isless(a.id, nid) ? (a.id, nid) : (nid, a.id)
+    if !(new_pair in pairs)
+      # We also need to check if our current pair is closer to each
+      # other than any pair using our first id already in the list,
+      # so we keep track of nn distances.
+      dist = pair_distance(a.pos, model[nid].pos, model.space.metric)
+
+      idx = findfirst(x->first(new_pair) == x, first.(pairs))
+      if idx == nothing
+        push!(pairs, new_pair)
+        push!(distances, dist)
+      elseif idx != nothing && distances[idx] > dist
+        # Replace this pair, it is not the true neighbor
+        pairs[idx] = new_pair
+        distances[idx] = dist
+      end
+    end
+  end
+end
+
+function pair_distance(pos1, pos2, metric)
+  if metric == :euclidean
+    @inbounds sqrt(sum(abs2.(pos1 .- pos2)))
+  elseif metric == :cityblock
+    @inbounds sum(abs.(pos1 .- pos2))
+  end
 end
 
 struct PairIterator{A}
--- a/src/core/continuous_space.jl
+++ b/src/core/continuous_space.jl
@@ -331,15 +331,33 @@ some pairwise interaction across all pairs of closest agents once
 (and does not want to trigger the event twice, both with `a1` and with `a2`, which
 is unavoidable when using `agent_step!`).
 
-Notice that the pairs are created by scanning each agent according to the given
+Notice that by default the pairs are created by scanning each agent according to the given
 `scheduler`. This is important, because this function does not match each agent with
 its absolute nearest neighbor. Imagine three agents A B C where the
 nearest neighbor of A is B but the nearest neighbor of B is C. If you start with A,
 you get the pair (A, B), but if you start with B you get (B, C).
+
+The keyword argument `method = :scheduler` provides two additional pairing scenarios
+- `:true`: where it is guaranteed that each agent is only paired with its true
+nearest neighbor.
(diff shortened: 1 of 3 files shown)


src/core/continuous_space.jl
```diff
@@ -331,15 +331,33 @@ some pairwise interaction across all pairs of closest agents once
...
+The keyword argument `method = :scheduler` provides two additional pairing scenarios
+- `:true`: where it is guaranteed that each agent is only paired with its true
+nearest neighbor.
+- `:all`: returns every pair of agents within the interaction radius `r` regardless of
+their nearest neighbor status.
 """
-function interacting_pairs(model, r, scheduler = model.scheduler)
-  #TODO: This can be optimized further I assume
+function interacting_pairs(model, r, scheduler = model.scheduler; method = :scheduler)
+  @assert method ∈ [:scheduler, :true, :all]
```
Datseris, line 347: Use `( )` instead of `[ ] `

src/core/continuous_space.jl
```diff
@@ -350,7 +368,20 @@ function interacting_pairs(model, r, scheduler = model.scheduler)
...
   end
-  return PairIterator(pairs, model.agents)
+end
+
+function all_pairs!(pairs::Vector{Tuple{Int, Int}}, model::ABM, r)
+  for id in keys(model.agents)
+    for nid in space_neighbors(model[id], model, r)
+      # Sort the pair to overcome any uniqueness issues
+      new_pair = isless(id, nid) ? (id, nid) : (nid, id)
+      !(new_pair in pairs) && push!(pairs, new_pair)
```
Datseris, line 378: Since you sort the pairs anyway, it might be more performant to use commands like `searchsortedfirst` instead of `in` and check whether `searchsortedfirst` returns length()+1 or something less.

Datseris: Ah no, what I said is incorrect, ou don't sort the pairs, you only sort the ID within the pairs. Sorting the outer vector as well would take too much time.

Libbum: OK, I'll look into that.

Datseris: no reason!

Datseris:
@lhupe can you use this in your example you are working on?

src/core/continuous_space.jl
Libbum: Yeah exactly. Wanted to avoid that.

Datseris:
These benchmarks are not really realistic and some where 100s or 1000s of agents exist in the database should be made. I Think this simple factor of 2 difference ins not true.

Libbum:
Current implementation:

```julia
using Agents, Random
Random.seed!(6548)
space = ContinuousSpace(2, extend = (10, 10), periodic = false, metric = :euclidean)
model = ABM(Agent6, space; scheduler = by_id)
for i in 1:10000                                                                                                                                                            
    add_agent_pos!(Agent6(i, (50*rand(), 50*rand()), (0.0, 0.0), 0), model)                                                                                                               
end 

julia> @btime interacting_pairs(model, 2.0).pairs
  5.346 s (2088735 allocations: 54.53 MiB)
  650-element Array{Tuple{Int64,Int64},1}:
  
julia> @btime interacting_pairs(model, 2.0; method = :true).pairs
  5.491 s (2154985 allocations: 95.69 MiB)
  650-element Array{Tuple{Int64,Int64},1}:
  
julia> @btime interacting_pairs(model, 2.0; method = :all).pairs
  62.462 s (2144994 allocations: 64.81 MiB)
307203-element Array{Tuple{Int64,Int64},1}:
```

src/core/continuous_space.jl
```diff
@@ -371,17 +371,47 @@ function scheduler_pairs!(pairs::Vector{Tuple{Int, Int}}, model::ABM, r, schedul
...
+        pairs[idx] = new_pair
+        distances[idx] = dist
+      end
+    end
+  end
+end
+
+function pair_distance(pos1, pos2, metric)
+  if metric == :euclidean
+    @inbounds sqrt(sum(abs2.(pos1 .- pos2)))
```
Datseris, line 411: `@inbounds` doesn't add anything here, there is no bound check happening. If you want more speed you have to use `@fastmath`.

Libbum: Ah, that was just copied from `nearest_neighbors`. Will check if `@fastmath` helps us out in this circumstance.

Libbum: fastmath doesn't yield any change, but bounds checking is now dropped.

Datseris:
Seems good to me. Let's wait for feedback from @lhupe before merging this.

Datseris:
Oh wow:
```
julia> @btime interacting_pairs(model, 2.0).pairs
  5.346 s (2088735 allocations: 54.53 MiB)
  650-element Array{Tuple{Int64,Int64},1}:
  
julia> @btime interacting_pairs(model, 2.0; method = :true).pairs
  5.491 s (2154985 allocations: 95.69 MiB)
  650-element Array{Tuple{Int64,Int64},1}:
```
if the `true` method is as fast as the scheduled one maybe it should be the default.

Libbum:
The question I have would be: Is there a reason to actually keep `:scheduler`? I don't really see a use case for it, but perhaps I overlook something.

If there's no reason to keep it, we can change the default to the current `:true` implementation, then remove the `method` keyword replacing it with an `all = false` keyword.

lhupe:
This method works perfectly for my model!
Thanks @Libbum!

Datseris:
@Libbum alright, let's go with the change you suggested and make `all = false` the default keyword.

Libbum:
028d2e3 drops the scheduler method as discussed, and fixes a bug when there are no nearest neighbors.

There are currently three broken collision tests to be checked before merging. They pass if we now use `interacting_pairs(; all = true)`, but I'm unsure if this is the behaviour the tests expect, or they if fact rely on the `:scheduler` implementation.

Libbum:
That was a bit pre-emptive. The social distancing example is definitely broken.

Datseris:
damnit

Libbum:
The collision tests are for a small radius and essentially test for the dynamics in the social distancing example. Setting all = true makes everything work again. I can just push that fix directly to master.

Libbum:
a07930a (on master) solves the mixup.

Datseris:
It is definitely incorrect to use `all` in social distancing. The dynamics are billiard ball collisions. It makes no sense for an agent to interact with more than one in one step.

PR #231, opened 2020-04-23
Agent space updates

Libbum:
Some spring cleaning for `core/agent_space_interaction.jl`.
There were a few `TODO`s in the codebase, which are now sorted:
1. An inefficient empty cells check. We already had a more performant version as a function. d7c952c
2. Complete refactor of add_agent!, basing everything on add_agent_pos! low-level methods

On top of that, I've added a few extra guards, narrowed type scope and altered a number of functions such that they are up to date with our current internal methods.

### DIFF
--- a/src/simulations/sample.jl
+++ b/src/simulations/sample.jl
@@ -41,7 +41,7 @@ function sample!(model::ABM{A, S}, n::Int, weight=nothing; replace=true,
       for t in 2:noccurances
         newagent = deepcopy(model.agents[id])
         newagent.id = n
-        add_agent_pos!(newagent, model)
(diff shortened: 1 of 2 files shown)


src/simulations/sample.jl
```diff
@@ -41,7 +41,7 @@ function sample!(model::ABM{A, S}, n::Int, weight=nothing; replace=true,
       for t in 2:noccurances
         newagent = deepcopy(model.agents[id])
         newagent.id = n
-        add_agent_pos!(newagent, model)
+        add_agent!(newagent, model)
```
Datseris, line 44: why was this line changed?

PR #233, opened 2020-04-23
add wolf example

itsdfish:
Here is the wolf sheep example. It is very close to complete. 

The main problem is that Mesa seems to produce more qualitative patterns in the results. Both typically result in all animals dying. However, Mesa has more runs in which sheep will live after the wolves die. I'm not sure why this occurs. 

I can have another look this weekend, unless someone else gets to it first.


Datseris:
@itsdfish I've added a different version of `by_breed` which is much simpler than your original version. In the commit message I wrote that it is more performant, but I am actually not sure about it anymore. Didn't test them. But since it is an example, clearer version is preferred.

Also, you can generalize this method to a model with arbitrary mixed agent types. If you want to try and generalize it we could provide `by_type` as a default scheduler.

itsdfish:
I tried generalizing this before, but ran into some issues. The following code is close, but I can't figure out how to fix [union_types](https://stackoverflow.com/questions/31643492/is-it-possible-to-get-collection-of-types-what-form-the-union)

```
union_types(x::Union) = union_types(x.a, x.b)
union_types(x::Type) = (x,)
union_types(a::Type, b::Union) = (a, union_types(b)...)
union_types(a::Union, b::Type) = (b, union_types(a)...)

function by_breed(model::ABM{A}) where {A}
    types = collect(union_types(A))
    c = [Int[], Int[], Int[]] # first=wolf, second=sheep
    for a in allagents(model)
        j = findfirst(x->x == typeof(a), types)
        push!(c[j], a.id)
    end
    for i in c; shuffle!(i); end
    shuffle!(c)
    return vcat(c...)
end
```
The problem with union_types is that `u = Union{...}`, u.a or u.b can be a union of a datatype. Strangely it's not consistent. 

Perhaps there is a better way to get an array of types from a union but I couldn't find one. Do you have any ideas?

itsdfish:
The problem above was that I needed yet another method. I'm curious if you can find a better approach. 

```
union_types(x::Union) = union_types(x.a, x.b)
union_types(x::Type) = (x,)
union_types(a::Type, b::Type) = (a, b)
union_types(a::Type, b::Union) = (a, union_types(b)...)
union_types(a::Union, b::Type) = (b, union_types(a)...)

function by_type(model::ABM{A}) where {A}
    types = collect(union_types(A))
    c = [Int[] for t in types]
    for a in allagents(model)
        j = findfirst(x->x == typeof(a), types)
        push!(c[j], a.id)
    end
    for i in c; shuffle!(i); end
    shuffle!(c)
    return vcat(c...)
end
```

examples/sheep_wolf.jl
```diff
@@ -0,0 +1,164 @@
...
+    n_grass = count(x->(typeof(x) == Grass) && x.fully_grown, agents)
+    push!(df, [step n_sheep n_wolves n_grass])
+end
+
+
+n_steps = 500
+Random.seed!(23182)
+model = initialize_model()
+results = DataFrame(step=Int[], n_sheep=Int[], n_wolves=Int[], n_grass=Int[])
+step!(model, agent_step!, x->model_step!(x, results), n_steps)
```
Libbum, line 158: This is not needed anymore with the new data collection system. You can replace this highlighted section with the following:

```julia
n_steps = 500
Random.seed!(23182)
model = initialize_model()
sheep(a) = typeof(a) == Sheep
wolves(a) = typeof(a) == Wolf
grass(a) = typeof(a) == Grass && a.fully_grown
results, _ = run!(model, agent_step!, n_steps; adata = adata)
```

examples/sheep_wolf.jl
```diff
@@ -0,0 +1,164 @@
...
+        push!(c[j], a.id)
+    end
+    for i in c; shuffle!(i); end
+    shuffle!(c)
+    return vcat(c...)
+end
+
+function initialize_model(;n_sheep=100, n_wolves=50, dims=(20,20), regrowth_time=30,
+    Δenergy_sheep=4, Δenergy_wolf=20, sheep_reproduce=.04, wolf_reproduce=.05)
+    space = GridSpace(dims)
```
Libbum, line 41: `space = GridSpace(dims, moore = true)`

At the moment, your random walk is just up, down, left, right.

examples/sheep_wolf.jl
```diff
@@ -0,0 +1,164 @@
...
+    end
+end
+
+function reproduce!(agent, model)
+    agent.energy /= 2
+    id = nextid(model)
+    energy = rand(1:agent.Δenergy*2) - 1
+    A = typeof(agent)
+    spawn = A(id, (0,0), energy, agent.reproduction_prob, agent.Δenergy)
+    add_agent!(spawn, model)
```
Libbum, line 139: It looks like the mesa example spawns lambs/cubs at the position of the mother. This here will place them randomly. Swap to:

```julia
spawn = A(id, agent.pos, energy, agent.reproduction_prob, agent.Δenergy) 
add_agent_pos!(spawn, model)
```

examples/sheep_wolf.jl
itsdfish: Thank you. I pushed the requested changes. 

I was wondering if there is a reason for using `function(variable)` as the default naming scheme? Unfortunately, this requires wrapping the variable in Symbol and cols in order to plot with StatsPlots: `cols(Symbol("count(sheep)"))`. It seems like `:function_variable` would be more concise.

Libbum: There's also [aggname](https://juliadynamics.github.io/Agents.jl/dev/api/#Agents.aggname) to help with this.

Exactly why we're using this convention&mdash;I'm unsure. It's probably a relic from the old data collection methods. We could have a discussion about that.

itsdfish: In that case, I'll open a separate issue.

Datseris:
So after these changes, are the two examples Mesa/Agents the same?

Libbum:
Almost (I think). 
- `grass.countdown = grass.regrowth_time` in `eat!(Sheep...)` doesn't seem to need to be there.
- energy checks for Wolf and Sheep (in their `agent_step!`) are `<=`, whereas Mesa has `<`.
- The above could be connected to how energy is calculated: `rand(1:(Δenergy_sheep * 2)) - 1`. Mesa uses `self.random.randrange(2 * self.wolf_gain_from_food)`. I don't have any of this installed, so can't verify if this is the same, or slightly off.

One thing additional I did notice just now:
`reproduce!` rolls a new random energy value for the offspring. Mesa imparts the same energy as the mother. Actually the only difference between the offspring and the mother is the `id` (`agent.energy /= 2` is set just previous to the 'birth')

Datseris:
I don't see a reason to make it **exactly** the same with Mesa. As long as our example is clear, we don't care about Mesa that much right?

itsdfish:
> * `grass.countdown = grass.regrowth_time` in `eat!(Sheep...)` doesn't seem to need to be there.

I think that is correct. From what I can tell, grass can only be eaten if it is fully grown. If it is fully grown, the sheep will eat it, and the countdown is reset to the regrowth time. Does that seem correct to you? 



> * energy checks for Wolf and Sheep (in their `agent_step!`) are `<=`, whereas Mesa has `<`.

I will correct this. 



> * The above could be connected to how energy is calculated: `rand(1:(Δenergy_sheep * 2)) - 1`. Mesa uses `self.random.randrange(2 * self.wolf_gain_from_food)`. I don't have any of this installed, so can't verify if this is the same, or slightly off.

This should produce the same range as Python: 0 to 2 * self.wolf_gain_from_food - 1



> don't see a reason to make it **exactly** the same with Mesa.

I agree. There are diminishing returns for making it exactly like Mesa. It might be close enough with the fixes I pushed.

Libbum:
I'm working on some improvements to the `by_type` scheduler, but they seem to be minor from the current implementation. We can probably open another issue for that though.

Datseris:
Tim I leave it to you to merge this as/when you see fit.

Libbum:
Todo:

- [ ] Inline documentation, explanation of the hows and whys.
- [ ] Add to examples in docs
- [ ] Perhaps some better figures

I'm fine taking a look at the second two, but would appreciate a start on the first one.

itsdfish:
I can complete the first point tomorrow morning.

Libbum:
image

I've tweaked some settings, refactored a few things and used the scheduler in #237. Will push the changes up once we have the `by_type` merged.

examples/predator_prey.jl
```diff
@@ -1,3 +1,28 @@
+# # Model of predator-prey dynamics
+
+# The predator-prey model emmulates the population dynamics of predator and prey animals who live in
+# a common ecosystem and compete over limited resources. This model is an agent-based analog to the
+# classic [Lotka-Volterra](https://en.wikipedia.org/wiki/Lotka%E2%80%93Volterra_equations) differential
+# equation model. This example illustrates how to develop models heterogeneous agents.
```
Datseris, line 6:
```suggestion
# equation model. This example illustrates how to develop models with heterogeneous agents.
```

examples/predator_prey.jl
```diff
@@ -148,6 +199,10 @@ grass(a) = typeof(a) == Grass && a.fully_grown
 adata = [(sheep, count), (wolves, count), (grass, count)]
 results, _ = run!(model, agent_step!, n_steps; adata = adata)
 
+# The plot shows the population dynamics over time. Initially, wolves become extent because they
```
Datseris, line 202:
```suggestion
# The plot shows the population dynamics over time. Initially, wolves become extinct because they
```

itsdfish:
I added inline documentation. I also thought predator_prey might be a better description of the model, but feel free to change. I tried incorporating the new scheduler, but git gave an error telling me it is not something I can merge.

Libbum:
@itsdfish, just waiting for CI to build, then I'm ready to merge this. I've made some minor chages&mdash;let me know if there's anything I've done you're not in agreement with.

itsdfish:
I think it looks good. Thanks for your help.

Datseris:
Should we add a simple plot here? with `plotabm`? A vector of agents is given, each can be a scatter plot with offset depending on type. Or one can be a much larger marker depending on type.

Libbum:
Sure, I can take a look at that.

Issue #235, opened 2020-04-24
by_type scheduler


itsdfish:
Here is my attempt to create a by_type scheduler. The main problem is that there does not appear to be an elegant way to extract types from a union. union_types was modified from a solution on stackoverflow.  

```
union_types(x::Union) = union_types(x.a, x.b)
union_types(x::Type) = (x,)
union_types(a::Type, b::Type) = (a, b)
union_types(a::Type, b::Union) = (a, union_types(b)...)
union_types(a::Union, b::Type) = (b, union_types(a)...)

function by_type(model::ABM{A}) where {A}
    types = collect(union_types(A))
    c = [Int[] for t in types]
    for a in allagents(model)
        j = findfirst(x->x == typeof(a), types)
        push!(c[j], a.id)
    end
    for i in c; shuffle!(i); end
    shuffle!(c)
    return vcat(c...)
end
```

_Originally posted by @itsdfish in https://github.com/JuliaDynamics/Agents.jl/pull/233#issuecomment-618682446_

Libbum:
It seems pretty performant if not 'elegant'. The only other example I found in the wild is an overload of `collect`:

```julia
Base.collect(t::Union{Type, DataType, Union{}}) = _collect(t, [])
_collect(t::Type, list) = t<:Union{} ? push!(list, t) : _collect(t.b, push!(list, t.a))
_collect(t::Union{DataType,Core.TypeofBottom}, list) = push!(list, t)
```

And comparing the two:
```julia
@btime collect(Union{Wolf, Sheep, Grass})
  330.342 ns (2 allocations: 128 bytes)
 
@btime collect(union_types(Union{Wolf, Sheep, Grass}))
41.291 ns (2 allocations: 144 bytes)
```
Of course for simplicity we could overload `Base.collect` here to work directly with `union_types`.

There are some minor allocation enhancements I'm testing, but we can start a PR around this I think. We'll need to add in documentation, tests etc. An explicit mention about a convention of union order when constructing models will need to be added (if you want your sheep to move first, they must come before wolves in your union type constructor)...

Something else to think about though: this is essentially `random_by_type`. Could we alter the method in such a way that it allows us to invoke `fastest`, `by_id`, `random_activation` etc, with this method only doing the union type sorting?

Datseris:
```julia
Base.collect(t::Union{Type, DataType, Union{}}) = _collect(t, [])
[...]

PR #237, opened 2020-04-24
Mixed method additions

Libbum:
Improve support for mixed models.
- Adds a `by_type` method (as discussed in #235)
- As a result, schedulers have been moved to their own file
- A method to capture types in a `Union` `union_types`
- We now iterate over subtypes of the union and provide warnings/errors for each.

We can bikeshed naming the scheduler functions, there's probably better ones to use.

### DIFF
--- a/src/core/schedule.jl
+++ b/src/core/schedule.jl
@@ -60,7 +60,7 @@ function property_activation(p::Symbol)
 end
 
 """
-    by_type(shuffle)
+    by_type(shuffle::Bool)
 Useful only for mixed agent models using `Union` types.
 Activate agents by type in order of appearance in the `Union`.
 To group by type, but randomize the type order, set `shuffle = true`.
--- a/src/core/schedule.jl
+++ b/src/core/schedule.jl
@@ -60,12 +60,14 @@ function property_activation(p::Symbol)
 end
 
 """
-    by_type(shuffle::Bool)
+    by_type(shuffle_types::Bool, shuffle_agents::Bool)
 Useful only for mixed agent models using `Union` types.
-Activate agents by type in order of appearance in the `Union`.
-To group by type, but randomize the type order, set `shuffle = true`.
+- Setting `shuffle_types = true` groups by agent type, but randomizes the type order.
+Otherwise returns agents grouped in order of appearance in the `Union`.
+- `shuffle_agents = true` randomizes the order of agents within each group, `false` returns
+the default order of the container (equivalent to [fastest](@ref)).
 """
-function by_type(shuffle::Bool)
+function by_type(shuffle_types::Bool, shuffle_agents::Bool)
     function by_union(model::ABM{A,S,F,P}) where {A,S,F,P}
         types = union_types(A)
         sets = [Integer[] for _ in types]
@@ -73,16 +75,18 @@ function by_type(shuffle::Bool)
             idx = findfirst(t -> t == typeof(agent), types)
             push!(sets[idx], agent.id)
         end
-        shuffle && shuffle!(sets)
+        shuffle_types && shuffle!(sets)
+        shuffle_agents && [shuffle!(set) for set in sets]
         vcat(sets...)
     end
 end
 
 """
-    by_type((C, B, A))
+    by_type((C, B, A), shuffle_agents::Bool)
 Activate agents by type in specified order (since `Union`s are not order preserving).
+`shuffle_agents = true` randomizes the order of agents within each group.
 """
-function by_type(order::Tuple{Type, Vararg{Type}})
+function by_type(order::Tuple{Type, Vararg{Type}}, shuffle_agents::Bool)
     function by_ordered_union(model::ABM{A,S,F,P}) where {A,S,F,P}
         types = union_types(A)
         if order != nothing
@@ -94,6 +98,7 @@ function by_type(order::Tuple{Type, Vararg{Type}})
             idx = findfirst(t -> t == typeof(agent), types)
             push!(sets[idx], agent.id)
         end
+        shuffle_agents && [shuffle!(set) for set in sets]
         vcat(sets...)
     end
 end
--- a/src/core/schedule.jl
+++ b/src/core/schedule.jl
@@ -1,0 +1,105 @@
+####################################
+# Schedulers
+####################################
+
+export random_activation, by_id, fastest, partial_activation, property_activation, by_type
+
+"""
+    fastest
+Activate all agents once per step in the order dictated by the agent's container,
+which is arbitrary (the keys sequence of a dictionary).
+This is the fastest way to activate all agents once per step.
+"""
+fastest(model::ABM) = keys(model.agents)
+
+function fastest(sets::Vector{Vector{Integer}})
+    vcat(sets...)
+end
+
+"""
+    by_id
+Activate agents at each step according to their id.
+"""
+function by_id(model::ABM)
+    agent_ids = sort(collect(keys(model.agents)))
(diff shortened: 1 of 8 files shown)


Libbum:
Hmm. Tests are working locally, but failing on travis. Will investigate that tomorrow.

Libbum:
Something must have been wrong with travis yesterday. Builds are fine now.

Datseris:
I think allowing `by_type` to get a `method` is a bit too much complexity. I feel such behavior should really come from a user unrolling their own scheduler.

I propose to change to `by_type(shuffle::Bool = false)` where for `true` it also shuffles the order of the types (as I Did originally) and for false it doesn't, and instead keeps types in order of the Union. We don't care about shuffling the inner vectors, once again, user-specific scheduler is better.

Libbum:
So all you think all `by_type` should do is:
```julia
model = ABM(Union{Agent0, Agent1}; scheduler=by_type)
for id in 1:10
    if rand(Bool)
        a0 = Agent0(id)
        add_agent!(a0, model)
    else
        a1 = Agent1(id, (0, 0))
        add_agent!(a1, model)
    end
end

julia> model.agents
Dict{Int64,Union{Agent0, Agent1}} with 10 entries:
  7  => Agent0(7)
  4  => Agent0(4)
  9  => Agent1(9, (0, 0))
  10 => Agent1(10, (0, 0))
  2  => Agent1(2, (0, 0))
  3  => Agent1(3, (0, 0))
  5  => Agent0(5)
  8  => Agent0(8)
  6  => Agent0(6)
  1  => Agent1(1, (0, 0))

julia> model.scheduler(model)
[7, 4, 5, 8, 6, 9, 10, 2, 3, 1]
```
or the same thing with `by_type(true)`, where it happens that the union order becomes `Agent0, Agent1`

```julia
julia> model.scheduler(model)
[9, 10, 2, 3, 1, 7, 4, 5, 8, 6]
```

?

Datseris:
No I didn't explain it good enough. Something like:
```
function by_type(bool=true)
funciton type_scheduler(model::ABM{A}) where {A}    
types = collect(union_types(A))
    c = [Int[] for t in types]
    for a in allagents(model)
        j = findfirst(x->x == typeof(a), types)
        push!(c[j], a.id)
    end
    if bool; shuffle!(c); end
    return vcat(c...)
end
```

Libbum:
The output I gave as an example is exactly the output one would expect from that implementation?

Datseris:
I guess. Very difficult to tell exactly what's happening, given how mixed the ids are. But The code I pasted is very clear in what it is supposed to do.

Libbum:
Well, that's a trivial implementation, so I can do that. Will be busy with other things for most of today though so this will be updated tomorrow.

Libbum:
(Plans have changed, I have some time now)

In addition to `shuffle`, I think it would be a good idea for users to be able to suggest an `order`&mdash;reason being that `Union`s are not order preserving:
```julia
julia> a = Union{Sheep,Wolf,Grass}
Union{Grass, Sheep, Wolf}
```

API:
```julia
by_type(shuffle::Bool; order::NTuple{Vararg{Type}}=nothing)
...
types = order == nothing ? union_types(A) : order
```

itsdfish:
> ```julia
> by_type(shuffle::Bool; order::NTuple{Vararg{Type}}=nothing)
> ```

My understanding is that the order of types is shuffled when shuffle is true (and the order of ids within a type is also shuffled). What happens if shuffle is true and an order is specified? Should there be a check to prevent such a situation?

Libbum:
Yeah, that's a complication I'm trying to figure out the best solution for. We want to

1. Group by type (default order)
2. Group by type & shuffle
3. Group by type (with set order)

May need a method keyword

Libbum:
I think multiple dispatch is the better solution here.
```julia
by_type(shuffle::Bool)
by_type(order::Tuple{Type, Vararg{Type}})
```
Thoughts?

itsdfish:
As far as I can tell, that looks like it should work. By having two methods, it is no longer possible to shuffle a specified order.  Is there a default behavior for fixed order?

Libbum:
From @Datseris above:
> We don't care about shuffling the inner vectors, once again, user-specific scheduler is better.

This new proposal wont shuffle the internals at all, so if users are wanting that kind of method they need to provide it.

itsdfish:
Ok. I guess I misunderstood. I thought you were trying to implement a hierarchical shuffling that shuffles types and members within types.

Libbum:
f40229f did that, but no longer seems to be the way forward.

Datseris:
the inner vectors are "shuffled" by default, because dictionary keys don't have order, and there is no deterministic way to know which key you will get. So inner shuffle is not necessary. If you want inner sorting by id, its too complicated to add function for that, better to roll specific scheduler.

src/core/schedule.jl
```diff
@@ -0,0 +1,99 @@
...
+    function by_property(model::ABM{A,S,F,P}) where {A,S,F,P}
+        ids = collect(keys(model.agents))
+        properties = [getproperty(model.agents[id], p) for id in ids]
+        s = sortperm(properties)
+        return ids[s]
+    end
+end
+
+"""
+    by_type(shuffle)
```
Datseris, line 63:
```suggestion
    by_type(shuffle::Bool)
```

src/core/schedule.jl
```diff
@@ -0,0 +1,99 @@
...
+        types = union_types(A)
+        sets = [Integer[] for _ in types]
+        for agent in allagents(model)
+            idx = findfirst(t -> t == typeof(agent), types)
+            push!(sets[idx], agent.id)
+        end
+        shuffle && shuffle!(sets)
+        vcat(sets...)
+    end
+end
```
Datseris, line 79: If Unions are not order preserving, then what does this method achieve? if `shuffle=false`, is it possible that I get different orderings if I call the scheduler with the same model?

Libbum: They're not unordered in the sense of a dict, they alphabetise their types.
`Union{B, C, A} -> Union{A, B, C}` always (at least from my testing). So you'll get repeatable groups when `shuffle=false`, just perhaps not the same order as you input your agents.

If you require `Wolf -> Sheep -> Grass`, you cannot initialise with `Union{Wolf, Sheep, Grass}` and hope the scheduler will give you that. It will always return `Grass -> Sheep -> Wolf` (i.e. alphabetical).

Hence the order method.

itsdfish:
> the inner vectors are "shuffled" by default, because dictionary keys don't have order, and there is no deterministic way to know which key you will get. So inner shuffle is not necessary.

This is partially true. However, dictionaries iterate in the same order each time:

```
julia> d = Dict(1=>"a",2=>"b", 3=>"c")
Dict{Int64,String} with 3 entries:
  2 => "b"
  3 => "c"
  1 => "a"

julia> for (k,v) in d
           println(k, "=>", v)
       end
2=>b
3=>c
1=>a

julia> for (k,v) in d
           println(k, "=>", v)
       end
2=>b
3=>c
1=>a

```
So  it introduces systematic variability into the simulation.

src/core/schedule.jl
Datseris: Okay, I get it.

Datseris:
Okay then. Let's add two arguments to `by_type`, one for sorting the outer types and one for shuffling the inner vectors.

Libbum:
We now have `shuffle_type::Bool` or `order::Tuple` methods, both with an additional parameter: `shuffle_agent::Bool`.

PR #239, opened 2020-04-28
Re-enable codecov

Libbum:
1c97912e689d2 seemed to kill off codecov, which was using the old way of integrating coverage anyhow. 

From the looks of it, everything is still set up to have it running, it's just disabled currently.


Libbum:
Hmm. I don't think I needed to add the `after_success` clause - the integration was just a little laggy.

PR #240, opened 2020-04-28
New syntax for aggname

### LINKED ISSUE #234: Discussion of naming convention for collected data
I was wondering whether it would be possible or desired to use `function_variable` rather than `function(variable)` as the naming convention of creating variables for data collection. For example, in the sheep wolf example:

```
results, _ = run!(model, agent_step!, n_steps; adata = adata)

pyplot()
@df results plot(:step, cols(Symbol("count(sheep)")), grid=false, xlabel="Step",
    ylabel="Population", label="Sheep")
```
it is necessary to wrap the variables around two functions. Furthermore, indexing columns is cumbersome. As far as I can tell, its not possible to index by `df.function_variable`.

Libbum:
Closes #234 by swapping `mean(weight)` to `:mean_weight` syntax.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -93,6 +93,7 @@ by_id
 random_activation
 partial_activation
 property_activation
+by_type
 ```
 
 ## Plotting
--- a/examples/predator_prey.jl
+++ b/examples/predator_prey.jl
@@ -246,14 +246,14 @@ results, _ = run!(model, agent_step!, n_steps; adata = adata)
 
 @df results plot(
     :step,
-    cols(aggname(sheep, count)),
+    :count_sheep,
     grid = false,
     xlabel = "Step",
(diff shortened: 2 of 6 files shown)


Datseris:
I forgot about this... remind me why we are doing this again? :D

Libbum:
The nail in the coffin (for you) was `dataframe.mean_weight` is possible, but `dataframe.Symbol(...)` or `dataframe.aggname(..)` is not.

Datseris:
God damn it , `mean(weight)` is so much more beautiful. Goes to show our world is "real" because we can't have perfection.

Libbum:
> Besides, you can always `rename!` as you wish.

:P 

Agreed though. `mean(weight)` is nicer, but runs into practicality issues.

Issue #100, opened 2019-12-10
Improve test suite 


Datseris:
The test suite of Agents.jl is not on par its base code or documentation. At the moment I think our test suite (located under `tests/`) suffers from the following problems:

1. Some of the _basic_ API is not tested. What is tested is the high level function `step!`. But good tests should be testing each of the individual basic functions that `step!` uses, e.g. if the scheduling works properly, if `move_agent!` does what it should, etc. Same goes for data collection.
2. There are too many model specific tests. It is fine to test one "real" model, but more than that is not necessary. In the end of the day the test suite is supposed to test if Agents.jl works, not if Schelling's model is a good approximation of society. Having too many tests that do the same thing just makes the test suite less clear.
3. There are no real tests about parallelization nor distributed computing.
4. There are no real tests about reproducibility of the end result, starting all the way from a model and going to the data collection, while setting `Random.seed!()` only once.

This issue is great for newcomers!

Datseris:
Notice that we now have coverall as well, to help identify untested parts: https://codecov.io/gh/JuliaDynamics/Agents.jl

Libbum:
Codecov seems to have died three months ago and not come back. I can't see any integration details with it here anymore, but I may not have access to that part of Settings.

Datseris:
do we have tests for parallel / replicates? that's the only thing missing from this list.

Libbum:
Not currently no, I can take a look at that at least.

Libbum:
Closing since #242 covered most of this. Parallel tests are now tracked in #243.

PR #242, opened 2020-04-29
A comprehensive test suite

### LINKED ISSUE #72: Increase test coverage

Libbum:
Closes #72 and #100 

This should bump our coverage up to 95% or above I'd guess. Fixed a number of small bugs and oversights in the process. 

Parallel cases have not been looked at, I'll tackle that at a later date. Maybe we should open a new issue specific to them.


Libbum:
Will push a few more fixes up soon after seeing the coverage gaps.

Libbum:
This is good to merge if you're happy with it.

src/core/agent_space_interaction.jl
```diff
@@ -40,30 +40,39 @@ kill_agent!(id::Integer, model) = kill_agent!(model[id], model)
     genocide!(model::ABM)
 Kill all the agents of the model.
 """
-genocide!(model::ABM) =
+genocide!(model::ABM{A,<:DiscreteSpace}) where {A} = _genocide!(model)
+genocide!(model::ABM{A,Nothing}) where {A} = _genocide!(model)
```
Datseris, line 44: Both methods actually do the same thing, why dispatch on space here?

src/core/agent_space_interaction.jl
```diff
@@ -40,30 +40,39 @@ kill_agent!(id::Integer, model) = kill_agent!(model[id], model)
...
     end
 
+
 """
     genocide!(model::ABM, n::Int)
 Kill the agents of the model whose IDs are larger than n.
 """
-function genocide!(model::ABM, n::Integer)
+genocide!(model::ABM{A,<:DiscreteSpace}, n::Integer) where {A} = _genocide!(model, n)
+genocide!(model::ABM{A,Nothing}, n::Integer) where {A} = _genocide!(model, n)
```
Datseris, line 58: Both methods actually do the same thing, why dispatch on space here?

src/core/agent_space_interaction.jl
```diff
@@ -40,30 +40,39 @@ kill_agent!(id::Integer, model) = kill_agent!(model[id], model)
...
     end
-end
 
 """
     genocide!(model::ABM, f::Function)
 Kill all agents where the function `f(agent)` returns `true`.
 """
-function genocide!(model::ABM, f::Function)
+genocide!(model::ABM{A,<:DiscreteSpace}, f::Function) where {A} = _genocide!(model, f)
+genocide!(model::ABM{A,Nothing}, f::Function) where {A} = _genocide!(model, f)
```
Datseris, line 70: Both methods actually do the same thing, why dispatch on space here?

src/core/agent_space_interaction.jl
Libbum: I kept getting ambiguous errors if we just set this as `genocide!(model)`. I fixed a bunch of continuous space issues after this though, so maybe those errors are dropped now. Give me a sec an I'll revert these to see...

Datseris: If it resolves ambiguity its okay.

Libbum: Yep. Here:
```julia
genocide!: Error During Test at /mnt/turtle/repos/Agents.jl/test/api_tests.jl:271
  Got exception outside of a @test
  MethodError: genocide!(::AgentBasedModel{Agent8,ContinuousSpace{typeof(Agents.defvel),Tuple{Int64,Int64}},typeof(fastest),Nothing}) is ambiguous. Candidates:
    genocide!(model::AgentBasedModel{A,#s60,F,P} where P where F where #s60<:ContinuousSpace) where A in Agents at /mnt/turtle/repos/Agents.jl/src/core/continuous_space.jl:214
    genocide!(model::AgentBasedModel) in Agents at /mnt/turtle/repos/Agents.jl/src/core/agent_space_interaction.jl:43
  Possible fix, define
    genocide!(::AgentBasedModel{A,S,F,P} where P where F where S<:ContinuousSpace) where A<:AbstractAgent
```

Call is:
```julia
  space2d = ContinuousSpace(2; periodic=true, extend=(1, 1))    
  model = ABM(Agent8, space2d)    
  attributes = (f1=true,f2=1)    
  for _ in 1:10 add_agent!(model, attributes...) end    
  genocide!(model)    
```
Ambiguous methods:
```julia
genocide!(model::ABM) = ... # agent_space_interaction.jl#43

genocide!(model::ABM{A,<:ContinuousSpace}) where {A} = ... # continuous_space.jl#214
```

How best to rectify that?

kavir1698: Maybe adding type info for genocide in discrete space?
```julia
# agent_space_interaction.jl#43
genocide!(model::ABM{A,<:DiscreteSpace}) where {A}
    for a in allagents(model)
        kill_agent!(a, model)
    end
end
```

Libbum: But then we don't have genocide for `Nothing`.

kavir1698: yeah, right!

On a different note, isn't massacre better than genocide?

Libbum:
```julia
genocide!(model)
decimate!(model) #where agent_ids.%10==0
massacre!(model) #where 20 < agent_ids < nagents(model)*0.1
```

Datseris: Let's minimize breaking changes by keeping `genocide!`. Other than that, I don't care as long as the name remains funny :D

PR #245, opened 2020-04-30
Update examples to current plotabm

Libbum:
Project.toml is tracking AgentsPlots#grid-align so that we can get a doc preview. Tested everything locally and it all seems fine. I've switched to pyplots for a lot of the examples, but may not be needed for all - there's a big runtime hit in some cases.


Libbum:
OK, don't merge this yet, there are issues at least in predator - prey with graphs not showing.

Datseris:
I use PyPlot.jl in DynamicalBilliards.jl, have a look at the travis script I use there. Oh, I also use it in DynamicalSystems.jl.

Libbum:
Hmm. I'm going to have to kill off some of these pyplot versions of things. Build time is too long with no output. Travis is killing the jobs.

Datseris:
We can configure Documenter to print a message for each page processed. I don't remember how to do it but it is possible. we can ask in `#documentation`.

examples/forest_fire.jl
```diff
@@ -20,21 +20,23 @@
...
-using Agents, Random
+using Agents, Random, AgentsPlots, Plots
+pyplot() # hide
 
 mutable struct Tree <: AbstractAgent
     id::Int
     pos::Tuple{Int,Int}
     status::Bool  # true is green and false is burning
 end
+nothing # hide
```
Libbum, line 31: I tried to step the matplotlib install from spilling its output here, but this didn't fix it. Any ideas?
https://juliadynamics.github.io/Agents.jl/previews/PR245/examples/forest_fire/

examples/game_of_life_2D_CA.jl
```diff
@@ -4,10 +4,9 @@
 
 # [Game of life on wikipedia](https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life).
 
-using Agents
-using AgentsPlots
-using Plots
-using Random
+using Agents, AgentsPlots, Plots
+using Random # hide
+gr() # hide
```
Libbum, line 9: This also seems to be noisy.

examples/forest_fire.jl
Datseris: Just plot something random before `makedocs` command I think. (with `pyplot()`)

Libbum: Great, that worked!

Libbum:
OK, I think this is ready to a point. Take a look around and see if you're happy with the rendering. Then, we should merge the AgentsPlots PR before pointing the Project.toml here back to #master

Datseris:
Just saw the changes, and everything is fine for me. I'm a bit busy with other projects these days, so feel free to merge the PRs both here and in AgentsPlots.

PR #249, opened 2020-05-12
add a move_agent! method that uses an argument as the velocity.

### LINKED ISSUE #248: move_agent! based on new position or an arbitrary velocity for ContinuousSpace models
Looking through the source code I see that there's no move_agent that uses some position or an arbitrary velocity for ContinuousSpace models. It is pretty trivial to, for example, add a move_agent! method with the signature:
```
move_agent!(agent::A, model::ABM{A,S,F,P}, vel::NTuple{D, N}, dt = 1.0) where {A <: AbstractAgent, S <: ContinuousSpace, F, P, D, N <: AbstractFloat}
```
If I understand correctly the reason is that there's simply no incentive and users are recommended to create a custom move_agent!. But doesn't this pattern repeat enough times so as to warrant being implemented in Agents.jl?

balenamiaa:
might close #248

### DIFF
--- a/src/core/continuous_space.jl
+++ b/src/core/continuous_space.jl
@@ -179,6 +179,24 @@ function move_agent!(agent::A, model::ABM{A, S, F, P}, dt = 1.0) where {A<:Abstr
 end
 
 """
+    move_agent!(agent::A, model::ABM{A, ContinuousSpace}, vel::NTuple{D, N}, dt = 1.0)
+Propagate the agent forwards one step according to vel.
+Also take care of periodic boundary conditions.
+
+For this continuous space version of `move_agent!`, the "evolution algorithm"
+is a trivial Euler scheme with `dt` the step size, i.e. the agent position is updated
+as `agent.pos += vel * dt`.
+"""
+function move_agent!(agent::A, model::ABM{A,S,F,P}, vel::NTuple{D, N}, dt = 1.0) where {A <: AbstractAgent, S <: ContinuousSpace, F, P, D, N <: AbstractFloat}
+      agent.pos = agent.pos .+ dt .* vel
+      if model.space.periodic
+        agent.pos = mod.(agent.pos, model.space.extend)
+      end
+      update_space!(model, agent)
+      return agent.pos
+end 
+
+"""
     update_space!(model::ABM{A, ContinuousSpace}, agent)
 Update the internal representation of continuous space to match the new position of
 the agent (useful in custom `move_agent` functions).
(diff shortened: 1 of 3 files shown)


Libbum:
I *think* those test failures are unrelated (will do some digging), but it would be good to have some tests for this method in [`api_tests`](https://github.com/JuliaDynamics/Agents.jl/blob/378ff77fc952d05ed641b38e9c882426d091ac52/test/api_tests.jl#L199) regardless.

src/core/continuous_space.jl
```diff
@@ -178,6 +178,24 @@ function move_agent!(agent::A, model::ABM{A, S, F, P}, dt = 1.0) where {A<:Abstr
   return agent.pos
 end
 
+"""
+    move_agent!(agent::A, model::ABM{A, ContinuousSpace}, vel::NTuple{D, N}, dt = 1.0)
+Propagate the agent forwards one step according to vel.
+Also take care of periodic boundary conditions.
```
Datseris, line 184:
```suggestion
Propagate the agent forwards one step according to `vel`.
```

src/core/continuous_space.jl
```diff
@@ -178,6 +178,23 @@ function move_agent!(agent::A, model::ABM{A, S, F, P}, dt = 1.0) where {A<:Abstr
...
 end
 
+"""
+    move_agent!(agent::A, model::ABM{A, ContinuousSpace}, vel::NTuple{D, N}, dt = 1.0)
+Propagate the agent forwards one step according to `vel` and the model's space.
+
+For this continuous space version of `move_agent!`, the "evolution algorithm"
+is a trivial Euler scheme with `dt` the step size, i.e. the agent position is updated
+as `agent.pos += vel * dt`.
+"""
```
Datseris, line 188: I don't think this information should be repeated.

Datseris: Duplicating information leads to less clear docs. It might sound counter intuitive, but it is in fact counter productive to say the same things again and again.

src/core/continuous_space.jl
```diff
@@ -180,11 +180,7 @@ end
 
 """
     move_agent!(agent::A, model::ABM{A, ContinuousSpace}, vel::NTuple{D, N}, dt = 1.0)
-Propagate the agent forwards one step according to `vel` and the model's space.
-
-For this continuous space version of `move_agent!`, the "evolution algorithm"
-is a trivial Euler scheme with `dt` the step size, i.e. the agent position is updated
-as `agent.pos += vel * dt`.
+Propagate the agent forwards one step according to `vel`, with dt as the time step,  and the model's space.
 """
```
balenamiaa, line 184: Should it be mentioned that update_vel! isn't called on `vel`? I think it's pretty obvious.

Datseris: It doesn't cost much: I propose to add a simple parenthesis: (`update_vel!` is not used).

Datseris:
@balenamiaa The only thing left is to add rudimentary tests for the method.

PR #251, opened 2020-05-12
Use propertynames instead of names

### LINKED ISSUE #250: Test failures with recent package versions
Tests were passing on my machine, but #249 shows string vs symbol failures in some tests on Travis.

Example failure:
```julia
Expression: names(df) == [:step, :id, :weight]
   Evaluated: ["step", "id", "weight"] == [:step, :id, :weight]
```

```julia
(Agents) pkg> up
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `https://github.com/JuliaRegistries/General.git`
   Updating `/mnt/turtle/repos/Agents.jl/Project.toml`
  [a93c6f00] ↑ DataFrames v0.20.2 ⇒ v0.21.0
  [093fc24a] ↑ LightGraphs v1.3.2 ⇒ v1.3.3
   Updating `/mnt/turtle/repos/Agents.jl/Manifest.toml`
  [324d7699] ↑ CategoricalArrays v0.7.7 ⇒ v0.8.0
  [34da2185] ↑ Compat v3.9.0 ⇒ v3.9.1
  [a10d1c49] ↑ DBInterface v2.0.0 ⇒ v2.1.0
  [a93c6f00] ↑ DataFrames v0.20.2 ⇒ v0.21.0
  [093fc24a] ↑ LightGraphs v1.3.2 ⇒ v1.3.3
  [bac558e1] ↑ OrderedCollections v1.1.0 ⇒ v1.2.0
  [69de0a69] ↑ Parsers v1.0.2 ⇒ v1.0.3
  [90137ffa] ↑ StaticArrays v0.12.2 ⇒ v0.12.3
```

Now I get the same issues.

Libbum:
Fixes #250.

Seems like this change is due to a bit of churn in DataFrames, [this commit](https://github.com/JuliaData/DataFrames.jl/commit/b1f675d89b63e5da4c6bdf0fb3f92bfab0c4d8c0) updates the docs at least concerning the adjustment. `names` will now output strings, although `propertynames` will still output symbols.

### DIFF
--- a/test/collect_tests.jl
+++ b/test/collect_tests.jl
@@ -54,7 +54,7 @@ end
     collect_agent_data!(df, model, props, 1)
     # Expecting weight values of all three agents. ID and step included.
     @test size(df) == (3, 3)
-    @test names(df) == [:step, :id, :weight]
+    @test propertynames(df) == [:step, :id, :weight]
     @test mean(df[!, :weight]) ≈ 0.3917615139
 
     props = [(:weight, mean)]
@@ -63,7 +63,7 @@ end
     # Activate aggregation. Weight column is expected to be one value for this step,
     # renamed mean(weight). ID is meaningless and will therefore be dropped.
     @test size(df) == (1, 2)
-    @test names(df) == [:step, :mean_weight]
+    @test propertynames(df) == [:step, :mean_weight]
     @test df[1, aggname(:weight, mean)] ≈ 0.3917615139
 
     # Add a function as a property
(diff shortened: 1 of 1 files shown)


Datseris:
Shouldn't the version of dataframes in compat be changed?

Libbum:
Definitely. Just checking how best to do that.

We currently have `DataFrames = "< 1, 1"`. Looks like well need to put 0.21.0 as a minimum since [b1f675d8](https://github.com/JuliaData/DataFrames.jl/commit/b1f675d89b63e5da4c6bdf0fb3f92bfab0c4d8c0#diff-c5cf4149fb7b9543a3f4cd7d9746163e)
is breaking for us in then sense that `names` changes and `propertynames`, whilst introduced in 0.19.3, changes its output over these versions.

Hard to say if this is seriously needed though for one function that's only used in testing not the collection methods themselves. We could add a test/Project.toml and set a low limit of 0.21.0 there, but then we break Julia 1.0 compatibility.

PR #253, opened 2020-05-13
Mixed model support in interacting_pairs

### LINKED ISSUE #252: Interacting_pairs with different types of agents
Hello,

New to Julia and Agents.jl.

I would like to know how to use Interacting_pairs with only agents with different types.

e.g. wolves and sheep interact only. No sheep with sheep or wolves with wolves

Thanks!
- Omar B

Libbum:
Closes #252 by supporting agent pairing separated by agent type in mixed models.

Continuing discussions from #252, whilst this is essentially a simple extension, so too are most of the `interacting_pairs` implementations that are already extant. In addition this change improves our mixed model support, so in general this is a win IMO.

### DIFF
--- a/src/core/continuous_space.jl
+++ b/src/core/continuous_space.jl
@@ -353,10 +353,11 @@ The argument `method` provides three pairing scenarios
   to only one pair. This functionality is useful e.g. when you want some agents to be
   paired "guaranteed", even if some other agents might be nearest to each other.
 - `:types`: For mixed agent models only. Return every pair of agents within radius `r`
-  (similar to `:all`), with an additional constraint that excludes neighbors with the same
-  type. For example, a mixed agent model of `Union{Sheep,Wolf,Grass}`, pairs of 
-  `(Sheep,Sheep)` will be excluded. To also exclude `(Wolf,Grass)` pairs, provide a
-  scheduler that samples only matchable parings.
+  (similar to `:all`), only capturing pairs of differing types. For example, a model of
+  `Union{Sheep,Wolf}` will only return pairs of `(Sheep, Wolf)`. In the case multiple
+  agent types, *e.g.* `Union{Sheep, Wolf, Grass}`, skipping pairings that involve
+  `Grass`, can be achived by a [`scheduler`](@ref Schedulers) that doesn't schedule `Grass`
+  types, *i.e.*: `scheduler = [a.id for a in allagents(model) of !(a isa Grass)]`.
 """
 function interacting_pairs(model::ABM, r::Real, method; scheduler = model.scheduler)
     @assert method ∈ (:scheduler, :nearest, :all, :types)
--- a/src/core/continuous_space.jl
+++ b/src/core/continuous_space.jl
@@ -355,7 +355,6 @@ The argument `method` provides three pairing scenarios
 - `:types`: For mixed agent models only. Return every pair of agents within radius `r`
   (similar to `:all`), only capturing pairs of differing types. For example, a model of
   `Union{Sheep,Wolf}` will only return pairs of `(Sheep, Wolf)`. In the case of multiple
-
   agent types, *e.g.* `Union{Sheep, Wolf, Grass}`, skipping pairings that involve
   `Grass`, can be achived by a [`scheduler`](@ref Schedulers) that doesn't schedule `Grass`
   types, *i.e.*: `scheduler = [a.id for a in allagents(model) of !(a isa Grass)]`.
--- a/src/core/continuous_space.jl
+++ b/src/core/continuous_space.jl
@@ -354,7 +354,8 @@ The argument `method` provides three pairing scenarios
   paired "guaranteed", even if some other agents might be nearest to each other.
 - `:types`: For mixed agent models only. Return every pair of agents within radius `r`
   (similar to `:all`), only capturing pairs of differing types. For example, a model of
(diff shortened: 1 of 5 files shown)


src/core/continuous_space.jl
```diff
@@ -352,16 +352,23 @@ The argument `method` provides three pairing scenarios
   agent is paired to its nearest neighbor. Similar to `:nearest`, each agent can belong
   to only one pair. This functionality is useful e.g. when you want some agents to be
   paired "guaranteed", even if some other agents might be nearest to each other.
+- `:types`: For mixed models only. Return every pair of agents within radius `r` (similar
```
Datseris, line 355: mixed models -> mixed agent models

src/core/continuous_space.jl
```diff
@@ -352,16 +352,23 @@ The argument `method` provides three pairing scenarios
   agent is paired to its nearest neighbor. Similar to `:nearest`, each agent can belong
   to only one pair. This functionality is useful e.g. when you want some agents to be
   paired "guaranteed", even if some other agents might be nearest to each other.
+- `:types`: For mixed models only. Return every pair of agents within radius `r` (similar
+  to `:all`), with an additional constraint that excludes neighbors with the same type.
+  For example, a mixed model of `Union{Sheep,Wolf,Grass}`, pairs of `(Sheep,Sheep)` will
+  be excluded. To also exclude `(Wolf,Grass)` pairs, provide a scheduler that samples only
```
Datseris, line 358: I don't immediatelly understand how this exclusion would work.

Libbum: Take a look at the tests, I have a model with `Union{Agent6, AgentU1, AgentU2}`, which wont match `(Agent6, Agent6)` or pairs of the other types. Then provide a scheduler that returns allowed matches, such that `(Agent6, AgentU1)` would not be possible.

The predator model, you would exclude `Grass` in the scheduler to only allow `(Sheep,Wolf)`.

Generally speaking you may have `Union{Road, House, River, Boat, Car, Bicycle}`, when you need only interactions on the road. Your scheduler returns ids of `Road, Car, Bicycle`, with interactions of `(Road, Car), (Road, Bicycle), (Car, Bicycle)` possible.

Datseris: I do not understand why interaction `(Car, Bicycle)` is not possible though....?

Libbum: It is, that would be the point of this. If you only wanted `(Road, Car), (Road, Bicycle)`, that would require two calls.

src/core/continuous_space.jl
```diff
@@ -352,16 +352,23 @@ The argument `method` provides three pairing scenarios
   agent is paired to its nearest neighbor. Similar to `:nearest`, each agent can belong
   to only one pair. This functionality is useful e.g. when you want some agents to be
   paired "guaranteed", even if some other agents might be nearest to each other.
+- `:types`: For mixed agent models only. Return every pair of agents within radius `r`
+  (similar to `:all`), with an additional constraint that excludes neighbors with the same
+  type. For example, a mixed agent model of `Union{Sheep,Wolf,Grass}`, pairs of 
+  `(Sheep,Sheep)` will be excluded. To also exclude `(Wolf,Grass)` pairs, provide a
+  scheduler that samples only matchable parings.
```
Datseris, line 359: In all honesty this all sounds very confusing for me, I think it needs to be documented simpler. 

What is "matchable pairings" ...?

(1) I don't think using `(Sheep, Sheep)` is a good idea. It specifies the thought process. It is not just `(sheep, Sheep)` pairs that are not allowed, it is also `(Wolf, Wolf)` and also `(Grass, Grass)`. What I recommend is to start the docstring with only a model `Union{Sheep, Wolf}` and say that only `(Sheep, Wolf)` pairs would be allowed.

(2) the second part of the docstring is unclear to me. Here we can say: In the case of having more agent types, e.g. `{Sheep, Wolf, Grass}`, and you want to _in general_ skip pairings that involve `Grass`, then simply provide a scheduler that doesn't schedule `Grass` types, i.e.: `scheduler = [a.id for a in allagents(model) of !(a isa Grass)]`.

src/core/continuous_space.jl
```diff
@@ -353,10 +353,11 @@ The argument `method` provides three pairing scenarios
   to only one pair. This functionality is useful e.g. when you want some agents to be
   paired "guaranteed", even if some other agents might be nearest to each other.
 - `:types`: For mixed agent models only. Return every pair of agents within radius `r`
-  (similar to `:all`), with an additional constraint that excludes neighbors with the same
-  type. For example, a mixed agent model of `Union{Sheep,Wolf,Grass}`, pairs of 
-  `(Sheep,Sheep)` will be excluded. To also exclude `(Wolf,Grass)` pairs, provide a
-  scheduler that samples only matchable parings.
+  (similar to `:all`), only capturing pairs of differing types. For example, a model of
+  `Union{Sheep,Wolf}` will only return pairs of `(Sheep, Wolf)`. In the case multiple
```
Datseris, line 357:
```suggestion
  `Union{Sheep,Wolf}` will only return pairs of `(Sheep, Wolf)`. In the case of multiple

```

Libbum: Bah, swing and a miss!

Libbum:
@omarbetancourt, how does this look for you?

omarbetancourt:
Thanks for that quick implementation guys!

Cheers!

PR #257, opened 2020-06-11
Models sub-module

Datseris:
The submodule that makes it convienent to use models.

Example:
```julia
using Agents

model, agent_step!, model_step! = Models.flocking()
step!(model, agent_step!, model_step!, 10)

using AgentsPlots
plotabm(model)
```

produced docs: https://juliadynamics.github.io/Agents.jl/previews/PR257/models/


Datseris:
@kavir1698 or @Libbum could you guys please give a quick review here? I'd like to merge this and _then_ the interactive app and then I'll start adding to the comparison table.

Datseris:
(by the way, I've added this to the JuliaDynamics website, to see how it would like as an example, and it is actually really good: https://juliadynamics.github.io/JuliaDynamics/ (go to the end and click ABM))

docs/src/models.md
```diff
@@ -0,0 +1,15 @@
+# Predefined Models
+Predefined agent based models exist in the `Models` submodule in the form of functions that return `model, agent_step!, model_step!` when called.
+They are the versions already explained in the Examples section of the docs.
```
Libbum, line 3: > They are the versions already explained in the Examples section of the docs.
Perhaps move this underneath the access example, and reword to:
The Examples section of the docs outline how to use and interact with each model.

Libbum:
This looks great. Works as expected. I think we can add some issues with the 'good first issue' for importing the rest of the examples into the framework rather than doing all of that right now.

A note on the julia dynamics page - looks sweet in chrome, but firefox has some issues for me with the tab links to swap to each example. I'm getting:
image

Datseris:
> This looks great. Works as expected. I think we can add some issues with the 'good first issue' for importing the rest of the examples into the framework rather than doing all of that right now.

Hah, this is exactly the reason I didn't put in all the models!

> A note on the julia dynamics page - looks sweet in chrome, but firefox has some issues for me with the tab links to swap to each example. I'm getting:

Hm, I also use firefox but never had a problem. Perhaps you can open an issue in its repo outlining how to reproduce?

kavir1698:
Everything works smoothly for me, including the JuliaDynamics page on FireFox. I suggest we use a different model on the JuliaDynamics page, because the disease spread model is visually similar to the billiards example. The flocking or bacteria examples are better.

Libbum:
OK - if both of you guys are fine with firefox then its probably me and some extension I've got configured. So don't worry about that.

Would second Ali's suggestion. Perhaps we could do the bacteria since it's quite visually striking.

Datseris:
Ugh, every damned time I use Plots.jl I get problems, some error here, some error there. Terrible!!! The very standard and straightforward plotting example we have in the docs for flocking now gives an unending stream of:
```

'c' argument looks like a single numeric RGB or RGBA sequence,
which should be avoided as value-mapping will have precedence in case its
length matches with 'x' & 'y'.  Please use a 2-D array with a single row if you
really want to specify the same RGB or RGBA value for all points.
```

grrrrr

I can't wait for Makie and MakieLayout to be merged and we switch to Makie!

PR #259, opened 2020-06-20
show interactive app in documentation

Datseris:


Libbum:
Would it be useful to consider a direct coupling between the new predefined Models and the interactive app?

```julia
model, agent_step!, model_step! = Models.forest_fire()
...
p1 = interactive_abm(model, agent_step!, model_step!, params;
ac = ac, as = 1, am = am, mdata = mdata, mlabels=mlabels)
```
Could become
```julia
p1 = interactive_abm(Models.forest_fire(), params;
ac = ac, as = 1, am = am, mdata = mdata, mlabels=mlabels)
```

Datseris:
yeah this 
```julia
p1 = interactive_abm(Models.forest_fire()..., params;
```
achieves it via the "splatting" operator. 

However, I think it is worth considering pointing out that the app _explicitly_ takes in the model and the two stepping functions, which are in any case the central 3 things one need to use e.g. `step!`. It somehow highlights the simplicity of usage...?

Libbum:
Fair points. I think we can flesh out an example later on that uses the splatting idea, but otherwise keep it as-is.

PR #261, opened 2020-07-14
DifferentialEquations integration documentation

Libbum:
A first pass at the DifferentialEquations.jl document. It's written in long form so everyone can follow, but feel free to identify places we can cut so that it's a little more concise and to the point.

I didn't add benchmarks for the last section - unsure if they're really necessary anyway, but I ran into some issues implementing them.

```julia
Random.seed!(759) #hide
modelcb = initialise_cb()
prob = ODEProblem(fish_stock!, [initial_stock], tspan, [max_population, 0.0])
@btime solve($prob, Tsit5(); callback = CallbackSet(PeriodicCallback(i -> fish!(i, $modelcb), 364), PeriodicCallback(i -> i.p[2] = 0.0, 365)))
```

Is perhaps the most representative of my constructions. This runs, but I hit a bunch of `maxiter` warnings with the solver here, and if I increase/decrease the `maxiter` value, then the benchmark timings change with it - dramatically.

So until that issue is resolved, the benchmark it unreliable.


Libbum:
https://juliadynamics.github.io/Agents.jl/previews/PR261/examples/diffeq/

This extends the time to build documentation by a fair chunk. Perhaps I can drop down to `OrdinaryDiffEq` to build faster, but I kind of wanted it to remain high level, so that people could choose any ODE type and not worry about it.

Datseris:
`OrdinaryDiffEq ` is what provides the ODE solvers, so you should still drop to OrdinaryDiffEq from DifferentialEquations.jl and they still have access to the same solvers.

For benchmarks, I think you have to set up the `@btime` macro to have a setup phase, otherwise it reuses the same integrator (and probably that's why you hit the maxiters mark?).

docs/make.jl
```diff
@@ -82,6 +83,7 @@ pages = [
     "Predefined Models" => "models.md",
     "API" => "api.md",
     "Interactive application" => "interact.md",
+    "Integration" => ["DifferentialEquations.jl" => "examples/diffeq.md"],
```
Datseris, line 86: Perhaps name this `Ecosystem Integration`?

examples/diffeq.jl
```diff
@@ -0,0 +1,385 @@
+# # Integrating Agents.jl with DifferentialEquations.jl
+#
+# Leveraging other best-in-class packages from the Julia ecosystem is one of the many
+# strengths Agents.jl provides over alternative ABMs.
+#
+# The [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl) package
+# [consistently outperforms](http://www.stochasticlifestyle.com/wp-content/uploads/2019/08/de_solver_software_comparsion.pdf)
```
Datseris, line 7: I don't think this fits here. It is not our place to discuss DE solver comparison, we only show how easy it is to use Agents with one of them.

Libbum:
> For benchmarks, I think you have to set up the `@btime` macro to have a setup phase, otherwise it reuses the same integrator (and probably that's why you hit the maxiters mark?).

Yeah, I'm using the 'global' method here (which prepends the `$` sign to variables). There's no change if I refactor to a setup implementation

```julia
@btime solve(prob, Tsit5(); callback = CallbackSet(PeriodicCallback(i -> fish!(i, modelcb), 364), PeriodicCallback(i -> i.p[2] = 0.0, 365))) setup=(modelcb = initialise_cb(), prob = ODEProblem(fish_stock!, [initial_stock], tspan, [max_population, 0.0]))
```
Anyway, I'm not sure it really adds much more value. It's going to run mostly the same as the integrator version, since I confirmed with Chris that what I'm doing there is [exactly what discrete callbacks do](https://discourse.julialang.org/t/handling-unstable-true-solution-issues-with-the-integrator-interface/41208/8?u=libbum).

Datseris:
Heh, I just had a brilliant idea...

Since the actual interaction is done at the integrator level, and since most functionality of DynamicalSystems.jl also occurs at the integrator level, I might be able to set up something fancy like "lyapunov exponents" for ABMs (which makes no theoretical sense by the way). I'll think about it!

Datseris:
> Anyway, I'm not sure it really adds much more value. It's going to run mostly the same as the integrator version, since I confirmed with Chris that what I'm doing there is exactly what discrete callbacks do.

Yeap as long as you also set `u_modified = true` in the integrator.

Libbum:
> Yeap as long as you also set `u_modified = true` in the integrator.

Ah, good to know. I'll do some checks with that.

Dropping down to `OrdinaryDiffEq` saves us 8 minutes of build time.

Datseris:
Also increment patch version in Project.toml so this can go into the stable documentation as well.

Libbum:
I thought `u_modified!` was only for `u`, but as you say - it's the entire system. We get a small difference by adding the guards, so better safe than sorry.

Libbum:
Will give Ali a little time to chime in if he wants. Will merge tomorrow.

kavir1698:
Hi, thank you. I will take a look later today.
________________________________

PR #263, opened 2020-07-22
Daisyworld with structs, and fill space

### LINKED ISSUE #260: FR: function that puts one agent in every grid point
Would be nice to create and expose a function to the user that creates and adds a specific agent to every grid point. For example the function could take a pre-initialized agent, and then make copies and put them into the grid.

Datseris:
Closes #260

This PR changes the Daisywolrd example to use `Land` for the surface. It also does some crucial bugfix to the documentation of `step!/run!` : in the case of `n::Function` the model_step function must be provided explicitly (there is no way to say whether the 3rd function is the model step or the n otherwise).

It also adds the `fill_space!` function and tests for it.

IT also does some documentation improvements on some inconsistencies I found while doing the change.

---

There are problems with the interactive application that I have to solve independently and will do so in another PR!


Datseris:
@Libbum one of the major troubles  I had here is that the stability of the daisyworld depends majorly on the starting seed. Did you have same experience when you made the original version, or did I code some rule wrongly in the translation?

Libbum:
To get a semistable situation where one or the other species don't just die out from exposure was certainly a problem. That's not an issue with your refactoring.

I've been watching the progress and haven't seen anything too out of place - will take a more thorough look now though.

Libbum:
Same thing happens in Netlogo, so at the very least it's an artefact of the implemented algorithm.

daisydie

Datseris:
Aha, in our interactive application we cannot add or remove agents. We should also put this in the comparison, or simply make the interactive applications of NetLogo to be the "best", and ours to have a lesser color.

examples/daisyworld.jl
```diff
@@ -26,15 +28,25 @@
...
-# ## Defining the agent type
+# ## Defining the agent types
 
-# The agent here is not so complex. We see it has three values (other than the required
+# `Daisy` has three values (other than the required
 # `id` and `pos` for an agent that lives on a [`GridSpace`](@ref). Each daisy has an `age`,
 # confined later by a maximum age set by the user, a `breed` (either `:black` or `:white`)
 # and an associated `albedo` value, again set by the user.
+# `Land` represents the surface. We could make `Land` also have an albedo field, but
+# in this world all surface has the same albedo and thus we make it a model parameter.
```
Libbum, line 41: world all surface -> world, the entire surface

examples/daisyworld.jl
```diff
@@ -26,15 +28,25 @@
...
+
+# Notice that the `Land` does not necessarily have to be an agent, and one could represent
+# surface temperature via a matrix (parameter of the model). This is done in an older version,
+# see file `examples/daisyworld_old.jl`. The old version has a slight performance advantage.
```
Libbum, line 45: Should we change this files name to be something like `daisyworld_matrix` or something? It's an older version sure, but for a new user that's not really relevant. What they're interested in is the choice between performance and usability.

Haven't confirmed, but my assumption is that the `_old` implementation will not show up in the docs now, but will stay in the examples directory for this link only. Is that right? If so - cool, that makes the most sense.

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
     albedo::Float64 # 0-1 fraction
 end
 
+mutable struct Land <: AbstractAgent
+    id::Int
+    pos::Tuple{Int,Int}
+    temperature::Float64
+end
+
+const DaisyWorld = ABM{Union{Daisy, Land}};
```
Libbum, line 69: This is a nice touch.

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
 # ## World heating
 
 # The surface temperature of the world is heated by its sun, but daisies growing upon it
 # absorb or reflect the starlight -- altering the local temperature.
 
-function suface_temperature!(node::Int, model::ABM{Daisy})
+function update_surface_temperature!(node::Int, model::DaisyWorld)
     ids = get_node_contents(node, model)
-    absorbed_luminosity = if isempty(ids)
+    absorbed_luminosity = if length(ids) == 1 # no daisy
```
Libbum, line 78: Perhaps an explicit comment here for the user. Rather than 'no daisy', how about 'We use `fill_space!` for our `Land` agents, thus we expect to have a minimum of 1 agent in each node (i.e. when no daisy has grown here)'

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
-function diffuse_temperature!(node::Int, model::ABM{Daisy}; ratio = 0.5)
-    neighbors = node_neighbors(node, model)
-    model.temperature[node] =
-        (1 - ratio) * model.temperature[node] +
-        ## Each neighbor is giving up 1/8 of the diffused
-        ## amount to each of *its* neighbors
-        sum(model.temperature[neighbors]) * 0.125 * ratio
+# In addition, temperature diffuses over time
+function diffuse_temperature!(node::Int, model::DaisyWorld)
+    ratio = get(model.properties, :ratio, 0.5) # diffusion ratio
```
Libbum, line 95: Why use this vs a keyword? If a user always constructs their model via the initialise function, they'll not have access to modify this ratio anymore right?

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+        surface_albedo = 0.4,
+        solar_change = 0.005,
+        solar_luminosity = 1.0, # initial luminosity
+        scenario = :default,
+    )
+
+    space = GridSpace(griddims, moore = true, periodic = true)
+    properties = @dict max_age surface_albedo solar_luminosity solar_change scenario
+    properties[:tick] = 0
+    ## create a scheduler that only schedules Daisies
```
Libbum, line 207: Helpful example!

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+    daisysched(model) = [a.id for a in allagents(model) if a isa Daisy]
+    model = ABM(Union{Daisy, Land}, space;
+        scheduler = daisysched, properties = properties, warn = false
```
Libbum, line 210: With all the mixed models we're building, I wonder if we should go back and take a look at this warning flag. Another issue for another time though.

test/interaction_tests.jl
```diff
@@ -106,3 +107,24 @@ end
...
+  mutable struct Daisy <: AbstractAgent
+      id::Int
+      pos::Tuple{Int, Int}
+      breed::String
+  end
+  mutable struct Land <: AbstractAgent
+      id::Int
+      pos::Tuple{Int, Int}
+      temperature::Float64
+  end
```
Libbum, line 121: Have to take these outside the `@testset` otherwise v1.0 tests will break.

Libbum:
Other than those comments, all looks good in the code. Since the doc build didn't happen I haven't seen what the figures etc now look like so I'm building offline. Will get back to you with comments there once its done.

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
-        propagate!(n, model)
+# ## Initialising Daisyworld
+
+# Here, we construct a function to initialize a Daisyworld. We use [`fill_space!`](@ref)
+# to fill the space with `Land` instances. Then, we need to know how many
+# daisies of each type to seed the planet with and what their albedo's are.
+# We also want a value for surface albedo, as well as solar intensity
+# (and we also choose between constant or time-dependent intensity with `scenario`).
+
+import StatsBase
```
Libbum, line 188: All other examples are using `Statistics`. Docs will fail on this at present.

Libbum: Ah, sample isn't in `Statistics`, we'll need to add `StatsBase` to `make.jl`

src/core/model.jl
```diff
@@ -249,3 +249,9 @@ Return an iterator over all agents of the model.
 """
 allagents(model) = values(model.agents)
 
+
+"""
+    allids(model)
+Return an iterator over all agent IDs of the model.
+"""
+allids(model) = keys(model.agents)
```
Libbum, line 257: You've got `allids` listed in the docs, but it's not exported here.

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
-        72 * log(absorbed_luminosity) + 80
-    else
-        80
-    end
+    local_heating = absorbed_luminosity > 0 ? 72 * log(absorbed_luminosity) + 80 : 80
     ## Surface temperature is the average of the current temperature and local heating.
-    model.temperature[node] = (model.temperature[node] + local_heating) / 2
+    T0 = model[ids[1]].temperature
+    model[ids[1]].temperature = (T0 + local_heating) / 2
 end
```
Libbum, line 91: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+# In addition, temperature diffuses over time
+function diffuse_temperature!(node::Int, model::DaisyWorld)
+    ratio = get(model.properties, :ratio, 0.5) # diffusion ratio
+    ids = space_neighbors(node, model)
+    meantemp = sum(model[i].temperature for i in ids if model[i] isa Land)/8
+    land = model[get_node_contents(node, model)[1]] # land at current node
+    ## Each neighbor land patch is giving up 1/8 of the diffused
+    ## amount to each of *its* neighbors
+    land.temperature = (1 - ratio)*land.temperature + ratio*meantemp
 end
```
Libbum, line 102: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
                 ## Seed a new daisy in one of those cells
-                seeding_place = rand(empty_neighbors)
-                add_agent!(seeding_place, model, agent.breed, 0, agent.albedo)
+                seeding_place = vertex2coord(rand(empty_neighbors), model)
+                a = Daisy(nextid(model), seeding_place, daisy.breed, 0, daisy.albedo)
+                add_agent_pos!(a, model)
             end
         end
     end
 end
```
Libbum, line 136: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
-nothing # hide
 
-# Now, we need to write the model and agent step functions for Agents.jl to advance
+# And if the daisies cross an age threshold, they die out.
+# Death is controlled by the `agent_step` function
+function agent_step!(agent::Daisy, model::DaisyWorld)
+    agent.age += 1
+    agent.age >= model.max_age && kill_agent!(agent, model)
+end
```
Libbum, line 143: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+
+# We also need to define a version for the `Land` instances
+# (the dynamics of the `Land` are resolved at model level)
+agent_step!(agent::Land, model::DaisyWorld) = nothing
+
```
Libbum, line 148: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+function model_step!(model)
+    for n in nodes(model)
+        update_surface_temperature!(n, model)
+        diffuse_temperature!(n, model)
+        propagate!(n, model)
+    end
+    model.tick += 1
+    solar_activity!(model)
+end
+
```
Libbum, line 162: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+            model.solar_luminosity += model.solar_change
         end
         if model.tick > 500 && model.tick <= 750
-            model.solar_luminosity -= 0.0025
+            model.solar_luminosity -= model.solar_change/2
         end
+    elseif model.scenario == :change
+        model.solar_luminosity += model.solar_change
     end
 end
```
Libbum, line 179: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
     end
-    model.tick += 1
-    solar_activity!(model)
-end
 
-function agent_step!(agent::Daisy, model::ABM{Daisy})
-    agent.age += 1
-    agent.age >= model.max_age && kill_agent!(agent, model)
+    return model
 end
```
Libbum, line 230: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+model = daisyworld()
+
+# To visualize we need to define the necessary functions for [`plotabm`](@ref).
+# The daisies will obviously be black or white, but the land will have a color
+# that reflects its temperature, with -50 darkest and 100 ᵒC brightest color
+
+daisycolor(a::Daisy) = a.breed
+const landcolor = cgrad(:thermal)
+daisycolor(a::Land) = landcolor[(a.temperature+50)/150]
+
```
Libbum, line 247: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+# And we plot daisies as circles, and land patches as squares
+daisyshape(a::Daisy) = :circle
+daisysize(a::Daisy) = 8
+daisyshape(a::Land) = :square
+daisysize(a::Land) = 13.5
+
```
Libbum, line 253: `nothing #hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+# Notice that we want to ensure that the `Land` patches are always plotted first.
+plotsched = by_type((Land, Daisy), false)
+
+plotkwargs = (
+    ac = daisycolor, am = daisyshape, as = daisysize,
+    scheduler = plotsched, aspect_ratio = 1, size = (600, 600), showaxis = false,
 )
-plot(p1, p2, p3, layout = (3, 1), size = (500, 800))
+
+p = plotabm(model; plotkwargs...)
```
Libbum, line 262: The land size is not correct, both here and in the next two plots

landissue

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+end
+gif(anim, "daisyworld.gif", fps = 3)
+
+# Running this animation for longer hints that this world achieves quasi-equilibrium
+# for some input parameters, where one `breed` does not totally dominate the other.
+# Of course we can check this easily through data collection.
+# Notice that here we have to define a function `breed` that gives the daisy's `breed`
+# field. We cannot use just `:breed` to automatically find it, because in this mixed
+# agent model, the `Land` doesn't have any `breed`.
+# %% # src
```
Libbum, line 284: Not sure what this command is meant to to, but currently it's being displayed in the documentation text.

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+agent_df, model_df = run!(model, agent_step!, model_step!, 1000; adata = adata)
+
+p = plot(agent_df[!, :step], agent_df[!, :black_breed], label = "black")
+plot!(p, agent_df[!, :step], agent_df[!, :white_breed], label = "white")
+plot!(p; xlabel = "tick", ylabel = "daisy count")
+
+# ## Time dependent dynamics
+# %% #src
+
+# To use the time-dependent dynamics we simple use the keyword `scenario = :ramp` during
```
Libbum, line 302: simple -> simply

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+# model creation. However, we also want to see how the planet surface temperature changes
+# and would be nice to plot solar luminosity as well.
+# Thus, we define in addition
+gettemperature(a) = a isa Land ? a.temperature : missing
+meantemperature(x) = mean(skipmissing(x))
+adata = [(breed, black), (breed, white), (gettemperature, meantemperature)]
+
```
Libbum, line 309: `nothing # hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+# And, to have it as reference, we also record the solar luminosity value
+mdata = [:solar_luminosity]
+
```
Libbum, line 312: `nothing # hide`

examples/daisyworld.jl
```diff
@@ -48,249 +60,269 @@ mutable struct Daisy <: AbstractAgent
...
+agent_df, model_df = run!(model, agent_step!, model_step!, 1000; adata = adata, mdata = mdata)
+
+p = plot(agent_df[!, :step], agent_df[!, :black_breed], label = "black")
+plot!(p, agent_df[!, :step], agent_df[!, :white_breed], label = "white")
+plot!(p; xlabel = "tick", ylabel = "daisy count")
+
+p2 = plot(agent_df[!, :step], agent_df[!, aggname(adata[3])], ylabel = "temperature")
+p3 = plot(model_df[!, :step], model_df[!, :solar_luminosity], ylabel = "L", xlabel = "ticks")
+
+plot(p, p2, p3, layout = (3, 1))
```
Libbum, line 325: Probably worth adding `, size = (500, 800)` to the arguments here. Otherwise the graph is a little small.

examples/daisyworld.jl
```diff
@@ -304,6 +336,4 @@ plot(p1, p2, p3, layout = (3, 1), size = (500, 800))
 #
 # Finally, as the sun fades back to normal levels, both the temperature and white daisy
 # population struggle to find equilibrium. The counterbalancing force of the black daisies
```
Libbum, line 338: This is no longer true with the scenario you have generated. This text refers to a daisyworld there the solar luminosity starts at 0.75 and it's temperature never exceeds around 40 degrees. 

Will need to either change this text to refer to the new plot, or change the parameters back.

Libbum:
OK, think I'm done with the comments for the moment. The only other issue I can see is that the example `daisyworld.gif` that ends up at the top of the document looks to be an old version when I build. Not sure if that's just cache or something else.

examples/daisyworld.jl
Datseris: Why do you add these `nothing # hide` statements?

examples/daisyworld.jl
Datseris: I have a suspicion that it may depend on screen resolution.... But I'll test again.

examples/daisyworld.jl
Datseris: It isn't, it was misstyped because you have to have no space between `#` and `src`

examples/daisyworld.jl
Datseris: If `ratio` is a model parameter, it is used here. What do you mean by "keyword"?

examples/daisyworld.jl
Datseris: I think the flag is fine. Not a big deal to add `warn = false`.

examples/daisyworld.jl
Datseris: Turns out it strongly depends on the backend (I use pyplot)

examples/daisyworld.jl
Libbum: So that the text reads better in the docs. Don't see a reason for things like
```julia
agent_step!(agent::Land, model::DaisyWorld) = nothing
```
```julia
agent_step! (generic function with 2 methods)
```

The `nothing #hide` drops the last bit.

examples/daisyworld.jl
Libbum: Yeah, I follow that. 

But you've come from a function that has `diffuse_temperature!(node::Int, model::ABM{Daisy}; ratio = 0.5)`, to taking a model parameter (fine).

The issue is your `daisyworld` constructor function (beginning line 191) doesn't have ratio as a property. 

The user would therefore need to read the function details to see that they could add such a property, vs the old way where they'd only need to look at the function definition. 

So the question I'm asking is if there is an advantage I'm not seeing to changing this from the keyword (`ratio = 0.5`) method.

examples/daisyworld.jl
Libbum: Ah, fair enough. We have pyplot available - some of the other examples are built with it, but this guy takes forever on Travis with it enabled.

examples/daisyworld.jl
Datseris: The argument goes both ways, because you can understand that this code has been "run". In general I have not a strong preference, but for tutorials it might be worth it to explicitly show what is run code and what is not.

Datseris:
> Still think the last plot is a little small & now that its caption has been removed, the page ends a little abruptly.

There wills soon be the interactive part there so not much worry.

examples/daisyworld.jl
fbanning: I agree with Datseris that this feedback from running the code might be really helpful for beginners working through the tutorials.

Issue #258, opened 2020-06-16
Add all example models to `Models`


Datseris:
PR #257 introduced the `Models` submodule, that allows for convenient loading of models defined in the examples. 

At the moment, not all example models are in this module, but it is straightforward to add them!

- [x] SIR model for the spread of COVID-19
- [x] Wealth distribution model
- [x] Wright-Fisher model of evolution
- [x] HK (Hegselmann and Krause) opinion dynamics model
- [x] Daisyworld
- [x] Predator-prey dynamics (Wolf, Sheep, Grass)
- [x] Bacterial Growth

fbanning:
Am currently working on this.

fbanning:
Almost done. Will open a PR once I've finished it.

I have one question about growing_bacteria.jl though. Why does the `SimpleCell` function have to be inside the struct?

``` julia
mutable struct SimpleCell <: AbstractAgent
    id::Int
    pos::NTuple{2,Float64}
    length::Float64
    orientation::Float64
    growthprog::Float64
    growthrate::Float64

    ## node positions/forces
    p1::NTuple{2,Float64}
    p2::NTuple{2,Float64}
    f1::NTuple{2,Float64}
    f2::NTuple{2,Float64}

    function SimpleCell(id, pos, l, φ, g, γ)
        a = new(id, pos, l, φ, g, γ, (0.0, 0.0), (0.0, 0.0), (0.0, 0.0), (0.0, 0.0))
        update_nodes!(a)
        return a
    end
end
```

Libbum:
It's a constructor for `SimpleCell` that updates the position value in a custom way (the cell needs some redundant information which assists the mitosis). If that didn't exist, you couldn't do:

```julia
add_agent!((5.0, 5.0), model, 0.0, 0.3, 0.0, 0.1)
add_agent!((6.0, 5.0), model, 0.0, 0.0, 0.0, 0.1)
```
then run the model and expect it to work properly.

Is it causing an issue for the example library? If so, feel free to push up the PR without this working, and we can discuss what a work around could be.

fbanning:
> Is it causing an issue for the example library?

Nope, all good. I was just curious because this kind of structure isn't explained in the tutorial.
Maybe it's worth the time to add an explainer here about why this was written like that.

In other news: I just submitted the PR, see #266 .

Datseris:
[...]

PR #266, opened 2020-07-31
Extend models submodule

fbanning:
Related to #258 

Did I miss anything?
Happy to receive feedback if anything should be changed. (But please be kind, it's my first PR for an open source project.)


Libbum:
A good start! Just a few more things to take care of.

- [ ] Alter the files under `examples/` so that our documentation updates itself. Just a simple one liner will do it. This is the change done to [forrest_fire](https://github.com/JuliaDynamics/Agents.jl/commit/bb9780bc9a9926da0f1329f59de98e2a6db8f76d#diff-603bac5915bb7504a05320eb1f87e647) for example.
- [ ] The tests are failing because you're re-defining `agent_step!` and a few other things in multiple files. Here's the [forrest_fire](https://github.com/JuliaDynamics/Agents.jl/commit/bb9780bc9a9926da0f1329f59de98e2a6db8f76d#diff-bc279cd292aec31c06934d4c7c5928f1) model diff. See how we name the function `forest_step!`? We also have `schelling_step!` etc.

Once those are fixed, I'll need to run the documentation build locally to check & verify the changes are all correct. For the most part it looks good though, thanks!

fbanning:
Great, I'll fix these things on monday. 👍

fbanning:
Considering the renaming of the stepping functions:
I would propose to rename all of them following the pattern `[name]_agent_step!` and `[name]_model_step!`. This will help to make clear which function is which because otherwise you would just have e.g. `forest_step!` (unclear if agent or model step).
Are you OK with this?

fbanning:
One question about https://github.com/fbanning/Agents.jl/blob/Extend-Models-submodule/src/models/wright-fisher.jl : 
Is it really necessary to have those two functions with different names `*_neutral` and `*_selection` ? I think it might be possible to just define the function twice with the same name - once with and once without `:trait`as a parameter.
If that's correct, this could simplify the module a bit.

Datseris:
@kavir1698 knows about the evolution model

Libbum:
> I would propose to rename all of them following the pattern `[name]_agent_step!` and `[name]_model_step!`

This is a good idea. Being explicit is useful here, users can re-name these functions anyway they like anyway. So go ahead with this.

fbanning:
Great. I've already integrated that in the commit 069389629574bdf84b6787083f8a825a4b23c138.

How come the Travis checks are still failing?

Libbum:
I don't have access to push to your branch. But the (current) Travis failure is due to the inclusion of `Distributions` in the SIR model. I think it's fine to add this as a dependency to the project.

Jump into the julia repl in your branch, 
```julia
] activate .
add Distributions
```
Then add the altered `Project.toml` and push it up. That should fix the current problem.

fbanning:
Ah OK, didn't think of that. I haven't worked with Travis before and didn't know where to look.

src/models/daisyworld.jl
```diff
@@ -0,0 +1,163 @@
+using Random
+import StatsBase
+import DrWatson: @dict
```
Libbum, line 3: This is the current Travis error. It's just a convenience function, so let's drop this dependency.

src/models/daisyworld.jl
```diff
@@ -0,0 +1,163 @@
...
+        solar_change = 0.005,
+        solar_luminosity = 1.0, # initial luminosity
+        scenario = :default,
+        seed = 165
+    )
+
+    Random.seed!(seed)
+    space = GridSpace(griddims, moore = true, periodic = true)
+    properties = @dict max_age surface_albedo solar_luminosity solar_change scenario
+    properties[:tick] = 0
```
Libbum, line 54:
```julia
properties = Dict(max_age => max_age,
 surface_albedo => surface_albedo,
 solar_luminosity => solar_luminosity,
 solar_change => solar_change,
 scenario => scenario,
 tick => 0)
```

Datseris:
The **documentation** dependencies are completely independent from the actual Agents.jl dependencies (see the file in `docs/Project.toml`) and there we can add any dependency, and I don't think we should remove any, especially the very handy `@dict` macro.

src/models/sir.jl
```diff
@@ -0,0 +1,182 @@
+using Distributions: Poisson, DiscreteNonParametric
+using DrWatson: @dict
```
Libbum, line 2: Drop this

src/models/sir.jl
```diff
@@ -0,0 +1,182 @@
...
+        migration_rates,
+        β_und,
+        β_det,
+        infection_period,
+        reinfection_probability,
+        detection_time,
+        death_rate,
+        Is,
+        seed
+    )
```
Libbum, line 65:
```julia
params = Dict(
        Ns => Ns,
       migration_rates  => migration_rates,
       β_und =>  β_und,
       β_det  => β_det,
        infection_period  => infection_period,
       reinfection_probability  => reinfection_probability,
        detection_time => detection_time,
       death_rate  => death_rate,
       ls => Is,
       seed => seed
    )```

Libbum:
@Datseris, but these are in the models module, not the docs.

kavir1698:
> One question about https://github.com/fbanning/Agents.jl/blob/Extend-Models-submodule/src/models/wright-fisher.jl :
Is it really necessary to have those two functions with different names *_neutral and *_selection ? I think it might be possible to just define the function twice with the same name - once with and once without :traitas a parameter.
If that's correct, this could simplify the module a bit and might demonstrate usage of methods/multiple dispatch for stepping functions.

"model step" functions have always had a single argument, the model object. I think it is better not to break this convention in this example.

Datseris:
> @Datseris, but these are in the models module, not the docs.

Sorry , you are right, and I agree with your suggestion to drop it there.

fbanning:
Ok, I will apply those changes in a moment.

Libbum:
@fbanning, is it possible for you to enable "Allow edits from maintainers" on your `Extend-Models-submodule` branch? It would help speed up these travis conflicts.

The next hurdle is
> invalid redefinition of constant PoorSoul

The sir and social distancing models both have the same struct name. I'd suggest keeping the name of the one in the sir model and changing the name in social distancing. Perhaps `SocietyMember`?

fbanning:
grafik

It's already enabled?

Libbum:
Cool, thanks. I'll try again if something else fails :)

Libbum:
I'm going to do some offline work to try and get this building. Will push up some fixes soon.

Libbum:
That should fix the build errors. Docs errors may still happen though...

Libbum:
Things to finalise here.

- [ ] Wealth model is the 1D version, in the example a 2D version also exists. @Datseris, @kavir1698: Should we add the 2D version into the model library or just state in the example file the 1D version is available in the model library only?
- [x] SIR model needs lightgraphs. Hopefully my last commit fixes that.
- [x] HK needs StatsBase for mean. As above.
- [ ] Should we rename `HK()` to `hk()`? The SIR model is `sir()`...
- [x] Types for Daisyworld and Predator-prey need to be exported. Fixed
- [x] SimpleCell needs to be fixed in the model and the example. 

Perhaps more. I'll update this section if and when I find them. Want to make sure these changes are correct first.

Datseris:
I would actually put the 2D version of the wealth model in the `Models`. :D

Libbum:
OK, I've gone through and verified all the changes work in the documentation, and validated that using the `Model` generated `model, agent_step!, model_step!` values can re-generate results from the example texts.

Where that hasn't been the case is in my comment above.

@fbanning: The Wealth Model changes can be done now, although the other issues will need comments from the other maintainers before we proceed.

That being said, this is essentially done now - most of this looks like minor changes and design choice.

Datseris:
* Test coverage is meaningless to me, since it counts whether a functions is **called** at the tests. Whether it is tested properly is a completely different story, so I just don't care about coverages.
* yes for using `reverse(sincos(agent.orientation))`
* I would be happy to have a game of life in `Models`.
* The examples explicitly define `Daisy` themselves. Maybe in the docs of e.g. `daisyworld` we should just say "use `using Agents.Models: Daisy, Land` to access agent types".
* Yes for naming all models with lower-case.

Libbum:
Having issues pushing changes at the moment, but all of these design decisions are taken care of.
@fbanning: I'll leave the [game of life](https://github.com/JuliaDynamics/Agents.jl/blob/master/examples/game_of_life_2D_CA.jl) inclusion up to you? Exactly the same as the other ones!

Libbum:
@kavir1698, @Datseris: Either of you have any more comments on this?

kavir1698:
Looks good to me

Datseris:
to me too

Libbum:
Nice catch on that last bit @fbanning. Thanks for your help on this one - not to complex, but a lot of finicky changes. Very much appreciated!

fbanning:
Glad I could help! :)

PR #267, opened 2020-08-01
add daisies interactive app

Datseris:


examples/daisyworld.jl
```diff
@@ -335,3 +336,65 @@ p2 = plot(agent_df[!, :step], agent_df[!, aggname(adata[3])], ylabel = "temperat
...
+
+
+# ## Interactive scientific research
+# Julia is an interactive language, and thus everything that you do with Agents.jl can be
+# considered interactive. However, we can do even better by using our interactive application.
+# In this example, we care about exploring some time-dependent forcing of the system, and
+# so far we have been pre-describing it. But with using the interactive application,
+# we can control by ourselves, in real time, how much to change the solar forcing.
+
+# So, let's use [`interactive_abm`](@ref)!
```
Libbum, line 348: This link is undefined. Do we have to make it `InteractiveChaos.interactive_abm`?

examples/daisyworld.jl
```diff
@@ -335,3 +336,65 @@ p2 = plot(agent_df[!, :step], agent_df[!, aggname(adata[3])], ylabel = "temperat
...
+#     mdata = mdata, adata = adata, alabels = alabels, mlabels = mlabels,
+#     scheduler = landfirst # crucial to change model scheduler!
+# )
+# ```
+
+```@raw html
+<video width="100%" height="auto" controls autoplay loop>
+<source src="https://raw.githubusercontent.com/JuliaDynamics/JuliaDynamics/master/videos/agents/daisies.mp4?raw=true" type="video/mp4">
+</video>
+```
```
Libbum, line 400: I can't tell why, but this raw block doesn't show up in the preview docs. Link seems to be there though?

Also, in the video you click update a lot but it's not that clear as to why. Perhaps an additional sentence or two here to describe what you're doing in the demo?

examples/daisyworld.jl
```diff
@@ -335,3 +336,65 @@ p2 = plot(agent_df[!, :step], agent_df[!, aggname(adata[3])], ylabel = "temperat
...
 
 plot(p, p2, p3, layout = (3, 1))
+
+
+# ## Interactive scientific research
+# Julia is an interactive language, and thus everything that you do with Agents.jl can be
+# considered interactive. However, we can do even better by using our interactive application.
+# In this example, we care about exploring some time-dependent forcing of the system, and
+# so far we have been pre-describing it. But with using the interactive application,
+# we can control by ourselves, in real time, how much to change the solar forcing.
```
Libbum, line 346: Suggestion:

In this example, rather than describing what solar forcing we want to investigate before hand, we use the interactive application, to control by ourselves, in real time, how much solar forcing is delivered to daisyworld.

examples/daisyworld.jl
```diff
@@ -335,3 +336,65 @@ p2 = plot(agent_df[!, :step], agent_df[!, aggname(adata[3])], ylabel = "temperat
...
+
+# So, let's use [`interactive_abm`](@ref)!
+
+# ```julia
+# using InteractiveChaos, Makie, Random
+# Random.seed!(165)
```
Libbum, line 352: Do we need the Random seed to be explicit here?

examples/daisyworld.jl
```diff
@@ -335,3 +336,65 @@ p2 = plot(agent_df[!, :step], agent_df[!, aggname(adata[3])], ylabel = "temperat
...
+# model = daisyworld(; solar_luminosity = 1.0, solar_change = 0.0, scenario = :change)
+# ```
+
+# Thankfully, we have already defined the necessary `adata, mdata` as well as the agent
+# color/shape/size functions, and we can re-use them for the interactive application.
+# Unfortunately, because `InteractiveChaos` uses a different plotting package, Makie.jl, we have
+# to redefine the plotting functions. However, in the near future where AgentsPlots.jl
+# will move to Makie.jl, this will not be necessary.
```
Libbum, line 360: No need to be negative about improvements! A suggested change:

Because `InteractiveChaos` uses a different plotting package: Makie.jl, the plotting functions we define here are slightly different to our normal routine.
AgentsPlots.jl is planning to also move to Makie.jl, so something similar to this syntax will be normal in the near future.

examples/daisyworld.jl
```diff
@@ -335,3 +336,65 @@ p2 = plot(agent_df[!, :step], agent_df[!, aggname(adata[3])], ylabel = "temperat
...
+# const landcolor = cgrad(:thermal)
+# daisycolor(a::Land) = to_color(landcolor[(a.temperature+50)/150])
+#
+# daisyshape(a::Daisy) = :circle
+# daisysize(a::Daisy) = 0.6
+# daisyshape(a::Land) = :rect
+# daisysize(a::Land) = 1
+# ```
+
+# The only difference is that we make a parameter container for surface albedo and
```
Libbum, line 374: The only difference is -> The only addition is ?

Libbum:
Looking good, have added some suggestions and some minor things that need fixing.

Also, I don't know if its where you started this branch from, or something else, but take a look at the doc build on Travis. The Schelling model completely fails to build on this branch. We'll need to identify the cause of that before merging.

Libbum:
Just pushed a fix for the Schelling issue. 1166d05ffca made a `s/properties/adata/g`, but that was a bit greedy. `properties` is still required for the `model`.

examples/daisyworld.jl
Datseris: Yes. Being explicitly is always better and I am not sure why the decision to hide `seed!` in the previous parts of the code was made.

Libbum: Simple enough to drop that in the docs then. We can make it consistent across the board.

Issue #247, opened 2020-05-03
possible test failure in upcoming Julia version 1.5 


rfourquet:
A PkgEval run for a Julia [pull request](https://github.com/JuliaLang/julia/pull/29240) which changes the generated numbers for `rand(a:b)` indicates that the tests of this package might fail in Julia 1.5 (and on Julia current master branch).

Also, you might be interested in using the new [StableRNGs.jl](https://github.com/rfourquet/StableRNGs.jl) registered package, which provides guaranteed stable streams of random numbers across Julia releases.

Apologies if this is a false positive. Cf. https://github.com/JuliaCI/NanosoldierReports/blob/ab6676206b210325500b4f4619fa711f2d7429d2/pkgeval/by_hash/52c2272_vs_47c55db/logs/Agents/1.5.0-DEV-87d2a04de3.log

Libbum:
Thanks for bringing this to our attention! I'll have a closer look.

Libbum:
OK, so the TLDR for this is that `Random.seed!(N)` in julia 1.5 may yield a different output, thus our tests will break when using it.

The underlying algorithm changes.  

We can keep an eye on this once we start testing against 1.5.

Libbum:
Travis only swapped to 1.5 today, so this hasn't been a problem up until now. 

I suggest we add `1.4.2` to our travis, and allow `1` to fail for the time being until we sort this out.

Libbum:
Closed in #270.

PR #270, opened 2020-08-06
Schelling app, fix docs and test suite

Datseris:
The dev docs don't display the Daisyworld app video I've added in a recent PR. Better fix this in this PR. I'll monitor the Travis build...


Datseris:
Oh yeah, some tests fail on 1.5 because of the changed RNG HAHA! Gotta fix that too!

Datseris:
Oh man, these there are A LOT of failures... this is gonna need some effort.

Libbum:
Don't alter the tests! We can use [StableRNGs](https://github.com/rfourquet/StableRNGs.jl) for our seeds and it should be fine to keep them as-is.

Libbum:
Getting all tests to pass here will Close #247. #271 has been closed in favour of this.

Datseris:
> I suggest a rewrite of the paramscan section with something a little less complex, like the example model already in collect_tests. Sound OK?

Yes, I actually also was wondering why we would have an entire forest fore definition file. Notice that we now have a `Models` module, we should have to define _any_ model in the tests.

> Scheduler tests are going to be complicated - many are testing the random activation portion of the scheduler, along-side the order of operations that currently depends on a seed. This could be one place that StableRNGs is useful?

Datseris:
Alright, I've made many things deterministic. Tests pass locally for me, let's see what happens online. I've changed the `interaction_tests.jl` file, but I am wondering. Do we need it? Isn't everything that it tests already tested in `api_tests.jl` ? @Libbum I don't remember what you told me during the video call :D

Libbum:
We can drop interaction_tests since it isn't really useful for us, but the paramscan testset in collect_tests need to be adapted. Currently they are commented out.

Datseris:
Uuuuh, why do we have test failings with 
```
Expression: aggname(x_position, length) == "length_x_position"

   Evaluated: "length_getfield(Main, Symbol(\"#x_position#120\"))()" == "length_x_position"
```

Datseris:
It was already announced that 1.6 will be the next LTS version of Julia and 1.0 is dropped. Shall we just drop the 1.0 tests already?

kavir1698:
> It was already announced that 1.6 will be the next LTS version of Julia and 1.0 is dropped. Shall we just drop the 1.0 tests already?

Yes!

Libbum:
aggname already has a [1.0 specific override](https://github.com/JuliaDynamics/Agents.jl/blob/245744702254ae9c9dfe8a54aee58f9786e750b3/src/simulations/collect.jl#L198). Weird that the behaviour has changed again???

If we drop 1.0 testing we can also drop this static guard.

Datseris:
Okay but then we should make the julia dependency of Agents.jl to be 1.5 and bump version to 3.4. Also, Why do we use `join([string(agg), string(k)], "_")` instead of just `string(agg)*"_"*string(k)`...?

Libbum:
Don't think there was any particular reason, just semantics. I would be a legacy remnant from when we used `mean(test)` and such with more complex joins. Your suggestion is about 3x faster, so it's a better solution now.

PR #271, opened 2020-08-07
WIP. Upgrade test suite for 1.5

### LINKED ISSUE #247: possible test failure in upcoming Julia version 1.5

Libbum:
Fixes #247 

The first commit *should* be a useful hot-patch for any active PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ os:
 
 julia:
   - 1
+  - 1.4.2
   - 1.0
   # - nightly
 
@@ -19,11 +20,11 @@ after_success:
 
 matrix:
   allow_failures:
-    - julia: nightly
+    - julia: 1
   fast_finish: true
   include:
     - stage: "Documentation"
-      julia: 1
+      julia: 1.4.2
       os: linux
       script:
         - julia -e 'using Pkg; ENV["PYTHON"]=""; py="PyCall"; Pkg.add(py); Pkg.build(); Pkg.add("PyPlot"); Pkg.add("StaticArrays")'


Libbum:
:hankey:

We're going to have issues here.
`StableRNGs` provides a consistent seed across all Julia versions by using a less than state-of-the-art RNG. Not so much of an issue when you're just testing.

Herein lies the rub though: if you don't supply the generator to **all** instances of `rand`/`seed!`/etc, then results are inconsistent.

```julia
rng = StableRNG(658)
...
Random.seed!(rng, 6546) 
agent = Agent7(3, 2, attributes...)
@test add_agent!(agent, model).pos == 9 # Test failed, Evaluated: 3 == 9
```
Fine, probably a different seed...
```julia
rng = StableRNG(658)
...
Random.seed!(rng, 6546) 
agent = Agent7(3, 2, attributes...)
@test add_agent!(agent, model).pos == 3 # Test failed, Evaluated: 7 == 3
```

:confused: Oh...

```julia
function add_agent!(agent::A, model::ABM{A,<:DiscreteSpace}) where {A<:AbstractAgent}
    agent.pos = correct_pos_type(rand(1:nv(model)), model)
    add_agent_pos!(agent, model)
end
```
There's a `rand` in our definition. Since we don't provide it with the stable `rng`, it will use the new 1.5 generator. A quick test verifies this:

```julia
function add_agent!(rng, agent::A, model::ABM{A,<:DiscreteSpace}) where {A<:AbstractAgent}
    agent.pos = correct_pos_type(rand(rng, 1:nv(model)), model)
    add_agent_pos!(agent, model)
end

...
@test add_agent!(rng,agent, model).pos == 3 #Test Passed (repeatable)
```

I really don't want to have to pass around a seed to all of our functions. That's ridiculous right? Especially just for test purposes.

If we don't want to do that, then `StableRNGs` doesn't help here at all. I would suggest as an alternative that we create a second set of tests, mirrored from the first which has results that originate from the same seeds as we have in the current test set, but using the new 1.5 generator. Keep the test sets functioning in parallel, then drop the old set entirely once the 1.6 LTS version of Julia is released.

Thoughts?

Libbum:
One alternative could be extending the base concept of the `model`:

```julia
struct AgentBasedModel{A<:AbstractAgent, S<:SpaceType, F, P, R}
    agents::Dict{Int,A}
    space::S
    scheduler::F
    properties::P
    generator::R
end
```
Where we keep `generator` as a default instance of the current Julia versions RNG, but then override it to a StableRNG in the tests.

Datseris:
Hi Tim,

I don't at all including an RNG in a model.

I think this failings with the change of RNGs is a good wake up call to make better tests. If you see my PR, I have started **altering** some tests; not just "fixing" them. A point to consider is that we can test things that are deterministic and not random. E.g. we can test that `move_agent!` changes the agents location, instead of testing whether it results in exactly `agent.pos == 5`. 

Initially I also thought of using StableRNGs, but you already noted the problems: you have to do it **everywhere** which is just not possible, and I'd much rather have the state of the art RNGs anyways.

I think we should just continue this in #270 and improve the tests to be more deterministic and test less the rng-dependent outcomes.

Libbum:
Cool. Yeah, sorry - I didn't actually read your test changes and assumed you were 'fixing' them. Your process is much better for sure.

PR #272, opened 2020-08-10
Optim example

kavir1698:


Libbum:
We'll need to finalise the test fixes in #270 before we can get a preview for this one, but I can take a look offline and comment soon.

examples/optim.jl
```diff
@@ -0,0 +1,299 @@
+# # Optimizing agent-based models
+
+# Agent-based models (ABMs) are computationally expensive and can have many parameters. Sometimes we need to fine-tune a model's parameters to a specific outcome. Brute-force algorithms can take too long for testing each parameter setting. Even if it was feasbile to run the model for every parameter setting, it would not be enough because ABMs are stochastic and the effect of a parameter setting should be derived from running the model several times and taking its average behavior.
```
Libbum, line 3: They don't need to be computationally expensive _per se_. 

Agent-based models (ABMs) can have many parameters, and sometimes we need to fine-tune a model's parameters to a specific outcome. A model with hundreds of parameters (or more) makes brute-force parameter scanning unfeasible. On top of that, ABMs are stochastic and the effect of setting just one parameter should be derived from running the model several times and assigning the average behavior.

examples/optim.jl
```diff
@@ -0,0 +1,299 @@
...
+
+# Here we show how to use optimizations based on evolutionary algorithms with ABMs ([BlackBoxOptim.jl](https://github.com/robertfeldt/BlackBoxOptim.jl)).
```
Libbum, line 5: Here we show how to use the evolutionary algorithms in [BlackBoxOptim.jl](https://github.com/robertfeldt/BlackBoxOptim.jl) with Agents.jl.

examples/optim.jl
```diff
@@ -0,0 +1,299 @@
...
+    for city in 1:C
+        inds = get_node_contents(city, model)
+        for n in 1:Is[city]
+            agent = model[inds[n]]
+            agent.status = :I # Infected
+            agent.days_infected = 1
+        end
+    end
+    return model
+end
```
Libbum, line 82: Can some/all of this be replaced by a call to our new Model Library?

examples/optim.jl
```diff
@@ -0,0 +1,299 @@
...
+function recover_or_die!(agent, model)
+  if agent.days_infected ≥ model.infection_period
+      if rand() ≤ model.death_rate
+          kill_agent!(agent, model)
+      else
+          agent.status = :R
+          agent.days_infected = 0
+      end
+  end
+end
```
Libbum, line 164: Can some/all of this be replaced by a call to our new Model Library?

examples/optim.jl
```diff
@@ -0,0 +1,299 @@
...
+# We let the optimization algorithm change parameters to minimize the number of infected individuals. Note that we can limit the allowed range for each parameter separately.
+
+result = bboptimize(cost, SearchRange = [(0.0, 1.0), (0.0, 1.0), (0.0, 1.0), (0.0, 1.0), (7.0, 13.0),(0.0, 1.0),(2.0, 6.0)], NumDimensions = 7, MaxTime=20)
+best_fitness(result)
+
+# The fraction of the infected is down to 11%. Parameter values that give this result are:
+
+best_candidate(result)
+
+# We notice that the death rate is 96%, and transmission rates have also increased, while reinfection probability is much smaller. When all the infected indiduals die, infection doesn't transmit. Let's modify the cost function to give us a better solution that doesn't involve so much death.
```
Libbum, line 221: Let's modify the cost function to give us a better solution that doesn't involve so much death.
-> Let's modify the cost function to also keep the mortality rate low.

examples/optim.jl
Datseris: All please. The focus of this docpage is well outside model creation.

examples/optim.jl
kavir1698: Models with much fewer parameters can take too long to explore the parameter space.

kavir1698: Size of the parameter space of a simple model with 10 parameters and 10 states per parameter is 10^10. Even if each simulation takes only one second, exploring the whole space would take more than 300 years.

examples/optim.jl
kavir1698: You had decided not to include the SIR model in the Model library.

examples/optim.jl
Libbum: Oh, certainly. Just that they are not *computationally expensive* as a fundamental aspect of them. We have plenty of examples that show that this is not the case.

examples/optim.jl
Libbum: Ah! True!!!

Libbum: Let's chat about this tomorrow then perhaps.

Datseris: :( that is very unfortunate. How necessary is it for you @kavir1698 to use this graph based SIR model for such an example? I just feel like of having several pages of system definitions will make reading this page much harder than it has to. The main point is how to use BlackBoxOptim.

examples/optim.jl
kavir1698: Well, comparing them with analytical solutions, they are almost always more expensive, aren't they?

Libbum: If you have one parameter `x` and want to test `for x in 0:10`.

kavir1698: > What does "state per parameter" mean?

It means that a parameter can take 10 different values, for example.

examples/optim.jl
kavir1698: Among the examples we have, this model is the best candidate for this analysis. Other ones either have too few parameters, or are just not that interesting. One solution is to put these definitions in a different file, and reference to that file?

examples/optim.jl
Datseris: Oooooh that is what you mean. So "states" = "possible values"

Datseris: I think using `state` here is kind of confusing, because "state space" is used (in the lingo of dynamical systems) to describe the entire state space (which for the SIR model is 1-dimensional set of three values :S, :I, :R) .

examples/optim.jl
Libbum: I second @kavir1698's analysis here @Datseris. Our other examples don't provide us with enough expressiveness.

Might be worth trying Ali. Take the [sir.jl](https://github.com/JuliaDynamics/Agents.jl/blob/409c359d2431dc2991076c0f497ee192259cf61c/src/models/sir.jl) file we dumped in #269 and import it with a single line. Something like:

```julia
include("sir.jl")
model, agent_step!, model_step! = sir()
```
Here we use the SIR model discussed in our [Examples documentation](https://juliadynamics.github.io/Agents.jl/stable/examples/sir/). For the complete implementation, see [`sir.jl`](./sir.jl).

Datseris: Sounds fair, but please replace the link above with `[SIR model for the spread of COVID-19](@ref)` so that the crosslink is generated automatically (and can work in any version of docs)

examples/optim.jl
kavir1698: Fair enough. Parameter state for parameter categories is commonly used in the context of graphical models. Tim understood me because he is also reading about such models.

examples/optim.jl
kavir1698: I am going to to create a new file for this model because some functions are slightly different. The general logic is the same, so we can refer the reader to the examples documentation.

examples/optim.jl
Datseris: If you feel "state" is more appropriate then feel free to use it instead.

examples/optim.jl
```diff
@@ -0,0 +1,219 @@
+# # Optimizing agent-based models
+
+# Agent-based models (ABMs) are computationally more expensive than analytical models, and can have many parameters. Sometimes we need to fine-tune a model's parameters to a specific outcome. Brute-force algorithms can take too long for testing each parameter setting. Even if it was feasbile to run the model for every parameter setting, it would not be enough because ABMs are stochastic and the effect of a parameter setting should be derived from running the model several times and taking its average behavior.
```
Datseris, line 3: > Agent-based models (ABMs) are computationally more expensive than analytical models

This sentence is straight-out false, since it is **absolute**, and for example weather models are also "analytical". But more importantly, what is the point of this sentence? Why is there a need to compare ABMs with analytical models here?

examples/optim.jl
```diff
@@ -0,0 +1,219 @@
...
+# Here we show how to use the evolutionary algorithms in [BlackBoxOptim.jl](https://github.com/robertfeldt/BlackBoxOptim.jl) with Agents.jl, to optimize the parameters of an epidemiological model (SIR). We explain this model in detail in [SIR model for the spread of COVID-19](@ref). For brevity here, we just import it.
+
+cd(@__DIR__) #src
+cd("../../../examples/") #src
+include("siroptim.jl") ## From the examples directory
+
+# Now we need to define a cost function. The cost function takes as agruments the model parameters that we want to tune, here migration rate, death rate, transmission rate when an infected person has been (not) detected (`β_det`, `β_und`), infection period, reinfection probability, and time until the infection is detected. The function returns one or more numbers as the objective to be minimized. Here, we try to minimize the number of infected people after 50 days.
+
+using BlackBoxOptim
+import Statistics: mean
```
Datseris, line 14:
```suggestion
using Statistics: mean
```

Datseris: `import` is used when you want to extend a method.

examples/optim.jl
```diff
@@ -0,0 +1,219 @@
...
+function cost(x)
+    migration_rate,
+    death_rate,
+    β_det,
+    β_und,
+    infection_period,
+    reinfection_probability,
+    detection_time = x
+    C = 3
+    params = create_params(
```
Datseris, line 25: What's `create_params`?

examples/optim.jl
```diff
@@ -0,0 +1,219 @@
...
+        Is = ones(Int, C),
+        β_und = [β_und for i in 1:C],
+    )
+
+    model = model_initiation(; params...)
+
+    infected_fraction(model) =
+        count(a.status == :I for a in values(model.agents)) / nagents(model)
+    _, data = run!(
+        model,
```
Datseris, line 43: there seems to be some missalignment here

Datseris: I am not sure why we can't directly pass `x` to model creation...? It seems we do a lot of unecessary `reinfection_probability, ... = x` and then again `reinfection_probability = reinfection_probability` why is there so much duplication?

Datseris:
Most codeblocks in the documentation do not show output, meaning that there is an error in the code.

Libbum:
I'm aware. Working on it offline. The documenter is trying to import the siroptim file from its base working directory.

examples/optim.jl
Libbum: It's something imported from the siroptim file.

examples/optim.jl
Libbum: Looking into better refactors of this and the other exposed functions from siroptim. If this needs to happen due to the optimisation steps, it should certainly be documented further.

Libbum:
33247d4 drops the `create_params` function and refactors a bit to make things simpler. One complexity to the code is that BBOptim can't use named tuples as parameters, so rather than being able to splat x  (`model = model_initiation(; x...)`),  the expansion needs to be explicit.

I'll work on the text now. There's a bit to do there to make it more reasonable.

Libbum:
I've done a bit of a re-write/edit of the text now. It's to the point that I now understand all the steps involved and could re-implement in on a model of my own. @kavir1698: can you check to see if I've changed anything that doesn't make sense now?
Both: take a read and let me know if you have comments. If not, I'm pretty sure it's ready to merge.

examples/optim.jl
kavir1698: An adverb, such as "often", would solve the objection. The sentence hints at a reason why ABM parameter spaces are more difficult explore and optimize. But now if you and Tim don't like the sentence, as he as already excluded it, I don't mind not having it there.

kavir1698:
> @kavir1698: can you check to see if I've changed anything that doesn't make sense now?

The changes make sense. Good to go.

PR #278, opened 2020-08-14
Allow conditional aggregation

### LINKED ISSUE #275: Conditional agent data aggregation?
Let's say I have some agents on a `GridSpace` and I have a fairly complicated model where my agents have a lot of properties.

Let's also say there is a property `a` that I realy care about and I want to data collect and aggregate, but I want to do it *conditionally*. Specifically, I want to know the average value of `a` for all agents whose x position is greater than 5. How do we do this with the current interface?

One way is to *not* aggregate agent data, and instead collect all agent positions and `a`. Then, do a DataFrames-based operation to collapse property `a` to its mean but only when some other column `pos` satisfies some other property.

Libbum:
Closes #275.

The only issue I see here is that we don't separate our naming convention. There's no indication that `mean_weight` is over all or just some agents. Should we extend aggname and our column naming function to include something like `:conditional_mean_weight`? 

## Benchmarks

Before changes:
```julia
julia> props = [(:weight, mean)]
1-element Array{Tuple{Symbol,typeof(mean)},1}:
 (:weight, Statistics.mean)

julia> @benchmark collect_agent_data!(df, model, props, 1)
BenchmarkTools.Trial: 
  memory estimate:  128 bytes
  allocs estimate:  8
  --------------
  minimum time:     220.822 ns (0.00% GC)
  median time:      247.047 ns (0.00% GC)
  mean time:        332.252 ns (18.49% GC)
  maximum time:     192.535 μs (99.86% GC)
  --------------
  samples:          10000
  evals/sample:     506
```

After:
```julia
julia> props = [(:weight, mean)]
1-element Array{Tuple{Symbol,typeof(mean)},1}:
 (:weight, Statistics.mean)

julia> @benchmark collect_agent_data!(df, model, props, 1)
BenchmarkTools.Trial: 
  memory estimate:  144 bytes
  allocs estimate:  9
  --------------
  minimum time:     214.961 ns (0.00% GC)
  median time:      229.436 ns (0.00% GC)
  mean time:        296.531 ns (12.78% GC)
  maximum time:     122.061 μs (99.77% GC)
  --------------
  samples:          10000
  evals/sample:     545
```

With conditional aggregation:
```julia
julia> ytest(agent) = agent.pos[2] > 5
ytest (generic function with 1 method)

julia> props = [(:weight, mean, ytest)]
1-element Array{Tuple{Symbol,typeof(mean),typeof(ytest)},1}:
 (:weight, Statistics.mean, ytest)

julia> @benchmark collect_agent_data!(df, model, props, 1)
BenchmarkTools.Trial: 
  memory estimate:  144 bytes
  allocs estimate:  6
  --------------
  minimum time:     151.653 ns (0.00% GC)
  median time:      167.491 ns (0.00% GC)
  mean time:        240.460 ns (18.29% GC)
  maximum time:     107.437 μs (99.81% GC)
  --------------
  samples:          10000
  evals/sample:     825
```


Libbum:
Just added a failing test to show a (contrived) example of what may happen if we don't do something about aggname.

src/simulations/collect.jl
```diff
@@ -174,38 +181,60 @@ function init_agent_dataframe(model::ABM, properties::Vector{<:Tuple})
...
+
+# Normal aggregates
+function _add_col_data!(col::AbstractVector{T}, property::Tuple{K,A}, agent_iter) where {T,K,A}
+    k, agg = property
+    res::T = agg(get_data(a, k) for a in agent_iter)
+    push!(col, res)
+end
+
+# Conditional aggregates
+function _add_col_data!(col::AbstractVector{T}, property::Tuple{K,A,<:Function}, agent_iter) where {T,K,A}
```
Datseris, line 235:
```suggestion
function _add_col_data!(col::AbstractVector{T}, property::Tuple{K,A,F}, agent_iter) where {T,K,A,F}
```

Limiting to `Function` excludes functors from being used (very unlikely, but also there is no gain in putting `<:Function` instead of `<:Any`)

Datseris:
My only beef with the current naming scheme `conditional_...` is that it doesn't allow 
```julia
ytest(agent) = agent.pos[2] > 5
ytest2(agent) = agent.pos[2] <= 5
props = [(:weight, mean, ytest), (:weight, mean, ytest2)]
```
Why not continue doing what we do already, and just add a third string to `props`, so `join(string(a) for a in tuple, '_')`

Libbum:
Sure, we can do that. I guess we already have the warning about anonymous functions in the docs. I'll drop the anonymous tests and change this.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["Ali Vahdati", "George Datseris", "Tim DuBois"]
-version = "3.4"
+version = "3.5"
```
Datseris, line 4: Can you please add a changelog entry?

Libbum: Yep. Forgot.

Libbum:
What about a naming convention like this: "mean_weight|x_pos", where you have agg_k|cond, so that you can know that this average is 'dependent on `x_pos`'?

Datseris:
It seems like a nice name, but I still have in my head the convseration we had some time ago with a user that convinced me to change the convention from `agg(k)` to `agg_k`... but I don't remember in which issue it was! Maybe something similar would happen in the current case of using `|`? I need to find this conversation, because I don't remember what was the problem.

Libbum:
Ah. Yes, we will have to do `mean_weight_x_pos`, since you cannot have some dataframe `result` and call `result.mean_weight|x_pos`.

Datseris:
> `result.mean_weight|x_pos`

But this is not valid syntax with respect to the dataframes API, right? Because I think their syntax now is `result[!, "mean_weight|x_pos"]`. In fact it could be that they moved into strings from symbols for such weird column names.

Libbum:
Was that last merge meant to be the other way? Or...?

PR #281, opened 2020-09-01
Deepcopy data when saving to DataFrame

### LINKED ISSUE #280: Problem collecting agent data when agent field is a Vector
I'm working on a model where agents have a vector dynamic state, not related with the position in space, that I wish to measure each step.
A simplified version of my code is
```julia
using Agents, LinearAlgebra, LightGraphs

mutable struct FakeAgent <: AbstractAgent
    id::Int
    pos::Int
    state::Vector{Float64}
end

function fakestep!(agent,model)
    i = agent.id
    j = rand(1:size(agent.state,1))
    x = randn()
    model[agent.id].state[j] += x
    return nothing
end

N = 2
model = ABM(FakeAgent, 
                      GraphSpace(complete_graph(N)); 
                      scheduler=random_activation)

K = 3
# initialize agents to have obvious states: model.agents[i].state = [i, i, i]
for i in 1:N
    a = FakeAgent(i,i, i*ones(K))
    add_agent_pos!(a,model)
end

(adf,_) = run!(model, fakestep!, 10; adata=[:state])

first(adf,6)
```
which gives
```julia
"""
6×3 DataFrame
│ Row │ step  │ id    │ state                       │
│     │ Int64 │ Int64 │ Array{Float64,1}            │
├─────┼───────┼───────┼─────────────────────────────┤
│ 1   │ 0     │ 1     │ [3.71784, -1.2705, 0.75856] │
│ 2   │ 0     │ 2     │ [0.69792, 2.55628, 1.05595] │
│ 3   │ 1     │ 1     │ [3.71784, -1.2705, 0.75856] │
│ 4   │ 1     │ 2     │ [0.69792, 2.55628, 1.05595] │
│ 5   │ 2     │ 1     │ [3.71784, -1.2705, 0.75856] │
│ 6   │ 2     │ 2     │ [0.69792, 2.55628, 1.05595] │
"""
```
Notice how the `state` vector is always the last state after all steps.
If I fix the step function:
```julia
function fakestep_fixed!(agent,model)
    i = agent.id
    j = rand(1:size(agent.state,1))
    x = randn()
    
    new_state = copy(agent.state)
    new_state[j] += x
    
    model[agent.id].state = new_state
    return nothing
end
```
then I get what is expected
```julia
"""
6×3 DataFrame
│ Row │ step  │ id    │ state                   │
│     │ Int64 │ Int64 │ Array{Float64,1}        │
├─────┼───────┼───────┼─────────────────────────┤
│ 1   │ 0     │ 1     │ [1.0, 1.0, 1.0]         │
│ 2   │ 0     │ 2     │ [2.0, 2.0, 2.0]         │
│ 3   │ 1     │ 1     │ [1.0, 1.0, 1.27255]     │
│ 4   │ 1     │ 2     │ [2.0, 2.0, 1.23249]     │
│ 5   │ 2     │ 1     │ [0.22367, 1.0, 1.27255] │
│ 6   │ 2     │ 2     │ [2.32241, 2.0, 1.23249] │
"""
```
I could, instead of rewriting the step function, use `mdata` and collect all the agents at once (works, but is not convenient).
I'm not sure if it is a bug or just a misuse of Julia or DataFrames.

(PS: Code running on Linux, with Julia 1.5.1 and Agents.jl v3.5.0)

Libbum:
Fixes #280 

Whilst we enabled users to save arbitrary data structures into the dataframe, we didn't test that too well. This PR performs a `deepcopy` whenever we take data from the model and store it in the dataframe upon collection.

### DIFF
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -5,8 +5,8 @@ export run!, collect_agent_data!, collect_model_data!,
 ###################################################
 # Definition of the data collection API
 ###################################################
-get_data(a, s::Symbol) = getproperty(a, s)
-get_data(a, f::Function) = f(a)
+get_data(a, s::Symbol) = deepcopy(getproperty(a, s))
+get_data(a, f::Function) = deepcopy(f(a))
 
 should_we_collect(s, model, when::AbstractVector) = s ∈ when
 should_we_collect(s, model, when::Bool) = when
--- a/test/collect_tests.jl
+++ b/test/collect_tests.jl
@@ -1,9 +1,14 @@
 @testset "DataCollection" begin
     function initialize()
         model = ABM(
-        Agent3,
-        GridSpace((10, 10));
-        properties = Dict(:year => 0, :tick => 0, :flag => false),
+            Agent3,
+            GridSpace((10, 10));
+            properties = Dict(
+                :year => 0,
+                :tick => 0,
+                :flag => false,
+                :container => Float64[],
+            ),
         )
         add_agent!((4, 3), model, 0.1)
         add_agent!((7, 5), model, 0.35)
@@ -36,9 +41,18 @@
         @test collect_model_data!(DataFrame(), model, nothing, 1) == DataFrame()
 
         props = [:weight]
-        @test sprint(show, "text/csv", describe(init_agent_dataframe(model, props), :eltype)) == "\"variable\",\"eltype\"\n\"step\",\"Int64\"\n\"id\",\"Int64\"\n\"weight\",\"Float64\"\n"
+        @test sprint(
+            show,
+            "text/csv",
+            describe(init_agent_dataframe(model, props), :eltype),
+        ) ==
+              "\"variable\",\"eltype\"\n\"step\",\"Int64\"\n\"id\",\"Int64\"\n\"weight\",\"Float64\"\n"
         props = [:year]
-        @test sprint(show, "text/csv", describe(init_model_dataframe(model, props), :eltype)) == "\"variable\",\"eltype\"\n\"step\",\"Int64\"\n\"year\",\"Int64\"\n"
+        @test sprint(
+            show,
+            "text/csv",
+            describe(init_model_dataframe(model, props), :eltype),
+        ) == "\"variable\",\"eltype\"\n\"step\",\"Int64\"\n\"year\",\"Int64\"\n"
     end
 
     @testset "aggname" begin
@@ -46,7 +60,7 @@
         @test aggname(:weight, mean) == "mean_weight"
         @test aggname(x_position, length) == "length_x_position"
         @test aggname((x_position, length)) == "length_x_position"
-        ypos(a) = a.pos[2]>5
+        ypos(a) = a.pos[2] > 5
         @test aggname((x_position, length, ypos)) == "length_x_position_ypos"
     end
 
@@ -139,7 +153,7 @@
             agent_step!,
             model_step!,
             365 * 5;
-            when_model = [1, 365*5],
+            when_model = [1, 365 * 5],
             when = false,
             mdata = [:flag, :year],
             adata = [(:weight, mean)],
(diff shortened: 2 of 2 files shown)


Libbum:
I don't know why the docs are failing to build here. Nothing has been changed.

Datseris:
`deepcopy` is a costly function and I don't really knows where it impacts performance... I would be reluctant to merge this before we do some benchmarks that show that the "standard" functionality has not been affected.

Libbum:
Sure. I'll do some runs later on and see what kind of hit this is. I'd assume it'd be minimal to none on our current tests, since it's only happening on the containers that are going into the DataFrame.

Datseris:
just benchmarking `run!` of one of the examples is a good start

Datseris:
> Allow primitives directly, and only deepcopy structures that require it.

I was thinking of that, the problem is, I don't know how to distinguish between the two in a general setting that includes user-defined types.

I am also thinking whether we should do this, instead of the user. Because to be honest the performance drop is more than I expected, it is about 5%.

Libbum:
How would we go about getting the user to do this though? If they're wanting a `Vector` as a result in the `DataFrame`, which comes from an agent, we don't have any form of API they could leverage for that at present.

The only thing that comes to mind would be to have another keyword on `run!` like `deepcopy = true`, which would default to `false`. Then have a discussion about this in the documentation.

Datseris:
> The only thing that comes to mind would be to have another keyword on run! like deepcopy = true, which would default to false. Then have a discussion about this in the documentation.

I agree with this. We can do better actually, and making a keyword whose value is either `identity, copy, deepcopy` (the last only being useful in the case of nested mutable structures, the second only possible if user has defined `copy` for their structs). Then the collection collects `f(thing)` , which for `identity` is just `thing`.

Datseris:
potential name for the keyword: `obtainer`

Libbum:
I will want to write some more test for this, but thought it best to push the current implementation, since I think it got a little more.. invasive than I anticipated.
Mostly in the low level, but perhaps this is not the best way to approach the issue. Thoughts?

src/simulations/collect.jl
```diff
@@ -1,12 +1,20 @@
 export run!, collect_agent_data!, collect_model_data!,
        init_agent_dataframe, init_model_dataframe, aggname,
-       should_we_collect
+       should_we_collect, Identity, Copy, DeepCopy
+
+abstract type Obtainer end
+struct Identity <: Obtainer end
+struct Copy <: Obtainer end
+struct DeepCopy <: Obtainer end
```
Datseris, line 8: Tim, why go through make *new* names, specifically, introduce new types? The user can directly give `copy` or `deepcopy` to a keyword `obtainer`, and we can just use this value to the functions. E.g. we have
```julia
get_data(a, f::Function, obtainer = identity) = obtainer(f(a))
get_data(a, s::Symbol, obtainer = identity) = obtainer(getproperty(a, s))
```

Datseris: and `obtainer = identity` becomes a keyword of `run!`.

Libbum: Oh. That's what you meant by identity! `identity()`, not just a keyword named identity. 

Yep. Sounds far better. I'll revert a bunch of this.

Datseris: haha, sorry, I should have been clearer!

src/simulations/collect.jl
```diff
@@ -1,20 +1,12 @@
...
 ###################################################
 # Definition of the data collection API
 ###################################################
-get_data(a, p, o::Type{O}) where {O<:Obtainer} = _get_data(a, p)
-get_data(a, p, o::Type{Copy}) = copy(_get_data(a, p))
-get_data(a, p, o::Type{DeepCopy}) = deepcopy(_get_data(a, p))
-_get_data(a, s::Symbol) = getproperty(a, s)
-_get_data(a, f::Function) = f(a)
+get_data(a, s::Symbol; obtainer = identity) = obtainer(getproperty(a, s))
+get_data(a, f::Function; obtainer = identity) = obtainer(f(a))
```
Datseris, line 9: For performance reasons, low-level functions that participate in hot loops should avoid the usage of keywords: `; -> ,`

src/simulations/collect.jl
```diff
@@ -65,10 +57,11 @@ By default both keywords are `nothing`, i.e. nothing is collected/aggregated.
   then data are collect if `s ∈ when`. Otherwise data are collected if `when(model, s)`
   returns `true`. By default data are collected in every step.
 * `when_model = when` : same as `when` but for model data.
-* `obtainer = Identity` : with 2 other options, `Copy` & `DeepCopy`. `Copy` is useful for
-when you wish to save a container (e.g. an `Array`) to the DataFrame, or a `struct` that
-you have implemented [Base.copy](https://docs.julialang.org/en/v1/base/base/#Base.copy) for.
-`DeepCopy` is for when you need to save a
+* `obtainer = identity` : the method used to transfer information from the `model` to
```
Datseris, line 60: I'd suggest:

"to transfer collected data to the `DataFrame`. Typically only change this to `copy` if some data are mutable containers (e.g. `Vector`) which change during evolution, or `deepcopy` if some data are nested mutable containers."

(but keep all hyperlinks, they rock!)

Datseris:
Seems good to me, only minor comments. Don't forget the CHANGELOG+Project.toml though!

Libbum:
I'll add a few more tests tomorrow, other than those, this is probably good to go, but the document build issue is going to cause us grief here too.

Libbum:
OK, problem is forest_fire.

Libbum:
:fireworks: :champagne: 
That fixed it!
Propose this guy is merged and we drop any travis/docs alterations other than the ones you've explicitly added from #282

PR #282, opened 2020-09-01
add docs for struct scheduler

Datseris:


Libbum:
This documentation build error is infuriating. I can't trace it.

Libbum:
FYI: JuliaPlots/Plots.jl#2945
Coming from [this line](https://github.com/JuliaDynamics/Agents.jl/blob/39390112e79c3bf21b47e50cdac8da470c34758a/examples/diffeq.jl#L373).

Libbum:
1.6.2 of Plots just got merged with a patch, so docs should work again now.

Datseris:
I've re-run tests but still fail (again at the docbuiild)

Datseris:
Nah, no, I'll try to make passing tests.

Libbum:
It's hitting a timeout now, not the Plots error. Try a restart perhaps. Otherwise we may need to increase the `travis_wait` value.

Datseris:
> Otherwise we may need to increase the travis_wait value

Where is this done?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["Ali Vahdati", "George Datseris", "Tim DuBois"]
-version = "3.5"
+version = "3.5.1"
```
Libbum, line 4: I'm working on #281 at the moment. We may need to bump to 3.6 with that.

Datseris: Alright, then let's first merge #281 and then I'll rebase this one to the master

Libbum:
Jeez. 45 minutes and the docs still fail. What's made them so heavy all of a sudden?

Project.toml
Libbum: Think it'll be fine with a patch bump with this new implementation.

Libbum:
:hankey: 

We exceed the travis limit now. We're going to have to find out whats taking so long to compile. Maybe separate it out into multiple jobs or something.

Datseris:
Its okay, focus on #281 , I'll finish this PR once the other is done.

PR #283, opened 2020-09-03
Create opinion model

kavir1698:
Here is a new model example I created today, using a model on hash.ai

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -80,6 +80,7 @@ pages = [
         "Daisyworld" => "examples/daisyworld.md",
         "Predator-Prey" => "examples/predator_prey.md",
         "Bacteria Growth" => "examples/growing_bacteria.md"
+        "Opinion spread" => "examples/opinion_spread.md"
         ],
     "Predefined Models" => "models.md",
     "API" => "api.md",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -79,7 +79,7 @@ pages = [
         "Flocking" => "examples/flock.md",
         "Daisyworld" => "examples/daisyworld.md",
         "Predator-Prey" => "examples/predator_prey.md",
-        "Bacteria Growth" => "examples/growing_bacteria.md"
+        "Bacteria Growth" => "examples/growing_bacteria.md",
         "Opinion spread" => "examples/opinion_spread.md"
         ],
     "Predefined Models" => "models.md",
--- a/examples/opinion_spread.jl
+++ b/examples/opinion_spread.jl
@@ -37,7 +37,6 @@ function adopt!(agent, model)
 	neighbor = rand(space_neighbors(agent, model))
 	matches = model[neighbor].opinion .== agent.opinion
 	nmatches = count(matches)
-	## Adopt a different opinion w/ calculated probability
 	if nmatches < model.nopinions && rand() < nmatches/model.nopinions
 		switchId = rand(findall(x-> x==false, matches))
 		agent.opinion[switchId] = model[neighbor].opinion[switchId]
--- a/examples/opinion_spread.jl
+++ b/examples/opinion_spread.jl
@@ -37,7 +37,7 @@ function adopt!(agent, model)
 	neighbor = rand(space_neighbors(agent, model))
 	matches = model[neighbor].opinion .== agent.opinion
 	nmatches = count(matches)
-	# Adopt a different opinion w/ calculated probability
+	## Adopt a different opinion w/ calculated probability
 	if nmatches < model.nopinions && rand() < nmatches/model.nopinions
 		switchId = rand(findall(x-> x==false, matches))
 		agent.opinion[switchId] = model[neighbor].opinion[switchId]
--- a/examples/opinion_spread.jl
+++ b/examples/opinion_spread.jl
@@ -1,0 +1,92 @@
+# # Opinion spread
+
+# This is a simple model of how an opinion spreads through a community. Each individual has a number of opinions as a list of integers. They can change their opinion by changing the numbers in the list.
+
+# Agents can change their opinion at each step. They choose one of their neighbors randomly, and adopt one of the neighbor's opinion. They are more likely to adopt their neighbors opinion if the share more opinions with each other.
+
+using Agents
+using AgentsPlots
+using Random
+
+# ## Building the model
+# ### 1. Model creation
+
+mutable struct Citizen <: AbstractAgent
+	id::Int
+	pos::Tuple{Int, Int}
+	stabilized::Bool
+	opinion::Array{Int, 1}
+	prev_opinion::Array{Int, 1}
+end
+
+function create_model(;dims=(10, 10), nopinions=3, levels_per_opinion=4)
+	space = GridSpace(dims, periodic=true, moore=true)
+	properties = Dict(:nopinions=>nopinions) 
+	model = AgentBasedModel(Citizen, space, scheduler=random_activation, properties=properties)
+	for cell in 1:nv(model)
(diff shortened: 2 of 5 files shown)


examples/opinion_spread.jl
```diff
@@ -0,0 +1,92 @@
...
+# Here is an animation that shows change of agent opinions over time. The first three opinions of an agent determines its color in RGB.
+levels_per_opinion = 3
+ac(agent) = RGB((agent.opinion[1:3] ./ levels_per_opinion)...)
+model = create_model(nopinions=3, levels_per_opinion=levels_per_opinion)
+anim = @animate for sp in 1:500
+  step!(model, agent_step!)
+  p = plotabm(model, ac=ac, as=12)
+  title!(p, "Step $(sp)")
+end
+gif(anim)
```
Datseris, line 92: If you save this, then you can re-use it at the start of the file similarly with how we show the animation at the start for e.g. Schelling

Datseris:
You can also consider putting this in the `Models`.

examples/opinion_spread.jl
```diff
@@ -0,0 +1,92 @@
...
+	prev_opinion::Array{Int, 1}
+end
+
+function create_model(;dims=(10, 10), nopinions=3, levels_per_opinion=4)
+	space = GridSpace(dims, periodic=true, moore=true)
+	properties = Dict(:nopinions=>nopinions) 
+	model = AgentBasedModel(Citizen, space, scheduler=random_activation, properties=properties)
+	for cell in 1:nv(model)
+		add_agent!(cell, model, false, rand(1:levels_per_opinion, nopinions), rand(1:levels_per_opinion, nopinions))
+	end
```
Libbum, line 28: `fill_space!` could be used here

examples/opinion_spread.jl
```diff
@@ -0,0 +1,92 @@
...
+
+# ## Running the model
+
+levels_per_opinion=4
+model = create_model(nopinions=3,levels_per_opinion=levels_per_opinion)
+
+"run the model until all agents stabilize"
+rununtil(model, s) = count(getproperty.(values(model.agents), :stabilized)) == prod(model.space.dimensions)
+
+agentdata, _ = run!(model, agent_step!, dummystep, rununtil, adata=[(:stabilized, count)])
```
Libbum, line 73: I think this highlights a gap in our `run!` constructors. We state that `model_step!` is optional, so `run!(model, agent_step!)` is possible, but when we want to overload `step`, it's necessary to use `dummystep` explicitly. Would be nice if we can manage to remove this.

examples/opinion_spread.jl
Libbum: Also, might be a good idea to break the loop with `rununtil`. For me the opinions stopped changing at around 240, but then the animation continues all the way to 500.

examples/opinion_spread.jl
Libbum: #284

docs/make.jl
```diff
@@ -80,6 +80,7 @@ pages = [
         "Daisyworld" => "examples/daisyworld.md",
         "Predator-Prey" => "examples/predator_prey.md",
         "Bacteria Growth" => "examples/growing_bacteria.md"
```
Libbum, line 82: Need a `,` at the end of this line otherwise the docs will fail

examples/opinion_spread.jl
kavir1698: I want each agent to have a different initial population, but `fill_space!` uses one set of random numbers for all agents.

Datseris: You can give a function of `position` to `fill_space!`.

examples/opinion_spread.jl
Datseris: `model_step!` is only optional in the case of `n` being an integer. If `n` is a function `model_step` is mandatory. I've changed this a while back in the docs as well. Did I forget to update the docs of `run!`? Because the docs of `step!` are clear about this.

Datseris: @Libbum see my comment above. I don't think #284 is necessary.

Libbum: Oh, then why is #284 possible? But yeah, `run!(model, agent_step! [, model_step!], n; kwargs...)` is what's listed currently.

Datseris: Technically it isn't. Your method `run!(model, agenstep, modelstep)` conflicts with `run!(model, agentstep, n)` in the case `n` is a function.

Datseris:
https://juliadynamics.github.io/Agents.jl/previews/PR283/examples/opinion_spread/

I don't see any figures, @kavir1698 I think you need to check the build logs

examples/opinion_spread.jl
```diff
@@ -0,0 +1,97 @@
...
+	return model
+end
+
+# ### 2. Stepping functions
+
+function adopt!(agent, model)
+	neighbor = rand(space_neighbors(agent, model))
+	matches = model[neighbor].opinion .== agent.opinion
+	nmatches = count(matches)
+	# Adopt a different opinion w/ calculated probability
```
Libbum, line 40: Needs ##

Datseris:
@kavir1698 I think it is because you have commentsi nside code snippets that start with `#` instead of `##`

kavir1698:
Yeah, right. Thanks for your input, guys!

examples/opinion_spread.jl
kavir1698: > You can give a function of position to fill_space!.

Good point. But then the loop is simpler, I guess.

src/models/opinion.jl
```diff
@@ -0,0 +1,55 @@
...
+Same as in [Opinion spread model](@ref).
+"""
+function opinion(;dims=(10, 10), nopinions=3, levels_per_opinion=4)
+	space = GridSpace(dims, periodic=true, moore=true)
+	properties = Dict(:nopinions=>nopinions) 
+	model = AgentBasedModel(Citizen, space, scheduler=random_activation, properties=properties)
+	for cell in 1:nv(model)
+		add_agent!(cell, model, false, rand(1:levels_per_opinion, nopinions), rand(1:levels_per_opinion, nopinions))
+	end
+	return model
```
Libbum, line 23: The `opinion_agent_step!` and a `dummy_model` need to be returned here too, yeah?

Libbum:
@kavir1698, you say that you got this off hash.ai essentially. Is there any reference to the algorithm used here? I'm wondering if it's just a toy model or comes with any valid reflections on behavioural studies.

kavir1698:
> @kavir1698, you say that you got this off hash.ai essentially. Is there any reference to the algorithm used here? I'm wondering if it's just a toy model or comes with any valid reflections from behavioural studies.

No, there is no references. Most of their models have no references.

src/models/opinion.jl
kavir1698: `dummy_model` is exported by the package. Do we need to return it here?

Libbum: See [here](https://github.com/JuliaDynamics/Agents.jl/blob/b5405d11da8931bf9f6063e4a91ff81219fcea35/src/models/forestfire.jl#L32). We need to return the model and both steps to keep the return parameters of the [model ecosystem](https://juliadynamics.github.io/Agents.jl/stable/models/) the same for all models within it.

Libbum: (but yes, `dummystep`, not `dummy_model`, sorry)

Libbum (approved):

Datseris:
ha this model is really cool!

PR #284, opened 2020-09-04
Cover run!(model, agent_step!, n) correctly.

Libbum:
We have `run!(model, agent_step! [, model_step!], n; kwargs...)` in the docs, but #283 requires
`run!(model, agent_step!, dummystep, n; kwargs...)` when `n::Function`. This rectifies that omission.

### DIFF
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -68,10 +68,13 @@ By default both keywords are `nothing`, i.e. nothing is collected/aggregated.
 """
 function run! end
 
-run!(model::ABM, agent_step!, n::Int = 1; kwargs...) =
+run!(model::ABM, agent_step!::Function, n::Int = 1; kwargs...) =
 run!(model::ABM, agent_step!, dummystep, n; kwargs...)
 
-function run!(model::ABM, agent_step!, model_step!, n;
+run!(model::ABM, agent_step!::Function, n::Function; kwargs...) =
+run!(model::ABM, agent_step!, dummystep, n; kwargs...)
+
+function run!(model::ABM, agent_step!::Function, model_step!::Function, n;
   replicates::Int=0, parallel::Bool=false, kwargs...)
 
   r = replicates
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -68,13 +68,13 @@ By default both keywords are `nothing`, i.e. nothing is collected/aggregated.
 """
 function run! end
 
-run!(model::ABM, agent_step!::Function, n::Int = 1; kwargs...) =
+run!(model::ABM, agent_step!, n::Int = 1; kwargs...) =
 run!(model::ABM, agent_step!, dummystep, n; kwargs...)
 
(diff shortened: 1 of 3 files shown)


src/simulations/collect.jl
```diff
@@ -68,10 +68,13 @@ By default both keywords are `nothing`, i.e. nothing is collected/aggregated.
 """
 function run! end
 
-run!(model::ABM, agent_step!, n::Int = 1; kwargs...) =
+run!(model::ABM, agent_step!::Function, n::Int = 1; kwargs...) =
 run!(model::ABM, agent_step!, dummystep, n; kwargs...)
 
-function run!(model::ABM, agent_step!, model_step!, n;
+run!(model::ABM, agent_step!::Function, n::Function; kwargs...) =
+run!(model::ABM, agent_step!, dummystep, n; kwargs...)
```
Datseris, line 75: This dissallows `agent_step!` to be a function like object, and is a big limitation. https://docs.julialang.org/en/v1.5/manual/methods/#Function-like-objects

Libbum: Ah, we can probably drop that function guard.

Libbum:
@Datseris: is there a quick example you could give as to when & how you would use `agent_step!` as a function-like-object? I can see how they're useful now in the Scheduler, but it's difficult to conceptualise a scenario here.

Datseris:
We should just change the docstring of `run!` to reflect https://juliadynamics.github.io/Agents.jl/stable/tutorial/#Agents.step!  . What it says now is incorrect, I forgot about it, sorry...

> is there a quick example you could give 

I don't have one now, but I also don't see why we should limit this. Explicitly requiring `model_step!` in the case of `n` is a function is a reasonable demand.

Libbum:
> but I also don't see why we should limit this

sure - makes sense

> We should just change the docstring of `run!`

Also fine, and I know we had this discussion before, but I can't see why its needed.

```julia
step!(model, agent_step!, n::Integer = 1)
step!(model, agent_step!, model_step!, n::Integer = 1)
step!(model, agent_step!, model_step!, n::Function)
```
means we can have
```julia
step!(::ABM, ::Function, ::Int)
step!(::ABM, ::Function, ::Function, ::Int)
step!(::ABM, ::Function, ::Function, ::Function)
```
There's no overlap if we also allow
```julia
step!(model, agent_step!, n::Function)
# i.e.
step!(::ABM, ::Function, ::Function)
```

Datseris:
That is not how default positional arguments work. 
```
step!(model, agent_step!, model_step!, n::Integer = 1)
```
defines **two** methods. 
```
step!(::ABM, ::Function, ::Function, ::Int)
step!(::ABM, ::Function, ::Function)
```
do you see how the later overwrites the method where `n` is a function *and* `modelstep` is not given?

Libbum:
julia> step!(model, agent_step!, model_step!, n::Integer = 1) = 5
step! (generic function with **2** methods)

Why have I never noticed this before! Thanks for pointing that out.

OK, I'm going to scrap this PR and just update the run! docs in master.

PR #285, opened 2020-09-04
Decide the order of agent and model activation

kavir1698:


src/simulations/collect.jl
```diff
@@ -66,6 +66,7 @@ By default both keywords are `nothing`, i.e. nothing is collected/aggregated.
   Both of these options have performance penalties.
 * `replicates=0` : Run `replicates` replicates of the simulation.
 * `parallel=false` : Only when `replicates>0`. Run replicate simulations in parallel.
+* `agents_first=true` : Wether to update agents first and then the model, or wise versa.
```
Libbum, line 69: Whether to update agents first and then the model, or vise versa

Libbum:
I've had the need for this before, so it would indeed be good to add. I'll need a bit to think about whether or not this is the best way of going about it though. Wonder if it could be part of the scheduler instead?

src/simulations/step.jl
```diff
@@ -32,22 +32,31 @@ dummystep(agent, model) = nothing
...
+        agent_step!(model.agents[index], model)
+      end
+      model_step!(model)
+    else
+      model_step!(model)
+      activation_order = model.scheduler(model)
+      for index in activation_order
+        haskey(model.agents, index) || continue
+        agent_step!(model.agents[index], model)
+      end
```
Datseris, line 53: This code snippet shouldn't be so complicated. You can do:
```julia
for ...
end
agents_first && model_step!(model)
```

Datseris:
HAHA this is sooooo timely man!!! I was just making this cloudworld and I have a model stepping function that actually makes clouds. And I always thought "daaamn, I would really like this function to preceede agent activation". Well, here it is! 

This needs tests, and the same option needs to be added to `run!, paramscan` and all other derivatives that do stepping.

kavir1698:
> HAHA this is sooooo timely man!!! I was just making this cloudworld and I have a model stepping function that actually makes clouds. And I always thought "daaamn, I would really like this function to preceede agent activation". Well, here it is!

Yeah, I just encountered the need too.

> This needs tests, and the same option needs to be added to `run!, paramscan` and all other derivatives that do stepping.

I have added the option to `run!`.

src/simulations/step.jl
```diff
@@ -32,22 +32,23 @@ dummystep(agent, model) = nothing
 until(ss, n::Int, model) = ss < n
 until(ss, n, model) = !n(model, ss)
 
-step!(model::ABM, agent_step!, n::Int=1) = step!(model, agent_step!, dummystep, n)
+step!(model::ABM, agent_step!, n::Int=1, agents_first::Bool=true) = step!(model, agent_step!, dummystep, n)
```
Libbum, line 35: I think this one, you need to pass `agents_first` through to the second function call yes?

src/simulations/step.jl
```diff
@@ -32,22 +32,23 @@ dummystep(agent, model) = nothing
...
     s += 1
   end
 end
 
-function step!(model, agent_step!, model_step!, n; kwargs...)
+function step!(model, agent_step!, model_step!, n, agents_first; kwargs...)
   @warn "`step!` with keyword arguments is deprecated. Use `run!` instead."
-  run!(model, agent_step!, model_step!, n; kwargs...)
+  run!(model, agent_step!, model_step!, n, agents_first; kwargs...)
 end
```
Datseris, line 54: I think we can remove this function now. we have deprecated it for long enough.

src/simulations/step.jl
kavir1698: Yes, I missed that.

Libbum:
Whilst we're all here. Two housekeeping things.
1. Can we perhaps turn off CI on travis-ci.org? It's running on travis-ci.com as well, seems overkill to be running 4 tests each time.
2. We may want to sort out a project board an timeline if we plan on doing a bunch of small but useful things like this. Otherwise we're going to have multiple releases a day...

kavir1698:
All tests pass locally. There is a single test failing online. I don't see why. It was failing even before these changes, when I added the opinion spread example.

Libbum:
My tests are failing locally. Ill dig deeper.

Datseris:
> Can we perhaps turn off CI on travis-ci.org? It's running on travis-ci.com as well, seems overkill to be running 4 tests each time.

We have to turn off `travis-ci.com` then, because that is the enterprise version.

> Otherwise we're going to have multiple releases a day...

That is totally fine for me. The registration system is automated.

Libbum:
> because that is the enterprise version
I'm not totally sure of that. I think they're depreciating .org.

kavir1698:
> My tests are failing locally. Ill dig deeper.

That single test fails locally for me too. I had forgotten to activate the project folder.

Libbum:
It was for the depreciation anyway, so since we dropped that function, we can drop the test.

[Details on the travis org/com split](https://devops.stackexchange.com/questions/1201/whats-the-difference-between-travis-ci-org-and-travis-ci-com)

Datseris:
I've just realized that we have turned off tests for windows and mac... @kavir1698 can you please re-enable them in the `.travis` file?

src/simulations/paramscan.jl
```diff
@@ -19,7 +19,8 @@ can take both model and agent constructor properties.
 All the following keywords are propagated into [`run!`](@ref).
 Defaults are also listed for convenience:
 `agent_step! = dummystep, n = 1, when = 1:n, model_step! = dummystep`,
-`step0::Bool = true`, `parallel::Bool = false`, `replicates::Int = 0`.
+`step0::Bool = true`, `parallel::Bool = false`, `replicates::Int = 0`,
+agents_first::Bool=true.
```
Datseris, line 23: this is not code highlighted

src/simulations/paramscan.jl
```diff
@@ -33,7 +34,7 @@ included in the output `DataFrame`.
 function paramscan(parameters::Dict{Symbol,}, initialize;
   n = 1, agent_step! = dummystep,  model_step! = dummystep,
   progress::Bool = true, include_constants::Bool = false,
-  kwargs...)
+  agents_first::Bool=true, kwargs...)
```
Datseris, line 37: You've added this here but is it used inside the source of `paramscan`?

kavir1698: Its a kwarg. It will be passed to `run!`. Here I just set a default value.

kavir1698: Or am I wrong?

kavir1698: I think it doesn't need to be added to `paramscan` at all.

Libbum: paramscan will pass all kwargs including this one to run, run will be the one that sets the default value. So yeah - not needed here.

Libbum (approved):

Libbum:
Huh. What is that windows error?
```julia
Mixed Scheduler: Test Failed at C:\Users\travis\build\JuliaDynamics\Agents.jl\test\scheduler_tests.jl:104
  Expression: unique([typeof(model[id]) for id = s1]) != unique([typeof(model[id]) for id = s2])
  Evaluated: DataType[Agent0, Agent1, Agent3, Agent2] != DataType[Agent0, Agent1, Agent3, Agent2]
```

kavir1698:
I say we drop windows tests. Windows builds are still in beta version.

Libbum:
That may be the case, but also this may be a windows specific error. Either of you have a windows machine you can run tests on (I don't)?

kavir1698:
I can check it on Windows later tonight

Datseris:
> I say we drop windows tests. Windows builds are still in beta version.

I test all other packages of JuliaDynamics on Windows on travis and I don't have problems there. I think this "beta" is kind of missleading. I would strongly recommend that we do the same here.

Libbum:
This is not a windows error, it's a product of our `Random.seed` removal.

```julia
model = init_mixed_model(scheduler = by_type(true, false))    
s1 = model.scheduler(model)    
s2 = model.scheduler(model)    
@test unique([typeof(model[id]) for id in s1]) != unique([typeof(model[id]) for id in s2])  
```

There will be cases when this random ordering of the types will be the same. Hasn't seem to fail on us for a while, but that's just luck. 

There are other tests in this section that will also fail at some stage if we don't seed it.
```julia
@test model[s1[1]].id ≠ model[s2[1]] || model[s1[2]].id ≠ model[s2[2]]
```

I'm quite confused by all this though&mdash;I thought all seeds were removed, but there's now at least some of them in the sections above in this file?

Since we dropped v1.0 support, one solution is to add back a few seeds in this section. Otherwise a lot of these checks may need to be reconsidered.

Datseris:
> Since we dropped v1.0 support, one solution is to add back a few seeds in this section. Otherwise a lot of these checks may need to be reconsidered.

Let's add the necessary seeds back.

Libbum:
OK, looks good now.

Datseris:
But there are still no tests of the new functionality. Even if it is a trivial 2-line addition, we should still add a test about it for future robustness.

kavir1698:
@Datseris would be enough to just run a sim with `agents_first=false`?

Datseris:
yes!

src/simulations/step.jl
```diff
@@ -32,22 +32,19 @@ dummystep(agent, model) = nothing
 until(ss, n::Int, model) = ss < n
 until(ss, n, model) = !n(model, ss)
 
-step!(model::ABM, agent_step!, n::Int=1) = step!(model, agent_step!, dummystep, n)
+step!(model::ABM, agent_step!, n::Int=1, agents_first::Bool=true) = step!(model, agent_step!, dummystep, n, agents_first)
 
-function step!(model::ABM, agent_step!, model_step!, n = 1)
+function step!(model::ABM, agent_step!, model_step!, n = 1, agents_first=true)
```
Datseris, line 37: The new argument `agents_first` should be reflected in the docstring of `step!` as well. At the moment it is only in `run!`.

src/simulations/step.jl
```diff
@@ -2,7 +2,7 @@ export step!, dummystep
 
 """
     step!(model, agent_step!, n::Integer = 1)
-    step!(model, agent_step!, model_step!, n::Integer = 1)
+    step!(model, agent_step!, model_step!, n::Integer = 1, , agents_first::Bool=true)
```
Datseris, line 5: double comma.

Also should be added to the version with `n::Function`.

test/collect_tests.jl
```diff
@@ -178,6 +178,17 @@
...
+            model,
+            agent_step!,
+            model_step!,
+            20;
+            agents_first=false,
+            when_model = [1, 20],
+            when = false,
+            mdata = [:flag, :year],
+            adata = [(:weight, mean)],
+        )
```
Libbum, line 191: These lines don't do any testing. The `@test` macro is the point at which a test is completed. A good test for this would be some model that counts agents with some flag = true in the model step, none of which are turned on until the first agent step. The test would be to run the model twice, one with agent, model stepping, then with model, agent stepping and compare the output.

PR #286, opened 2020-09-05
Sugarscape

kavir1698:
This is an important example to have.


docs/make.jl
```diff
@@ -68,6 +68,7 @@ pages = [
     "Introduction" => "index.md",
     "Tutorial" => "tutorial.md",
     "Examples" => [
+        "Sugarscape" => "examples/sugarscape.md",
```
Datseris, line 71: Because schelling is the example that we use to demonstrate our tutorial, can we keep the Schelling example as the topmost on this list?

kavir1698: Yeah, I thought since Schelling is already covered in the Tutorial, we can put a more important model on the top.

Libbum:
I'll take a better look at this tomorrow, but that last commit got the figures building at least.
This LinearAlgebra build error that we have do to some conflict in BBOptim is now spilling out onto [wolf/sheep](https://juliadynamics.github.io/Agents.jl/previews/PR286/examples/predator_prey/). I tried to suppress it, but no luck yet.

kavir1698:
Shall we merge this?

Libbum:
I haven't had a chance to look at it yet sorry. Its on the list of things to do. From skimming it I'll probably have some comments.

Datseris:
I've had a look at it here: https://juliadynamics.github.io/Agents.jl/previews/PR286/examples/sugarscape/ looks good.

Datseris:
Sorry Tim, I should have read your message in more detail. The warnings are still there fore Linear Algebra, but at least all code of the Wolf Sheep is properly running (figures and numbers are shown).

PR #289, opened 2020-09-10
Guard for extend in ContinouousSpace

### LINKED ISSUE #287: Continuous space initialization allows extend dimension mismatch
Currently there's no check to see if `length(extend) == D` for `ContinuousSpace`.

So users can do weird things like this:
```
ContinuousSpace(2, extend=(1,2,3,4))
ContinuousSpace(2, extend=())
```

You can even make extend some other types which don't make sense:
```
ContinuousSpace(2, extend="woohoo")
ContinuousSpace(1, extend=[[0,1,3]])
ContinuousSpace(1, extend=0)
```

The user should probably be alerted to any problematic argument definition during initialization.

Libbum:
Closes #287

### DIFF
--- a/src/core/continuous_space.jl
+++ b/src/core/continuous_space.jl
@@ -40,7 +40,7 @@ Notice that if you need to write your own custom `move_agent` function, call
 
 ## Keywords
 * `periodic = true` : whether continuous space is periodic or not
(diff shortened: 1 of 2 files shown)


Datseris:
cool; bump patch version and merge away!

Libbum:
Will bump with #290.

PR #295, opened 2020-09-15
Add `get_node_agents` method for list of nodes/agents

### LINKED ISSUE #294: `get_node_contents` method for list of nodes
In a `GridSpace` model I would like to write

``` julia
node_neighbors(agent.pos, model) |> 
neighbors -> get_node_agents(neighbors, model)
```

but there is no method for `get_node_agents` (and respectively for `get_node_contents`) to get all the agents on a list of nodes.

Is there a way to collect agents from a list? If yes, what is it called? If not, would it make sense to implement such a method?

Here's the error message for your further information:

``` julia
ERROR: LoadError: MethodError: no method matching get_node_contents(::Array{Tuple{Int64,Int64},1}, ::AgentBasedModel{MyAgent,GridSpace{SimpleGraph{Int64},2,Int64},typeof(random_activation),Dict{Symbol,Any}})
Closest candidates are:
  get_node_contents(::Integer, ::AgentBasedModel{A,var"#s57",F,P} where P where F where var"#s57"<:Agents.DiscreteSpace) where A at C:\Users\Frederik Banning\.julia\packages\Agents\kEF2y\src\core\discrete_space.jl:342
  get_node_contents(::AbstractAgent, ::AgentBasedModel{A,var"#s57",F,P} where P where F where var"#s57"<:Agents.DiscreteSpace) where A at C:\Users\Frederik Banning\.julia\packages\Agents\kEF2y\src\core\discrete_space.jl:349
  get_node_contents(::Tuple, ::AgentBasedModel{A,var"#s57",F,P} where P where F where var"#s57"<:Agents.DiscreteSpace) where A at C:\Users\Frederik Banning\.julia\packages\Agents\kEF2y\src\core\discrete_space.jl:351
```

fbanning:
Closes #294

### DIFF
--- a/src/core/discrete_space.jl
+++ b/src/core/discrete_space.jl
@@ -355,17 +355,19 @@ end
 
 """
     get_node_agents(x, model) → agents
+
 Same as `get_node_contents(x, model)` but directly returns the list of agents
 instead of just the list of IDs.
+
+    get_node_agents(x::Array{Any, 1}, model) → agents
+
+Call `get_node_contents(x, model)` by broadcasting a list of nodes or agents.
+Useful to obtain a list of agents with their relative direction from starting 
+point `x`.
 """
 get_node_agents(x, model::ABM{A,<:DiscreteSpace}) where {A} = [model[id] for id in get_node_contents(x, model)]
 
-"""
-    get_node_agents(x::Array{NTuple{Int}, 1}, model) → agents
-Same as `get_node_agents(x, model)` but iterates over a list of nodes instead
-of just a single node.
-"""
-function get_node_agents(x::Array{NTuple{Int}, 1}, model::ABM{A,<:DiscreteSpace}) where {A}
+function get_node_agents(x::Array{Any, 1}, model::ABM{A,<:DiscreteSpace}) where {A}
   [model[id] for id in get_node_contents.(x, Ref(model))]
 end
 
--- a/src/core/discrete_space.jl
+++ b/src/core/discrete_space.jl
@@ -362,31 +362,15 @@ instead of just the list of IDs.
 get_node_agents(x, model::ABM{A,<:DiscreteSpace}) where {A} = [model[id] for id in get_node_contents(x, model)]
 
 """
-    get_node_agents(x::Vector{Tuple{Int, Int}}, model) → agents
+    get_node_agents(x::Vector{NTuple{N, Int}}, model) → agents
 
-Call `get_node_agents(x, model)` by broadcasting a list of nodes or agents on a 
-2D grid. Useful to obtain a list of agents with their relative direction from 
-starting point `x`.
-By default, nodes outside the grid will be omitted and a list of agents in the 
-following directions and order is returned: [N, W, E, S] 
-If `moore == true`, the order of the list will be [NW, N, NE, W, E, SW, S, SE].
-If `periodic == true` the list will include neighboring agents on the other end 
-of the grid (e.g. if there's no neighbor to the west of the observed node `x`, the node 
-farthest to the east in the same row on the grid is returned).
+Obtain a list of agents with their relative direction from starting point `x`.
+`x` may either be a list of nodes or a list of agents.
+On a 2D grid for example, agent lists will be returned in a [N, W, E, S] pattern,
+and [NW, N, NE, W, E, SW, S, SE] if `moore = true`, as long as each direction does
+not exceed the limits of the grid.
 """
-get_node_agents(x::Vector{Tuple{Int, Int}}, model::ABM{A,<:DiscreteSpace}) where {A} = get_node_agents.(x, Ref(model))
-
-"""
-    get_node_agents(x::Vector{Tuple{Int, Int, Int}}, model) → agents
-
-Call `get_node_agents(x, model)` by broadcasting a list of nodes or agents on a 
-3D grid. Useful to obtain a list of agents with their relative direction from 
-starting point `x`.
-By default, nodes outside the grid will be omitted and a list of agents in the 
-following directions and order is returned: `PLACEHOLDER`
-"""
-get_node_agents(x::Vector{Tuple{Int, Int, Int}}, model::ABM{A,<:DiscreteSpace}) where {A} = get_node_agents.(x, Ref(model))
-
+get_node_agents(x::Vector{NTuple{N, Int}}, model::ABM{A,<:DiscreteSpace}) where {A} = get_node_agents.(x, Ref(model))
 
 @deprecate id2agent(id::Integer, model::ABM) model[id]
 
--- a/src/core/discrete_space.jl
+++ b/src/core/discrete_space.jl
@@ -367,9 +367,7 @@ point `x`.
 """
 get_node_agents(x, model::ABM{A,<:DiscreteSpace}) where {A} = [model[id] for id in get_node_contents(x, model)]
 
-function get_node_agents(x::Array{Any, 1}, model::ABM{A,<:DiscreteSpace}) where {A}
-  [model[id] for id in get_node_contents.(x, Ref(model))]
-end
+get_node_agents(x::Array{Tuple{Int, Int}, 1}, model::ABM{A,<:DiscreteSpace}) where {A} = get_node_agents.(x, Ref(model))
 
 @deprecate id2agent(id::Integer, model::ABM) model[id]
 
--- a/src/core/discrete_space.jl
+++ b/src/core/discrete_space.jl
@@ -370,7 +370,7 @@ On a 2D grid for example, agent lists will be returned in a [N, W, E, S] pattern
 and [NW, N, NE, W, E, SW, S, SE] if `moore = true`, as long as each direction does
 not exceed the limits of the grid.
 """
-get_node_agents(x::Vector{NTuple{N, Int}}, model::ABM{A,<:DiscreteSpace}) where {A} = get_node_agents.(x, Ref(model))
+get_node_agents(x::Vector{NTuple{N, Int}}, model::ABM{A,<:DiscreteSpace}) where {A,N} = get_node_agents.(x, Ref(model))
 
 @deprecate id2agent(id::Integer, model::ABM) model[id]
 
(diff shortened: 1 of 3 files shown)


Libbum:
Will be good to add some tests for these new methods [here](https://github.com/JuliaDynamics/Agents.jl/blob/5d87e26022fe9920f78cdfc27dd905017940a47b/test/space_test.jl#L166).

fbanning:
I'm not sure about the order of the arrays in the test. Grid is 3x3 and agents are on (2,2) and (3,2). Is this correct?

Datseris:
I'll give a review once we have passing tests :)

fbanning:
Yes, I'll fix this locally and push afterwards. :D

Libbum:
I'm not sure why that test shows up as still being queued. I've restarted it twice & it passes on Travis. So we can ignore that yet-to-build notification

src/core/discrete_space.jl
```diff
@@ -354,12 +354,21 @@ function get_node_contents(coords::Tuple, model::ABM{A,<:DiscreteSpace}) where {
 end
 
 """
-    get_node_agents(x, model)
+    get_node_agents(x, model) → agents
+
 Same as `get_node_contents(x, model)` but directly returns the list of agents
 instead of just the list of IDs.
+
+    get_node_agents(x::Array{Any, 1}, model) → agents
```
Libbum, line 362: You can use `Vector{Any}` as a cleaner shorthand for `Array{Any, 1}`.

src/core/discrete_space.jl
```diff
@@ -354,12 +354,21 @@ function get_node_contents(coords::Tuple, model::ABM{A,<:DiscreteSpace}) where {
...
+
+Call `get_node_contents(x, model)` by broadcasting a list of nodes or agents.
+Useful to obtain a list of agents with their relative direction from starting 
+point `x`.
 """
 get_node_agents(x, model::ABM{A,<:DiscreteSpace}) where {A} = [model[id] for id in get_node_contents(x, model)]
 
+get_node_agents(x::Array{Tuple{Int, Int}, 1}, model::ABM{A,<:DiscreteSpace}) where {A} = get_node_agents.(x, Ref(model))
```
Libbum, line 370: `Array{Tuple{Int, Int}, 1}` will only allow 2D grids here right?

src/core/discrete_space.jl
```diff
@@ -354,12 +354,21 @@ function get_node_contents(coords::Tuple, model::ABM{A,<:DiscreteSpace}) where {
...
 """
-    get_node_agents(x, model)
+    get_node_agents(x, model) → agents
+
 Same as `get_node_contents(x, model)` but directly returns the list of agents
 instead of just the list of IDs.
+
+    get_node_agents(x::Array{Any, 1}, model) → agents
+
+Call `get_node_contents(x, model)` by broadcasting a list of nodes or agents.
```
Libbum, line 364: Should this be "Call `get_node_contents(x, model)`"?

src/core/discrete_space.jl
```diff
@@ -354,12 +354,21 @@ function get_node_contents(coords::Tuple, model::ABM{A,<:DiscreteSpace}) where {
...
+Useful to obtain a list of agents with their relative direction from starting 
+point `x`.
```
Libbum, line 366: Relative direction starting from point x is perhaps not clear enough. Are we going N, E, S, W for example?

test/space_test.jl
```diff
@@ -203,12 +203,14 @@ end
...
     a = add_agent!((2, 2), gridspace, rand())
-    add_agent!((3, 2), gridspace, rand())
+    b = add_agent!((3, 2), gridspace, rand())
     @test space_neighbors((1, 2), gridspace) == [1]
     @test sort!(space_neighbors((1, 2), gridspace, 2)) == [1, 2]
     @test_throws MethodError space_neighbors((1, 2), gridspace, 1.5)
     @test sort!(space_neighbors((2, 2), gridspace)) == [1, 2]
     @test space_neighbors(a, gridspace) == [2]
+    @test get_node_agents(node_neighbors(a, gridspace), gridspace) == [[], [], [b], []]
+    @test get_node_agents(node_neighbors(b, gridspace), gridspace) == [[], [a], []]
```
Libbum, line 213: This is confusing. How do I receive 4 directions for `a`, but only 3 directions for `b`?

Also, since this is explicitly returning directional data, there should be some tests when the space has `moore = true`, as well as a verification that `periodic = true` finds neighbors correctly.

src/core/discrete_space.jl
fbanning: Hm. Technically it calls `get_node_agents` which in turn calls `get_node_contents`. I guess the broadcasting happens with `get_node_agents` so it would make more sense to have that there.

test/space_test.jl
fbanning: > How do I receive 4 directions for `a`, but only 3 directions for `b`?

Because the grid is 3x3 and agent b is at the edge of the grid which returns 3 neighboring nodes.

Sure, we can also test for moore neighorhood.

src/core/discrete_space.jl
fbanning: I think it's in line with the normal [von Neumann neighborhood](https://upload.wikimedia.org/wikipedia/commons/thumb/e/e8/Von_Neumann_neighborhood.svg/1024px-Von_Neumann_neighborhood.svg.png) resulting in a `(N, W, E, S)` Tuple. I can add a quick note on that to the docstring.

src/core/discrete_space.jl
fbanning: Yes. Do you think that's too restrictive? My reasoning was that allowing 3D grid positions would result in a more complicated docstring. Maybe we can introduce a second method for `Array{Tuple{Int, Int, Int}, 1}` with its own docstring section so that the specifics of 3D grids can be discussed there?

src/core/discrete_space.jl
fbanning: Sure

src/core/discrete_space.jl
fbanning: I've never worked with 3D grids on Agents.jl. Can anyone of you fill out the `PLACEHOLDER` that I've inserted in the docstring of the 3D method, please? @Libbum or @Datseris maybe?

kavir1698: Why not keeping it general `Array{NTuple{N, Int}, 1} where N`

fbanning: For the method itself it might make sense to keep it as general as possible. For documentation purposes I think it's not so helpful to conflate the methods in one big docstring block. Maybe that's how it is regularly done and if so, I'm of course OK with that as well. However, being more explicit is better here, I think.

fbanning:
- [ ] Add tests for periodic grid
- [ ] Replace `PLACEHOLDER` in 3D method docstring

src/core/discrete_space.jl
Libbum: The constructor that @kavir1698 suggested is the way this should go.

As for doc strings: we want to make them as short and direct as possible. Since we already speak about agent positions a lot in the greater documentation, it's not as important to list the explicit possibilities. As an aside: we can also have 0D and 1D grids, so this further exacerbates the issue.

I would suggest 
```
    get_node_agents(x::Vector{Any}, model) → agents

Obtain a list of agents with their relative direction from starting point `x`.
`x` may be either a list of nodes, or a list of agents.
On a 2D grid for example, agent lists will be returned in a [N, W, E, S] pattern,
and [NW, N, NE, W, E, SW, S, SE] if `moore = true`.
```

test/space_test.jl
Libbum: Ah. I see. I will amend my comment above to reflect this.

src/core/discrete_space.jl
fbanning: Maybe this is too offtopic but I wonder how do neighbors look in a 0D or 1D grid?

Libbum: 1D at least you would have left and right. For 0D, I can't think of a case. On the other hand, just because we can't does not mean it doesn't exist. We had a discussion about this in #284 concerning a different issue. We want the framework to not get in the way if a user identifies such a need.

src/core/discrete_space.jl
```diff
@@ -362,31 +362,15 @@ instead of just the list of IDs.
...
-
-Call `get_node_agents(x, model)` by broadcasting a list of nodes or agents on a 
-3D grid. Useful to obtain a list of agents with their relative direction from 
-starting point `x`.
-By default, nodes outside the grid will be omitted and a list of agents in the 
-following directions and order is returned: `PLACEHOLDER`
-"""
-get_node_agents(x::Vector{Tuple{Int, Int, Int}}, model::ABM{A,<:DiscreteSpace}) where {A} = get_node_agents.(x, Ref(model))
-
+get_node_agents(x::Vector{NTuple{N, Int}}, model::ABM{A,<:DiscreteSpace}) where {A} = get_node_agents.(x, Ref(model))
```
Libbum, line 373: Just need to change to `where {A,N}` here.

fbanning: Silly me.

Libbum:
I think the implementation & docstrings are good now. Some quick test coverage for the other sections as you identified and this will be good to go!

test/space_test.jl
```diff
@@ -234,6 +234,41 @@ end
         [[], [], [a], [], []]
     end
 
+    gridspace_periodic = ABM(Agent3, GridSpace((3, 3); periodic = true))
+    @test node_neighbors((1, 1), gridspace_periodic) == [(1, 3), (3, 1), (2, 1), (1, 2)]
```
fbanning, line 238: The tests fail whenever the periodic grid comes into play. This is because the order changes from [N, W, E, S] to [E, W, N, S]. Anyone got an idea why that happens?

This applies to tests at LOC 238, 241, 248 as well as 254, 260, 267.

Libbum: I'd have to take a deeper look, but my guess would be that we don't respect this order when grabbing the wrap-around cells. That's something we'll need to take care of if so. Should have some time for this tomorrow.

Datseris:
@fbanning `GridSpace` has been re-writen from scratch, see https://github.com/JuliaDynamics/Agents.jl/pull/306 . 

Many inconsistencies will soon be eliminated in a new PR. I am closing this in favor of that one, which will also have the `get_node_agents` implementation you need.

PR #298, opened 2020-09-15
Various software Benchmarking comparison

Libbum:
These are the runs that need to be compared across our four frameworks. They need to be further validated.

Pretty sure we could get these running on CI if we were interested in doing so.

An accompanying documentation page should also be built. It could probably be self generating if we were to sort out the CI integration. Otherwise a static page would be fine I guess.

# Current Status (fac4388)
|| Agents | Mesa | Netlogo | MASON |
|---|---|---|---|---|
|WolfSheepGrass|159.115 ms|1129.455 ms|332.028 ms|NA|
||1|7.1x|2.1x|NA|
|(LOC)|139|273|137 (871)||
|Flocking|98.614 ms|2932.621 ms|1014.527 ms**|206.914 ms|
||1|29.7x|10.3x|2.1x|
|(LOC)|66|120|82 (689)|369|
|ForestFire|24.638 ms|718.260 ms|101.597 ms|NA|
||1|29.1x|4.1x|NA|
|(LOC)|27|61|43 (545)||
|Schelling|13.783 ms|434.652 ms|110.210 ms|197.103 ms|
||1|31.5x|8.0x|14.3x|
|(LOC)|34|63|68 (732)|248|

Note: For LOC, we use the following convention: Code is formatted using standard practices & linting for the associated language. Documentation strings and in-line comments (residing on lines of their own) are discarded, as well as any benchmark infrastructure. NetLogo is assigned two values since its files have a code base section and an encoding of the GUI. Since many parameters live in the GUI, we must take this into account. Thus `375 (785)` in a NetLogo count means 375 lines in the code section, 785 lines total in the file.

** Netlogo has a different implementation to the other three frameworks here. It cheats a little so a true comparison would ultimately see a slower result here.


Libbum:
@kavir1698 Can you maybe push those changes to Wolf sheep to the `src/models` directory in this branch?

kavir1698:
Here are relative benchmark results for wolf sheep:
Agents: 1
Mesa: 5.7

test/benchmark/Mesa/WolfSheep/benchmark.py
```diff
@@ -17,7 +17,6 @@
 wolfsheepmodel = WolfSheep()
 """
 
-for a in range(0, 10):
-    tt = timeit.timeit('wolfsheepmodel.run_model()', setup=setup, number=1)
-    print(tt)
-
+tt = timeit.Timer('wolfsheepmodel.run_model()', setup=setup)
+a = min(tt.repeat(100, 1))
```
Libbum, line 21: I was having some real issues with the documentation here. 
- What is the default unit of `timeit`?
- `timeit.timeit` does repeats internally, defaults to 100 runs. Does `timeit.Timer` not do that? Then you just control it by the `tt.repeat` clause?

kavir1698: Timer is just an object but doesn't run. We can explicitly control the number of repeats.

Libbum:
I'm only seeing Mesa: 1.6, which is no change from before these refactors.

kavir1698:
> 
> 
> I'm only seeing Mesa: 1.6, which is no change from before these refactors.

That's weird. There should be something that we are doing differently. Are you taking the minimum time?

Libbum:
Yep. I'm going to write a wrapper script so we can both be on the same page.

kavir1698:
Here is what I run in Julia
```julia
using BenchmarkTools
using Agents
mutable struct Sheep <: AbstractAgent
    id::Int
    pos::Tuple{Int,Int}
    energy::Float64
end

mutable struct Wolf <: AbstractAgent
    id::Int
    pos::Tuple{Int,Int}
    energy::Float64
end

mutable struct Grass <: AbstractAgent
    id::Int
    pos::Tuple{Int,Int}
    fully_grown::Bool
    countdown::Int
end

sheep(a) = typeof(a) == Sheep
wolves(a) = typeof(a) == Wolf
grass(a) = typeof(a) == Grass && a.fully_grown
a = @benchmark data = run!(model, Models.predator_prey_agent_step!, 100, adata=adata, when=1:100) setup=(model= Models.predator_prey()[1]) samples=100
```

For Mesa, I just run `python3 benchmark.py`

Libbum:
Oh. Does the Mesa run save information somewhere? And that's only 100 steps?


Results from `./runall.sh`
> Agents.jl WolfSheep (ms): 49.608971999999994
> Agents.jl Flocking (ms): 1470.378458
> Mesa WolfSheep (ms): 81.00651299992023
> Mesa Flocking (ms): 8052.814532999946

kavir1698:
No, it's actually 200 steps and keeps track of the number of each agent type. The ratio is now 3.1 for me.

Libbum:
Oh, I see. This is making more sense now.
- [ ] We need to increase the Mesa step count to 500 (I missed that).
- [ ] Mesa is only printing counts, not collecting them. I think we're better off using step! for these benchmarks. 
- [ ] Need to confirm that NetLogo is not data collecting (I don't think it is)

Libbum:
I confirmed that mesa is not collecting data by removing all `datacollection` instances. Timing is the same. I'm now getting Mesa: 2.75 from the timings in `runall.sh`

kavir1698:
> 
> 
> I confirmed that mesa is not collecting data by removing all `datacollection` instances. Timing is the same. I'm now getting Mesa: 2.75 from the timings in `runall.sh`

So we converged. I get Agents to be about 3 times faster as well. Let's keep the number of steps lower for all packages, say 200. With 500 steps, Mesa benchmarks take too long to finish for me.

Libbum:
I just re-did the gridspace runs, turns out that my profiler for NetLogo was not running what I expected it to, so that has been fixed.

Numbers are in:
| Agents | Mesa | Netlogo |
|---|---|---|
|21.848 ms|81.491 ms|264.92 ms|
|1|3.73x|12.14x|

Datseris:
Tim, I am really confused by these numbers. How can it be that the Mesa comparison is like that? Our _initial comparisons_ from version 2.0 showed we are like 11x faster: https://juliadynamics.github.io/Agents.jl/dev/mesa/

Now we have done a massive performance update and we are only 4x faster? This cannot be. @kavir1698

Datseris:
I think we should put the Forest fire example as well in the paper, there we don't deal with the type instabilities of mixed type and Julia can really shine. We already have the code for it anyway. We don't have to compare with every software, just with the Mesa for this entry.

Libbum:
I was thinking of adding that and Schelling (since the latter is in all 4 frameworks).

Can't comment on the 2.0 values. Will need to look the baselines for those first.

Datseris:
Version 2.0 was the old `GridSpace`, hadn't changed at all until now.

Libbum:
I understand, but it was flocking and forest fire. We are 6.8x on flocking currently, but that's continous space and I haven't run forest fire yet.

Datseris:
> Pretty sure we could get these running on CI if we were interested in doing so.

We could get PkgBenchmark running on CI, the original post by `@tkf` that described how to do so wasn't that complicated.

p.s.; thanks for taking the effort to make the table in the original post.

Libbum:
Yeah, I think that'll be a side project to this one though. I've had to invoke some pretty nasty stuff to get reproducible numbers out of NetLogo. Don't think it'll be helpful to put this on CI now. But I do plan to have the machinery done so that someone who wants to install everything and run the numbers again can do it from one script.

Numbers are going to be up and down a bit until I find a useful parameter range for everything. Will ping you both once things settle down. 

(We're waiting on #313 for this anyhow)

Datseris:
I meant to run PkgBenchmark on CI to report only our own benchmarks, in new PRs for example. But actually it probably doesn't matter that much, we can run them manually on a performance related PR.

We don't care what NetLogo does I think i'll never even utter its name once we have numbers for the paper!

Libbum:
Guys I'm having some conceptual problems with forest_fire for the benchmarks. Is it OK if I alter how it works? We spontaneously ignite fire with some probability at random locations. Mesa and NetLogo ignite a row on the grid (`x=0`).

kavir1698:
> Guys I'm having some conceptual problems with forest_fire for the benchmarks. Is it OK if I alter how it works? We spontaneously ignite fire with some probability at random locations. Mesa and NetLogo ignite a row on the grid (`x=0`).

Sure. We would then need to update the model description too.

Datseris:
fine for me.

Libbum:
I need to do a Schelling case for MASON, but otherwise this is waiting on #313 for Flocking tests now.

kavir1698:
>I need to do a Schelling case for MASON, but otherwise this is waiting on #313 for Flocking tests now.

I'm working on it

Libbum:
travis-ci.org is becoming quite unreliable, and it's not that great that we can't manually restart those runs when they fail because it can't figure out what DNS it wants to use.

Libbum:
234ab73 adds a writeup that replaces the Mesa Comparison section. Flocking data needs to be added, but would be good to get some comments on the language.

Datseris:
I would also add that we upload the paper as pdf in the `paper` folder (and remove all old paper related files), once we submit it.

Libbum:
For sure. I left out any more explicit comparisons other than runtime and LOC since I think it's better we reference the paper for that once it's done.

Libbum:
This is now good to merge unless there are comments.

Libbum:
That test passes, it's just travis-ci.org being useless at downloading stuff again.

PR #302, opened 2020-09-16
Nearest neighbor fix1

### LINKED ISSUE #300: weirdness with nearest_neighbor and interacting_pairs
I'm trying to debug a few things here but they're interrelated in some way.

First thing, shouldn't this test pass?
```julia
@testset "nearest neighbor" begin 
  mutable struct Agent9 <: AbstractAgent
    id::Int
    pos::NTuple{2,Float64}
    vel::NTuple{2,Float64}
    f1::Union{Int,Nothing}
  end
  space = ContinuousSpace(2; periodic = true, extend = (1,1), metric = :euclidean)
  model = ABM(Agent9, space)
  pos = [(0.0, 0.0),(0.2, 0.0),(0.2, 0.2),(0.5, 0.5)]
  for i in pos
    add_agent!(i,model,(0.0,0.0),nothing)
  end

  for agent in values(model.agents)
    agent.f1 = nearest_neighbor(agent, model, sqrt(2)).id
  end

  @test model.agents[1].f1 == 2
  @test model.agents[2].f1 == 1
  @test model.agents[3].f1 == 2 ## This is evaluated as 1
  @test model.agents[4].f1 == 3
end
```

The second to last test is evaluated to 1 and fails.  So I went to investigate `nearest_neighbor` and seemed to find a weird line that I can't understand why it's present (line 293 in `continuous_space.jl`):
```julia
function nearest_neighbor(agent, model, r)
  n = space_neighbors(agent, model, r)
  length(n) == 0 && return nothing
  d, j = Inf, 0
  for i in 1:length(n)
    if model.space.metric == :euclidean
      @inbounds dnew = sqrt(sum(abs2.(agent.pos .- model[n[i]].pos)))
    elseif model.space.metric == :cityblock
      @inbounds dnew = sum(abs.(agent.pos .- model[n[i]].pos))
    end
    _, j = findmin(d) ## Why is this line here?
    if dnew < d
      d, j = dnew, i
    end
  end
  return @inbounds model[n[j]]
end
```
Won't this line always just set `j=1` since `length(dnew)` is always `1`? This is problematic if the last `dnew` in your for loop isn't `<d` because then `j` is reset to 1. 

For example, if I look at what `nearest_neighbor` is doing in my example above:
```julia
## I copy and pasted nearest_neighbor and added some @shows
agent = model.agents[3]
n = Agents.space_neighbors(agent, model, sqrt(2))
length(n) == 0 && return nothing
d, j = Inf, 0
for i in 1:length(n)
    if model.space.metric == :euclidean
        @inbounds dnew = sqrt(sum(abs2.(agent.pos .- model[n[i]].pos)))
    elseif model.space.metric == :cityblock
        @inbounds dnew = sum(abs.(agent.pos .- model[n[i]].pos))
    end
    _, j = findmin(d) ### I think this is where the trouble is--if the last dnew > d, then j will always be 1. Basically this line always resets j to 1 i think... I don't know why it's here
    @show dnew, i
    if dnew < d
        d, j = dnew, i
    end
    @show d, j
end
@show @inbounds model[n[j]]

## Returns this
(dnew, i) = (0.28284271247461906, 1)
(d, j) = (0.28284271247461906, 1)
(dnew, i) = (0.2, 2)
(d, j) = (0.2, 2)
(dnew, i) = (0.4242640687119285, 3)
(d, j) = (0.2, 1) ## j shouldn't be reset to 1 here
#= /Users/.../mytest.jl:379 =# @inbounds(model[n[j]]) = Agent9(1, (0.0, 0.0), (0.0, 0.0), nothing)
```

Modifying line 286 to `d, j = Inf, 1` and removing line 293 (`_, j = findmin(d)`) fixes my test...
```julia
(dnew, i) = (0.28284271247461906, 1)
(d, j) = (0.28284271247461906, 1)
(dnew, i) = (0.2, 2)
(d, j) = (0.2, 2)
(dnew, i) = (0.4242640687119285, 3)
(d, j) = (0.2, 2) ## All better
#= /Users/.../mytest.jl:378 =# @inbounds(model[n[j]]) = Agent9(2, (0.2, 0.0), (0.0, 0.0), nothing)
```
...but now there are two `interacting_pairs` tests that fail:
https://github.com/JuliaDynamics/Agents.jl/blob/d6dc074bf63a42213ac84adc9284567fcae73746/test/continuousSpace_tests.jl#L142
and 
https://github.com/JuliaDynamics/Agents.jl/blob/d6dc074bf63a42213ac84adc9284567fcae73746/test/continuousSpace_tests.jl#L145

But I think these tests are wrong, because shouldn't they both be returning the same value, and shouldn't that value be `2`, since that's the most pairs you can possibly have if both `:scheduler` and `:nearest` are supposed to restrict agents to be in at most 1 pair, and there are only 4 agents?

I haven't tried debugging `interacting_pairs` any further, but can anyone on here call me out if I'm going crazy, or verify what I'm saying?

hbsmith:
Fixes #300.

Note that I updated the `interacting_pairs` tests on line 142 to be `2`, which I believe is correct, but I also believe the test on line 145 should remain at `2`, and now it is failing (for I think an unrelated reason--again see #300 for details).


Datseris:
Tests fail here, and also @hbsmith can you please rebase to current master?

Libbum:
I'll rebase it and check out the failing tests. There is perhaps some nuance to this.

Libbum:
The test that @hbsmith left for us was incorrect due to the bug found in #300. There are three pairs that interact when using `:nearest`, since the distance between 1&ndash;2 and 2&ndash;3 are the same. Thus the correct result for this test is `3`, not `2.

This should all pass now. If so, it's good to merge.

Libbum (approved):

hbsmith:
@Libbum @Datseris So the implementation is by-design, even though the doc string for `interacting_pairs(:nearest)` says:
"Each agent can only belong to one pair, therefore if two agents share the same nearest neighbor only one of them (sorted by id) will be paired." ?
https://github.com/JuliaDynamics/Agents.jl/blob/46effae1898e8fda7afecf272cdc25eccbda683f/src/spaces/continuous_space.jl#L348

hbsmith:
Separately but related, I'm still not sure how these two tests 
https://github.com/JuliaDynamics/Agents.jl/blob/46effae1898e8fda7afecf272cdc25eccbda683f/test/continuousSpace_tests.jl#L141
https://github.com/JuliaDynamics/Agents.jl/blob/46effae1898e8fda7afecf272cdc25eccbda683f/test/continuousSpace_tests.jl#L144
are passing, because when I try to replicate the preceding code in the REPL, I get three pairs for the first test (`:scheduler`) and two for the second (`:nearest`)

Libbum:
We had a long discussion about this when it was implemented. Basically there is more than one way to classify a pair and in many cases the most rigorous is not what is necessary. 

So it may be that I overlooked some of the nuances there when merging this. I'll open a new discussion and we'll make sure everything makes sense.

PR #304, opened 2020-09-16
Declare simple space API and make maxid part of model

Datseris:
Making the id part of the model was a simple fix.

In the remaining I want to re-structure the source code so that it is clear exactly what is the space API and exactly which parts are necessary to be implemented for each space individually. While making the new version of `GridSpace`  I realized that _most_ things are duplicated without reason. I'll do this soon here.


Datseris:
@Libbum and @kavir1698 please prioritize a basic working integration of `PkgBenchmark`, to make such PRs easy to see performance improvements.

src/core/model.jl
```diff
@@ -37,21 +38,15 @@ space.
...
-
+abstract type AbstractSpace end
 SpaceType=Union{Nothing, AbstractSpace}
 
 struct AgentBasedModel{A<:AbstractAgent, S<:SpaceType, F, P}
     agents::Dict{Int,A}
     space::S
     scheduler::F
     properties::P
+    maxid::Vector{Int}
```
kavir1698, line 49: Why is `maxid` not an `Int`?

Datseris: Yeah this is something to consider. Because the model type is not mutable and therefore you can't change it. Once this PR is done, and once we have benchmarking capabilities, I'll test a version where the model is mutable and `maxid` is Integer.

Datseris:
This is absolutely insane. Turns out, with good design, to create a new space type that satisfies the basic API you only need four functions:
```julia
"""
    random_position(model) → pos
Return a random position in the model's space (always with appropriate Type).
"""
random_position(model) = notimplemented(model)

"""
    move_agent!(agent [, pos], model::ABM) → agent

Move agent to the given position, or to a random one if a position is not given.
`pos` must be the appropriate position type depending on the space type.
"""
move_agent!(agent::A, pos, model::ABM{A}) where {A<:AbstractAgent} = notimplemented(model)

"""
Add the agent to the underlying space structure at the agent's own position.
This function is called after the agent is already inserted into the model dictionary
and changing max id has been taken care of. This function is NOT part of the public API.
"""
add_agent_to_space!(agent, model) = notimplemented(model)

"""
    remove_agent_from_space!(agent, model)
Remove the agent to the underlying space structure.
This function is called after the agent is already removed from the model dictionary
This function is NOT part of the public API.
"""
remove_agent_from_space!(agent, model) = notimplemented(model)

```
have a look at the file `sapace_interaction_API.jl` ... Everything else can be done automatically with the above!!! This is crazy reduction in the complexity of adding a new space type.

We should add a "developer documentation" I think.

src/core/space_interaction_API.jl
```diff
@@ -0,0 +1,168 @@
...
+        k > n && kill_agent!(v, model)
+    end
+end
+
+"""
+    genocide!(model::ABM, f::Function)
+Kill all agents where the function `f(agent)` returns `true`.
+"""
+genocide!(model::ABM, f::Function)
+    for a in allagents(model)
```
Libbum, line 92: Not sure why, but I can't compile this branch as it stands. 
```julia
ERROR: LoadError: LoadError: UndefVarError: model not defined
Stacktrace:
 [1] top-level scope at /mnt/turtle/repos/Agents.jl/src/core/space_interaction_API.jl:92
 [2] include(::Function, ::Module, ::String) at ./Base.jl:380
 [3] include at ./Base.jl:368 [inlined]
 [4] include(::String) at /mnt/turtle/repos/Agents.jl/src/Agents.jl:1
 [5] top-level scope at /mnt/turtle/repos/Agents.jl/src/Agents.jl:14
 [6] include(::Function, ::Module, ::String) at ./Base.jl:380
 [7] include(::Module, ::String) at ./Base.jl:368
 [8] top-level scope at none:2
 [9] eval at ./boot.jl:331 [inlined]
 [10] eval(::Expr) at ./client.jl:467
 [11] top-level scope at ./none:3
in expression starting at /mnt/turtle/repos/Agents.jl/src/core/space_interaction_API.jl:92
in expression starting at /mnt/turtle/repos/Agents.jl/src/Agents.jl:14
```

src/core/space_interaction_API.jl
```diff
@@ -0,0 +1,168 @@
...
+    for (k, v) in model.agents
+        k > n && kill_agent!(v, model)
+    end
+end
+
+"""
+    genocide!(model::ABM, f::Function)
+Kill all agents where the function `f(agent)` returns `true`.
+"""
+genocide!(model::ABM, f::Function)
```
Libbum, line 91: function genocide!

src/core/space_interaction_API.jl
```diff
@@ -0,0 +1,168 @@
...
+end
+
+#######################################################################################
+# Adding agents (also space agnostic)
+#######################################################################################
+"""
+    add_agent_pos!(agent::AbstractAgent, model::ABM) → agent
+Add the agent to the `model` at the agent's own position.
+"""
+function add_agent_pos!(agent::A, model::ABM)
```
Libbum, line 110: where {A<:AbstractAgent}

src/Agents.jl
```diff
@@ -11,9 +11,11 @@ import Base.length
 # Core structures of Agents.jl
 include("core/model.jl")
 include("core/schedule.jl")
-include("core/discrete_space.jl")
-include("core/continuous_space.jl")
-include("core/agent_space_interaction.jl")
+include("core/space_interaction_API.jl")
+
+include("space/nothing.jl")
```
Libbum, line 16: This file hasn't been added yet

Datseris:
@Libbum don't review yet, this is Work In Progress. I'll let you know when it is ready to review (no reason to waster your time now :P )

Libbum:
Was just trying to make this my test PkgBenchmark branch :)

Datseris:
Alright @kavir1698 and @Libbum I've spent all my working day on this and the new space type :D There are still some tests failing, and if you spot a simple mistake please post it here or push directly. I now have to turn to some other projects and will return to this when I find some extra time again.

But the amount of simplification of the source is really astonishing. We used to define `add_agent!` for every space...

src/core/model.jl
Libbum: Why not `maxid::Ref{Int}`?

Then call or set with `model.maxid[]`.

src/core/model.jl
```diff
@@ -108,56 +106,57 @@ function AgentBasedModel(
...
+Return a valid `id` for creating a new agent with it.
+"""
+nextid(model::ABM) = model.maxid[1] + 1
+
 """
     getproperty(model::ABM, prop::Symbol)
+    model.prop
 
-Return a property from the current `model`. Possibilities are
+Return a property from the current `model`, assuming the model properties are i  Possibilities are
```
Libbum, line 154: Think something got cut off in this docstring addition.

Datseris: fixed!

src/core/space_interaction_API.jl
```diff
@@ -0,0 +1,186 @@
...
+Move agent to the given position, or to a random one if a position is not given.
+`pos` must be the appropriate position type depending on the space type.
+"""
+move_agent!(agent::A, pos::ValidPos, model::ABM) where {A<:AbstractAgent} = notimplemented(model)
+
+"""
+    add_agent_to_space!(agent, model)
+Add the agent to the underlying space structure at the agent's own position.
+This function is called after the agent is already inserted into the model dictionary
+and changing max id has been taken care of. This function is NOT part of the public API.
```
Libbum, line 40: Suggestion
and changing max id has been taken care of -> and the `maxid` has been updated

src/core/space_interaction_API.jl
```diff
@@ -0,0 +1,186 @@
...
+"""
+add_agent_to_space!(agent, model) = notimplemented(model)
+
+"""
+    remove_agent_from_space!(agent, model)
+Remove the agent to the underlying space structure.
```
Libbum, line 46: Remove the agent *from*

src/spaces/discrete_space.jl
```diff
@@ -300,6 +304,167 @@ end
...
+    nn = coord2vertex(agent.pos, model)
+    push!(model.space.agent_positions[nn], agent.id)
+    return agent
+end
+
+
+#######################################################################################
+# Extra space-related functions dedicated to discrete space
+#######################################################################################
+export add_agent_single!, move_agent_single!, fill_space!, move_agent_single!
```
Libbum, line 355: move_agent_single! is duplicated here

Datseris:
All tests pass. This refactoring paved the way not only for a new and more performant discrete and continuous space, but also for the open street map space, the planet space, or any other type of space. We should also state in the comparison table that we have a clear API for new space types and thus we should have an entry "extendable spaces" or "new spaces possible" or whatever, where we have "Yes".

Datseris:
Notice the following change: **I now explicitly state in the docs of `add_agent!` that `pos` MUST match the space position type**. Before hand we had a method that allowed `pos` to be `Int` for `GridSpace`, which I think was sloppy. Not that it matters much, because `GridSpace` will likely go to the trash, but good to know.

src/spaces/continuous_space.jl
```diff
@@ -1,482 +1,465 @@
...
+
+  @assert metric ∈ (:cityblock, :euclidean)
+  @assert typeof(extend) <: NTuple{D} "extend must be a tuple of length(D)"
+  @assert all(typeof(e) <: Real && e > 0 for e in extend) "extend must contain positive integers or floating point numbers"
+  db, q, q2, q3, q4, q5 = prepare_database(D)
+  ContinuousSpace(D, update_vel!, periodic, extend, metric, db, q, q2, q3, q4, q5)
+end
+
+# Deprecate Space constructor
+@deprecate Space(D::Int, update_vel!::Function) ContinuousSpace(D::Int, update_vel!::Function)
```
Libbum, line 68: Can drop this

Libbum:
This is super elegant, I love it.

The only outlier here is that of the `Vector` choice for `maxid`. I think it's best to keep the main struct immutable if possible, so I'd vote to swapping that out to `Ref`; although maybe I'm missing a performance issue with that choice.

Libbum (approved):

Datseris:
> Oh, and I guess there may be some work to separate Graph and Grid Space from a single Discrete file (if perhaps we want to add in Hex grids or what-not). But that's for another day I think.

In the PR with the "new" version of "GridSpace" that will be array based, hexagonal spaces are not possible, as Julia arrays are by construction rectangular... 

However, a new type `HexSpace` that is also based on a graph just like our current `GridSpace` becomes straightforward due to how little amount of new code is necessary!

> I think it's best to keep the main struct immutable if possible, so I'd vote to swapping that out to Ref; although maybe I'm missing a performance issue with that choice.

Sure I can swap it for `Ref`, I don't think the performance here matters much because it has been reduced to O(1) operation.

Datseris:
(I should say that I am not convinced about the "real world" usage of hexagonal spaces btw... are they _really_ useful in practice? What arguments would an actual scientist have to use a hex space instead of grid?)

Libbum:
(The hexagon is the highest order polygon which is capable of covering an area completely by itself. The nearest neighbourhood in a hexagonal grid is simpler and less ambiguous than in a rectangular grid. Therefore movement paths and connectivity are actually easier to work with / efficient.)

Datseris:
Wait, did Libbum say that you wanted to turn this PR into the benchmark PR...? Welp, sorry about that! :D Once you have the PkgBenchmarks you can apply them to the Array Space coming soon!

Libbum:
It doesn't really matter, I can target any commit hash regardless. Just thought since you fixed the maximum(keys) issue, there would be a sane case for `judge` to compare.

That being said, it's a real pain to judge anything that comes *before* the benchmark is implemented. So we'll just go with whatever we change in the future!

PR #306, opened 2020-09-17
[WIP] New grid space

Datseris:
I'm opening this PR just to show some quick benchmarks and proof of concept. Running the file `benchmarks/forest.jl` yields:
```
Times of old grid
Move agent
  62.602 ns (0 allocations: 0 bytes)
Space neighbors
  391.452 ns (5 allocations: 352 bytes)
Iterate over space neighbors
  1.090 μs (32 allocations: 928 bytes)
Iterate over position space neighbors
  975.000 ns (34 allocations: 976 bytes)
node neighbors
  166.224 ns (4 allocations: 368 bytes)
Add agent
  143.627 ns (1 allocation: 48 bytes)


Times of new grid
Move agent
  57.857 ns (0 allocations: 0 bytes)
Space neighbors
  3.199 ns (0 allocations: 0 bytes)
Iterate over space neighbors
  141.575 ns (8 allocations: 512 bytes)
Iterate over position space neighbors
  19.538 ns (0 allocations: 0 bytes)
Add agent
  139.585 ns (1 allocation: 48 bytes)
node neighbors
  2.399 ns (0 allocations: 0 bytes)
```

So, move and add remain at same speed (which was expected, identical operations happen under the hood), but finding neighbors gets 1 full order of magnitude performance for `agents` and THREE ORDERS OF MAGNITUTDE (!!!) performance for `node`.

My todo's:

- [x] Generalize to N dimensions
- [x] Allow periodic spaces
- [x] Add `space_neighbors` to API declaration of continuous space
- [ ] Rename array space to grid space, delete everything in grid space, comletely separate gridspace from `GraphSpace`.


Datseris:
@kavir1698 I will need an explicit definition from you of what is a "moore" neighborhood. What is moore neighborhood in 5-dimensional space?

I'd propose that we completely rework the way "radius" works in GridSpaces, so that we only have either Euclidean radius or Cityblock (with the euclidean radius only counting "Cells" entirely within the circle, as we do it in TimeseriesPrediction.jl).

How are neighborhoods on grids represented in other software?

Datseris:
in TImeseriesPrediction we use this function for euclidean selection of cells in D dimensions:
```julia
"""
    indices_within_sphere(r, Φ) → β
Return all cartesian indices within a hypersphere or radius `r` and dimension `Φ`.
"""
function indices_within_sphere(radius, dimension)
    #Floor radius to nearest integer. Does not lose points!
    r = floor(Int,radius)
    #Hypercube of indices
    hypercube = CartesianIndices((repeat([-r:r], dimension)...,))
    #Select subset of hc which is in Hypersphere
    βs = [ β for β ∈ hypercube if norm(β.I) <= radius ]
end
```

I propose that we should only have this method and the cityblock method (which is trivial because one selects just all cells having any index within +/- r in any dimension)

Datseris:
@fbanning perhaps you can also chime in here.

Libbum:
moore and cityblock are the same thing. A 5D Moore neighborhood would be any cell within a 5D hypersphere with radius 1

Is it worth considering [Distances.jl](https://github.com/JuliaStats/Distances.jl) again at this point in time?

Datseris:
> A 5D Moore neighborhood would be any cell within a 5D hypersphere with radius 1

Then Moore and Cityblock are different. Cityblock is any cell contained within a **hypercube** whose size is 2*1 across each dimension and whose barycenter is the origin.

Datseris:
> Is it worth considering Distances.jl again at this point in time?

I don't think so. The generality will come at the cost of learning curve: is it really the case that people would want other types of distances for GridSpace?

Libbum:
Sorry, I meant hypercube

Datseris:
The complexity of the neighborhood in the MASON manual is **completely unecessary**. I can see it has six possible options for a rectangular grid. This is just not reasonable, if you need this precision in the distance, you go for continuous space. We have continuous space and thus we shouldn't make the grid space so detailed with respect to distances...

I still vote for only two options: Euclidean and Cityblock, and we just clarify which points are considered in the Euclidean (whether a partially covered cell is included or not)

benchmarks/forest.jl
```diff
@@ -0,0 +1,130 @@
...
+function forest_model_step_array!(forest)
+    for pos in positions(forest, :random)
+        nc = get_node_contents(pos, forest)
+        ## the cell is empty, maybe a tree grows here
+        if length(nc) == 0
+            rand() ≤ forest.p && add_agent!(pos, forest, true)
+        else
+            tree = forest[nc[1]] # by definition only 1 agent per pos
+            if tree.status == false  # if it is has been burning, remove it.
+                kill_agent!(tree, forest)
```
kavir1698, line 36: Is `kill_agent!` updated to use the new `remove_agent_from_space!` function?

Datseris: Yes: https://github.com/JuliaDynamics/Agents.jl/blob/master/src/core/space_interaction_API.jl#L63-L67

`kill_agent!` is now space agnostic, same as `genocide!`.

kavir1698:
I also vote for keeping the distances simple.

benchmarks/forest.jl
kavir1698: I just found that it is.

fbanning:
Not much to add here. Let's keep it simple. Once any specific cases from users pop up, this enhancement can be opened for discussion again.

Datseris:
So, the current implementation:

- Works for any dimensionality
- Works for periodic 
- Is ABSURDLY FAST

Check this out, latest benchmarks:
```
Times of new grid
Move agent
56.880 ns (0 allocations: 0 bytes)
Space neighbors
  3.299 ns (0 allocations: 0 bytes)
Iterate over space neighbors
  27.363 ns (1 allocation: 64 bytes)
Iterate over position space neighbors
  10.309 ns (0 allocations: 0 bytes)
Add agent
  141.631 ns (1 allocation: 48 bytes)
node neighbors
  2.299 ns (0 allocations: 0 bytes)
```

I've used some crazy optimizations from TimeseriesPrediction.jl which pre-calculate everything but the last step for any neighborhood. SO now, when the function neighborhood is called, we check whether this _kind_ of neighborhood has been calculated. If not, we pre0initialize a bunch of stuff. Then, we do the calculation. Initializing once takes about 100-200ns (i.e. as much as it took before this improvement). After that, it takes about 10 ns, which means we get an 50-fold increase in finding neighbors with respect to the previous space, even though we can now support any dimensionality.

Ah, also, the source code is so much simpler than the current GridSpace, with this honestly crazy position-by-position creation of the space.

Datseris:
I won't be doing tests for this space, because it will replace the GridSpace. I don't know whether I should replace GridSpace in this PR or merge this in and replace GridSpace in a new PR. 

?

Datseris:
Actually, I won't do any GridSpace replacing until we have a PkgBenchmarking functinality merged. I'll make sure tests pass here, but leave ArraySpace as a separate entity for now.

Libbum:
:godmode: 

:hole:-y :hankey: 

That's totally insane.

I'm back from teaching, so will be adding the neighbour searches in the benchmarks soon. I'd appreciate the benchmarks to be merged first so we can track the changes.

kavir1698:
I am excited about these changes!

Datseris:
Alright, "tests pass" (this PR actually altered nothing of existing source, but only re-organized some things. ArraySpace is in a new file).

I'll merge this, and leave the replacing `GridSpace` in a new PR.

Datseris:
This is devastating. When I merged master into this branch, I can now clearly see I haven't committed the actual file `array.jl`. All this progress I had that produced these amazing benchmarks is now lost. I'll have to redo it. I cannot find the latest code anywhere I've searched every fucking commit. FUCK

Libbum:
Holy shit! So you don't think it was pushed up at all from your machine?

Libbum:
I-have an `array.jl` here though?

Datseris:
Everything is lost and this is really sad. You will notice that in this commit https://github.com/JuliaDynamics/Agents.jl/pull/306/commits/e43e0870321cdc5285413ea15d9b9cc014eb2ff2  there are no changes for the `array.jl` file. Even though clearly all this performance work went into that file :P but yet the commit has no changes. And then a remote merge happened. I just can't find it.

What is really worse is that I've tried to recreate something from memory but I get horrible timings...
```
Times of new grid
Move agent
  67.178 ns (0 allocations: 0 bytes)
Space neighbors
  285.354 ns (7 allocations: 256 bytes)
Iterate over space neighbors
  633.918 ns (20 allocations: 640 bytes)
Iterate over position space neighbors
  553.718 ns (20 allocations: 592 bytes)
Add agent
  172.555 ns (1 allocation: 48 bytes)
node neighbors
  179.690 ns (4 allocations: 144 bytes)

```
I was such a mistake to work on this last night, after I've had so many beers god fucking damn it.

Datseris:
Making some progress, 
```
Times of new grid
Move agent
  58.656 ns (0 allocations: 0 bytes)
Space neighbors
  49.140 ns (1 allocation: 224 bytes)
Iterate over space neighbors
  102.966 ns (3 allocations: 320 bytes)
Iterate over position space neighbors
  69.081 ns (1 allocation: 224 bytes)
Add agent
  172.973 ns (1 allocation: 48 bytes)
node neighbors
  55.657 ns (1 allocation: 224 bytes)

# versus (on the current machine I run this)

Times of old grid
Move agent
  63.877 ns (0 allocations: 0 bytes)
Space neighbors
  455.330 ns (5 allocations: 288 bytes)
Iterate over space neighbors
  561.828 ns (8 allocations: 352 bytes)
Iterate over position space neighbors
  484.108 ns (10 allocations: 400 bytes)
node neighbors
  177.869 ns (4 allocations: 368 bytes)
Add agent
  165.836 ns (1 allocation: 48 bytes)
```

So, even if I can't manage to reach the original performance, this is still worth it (and it is also any dimensions still)

Libbum:
Have you tried wading through `git reflog` for a dangling commit?

Datseris:
I don't know what `reflog` is :(

Libbum:
Just run `git reflog` and scroll through until you find the area of 
```
0654e5d possible to initialize array space 
e43e087 everything seems to work and benchmarks show INSANE RESULTS
```
If there's something that you DID commit but it was in some detached state, then you'll find it there. Checkout the commit hash directly and recover.

(I can tell you based on my local history, it never got pushed remotely)

Datseris:
I think I never pressed the "save" button on my editor................

Libbum:
Oooohhhh....

Datseris:
Anyways, I think I've actually reached original performance. The benchmarks I've pasted originally with the 20 nanoseconds where in a version where many agents have died and only 1 agent was neighbor of 1 agent. I'll push soon! (reallyy careful this time)

PR #309, opened 2020-09-18
Add note about model stepping

### LINKED ISSUE #308: How to get current step during model run
I'd like to use the current step during a model run (in my case to use it in a stepping function). What's the best way to do this?

Libbum:
Closes #308

### DIFF
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -65,11 +65,12 @@ dummystep
 ```
 
 !!! note
-    If you are coming from other ABM frameworks, it may be confusing that Agents.jl
-    does not store the current step somewhere. Since we give users total control of
-    their models, this is something that must be implemented yourself if needed.
+    Notice that the current step number is not explicitly given to the [`model_step!`](@ref)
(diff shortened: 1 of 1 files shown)


Datseris:
I don't like the phrasing because it seems like we are doing something wrong. Specially the word "confusing" is not something that should be used in this context (and also, there is no reason to comment on whether this is confusing or not, just state that we don't store the step).
Actually we do store the step, and we allow it for e.g. the `when` function or the `until` function so saying "we don't store the step" can be misleading. What we do is not give the step to the `model_step!` function, so we should be clear and precise with our wording and just state what is happening.

Here is what I propose:

```
Notice that the current step number is not explicitly given to the `model_step!` function,
because this is useful only for a small subset of ABMs.
If you need the step information, it is trivial to implement this by adding a
counting parameter into the model `properties`, and incrementing it by 1 each time `model_step!` is called. 
An example can be seen in the `model_step!` function of [Daisyworld](@ref),
where a `tick` is increased at each step.
```

PR #299, opened 2020-09-15
[WIP] Package benchmarks

Libbum:
WIP. Have just added some space building for now.

@Datseris: Do you usually invoke this manually? Is there some github trigger/action we could use in-conjunction to alert us that a new PR will alter performance?

Closes #297

PR #311, opened 2020-09-18
Complete rewrite of GridSpace

Datseris:
This PR, once done, will:

- Completely re-write the initial version of `GridSpace`
- Increase performance of discrete space simulations by about a full order of magnitude (see below)
- Separate graph and grid spaces into their own file (they have nothing in common anymore)
- Adapt models that use discrete space into the new changes.
- Increment version to 4.0 and start keeping track of breaking changes in the CHANGELOG

Current benchmarks of running the `benchmark/forest.jl` file, which has a *high density* forest (and thus many neighbors):

```
Times of old grid
Full model stepping
  208.446 ms (3387187 allocations: 132.57 MiB)
Move agent
  47.321 ns (0 allocations: 0 bytes)
Space neighbors
  412.500 ns (4 allocations: 304 bytes)
Iterate over space neighbors
  1.250 μs (19 allocations: 640 bytes)
Iterate over position space neighbors
  1.170 μs (19 allocations: 656 bytes)
Iterate over space neighbors2
  150.300 μs (2638 allocations: 78.63 KiB)
node neighbors
  159.250 ns (3 allocations: 304 bytes)
Add agent
  150.408 ns (1 allocation: 48 bytes)


Times of NEW grid
Full model stepping
  37.286 ms (552779 allocations: 17.85 MiB)
Space neighbors
  16.432 ns (0 allocations: 0 bytes)
Iterate over space neighbors
  281.419 ns (9 allocations: 432 bytes)
Iterate over position space neighbors
  108.368 ns (0 allocations: 0 bytes)
Iterate over space neighbors2
  29.300 μs (900 allocations: 42.19 KiB)
node neighbors
  15.630 ns (0 allocations: 0 bytes)
Move agent
  41.574 ns (0 allocations: 0 bytes)
Add agent
  142.973 ns (1 allocation: 48 bytes)
```

There is one problem I simply can't solve yet... Why is the "iterate over space neighbors" benchmark allocating? In short, why is the function 
```julia
function space_neighbors(agent::A, model::ABM{A}, args...; kwargs...) where {A<:AbstractAgent}
    all = space_neighbors(agent.pos, model, args...; kwargs...)
    id::Int = agent.id
    Iterators.filter(i -> i ≠ id, all)
end
```
resulting in allocations when iterated over...? I really have no god damn clue.


Libbum:
Looks like you fixed the allocation problem yeah? I'm not seeing anything here.

Datseris:
No I still get same numbers ...? :D

Libbum:
OK, maybe I'm using the wrong function. The PR doesn't compile for me in the current state, so I was fudging some things perhaps.

What are you running for tests at the moment?

Datseris:
So far I haven't been running any tests because I was doing major refactorings which of course broke some things. But now everything is done. The old `GridSpace` is dead, and the new `ArraySpace` just became `GridSpace`. It is a completely separated entity, but they both share a discrete space api (which is now in a new dedicated file).

We are now faced with the monumental task of making sure that tests pass :P An of course also deleting any tests related with e.g. `vertex2coord`, or e.g. testing that you can find neighbors in `GridSpace` using `Int` as position type.

But I have to leave now for about three hours :( I'll return to this later today!

Libbum:
Sure thing. I'll probably be out for a while today as well, but if you're not touching anything now I'll start on the tests.

Libbum:
const DiscreteSpace <: AbstractSpace

What do you need here? That doesnt compile.

Datseris:
Please do then! I've only started writing the Changelog so we don't forget. I also need to add a deprecation to GridSpace for the keyword `moore`.

Libbum:
Other than it not being exported, it looks fine to me:
```julia
julia> mutable struct Agent3 <: AbstractAgent    
           id::Int    
           pos::Tuple{Int,Int}    
           weight::Float64    
       end
julia> model = ABM(Agent3, GridSpace((100, 100)))
AgentBasedModel with 0 agents of type Agent3
 space: Array space with size (100, 100) and metric=chebyshev
 scheduler: fastest

julia> for _ in 1:10000
       add_agent!(model, 5.5)
       end
julia> agent = random_agent(model)
Agent3(4259, (62, 65), 5.5)
julia> collect(Agents.space_neighbors(agent, model))
19-element Array{Int64,1}:
 2175
 4336
 2522
 4122
 9916
 4683
  122
 5119
 1720
 2979
 7202
  809
 1267
 5972
 8238
 3596
 7280
 3870
 7772
julia> @benchmark Agents.space_neighbors($agent, $model)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     17.238 ns (0.00% GC)
  median time:      17.515 ns (0.00% GC)
  mean time:        18.118 ns (0.00% GC)
  maximum time:     45.747 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998
```

Datseris:
> julia> @benchmark Agents.space_neighbors($agent, $model)

heh that's not the correct benchmark. You have to benchmark _iterating_ over this output. Like I do in the `benchmark/forest` file. 

Just generating the iterator is instant with the new space ;)

Libbum:
Right. OK&mdash;thanks.

Libbum:
It's not us, its just allocations in the [Filter](https://github.com/JuliaLang/julia/blob/e37876793f77b516c30000b66b373e71b1e462e5/base/iterators.jl#L415-L457)
```julia
julia> @benchmark Iterators.filter(isodd, [1, 2, 3, 4, 5])
BenchmarkTools.Trial: 
  memory estimate:  128 bytes
  allocs estimate:  1
  --------------
  minimum time:     25.106 ns (0.00% GC)
  median time:      69.805 ns (0.00% GC)
  mean time:        80.770 ns (16.14% GC)
  maximum time:     65.389 μs (99.90% GC)
  --------------
  samples:          10000
  evals/sample:     996
```

Datseris:
ah excellent this has been tormenting me so much...!!! (it is still faster to use `filter` than to collect the IDs BTW, I've benchmarked that)

src/spaces/graph.jl
```diff
@@ -23,33 +23,42 @@ end
...
 # Agents.jl space API
 #######################################################################################
-random_position(model::ABM{<:AbstractAgent, <:GraphSpace}) = rand(1:nv(model))
+random_position(model::ABM{<:AbstractAgent,<:GraphSpace}) = rand(1:nv(model))
 
-function remove_agent_from_space!(agent::A, model::ABM{A,<:GraphSpace}) where {A<:AbstractAgent}
+function remove_agent_from_space!(
+    agent::A,
+    model::ABM{A,<:GraphSpace},
+) where {A<:AbstractAgent}
```
Datseris, line 50: Can you configure your editor / linter to ident these one level? it is intuitive to see a code block closing and the next to start at a higher identation level.

Libbum:
We don't have `has_empty_nodes` anymore, so one test has changed to `length(find_empty_nodes) == 0`. (not sure if this was an oversight or not

Having issues with this last one from api_tests.jl:
```julia
model = ABM(Agent1, GridSpace((5,5)))
add_agent!((1,3), model)
add_agent!((1,3), model)
add_agent!((5,2), model)
for agent in get_node_contents((1,3), model)
    kill_agent!(agent, model)
end
@test nagents(model) == 1
```
But returns 2.

```julia
for agent in get_node_contents((1,3), model)
           println(agent)
           println(model.agents)
           #kill_agent!(agent, model)         
           #println(agent)                    
           #println(model.agents)
         end      
1
Dict{Int64,Agent1}(2 => Agent1(2, (1, 3)),3 => Agent1(3, (5, 2)),1 => Agent1(1, (1, 3)))
2
Dict{Int64,Agent1}(2 => Agent1(2, (1, 3)),3 => Agent1(3, (5, 2)),1 => Agent1(1, (1, 3)))
```

```julia
julia> for agent in get_node_contents((1,3), model)
           println(agent)
           println(model.agents)
           kill_agent!(agent, model)          
           println(agent)                     
           println(model.agents)
         end    
1
Dict{Int64,Agent1}(2 => Agent1(2, (1, 3)),3 => Agent1(3, (5, 2)),1 => Agent1(1, (1, 3)))
1
Dict{Int64,Agent1}(2 => Agent1(2, (1, 3)),3 => Agent1(3, (5, 2)))
```

Manually, the `get_node_contents` and `kill_agent!` calls work for both agents that need to be removed, but the iteration fails if we take one agent away...

Datseris:
That's extremely weird... I have 15 minutes now and I am investigating.. Will post if I find something.

Datseris:
Same test fails for GraphSpace:
```julia
mutable struct Agent2 <: AbstractAgent
    id::Int
    pos::Int
end

model = ABM(Agent2, GraphSpace(complete_digraph(10)))
add_agent!(1, model)
add_agent!(1, model)
add_agent!(2, model)
for agent in get_node_contents(1, model)
    kill_agent!(agent, model)
end
@test nagents(model) == 1

```
what the hell did I change that leads to this...? How did we do it before? Like at stable version 3.7?

Datseris:
`get_node_contents` is the same as before: it returns the underlying entry which is Vector{Int}

Libbum:
Ah. It's a combination of issues.

The test WAS `get_node_agents`, not `get_node_contents`, but I changed it since there's no `get_node_agents` now. It's difficult to know what you've killed off due to the changes and what has just accidentally been left out.

I'll add a  `get_node_agents` back in.

Datseris:
Wait, why do we need `get_node_agents`...? (p.s. I've pushed a new api function `kill_agents!` to avoid doing this kind of loops - doesn't solve the problem of course)

Datseris:
I really don't know how this test used to pass, but I now realize that it is not a "true" issue. Because iteration over `node_contents` is dynamic, if it has two elements and you remove one INSIDE the for loop, then at the end of the for loop there is only one element at the first step of the loop. But if there is only one element, the loop will of course terminate after the first step, because it has exhausted the iterator.

Libbum:
I guess we don't, but it's the reason why this works in 3.7 and not now.

```julia
get_node_agents(x, model::ABM{A,<:DiscreteSpace}) where {A} = [model[id] for id in get_node_contents(x, model)]
```
So it instantiates a new list, rather than a pointer to model contents.

Datseris:
I think it is best to just do a copy in the test. We should also add a notice somewhwere central in the docs that now pretty much all iterators from Agents.jl are fully dynamic and their state would change if a user adds or removes an agent. Thus, it is best to `copy` or `collect` the iteration results to have a stored version of the "initial condition". What do you think?

Libbum:
Agreed.

Datseris:
I'm heading out now, will get around this tonight if you haven't do so! (meanwhile feel free to add copy when necessary to make tests passing)

Libbum:
`space_neighbors` in both Grid and Graph needs some though. Half of it is now iterators, half is not. Some isn't even implemented either. 

The problem I see with the iterator case is that many of the constructions are complex, thus when you go to collect, the timing ends up being quite a lot more than what you would have with the `vcat` solution.

I'm not too sure what the best option is there without further discussion, so I'm going to move on from there and come back to `space_test` later.

Libbum:
Well actually, that's all the tests sorted now apart from `space_neighbors`.

Datseris:
@Libbum there is space neighbors for agents, please see the `space_interaction_API.jl`:
```julia
"""
    space_neighbors(agent::AbstractAgent, model::ABM, r=1)

Same as `space_neighbors(agent.pos, model, r)` but the iterator *excludes* the given
`agent`'s id.
"""
function space_neighbors(agent::A, model::ABM{A}, args...; kwargs...) where {A<:AbstractAgent}
    all = space_neighbors(agent.pos, model, args...; kwargs...)
    id::Int = agent.id
    Iterators.filter(i -> i ≠ id, all)
end
```

Notice how it is not just equivalent to `space_neighbors(agent.pos, ...)` because you need to exclude the calling agent.

In fact, many functions that were initially implemented space by space are now space agnostic in file `space_interaction_API.jl`. If you crossed an error that some method doesn't exist, it is probably because I incorrectly specified the abstract types (e.g. writing `A` instead of `AbstractAgents`)

Libbum:
Think I had my wires crossed on that one, too much dirty/half commits when testing. Yeah, seems to be there now.

A few other notes:

- `node_neighbors(pos, graphspace, 2)` is pretty bad performance-wise, we need to look at that.
- Default for grids is now chebyshev & periodic. That may make sense, but previously we had not periodic and euclidean as defaults. Should at least put this in the changelog if it stays that way.

Datseris:
Yeah any part of the code that is pre-1.0 has suboptimal performance; it was written too long ago and never benchmarked either. But we should take care off graph space performance in another PR. I'll make tests here passing, adjust docs accordingly and then merge.

CHANGELOG.md
```diff
@@ -1,14 +1,19 @@
...
+  (see Breaking changes).
 
 ## Breaking changes
-- Keyword `moore` of `GridSpace` doesn't exist anymore. Use `metric` instead.
-
+All changes in this section (besides changes to default values) are deprecated and
+therefore are not "truly breaking".
 
+- Keyword `moore` of `GridSpace` doesn't exist anymore. Use `metric` instead.
+- Default arguments for `GridSpace` are not `periodc = false, metric = :chebyshev`.
```
Libbum, line 15: not -> now

docs/make.jl
```diff
@@ -1,4 +1,7 @@
 cd(@__DIR__)
+using Pkg; Pkg.activate(@__DIR__)
+const CI = get(ENV, "CI", nothing) == "true"
+CI && (ENV["GKSwstype"] = "100")
```
Libbum, line 4: Out of curiosity, what is GKSwstype ?

Datseris: No clue, its so using GR works on Travis...

Datseris:
running benchmark `judge` now...

docs/make.jl
Libbum: Fair enough.

Datseris:
Which commit am I supposed to compare to with `judge`...? @Libbum do you think you could dfo the benchmark comparison, pretty please...?

Libbum:
Anything after 03cb5b11f0 I'm guessing (since that's when the benchmarks landed). But yeah, I'll run it.

I'm almost done with the graph enhancements too.

Datseris:
> I'm almost done with the graph enhancements too. 

Awesome! but please open a new PR after we merge this (i.e. don't push them here)

Libbum:
> Awesome! but please open a new PR after we merge this (i.e. don't push them here)

Oh. Just did. I'll revert.

Datseris:
After the heart attack I got from loosing code I am trying to keep PRs as self contained as possible :D This one is pretty massive because it does the complete replacement of GridSpace, but anything not absolutely necessary here to get passing tests, let's do it in another PR.

Libbum:
Sure :) I'm running the judge now, will post it here when its done.

Libbum:
We've left a trail of semi-broken pieces through master recently. Just trying to find a commit that doesn't break one or another of the benchmarks to use as a baseline.

Datseris:
Can you just post the latest benchmarks here and we just compare with https://github.com/JuliaDynamics/Agents.jl/pull/299#issuecomment-694779393 just by looking at it? (because then they will come from the same computer)

Libbum:
Sure, give me 10 minutes for that, we'll do something more official later.

Libbum:
```julia
Benchmarkresults:
    Package: /mnt/turtle/repos/Agents.jl
    Date: 19 Sep 2020 - 23:10
    Package commit: 9a1b3b
    Julia commit: 96786e
    BenchmarkGroup:
        5-element BenchmarkTools.BenchmarkGroup:
          tags: ["Agents"]
          "space_creation" => 3-element BenchmarkTools.BenchmarkGroup:
                  tags: ["graph", "grid", "continuous"]
                  "grid" => Trial(6.372 ms)
                  "graph" => Trial(2.242 ms)
                  "continuous" => Trial(119.315 μs)
          "model" => 2-element BenchmarkTools.BenchmarkGroup:
                  tags: ["initialise", "initialise_union"]
                  "initialise_union" => 3-element BenchmarkTools.BenchmarkGroup:
                          tags: ["graph", "grid", "continuous"]
                          "grid" => Trial(5.950 μs)
                          "graph" => Trial(5.616 μs)
                          "continuous" => Trial(6.215 μs)
                  "initialise" => 3-element BenchmarkTools.BenchmarkGroup:
                          tags: ["graph", "grid", "continuous"]
                          "grid" => Trial(1.047 μs)
                          "graph" => Trial(1.052 μs)
                          "continuous" => Trial(1.508 μs)
          "grid" => 6-element BenchmarkTools.BenchmarkGroup:
                  tags: ["add", "add_union", "move", "neighbors", "node"]
                  "add" => 7-element BenchmarkTools.BenchmarkGroup:
                          tags: ["agent", "agent_pos", "agent_single", "create", "create_pos", "create_single", "create_fill"]
                          "create_single" => Trial(12.058 ms)
                          "create" => Trial(1.108 μs)
                          "agent_single" => Trial(12.152 ms)
                          "create_fill" => Trial(22.557 μs)
                          "create_pos" => Trial(46.000 ns)
                          "agent" => Trial(236.174 ns)
                          "agent_pos" => Trial(20.557 ns)
                  "add_union" => 4-element BenchmarkTools.BenchmarkGroup:
                          tags: ["agent", "agent_pos", "agent_single", "agent_fill"]
                          "agent_single" => Trial(12.075 ms)
                          "agent_fill" => Trial(22.651 μs)
                          "agent" => Trial(212.298 ns)
                          "agent_pos" => Trial(19.349 ns)
                  "collect" => 2-element BenchmarkTools.BenchmarkGroup:
                          tags: ["init_agent", "store_agent"]
                          "store_agent" => Trial(1.343 ms)
                          "init_agent" => Trial(1.625 μs)
                  "neighbors" => 6-element BenchmarkTools.BenchmarkGroup:
                          tags: ["space_pos", "space_agent", "space_pos_iterate", "space_agent_iterate", "node_pos", "node_agent"]
                          "space_agent_iterate" => Trial(426.192 μs)
                          "node_pos" => Trial(20.490 ns)
                          "node_agent" => Trial(25.000 ns)
                          "space_pos" => Trial(23.000 ns)
                          "space_pos_iterate" => Trial(87.903 μs)
                          "space_agent" => Trial(25.000 ns)
                  "node" => 1-element BenchmarkTools.BenchmarkGroup:
                          tags: ["contents", "agents"]
                          "contents" => Trial(2.305 ns)
                  "move" => 3-element BenchmarkTools.BenchmarkGroup:
                          tags: ["random", "pos", "single"]
                          "random" => Trial(66.530 ns)
                          "pos" => Trial(23.147 ns)
                          "single" => Trial(3.443 μs)
          "graph" => 6-element BenchmarkTools.BenchmarkGroup:
                  tags: ["add", "add_union", "move", "neighbors", "node"]
                  "add" => 6-element BenchmarkTools.BenchmarkGroup:
                          tags: ["agent", "agent_pos", "agent_single", "create", "create_pos", "create_single"]
                          "create_single" => Trial(287.679 μs)
                          "create" => Trial(943.000 ns)
                          "agent_single" => Trial(333.632 μs)
                          "create_pos" => Trial(44.000 ns)
                          "agent" => Trial(96.560 ns)
                          "agent_pos" => Trial(18.715 ns)
                  "add_union" => 3-element BenchmarkTools.BenchmarkGroup:
                          tags: ["agent", "agent_pos", "agent_single"]
                          "agent_single" => Trial(280.182 μs)
                          "agent" => Trial(95.924 ns)
                          "agent_pos" => Trial(18.695 ns)
                  "collect" => 2-element BenchmarkTools.BenchmarkGroup:
                          tags: ["init_agent", "store_agent"]
                          "store_agent" => Trial(38.008 μs)
                          "init_agent" => Trial(1.669 μs)
                  "neighbors" => 6-element BenchmarkTools.BenchmarkGroup:
                          tags: ["space_pos", "space_agent", "space_pos_iterate", "space_agent_iterate", "node_pos", "node_agent"]
                          "space_agent_iterate" => Trial(30.729 μs)
                          "node_pos" => Trial(18.937 ns)
                          "node_agent" => Trial(20.900 ns)
                          "space_pos" => Trial(1.422 μs)
                          "space_pos_iterate" => Trial(29.446 μs)
                          "space_agent" => Trial(1.640 μs)
                  "node" => 2-element BenchmarkTools.BenchmarkGroup:
                          tags: ["contents", "agents"]
                          "nodes" => Trial(12.000 ns)
                          "contents" => Trial(1.731 ns)
                  "move" => 3-element BenchmarkTools.BenchmarkGroup:
                          tags: ["random", "pos", "single"]
                          "random" => Trial(35.271 ns)
                          "pos" => Trial(22.503 ns)
                          "single" => Trial(114.000 ns)
          "continuous" => 5-element BenchmarkTools.BenchmarkGroup:
                  tags: ["add", "add_union", "move", "neighbors"]
                  "add" => 4-element BenchmarkTools.BenchmarkGroup:
                          tags: ["agent", "agent_pos", "create", "create_pos"]
                          "create" => Trial(5.042 μs)
                          "create_pos" => Trial(2.202 μs)
                          "agent" => Trial(8.720 μs)
                          "agent_pos" => Trial(5.694 μs)
                  "add_union" => 2-element BenchmarkTools.BenchmarkGroup:
                          tags: ["agent", "agent_pos"]
                          "agent" => Trial(8.708 μs)
                          "agent_pos" => Trial(6.095 μs)
                  "collect" => 2-element BenchmarkTools.BenchmarkGroup:
                          tags: ["init_agent", "store_agent"]
                          "store_agent" => Trial(12.961 ms)
                          "init_agent" => Trial(1.944 μs)
                  "neighbors" => 6-element BenchmarkTools.BenchmarkGroup:
                          tags: ["space_pos", "space_agent", "space_pos_iterate", "space_agent_iterate", "nearest", "interacting"]
                          "space_agent_iterate" => Trial(24.876 ms)
                          "nearest" => Trial(10.214 ms)
                          "space_pos" => Trial(15.107 ms)
                          "space_pos_iterate" => Trial(24.022 ms)
                          "interacting" => Trial(208.912 s)
                          "space_agent" => Trial(9.891 ms)
                  "move" => 2-element BenchmarkTools.BenchmarkGroup:
                          tags: ["update", "vel"]
                          "update" => Trial(2.240 μs)
                          "vel" => Trial(2.184 μs)
```

Libbum:
Numbers are looking good mostly across the board. `move_agent_single!` may need to be checked, and the continuous space results are up in some places down in others&mdash;they're less of a concern though. Think we can merge.

src/spaces/grid.jl
```diff
@@ -0,0 +1,241 @@
...
+        space.hoods[r]
+    else
+        initialize_neighborhood!(space, r)
+    end
+    _grid_space_neighborhood(α, space, hood)
+end
+
+function _grid_space_neighborhood(
+    α::CartesianIndex,
+    space::GridSpace{D,true},
```
Libbum, line 164: What is this constructor `GridSpace{D,true}` vs the `GridSpace{D,false}` below? (I'm wondering if this function is ever called & perhaps should have been dropped with the rest of the TimeseriesPrediction code we aren't using)

Datseris: It is correct there. the `true` means periodic. THere is one more below with `false` for non-periodic.

Issue #301, opened 2020-09-16
API Clarifications


Datseris:
I think the names we have chosen for all these `node_neighbors` and `space_neighgbhors` are not optimal and can be confusing. I'd suggest the alternatives:

* `neighborhing_agents`
* `neighborhing_positions`

or even better, to save space and not having to type one of the most difficult words in the english lexicon:

* `nearby_agents`
* `nearby_positions`

which can deprecate the existing names.

Datseris:
Further suggestioN: rename `nodes` function into `positions` that iterates over the "positions" of the space, which are either Int or Tuple.

Datseris:
Also: `get_node_contents` -> `agents_at_pos`

Libbum:
I'd agree with all these changes. We'd need depreciation guards. Most importantly, this would bump us to 4.0 yes? It's a breaking change.

Datseris:
Not with deprecations. You can deprecate `@deprecate get_node_contents agents_in_pos` and nothing breaks. However depending on the changes to space we might go into a new version. I am working on the new space now...

Libbum:
Let's do this once #295 is merged, provided @kavir1698 is on board.

kavir1698:
Yeah, I agree with the changes. The new names are easier to remember.

fbanning:
If I may add something to the discussion: I found these names to be confusing as well. `nearby_agents` would intuitively make a lot more sense. I would argue that `nearby_positions` would not be in line with the standard naming for the positional agent field `pos` and therefore `nearby_pos` could make more sense here.

Datseris:
Good point but there is another difficulty: `nearby_pos` hints at singular, and I would imagine it would return a single, random, nearby position.

fbanning:
Maybe it might also be interesting to have a look at how other APIs did it. Not saying that how others did it would be the best approach, but maybe it might help with transitions (no need to "reinvent the wheel" when it comes to function naming).

For example NetLogo (which is widely used in the ABM community as you all might know) has `neighbors` (moore) and `neighbors4` (von Neumann). See [their documentation](http://ccl.northwestern.edu/netlogo/docs/dictionary.html#neighbors) for this. They both return a list of neighboring patches (which equals to a list of `pos` in our case).

Of course we don't need to differentiate between Moore and von Neumann, so it will be a bit simpler in our case. Following this, we could also name our function simply `neighbors(x; include_self = false)` which might return either

[...]

PR #315, opened 2020-09-20
clean up grid space

Datseris:
unfortunately even though I ported most code from TimeseriesPrediction.jl, we can't use it due to design. We still use a lot of stuff though.

### DIFF
--- a/benchmark/forest.jl
+++ b/benchmark/forest.jl
@@ -3,14 +3,15 @@ using BenchmarkTools
 using Random
 
 # Define new space model
-mutable struct Tree <: AbstractAgent    id::Int
+mutable struct Tree <: AbstractAgent
+    id::Int
     pos::Tuple{Int,Int}
     status::Bool  # true is green and false is burning
 end
-function forest_fire_array(; f = 0.02, d = 1.0, p = 0.01,
-    griddims = (20, 20), seed = 111, periodic = true)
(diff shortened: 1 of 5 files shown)


Datseris:
also add `has_empty_nodes` which I have forgotten.

Libbum:
`get_node_agents` is also missing (that was the one I should have added, but actually added the space_neighbor function yesterday)

Libbum:
I didn't absorb too much of that `inner` was doing, are we loosing performance by removing it, or just not gaining something additional since it doesn't work in this context?

Datseris:
I turned out not using it because it made us lose performance by using it. It's a bit complicated to explain why and since we ditch it I'll not go into details ... :P

Datseris:
`get_node_agents` is just `(mode[id] for id in get_node_contents(pos, model))`, right?

Datseris:
I don't know whether we should have it, or if we should, then we should also have same version for `space_neighbors`, that returns an agent iterator instead of id iterator...?

Libbum:
That's it yeah. We added it from user feedback if I recall, but I don't think it really needs to be around. We use it in three examples where it doesn't truly need to be there. I wouldn't go complicating space_neighbors anymore at the moment. Let's do #301, which will need a documentation overhaul too, see if it really needs to remain.

PR #316, opened 2020-09-20
Graph neighbor improvements (Discussion)

### LINKED ISSUE #312: Finding neighbors in GraphSpace (for r>1) has poor performance.

Libbum:
Closes #312 

I've done a number of iterations on this. Whilst it's not the best performing in every situation, it aligns with our output requirements for these functions (iterators, not vectors) and provides the best all round numbers that I can find.

The big discrepancy is that the LightGraphs implementation that returns a vector must be wrapped in an iterator, and that increases the time spent here. Any ideas how to mitigate this? (we could state that graph space returns vectors and grid space iterators, but that's not a clean solution).

(tests have not been updated, that will come once we get an agreed upon implementation)

```julia
model = ABM(Agent5, GraphSpace(path_graph(500)))
for _ in 1:2000 add_agent!(model, rand()) end
a = random_agent(model)
pos = a.pos
```

Baseline
```julia
@btime node_neighbors($pos, $model) # vector
  19.265 ns (0 allocations: 0 bytes)

@btime node_neighbors($pos, $model, 20) # vector
  126.809 μs (2110 allocations: 85.63 KiB)

@btime space_neighbors($pos, $model) # vector
  169.447 ns (3 allocations: 304 bytes)
  
@btime space_neighbors($pos, $model, 20) # vector
  131.059 μs (2113 allocations: 87.36 KiB)

@btime space_neighbors($a, $model) #iter
  317.117 ns (5 allocations: 352 bytes)
  
@btime space_neighbors($a, $model, 20) #iter
  131.335 μs (2115 allocations: 87.41 KiB)
```

Current
```julia
@btime node_neighbors($pos, $model) #iter 
  84.422 ns (1 allocation: 32 bytes)

@btime node_neighbors($pos, $model, 20) #iter 
  102.718 μs (1270 allocations: 44.53 KiB)

@btime space_neighbors($pos, $model) #iter 
  286.460 ns (6 allocations: 272 bytes)

@btime space_neighbors($pos, $model, 20) #iter 
  103.476 μs (1270 allocations: 44.53 KiB)

@btime space_neighbors($a, $model) #iter 
  458.462 ns (9 allocations: 432 bytes)

@btime space_neighbors($a, $model, 20) #iter 
  103.711 μs (1270 allocations: 44.53 KiB)
```

George: an implementation with vectors:
```julia
function node_neighbors_vec(    
    node_number::Integer,    
    model::ABM{A,<:GraphSpace},    
    radius::Integer;    
    kwargs...,    
) where {A}    
    output = copy(collect(node_neighbors(pos, model; kwargs...)))
    for _ in 2:radius 
        newnns = (collect(node_neighbors(nn, model; kwargs...)) for nn in output)
        append!(output, reduce(vcat, newnns))
        unique!(output)
    end
    Iterators.filter(i->i!=node_number, (n for n in output))    
end

@btime node_neighbors_vec($pos, $model, 20)  
  112.217 μs (1388 allocations: 195.80 KiB)
```

We're forced to copy and collect a lot here. One possible improvement could be to take most of the meat out of the `r=1` function that calls `neighborfn` and put all of that behind a function guard, where the only purpose of node_neighbors is to wrap the array in an iterator. This new function can be called internally from the node_neighbors_vec function, which drops the need to `collect` so many times. The `copy` will still be needed though I think.

Any suggestions anywhere?

### DIFF
--- a/src/spaces/graph.jl
+++ b/src/spaces/graph.jl
@@ -38,7 +38,6 @@ function Base.show(io::IO, space::GraphSpace)
     )
 end
 
-
 #######################################################################################
 # Agents.jl space API
 #######################################################################################
@@ -91,7 +90,13 @@ nodes(model::ABM{<:AbstractAgent,<:GraphSpace}) = 1:nv(model)
 #######################################################################################
 function space_neighbors(pos::Int, model::ABM{A,<:GraphSpace}, args...; kwargs...) where {A}
     nn = node_neighbors(pos, model, args...; kwargs...)
-    Iterators.flatten(model.space.s[i] for i in Iterators.flatten((pos,nn)))
+    # This call is faster than reduce(vcat, ..), or Iterators.flatten
+    vcat(model.space.s[pos], model.space.s[nn]...)
+end
+
+function space_neighbors(agent::A, model::ABM{A,<:GraphSpace}, args...; kwargs...) where {A<:AbstractAgent}
+    all = space_neighbors(agent.pos, model, args...; kwargs...)
+    filter!(i -> i ≠ agent.id, all)
 end
 
 function node_neighbors(
@@ -109,7 +114,7 @@ function node_neighbors(
     else
         LightGraphs.all_neighbors
     end
-    Base.Generator(Int, neighborfn(model.space.graph, node_number))
+    neighborfn(model.space.graph, node_number)
 end
 
 function node_neighbors(
@@ -118,10 +123,12 @@ function node_neighbors(
     radius::Integer;
     kwargs...,
 ) where {A}
-    neighbor_nodes = Set(node_neighbors(node_number, model; kwargs...))
+    output = copy(node_neighbors(node_number, model; kwargs...))
     for _ in 2:radius
-        newnns = (node_neighbors(nn, model; kwargs...) for nn in neighbor_nodes)
-        union!(neighbor_nodes, newnns...)
+        newnns = (node_neighbors(nn, model; kwargs...) for nn in output)
+        append!(output, reduce(vcat, newnns))
+        unique!(output)
     end
-    Iterators.filter(i->i!=node_number, Base.Generator(Int, neighbor_nodes))
+    filter!(i -> i != node_number, output)
 end
+
--- a/src/spaces/graph.jl
+++ b/src/spaces/graph.jl
@@ -91,11 +91,7 @@ nodes(model::ABM{<:AbstractAgent,<:GraphSpace}) = 1:nv(model)
 #######################################################################################
 function space_neighbors(pos::Int, model::ABM{A,<:GraphSpace}, args...; kwargs...) where {A}
     nn = node_neighbors(pos, model, args...; kwargs...)
-    # TODO: Use flatten here or something for performance?
-    # `model.space.s[nn]...` allocates, because it creates a new array!
-    # Also `vcat` allocates a second time
-    # We include the current node in the search since we are searching over space
-    vcat(model.space.s[pos], model.space.s[nn]...)
+    Iterators.flatten(model.space.s[i] for i in Iterators.flatten((pos,nn)))
 end
 
 function node_neighbors(
@@ -113,7 +109,7 @@ function node_neighbors(
     else
         LightGraphs.all_neighbors
     end
-    neighborfn(model.space.graph, node_number)
+    Base.Generator(Int, neighborfn(model.space.graph, node_number))
 end
 
 function node_neighbors(
@@ -123,25 +119,9 @@ function node_neighbors(
     kwargs...,
 ) where {A}
     neighbor_nodes = Set(node_neighbors(node_number, model; kwargs...))
-    included_nodes = Set()
-    for rad in 2:radius
-        templist = Vector{Int}()
-        for nn in neighbor_nodes
-            if !in(nn, included_nodes)
-                newns = node_neighbors(nn, model; kwargs...)
-                for newn in newns
-                    push!(templist, newn)
-                end
-            end
-        end
-        for tt in templist
-            push!(neighbor_nodes, tt)
-        end
+    for _ in 2:radius
+        newnns = (node_neighbors(nn, model; kwargs...) for nn in neighbor_nodes)
+        union!(neighbor_nodes, newnns...)
     end
-    nlist = collect(neighbor_nodes)
-    j = findfirst(a -> a == node_number, nlist)
-    if j != nothing
-        deleteat!(nlist, j)
-    end
-    return nlist
+    Iterators.filter(i->i!=node_number, Base.Generator(Int, neighbor_nodes))
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - `GraphSpace` and `GridSpace` are completely separated entities, reducing complexity of source code dramatically, and removing unnecessary functions like `vertex2coord`.
 - Many things have been renamed to have clearer name that indicates their meaning
   (see Breaking changes).
+- Performance increase of finding neighbors in GraphSpace with r > 1.
 
 ## Breaking changes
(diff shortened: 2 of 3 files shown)


Datseris:
> Whilst it's not the best performing in every situation, it aligns with our output requirements for these functions (iterators, not vectors) and provides the best all round numbers that I can find.

Being an iterator is _not_ a requirement. It is used whenever it is beneficial for performance, but you shouldn't sacrifice performance for it.

src/spaces/graph.jl
```diff
@@ -91,11 +91,7 @@ nodes(model::ABM{<:AbstractAgent,<:GraphSpace}) = 1:nv(model)
 #######################################################################################
 function space_neighbors(pos::Int, model::ABM{A,<:GraphSpace}, args...; kwargs...) where {A}
     nn = node_neighbors(pos, model, args...; kwargs...)
-    # TODO: Use flatten here or something for performance?
-    # `model.space.s[nn]...` allocates, because it creates a new array!
-    # Also `vcat` allocates a second time
-    # We include the current node in the search since we are searching over space
-    vcat(model.space.s[pos], model.space.s[nn]...)
+    Iterators.flatten(model.space.s[i] for i in Iterators.flatten((pos,nn)))
```
Datseris, line 94: perhaps it is more performant to do `push!(nn, pos)` instead of `Iterators.flatten((pos, nn))`?

src/spaces/graph.jl
```diff
@@ -113,7 +109,7 @@ function node_neighbors(
     else
         LightGraphs.all_neighbors
     end
-    neighborfn(model.space.graph, node_number)
+    Base.Generator(Int, neighborfn(model.space.graph, node_number))
```
Datseris, line 112: Why? `neighborfn` doesn't return Integers?

src/spaces/graph.jl
Libbum: You have to `copy(nn)` to do this, leading to more time and allocations.

src/spaces/graph.jl
Libbum: It does, but not an iterator.

Libbum:
> Being an iterator is _not_ a requirement. It is used whenever it is beneficial for performance, but you shouldn't sacrifice performance for it.

In `space_interaction_API.jl`:

```julia
"""
    space_neighbors(position, model::ABM, r=1; kwargs...) → ids
Return an iterator of the ids of the agents within "radius" `r` of the given `position`
(which must match type with the spatial structure of the `model`).
"""

"""
    node_neighbors(position, model::ABM, r=1; kwargs...) → positions
Return an iterator of all positions within "radius" `r` of the given `position`
(which excludes given `position`).
"""
```

These need to be changed then. If so, I can make these better.
Problem is, graph space will return a vectors, grid space will return iterators. It's not consistent.

src/spaces/graph.jl
Datseris: `node_neighbors` is the result of finally calling `LightGraph.neighbors`, which returns `Vector{Int}`, right?

Datseris:
> These need to be changed then. If so, I can make these better.

The statement is not incorrect. A vector is also an iterator because it can be iterated over. EDIT: okay, yes, the statement needs to say _iterable_ instead of _iterator_

> Problem is, graph space will return a vectors, grid space will return iterators. It's not consistent.

This is not a problem. Actually, why would this be a problem...? If you can avoid allocations you do, if you can't then there isn't much to be done anyway. Julia as a whole operates on this level:

```julia
julia> x = 1:100
1:100

julia> x[1:20] # iterator
1:20

julia> x = collect(x)
100-element Array{Int64,1}:
   1
   2
   3
   4
   ⋮
  96
  97
  98
  99
 100

julia> x[1:20] # iterator
20-element Array{Int64,1}:
  1
  2
  3
 ...
 16
 17
 18
 19
 20
```
I have never seen any function documentation string that makes a distinction between getting in a range or a vector.

---

Okay, you are correct in the sense that we have to change the word _iterator_ to _iterable_.

src/spaces/graph.jl
Libbum: Yes, but the call is by reference, so when you start pushing numbers into `nn` you're adding nodes to the graph and things stop working very quickly.

Libbum:
> we have to change the word *iterator* to *iterable*.

Sure. I'll update the docs & benchmarks to reflect this. Results will be much better.

src/spaces/graph.jl
Datseris: What??? Are you sure that `LightGraph.neighbors` returns a reference? Because calling `LightGraph.neighbors` **allocates**.

Libbum:
New benchmarks:
```julia
@btime node_neighbors($pos, $model)
  19.157 ns (0 allocations: 0 bytes)

@btime node_neighbors($pos, $model, 20)
  58.307 μs (586 allocations: 145.81 KiB)

@btime space_neighbors($pos, $model)
  163.347 ns (3 allocations: 288 bytes)
  
@btime space_neighbors($pos, $model, 20)
  58.402 μs (589 allocations: 147.67 KiB)
  
@btime space_neighbors($a, $model)
  208.434 ns (4 allocations: 304 bytes)
  
@btime space_neighbors($a, $model, 20)
  59.973 μs (590 allocations: 147.69 KiB)
```

I don't think we'll get better than this, but feel free to poke around.

src/spaces/graph.jl
Libbum: Haven't gone into the depths of the implementation, but we certainly hit the mutation issue that made us do #281, so yeah&mdash;it's an issue that needs to be considered.

src/spaces/graph.jl
```diff
@@ -91,7 +90,13 @@ nodes(model::ABM{<:AbstractAgent,<:GraphSpace}) = 1:nv(model)
 #######################################################################################
 function space_neighbors(pos::Int, model::ABM{A,<:GraphSpace}, args...; kwargs...) where {A}
     nn = node_neighbors(pos, model, args...; kwargs...)
-    Iterators.flatten(model.space.s[i] for i in Iterators.flatten((pos,nn)))
+    # This call is faster than reduce(vcat, ..), or Iterators.flatten
+    vcat(model.space.s[pos], model.space.s[nn]...)
```
Datseris, line 94: Perhaps
```julia
x = copy(model.space.s[pos])
for z in nn
append!(x, model.space.s[z]
end
return x
```
is the fastest version...? I don't really know the performance caveats of `vcat` (wtherhr it performns well for many arguments0

src/spaces/graph.jl
```diff
@@ -91,7 +90,13 @@ nodes(model::ABM{<:AbstractAgent,<:GraphSpace}) = 1:nv(model)
...
+end
+
+function space_neighbors(agent::A, model::ABM{A,<:GraphSpace}, args...; kwargs...) where {A<:AbstractAgent}
```
Datseris, line 97: Why did you have to explicitly add this method? Isn't it defined already in the space API?

Libbum: No, that uses `Iterators.filter`, which is good for when you want to filter on an existing iterator (used in `GridSpace`), but if you want to drop something out of a `Vector`, `filter!` is faster.

Datseris: Of course, awesome.

src/spaces/graph.jl
Libbum: Nope, `vcat` is faster. 
Commited version:
```julia
@btime space_neighbors($pos, $model, 20)
  58.742 μs (589 allocations: 147.55 KiB)

@btime space_neighbors($pos, $model, 100)
  1.950 ms (10328 allocations: 9.44 MiB)
```
Append version:
```julia
@btime space_neighbors2($pos, $model, 20)
  61.719 μs (632 allocations: 149.75 KiB)

@btime space_neighbors2($pos, $model, 100)
  2.066 ms (10490 allocations: 9.45 MiB)
```

PR #322, opened 2020-09-25
Compartment space

### LINKED ISSUE #313: New continuous space specification: CompartmentSpace
PR https://github.com/JuliaDynamics/Agents.jl/pull/311 altered the underlying GridSpace fundamentally, and now enables us to move forwards with a new version of Continuous Space which I believe will have large performance benefits over the existing one. This is to be proven of course, but given that just `move_agent!` [takes micro seconds instead of nanoseconds in continuous space](https://github.com/JuliaDynamics/Agents.jl/pull/299#issuecomment-694779393), it is pretty clear that we will get at least a full order of magnitude speedup. 

The new space will be called CompartmentSpace until we its PR is fully operation. If it indeed has better performance in every aspect, it will replace the current continuous space entirely.

# Core idea
The agent and position types in CompartmentSpace are as were before. The new space now has as part of its field a `GridSpace` instance. A number `x::Float64` defines the grid spacing with respect to real space. e.g. if the user creates a CompartmentSpace with real dimensions (10.0, 20.0) and the grid spacing is 0.1, the the underlying GridSpace will have size (100, 200). Internal test ensures `x` is such that `10.0 / x == Integer`.

All agents in the CompartementSpace are being kept track of in the underlying GridSpace as well. This means that our `add_agent` and `move_agent` functions take care to first update the agents floating type position, but then also update its position in the underlying grid space if e.g. the `move` made the agent move enough to change grid cell.

`move_agent` obtains two versions: `move_agent(agent, model, dt::Real)` and `move_agent(agent, pos, model)`. The first one simple calculates what is the next position of the agent and calls the later version. The later version checks if the agent changes grid cell or not.

**Our new space specification protocol in the file `src/core/space_interaction_API.jl` makes it clear on how to implement a new space**.

# Finding neighbors
Now, main part we care about is `space_neighbors(position, model, r)` (soon to be renamed to `nearby_agents`). Here is the gist of it.

1. `space_neighbors` has a keyword argument `exact = false`. Just like how the space_neighbors of GraphSpace also has a keyword argument.
1. If `exact=false` we do as follows (please see image below)
    1. First we calculate the displacement of the given position from the center of the grid cell, `δ`, given by blue color below
    1. Calculate  `r' = ceil((r+δ)/x)`, which covers the maximum possible further cell the search radius cloud reach.
    2. Using the underlying grid space, find all cells that are within `r'`.
    3. Return all agents within those cells, exactly like it is done in GridSpace (see its implementation of `space_neighbors(pos, model)`
1. If `exact = true`, then 
    1. We have to split the collection of grid cell locations into two steps: in the first step we collect all grid cells that are **absolutely guaranteed** to be entirely contained within the search radius. These are with green color below and are easy to find analytically. All agent ids in those cells are part of the output
    2. We separately collect the cells that are in purple color, which also can be found analytically where they should be. 
    3. For all agents in the purple cells, we iterate using a formal call to `filter!`, according the distance of the agents' position and the given position in `space_agents`.

image

The image is for euclidean distance, same can be done for Chebyshev distance. Although, to be frank, continuous space and Chebyshev metric don't make much sense to me and perhaps we should drop Chebyshev support.

(I only added colors to the upper half of the image, too lazy to put everywhere!)

What do you think? Don't forget that our new GridSpace is **very fast**.

kavir1698:
closes #313 
WIP
This is a first try to make the compartment space. @Datseris is it along the lines of your suggestion?


Libbum:
From the surface, it looks pretty good so far. Seems like we may be able to remove some of the neighbor functions and put them in the space API file (since they are essentially copies of many of the `DiscreteSpace` functions).

src/Agents.jl
```diff
@@ -17,7 +17,8 @@ include("spaces/nothing.jl")
 include("spaces/graph.jl")
 include("spaces/grid.jl")
 include("spaces/discrete.jl")
-include("spaces/continuous_space.jl")
+# include("spaces/continuous_space.jl")
```
Datseris, line 20: Shouldn't we keep this running? How else will we benchmark this new space with the old? Please keep everything related to the old space. At the end of this PR you can do a massive rename CompartmentSpace -> ContinuousSpace and delete old files if need be, but for now you should set up a benchmark that creates the same model (e.g. SIR billiards) and benchmarks the same functions like move_agent, add_agent, nearby_agents, similarly with how I did it for the forest fire in my PR. Example: https://github.com/JuliaDynamics/Agents.jl/pull/311/commits/b5ea865ee5eea0f03ef9d1e2c3be29fd90e73bfd

src/spaces/compartment.jl
```diff
@@ -0,0 +1,188 @@
+export CompartmentSpace
+
+struct CompartmentSpace{D,F} <: AbstractSpace
+  grid::GridSpace
```
Datseris, line 4: GridSpace is type unstable because its type parameters are not stated explicitly as in the definition of the type in `grid.jl`.

src/Agents.jl
Libbum: I'm working on PkgBenchmark integration a little more at the moment. Will try to incorporate something regarding this.

Datseris:
Yes this is exactly what I had in mind. I went through the code for moving and seems exactly how it should be. Later tonight I will review the finding neighbors code. In the meantime what I'd like to see is something similar to what I did in my new GridSpace PR, where I kept the previous space as is and simply made a second version of forest fire and `@btime`d the basic functions.

Can you please do the same for the flocking model?

src/spaces/compartment.jl
```diff
@@ -0,0 +1,188 @@
...
+  update_vel!::F
+  dims::NTuple{D, Int}
+  periodic::Bool
+  D::Int
```
Datseris, line 8: D as a field is not needed anymore, and it is anyway part of the type parameter of the type so it is accessible without being an explicit field.

kavir1698:
> Can you please do the same for the flocking model?

Sure! I temporarily commented out the `ContinuousSpace` to not see the doc replacement warnings.

src/spaces/compartment.jl
```diff
@@ -0,0 +1,188 @@
+export CompartmentSpace
+
+struct CompartmentSpace{D,F} <: AbstractSpace
+  grid::GridSpace
+  update_vel!::F
+  dims::NTuple{D, Int}
+  periodic::Bool
```
Datseris, line 7: `periodic` becomes a type parameter, because it is the second type parameter of the field `grid`, since `GridSpace{D, periodic}`.

src/spaces/compartment.jl
```diff
@@ -0,0 +1,188 @@
...
+Propagate the agent forwards one step according to `vel` and the model's space, with `dt` as the time step. (`update_vel!` is not used)
+"""
+function move_agent!(agent::A, model::ABM{A,S,F,P}, vel::NTuple{D, X}, dt = 1.0) where {A <: AbstractAgent, S <: CompartmentSpace, F, P, D, X <: AbstractFloat}
+  pos = agent.pos .+ dt .* vel
+  if model.space.periodic
+    pos = mod.(pos, model.space.dims)
+  end
+  move_agent!(agent, pos, model)
+  return agent.pos
+end
```
Datseris, line 87: I would vote that we drop this method in 4.0, it seems obsolete to me. A user could always just do `newpos = agent.pos .+ newvel .* dt` and then `move_agent!(a, newpos, model)`.

Datseris: Periodicity then should be taken care of in the lowest-level implementation, which is `move_agent!(agent, pos, model)`.

src/spaces/compartment.jl
```diff
@@ -0,0 +1,188 @@
...
+
+Notice that if you want the agent to instantly move to a specified position, do
+`move_agent!(agent, pos, model)`.
+"""
+function move_agent!(agent::A, model::ABM{A, <: CompartmentSpace}, dt::Real = 1.0) where {A <: AbstractAgent}
+  model.space.update_vel!(agent, model)
+  pos = agent.pos .+ dt .* agent.vel
+  if model.space.periodic
+    pos = mod.(pos, model.space.dims)
+  end
```
Datseris, line 71: we should move this to `move_agent!(agent, pos, model)` to ensure that periodicity is taken care of at only one, central, place in the source code.

src/spaces/compartment.jl
```diff
@@ -0,0 +1,188 @@
...
+## Keywords
+Keyword arguments are space-specific.
+For `GraphSpace` the keyword `neighbor_type=:default` can be used to select differing
+neighbors depending on the underlying graph directionality type.
+- `:default` returns neighbors of a vertex. If graph is directed, this is equivalent
+  to `:out`. For undirected graphs, all options are equivalent to `:out`.
+- `:all` returns both `:in` and `:out` neighbors.
+- `:in` returns incoming vertex neighbors.
+- `:out` returns outgoing vertex neighbors.
+"""
```
Datseris, line 140: This documentation string should not be duplicated here, right?

Libbum:
As an aside, I've been looking at our implementation of Flocking and it's pretty outdated (and in some cases performance has certainly not been considered). I've changed the implementation of ForestFire and Schelling in #298 to reflect our new API as well as simplify the logic. Planning to do the same with Flocking once I've aligned the benchmarking implementations correctly with the other three tools. 

I guess it wont matter what you benchmark against for the space changes Ali, this is just to inform you that I expect we can eek out some more speed in other places.

src/spaces/compartment.jl
```diff
@@ -0,0 +1,188 @@
...
+    node_neighbors(position, model::ABM, r=1; kwargs...) → positions
+
+Return an iterator of all positions within "radius" `r` of the given `position`
+(which excludes given `position`).
+The `position` must match type with the spatial structure of the `model`).
+
+The value of `r` and possible keywords operate identically to [`space_neighbors`](@ref).
+"""
+node_neighbors(position, model, r=1; exact=false) = notimplemented(model)
```
Datseris, line 177: This shouldn't be here at all. Why duplicate it? the existing implementation in the API file already does this.

src/spaces/compartment.jl
kavir1698: No, I will fix the doc strings.

src/spaces/compartment.jl
Libbum: This was implemented for #248. I don't know if its still relevant (personally I never thought it was needed).

src/spaces/compartment.jl
```diff
@@ -0,0 +1,188 @@
...
+function positions(model::ABM{<:AbstractAgent,<:CompartmentSpace})
+  x = CartesianIndices(model.space.grid.s)
+  return (Tuple(y) for y in x)
+end
+
+function ids_in_position(pos::ValidPos, model::ABM{<:AbstractAgent,<:CompartmentSpace})
+  return model.space.grid.s[pos...]
+end
+
+cell_center(pos::Tuple) = getindex.(modf.(pos), 2) .+ 0.5
```
Datseris, line 116: I think this will lead to problems for continuous spaces that cover negative numbers. Depending on the application, it might be sensible for a user that their continuous space spans -10 to 10. We know how to handle these cases because of work we have done in DynamicalSystems.jl, and I will suggest a solution tonight.

kavir1698: Yes, I didn't consider negative spaces.

kavir1698: Previously, we only allowed positive spaces. I am not sure if negative spaces are needed.

src/spaces/compartment.jl
```diff
@@ -1,29 +1,25 @@
 export CompartmentSpace
 
-struct CompartmentSpace{D,F} <: AbstractSpace
-  grid::GridSpace
+struct CompartmentSpace{G,F,D} <: AbstractSpace
+  grid::G
```
Datseris, line 4: I can see from the definition of a GridSpace that https://github.com/JuliaDynamics/Agents.jl/blob/master/src/spaces/grid.jl#L19  `D, P` are the parameters of the grid space. **These** should be made type parameters of the CompartmentSpace, whose field should then type `::GridSpace{D, P}`. Also, this doesn't duplicate `D`.

Datseris:
Can you please use 4 spaces per level of indentation, consistent with the rest of the library? (or at least most of the rest, there are still some code remnants from 1.0)

Datseris:
@kavir1698 what do the benchmarks show?

kavir1698:
```
Times of COMPARTMENT space
Full model 500 steps
  122.369 ms (448919 allocations: 25.64 MiB)
Nearby Agents
  36.693 ns (0 allocations: 0 bytes)
nearby positions
  37.097 ns (0 allocations: 0 bytes)
Move agent
  53.455 ns (0 allocations: 0 bytes)
Add agent
  2.356 μs (41 allocations: 1.56 KiB)
```

src/spaces/compartment.jl
```diff
@@ -0,0 +1,143 @@
...
+    (which must match type with the spatial structure of the `model`).
+    """
+    function space_neighbors(pos, model::ABM{<:AbstractAgent,<:CompartmentSpace}, r=1; exact=false)
+        center = cell_center(pos)
+        focal_cell = ceil.(Int, center)
+        if exact
+            newr = ceil.(Int, r)
+            corner_of_largest_square_in_circle = floor.(Int, center .+ newr)
+            max_distance = corner_of_largest_square_in_circle[1] - focal_cell[1]
+            final_ids = Int[] # TODO make it an iterator
```
kavir1698, line 114: @Datseris Do you know how to make this an iterator?

Datseris:
```
Times of COMPARTMENT space
Full model 500 steps
  122.369 ms (448919 allocations: 25.64 MiB)
```
how does this compare to the existing version? (also, why is add_agent so slow...? should be same speed as the GridSpace)

Datseris:
Have you produced videos of flocking? is your implementation correct? does it look sufficiently like the previous one based on ContinuousSpace?

Libbum:
Values from my machine:
`benchmark/flocking.jl` on this branch
```
Times of COMPARTMENT space
Full model stepping
  103.174 ms (463631 allocations: 26.62 MiB)
Nearby Agents
  27.744 ns (0 allocations: 0 bytes)
nearby positions
  27.461 ns (0 allocations: 0 bytes)
Move agent
  47.440 ns (0 allocations: 0 bytes)
Add agent
  2.054 μs (41 allocations: 1.56 KiB)
```

Master branch, same benchmarks, but using `Models.flocking()` (ContinuousSpace)
```
Full model stepping
   590.375 ms (2850069 allocations: 145.73 MiB)
Nearby Agents
   2.098 μs (17 allocations: 1.11 KiB)
nearby positions <- Does not make sense in ContinuousSpace

Move agent
  3.025 μs (14 allocations: 944 bytes)
Add agent
  8.765 μs (90 allocations: 4.11 KiB)
```

src/spaces/compartment.jl
Libbum: I'm fairly certain you cannot append to an iterator, so making this one doesn't make sense with this current implementation.

src/spaces/compartment.jl
```diff
@@ -0,0 +1,143 @@
...
+    function move_agent!(agent::A, model::ABM{A, <: CompartmentSpace}, dt::Real = 1.0) where {A <: AbstractAgent}
+        model.space.update_vel!(agent, model)
+        pos = agent.pos .+ dt .* agent.vel
+        move_agent!(agent, pos, model)
+        return agent.pos
+    end
+    
+    #######################################################################################
+    # %% Neighbors and stuff
+    #######################################################################################
```
Datseris, line 73: Why is all this piece of code idented 4 spaces?!

Datseris (changes requested):
The PR misses a critical component: the grid spacing `x` which should be a parameter of CompartmentSpace and allows tuning the density of the covering by the underlying grid space.

Importantly, this doesn't have tests, so I'd like to see tests first before I review further.

kavir1698:
> 
> 
> The PR misses a critical component: the grid spacing `x` which should be a parameter of CompartmentSpace and allows tuning the density of the covering by the underlying grid space.
> 
> Importantly, this doesn't have tests, so I'd like to see tests first before I review further.

Spacing is given as a parameter to build the space. I don't see why it would be needed anywhere else because we don't mutate the grid or recreate it after the model is built.

kavir1698:
flock

Datseris:
> Spacing is given as a parameter to build the space. I don't see why it would be needed anywhere else because we don't mutate the grid or recreate it after the model is built.

Ah, you are correct, I've missed that. (was searching `x` in atom)

src/spaces/compartment.jl
```diff
@@ -18,7 +18,10 @@ function random_position(model::ABM{A, <:CompartmentSpace{D}}) where {A,D}
     pos = Tuple(rand(D) .* model.space.dims)
 end
 
-pos2cell(pos::Tuple) = ceil.(Int, pos)
+function pos2cell(pos::Tuple) 
+    any(x-> x <= 0, pos) && throw("Position should be positive")
```
Datseris, line 22: This is too costly to be done every single in every call to `nearby_...`, I would suggest to instead state that positions can only be positive in the space documentation string.

src/spaces/compartment.jl
```diff
@@ -0,0 +1,379 @@
...
+            if !any(x-> x> max_distance, abs.(cell .- focal_cell)) && max_distance > 1 # certain cell
+                ids = ids_in_position(cell, model)
+                final_ids = vcat(final_ids, ids)
+            else # uncertain cell
+                ids = copy(ids_in_position(cell, model))
+                filter!(i -> sqrt(sum(abs2.(pos .- model[i].pos))) ≤ r, ids)
+                final_ids = vcat(final_ids, ids)
+            end
+        end
+        return final_ids
```
Datseris, line 157: This part of the code has bad performance. Instead of continuously re-creating `final_ids` by vcat-ing it with itself and something else we should follow the [Julia performance tips](https://docs.julialang.org/en/v1/manual/performance-tips/) and reduce allocations. 

The `ids` you find in this loop should be added to `final_ids` using `append!` .

Datseris:
The code seems reasonable! Here is what I'd like to see:

1. Passing tests.
2. Plots (because otherwise it is very hard to check whether the code does exactly what it should do): Initialize a compartment space with let's say 10x10 grid, and add agents at random positions. Then, choose some agent randomly, and some r. Plot the agent in red color and a circle at its position. Plot all its neigbhors with blue color. Plot all remaining agents of the model with green color or whatever. Enable "grid" in the plot, so that the compartments are shown. Then we can check whether the "exact" version is correct. (the inexact version is definitely correct).

kavir1698:
Tests pass. On travis everything is failing, even for other spaces.

Libbum:
Tests are failing because there are a number of functions that are defined in continuous space and subsequently overwritten in compartment space.

kavir1698:
I will comment out continuous space then.

kavir1698:
I have created a file to visually inspect the `nearby_ids` file. Looks fine to me. You can check yourself.

kavir1698:
BTW, you would have to add `CompartmentSpace` to allowed spaces in AgentsPlots (`real.jl` line 20) to see the plot.

Datseris:
why don't you paste some pictures here...?

Datseris:
Can you report the final benchmarks please?

Datseris:
I've found a case (after several random plottings) that had a mistake:

image

yellow agent in the top right of the circle is not included

kavir1698:
It's the `grid_space_neighborhood` that doesn't include some cells. In the figure below, blue points outside of the circle are detected when `exact=false`. Still some diagonal cells are not included.

tmp

Libbum:
That graph is quite confusing. Is the circle identifying `r=2`? Looks closer to 1.5. 

I haven't been keeping track of development here too closely, but there seems to be some bugs to iron out still. I may be able to makes some checks & suggestions today, time permitting.

kavir1698:
> That graph is quite confusing. Is the circle identifying `r=2`? Looks closer to 1.5.
> 
> I haven't been keeping track of development here too closely, but there seems to be some bugs to iron out still. I may be able to makes some checks & suggestions today, time permitting.

The circle is "r=1.4" and its ceiling is 2

Datseris:
I think you have to make further adjustments to the algorithm I've proposed, since you represent box coordinates as the ceiling of the position, which I have not taken into account. Now you have the plotting code to check, so once you have a correct version I'll review again.

kavir1698:
This all is because I expected the `grid_space_neighborhood` to return all cells that are the even partially covered in a circle from `pos`. But it only returns cells that are fully within the circle. See the figure below:

```julia
points = collect(Agents.grid_space_neighborhood(CartesianIndex((9, 14)), model, 2))
scatter(points,legend=false)
```

tmp

Cells (8,16), (11,3), (11,16) are partially within the circle of radius 2 from (9, 14), but are not picked here.

Datseris:
In the grid space the word "covering" and "partially covered" does not have meaning, because the space is discrete. Each "cell" has zero spatial extent. In the documentation string of GridSpace, the algorithm is clear: it picks cells whose euclidean distance of Cartesian indices is less than or equal to the given value.

In the continuous space where covering makes sense, the search radius has to be increased so that all cells that would be "partially covered" are included in the search. You'd have to think about how this interplays with the way you defined the center of a cell, where you use `ceil` and thus the "coordinates" of each cell are in its top right corner instead of its center (which is what I based my description in #313).

kavir1698:
Right. I think the solution is to find `cityblock` neighbors, same as we did with `ContinuousSpace`.

Datseris:
This is _a_ solution, not _the_ solution. It also is a solution of low performance in the case of using Euclidean norm. I've already provided info on what should be done in #313 , why don't you want to do what was said there? You need to find a radius that includes all cells that can possibly be reached by the current cell and given radius, and you need to find another radius that it is absolutely guaranteed that every cell that will be included in this radius is guaranteed to contain true (exact) neighbors. I've colored these cases with green and purple color. 

I don't see why this would be hard to do, but if I am missing somethin please explain it to me.

Libbum:
I have a second question concerning this implementation that might need some consideration:
How do users know what size `spacing` is appropriate? Once things have settled down here I think we need to do some benchmarking to ascertain some form of gridsize/agent scaling and provide a helper function that assigns a value of `spacing` automatically, then provide the *option* of overriding that estimate.

kavir1698:
> 
> 
> This is _a_ solution, not _the_ solution. It also is a solution of low performance in the case of using Euclidean norm. I've already provided info on what should be done in #313 , why don't you want to do what was said there? You need to find a radius that includes all cells that can possibly be reached by the current cell and given radius, and you need to find another radius that it is absolutely guaranteed that every cell that will be included in this radius is guaranteed to contain true (exact) neighbors. I've colored these cases with green and purple color.
> 
> I don't see why this would be hard to do, but if I am missing somethin please explain it to me.

Because that solution is less performant. The smallest square encompassing a circle will include fewer cells than a circle with a larger radius that will fully cover all the cells in the smaller circle.

kavir1698:
> 
> 
> I have a second question concerning this implementation that might need some consideration:
> How do users know what size `spacing` is appropriate? Once things have settled down here I think we need to do some benchmarking to ascertain some form of gridsize/agent scaling and provide a helper function that assigns a value of `spacing` automatically, then provide the _option_ of overriding that estimate.

Yes, I think that will be a function of agent density on space.

Libbum:
The tests here are failing locally for me. I'll try to fix them.

src/spaces/compartment.jl
```diff
@@ -144,7 +146,7 @@ function nearby_ids(pos::ValidPos, model::ABM{<:AbstractAgent,<:CompartmentSpace
             max_distance = r
         end
         final_ids = Int[]
-        allcells = grid_space_neighborhood(CartesianIndex(focal_cell), model, newr)
+        allcells = grid_space_neighborhood(CartesianIndex(focal_cell), model, newr+1)
```
Datseris, line 149: Notice that the extra radius is not 1 but `sqrt(D)` with `D` the dimensionality of the space

Libbum:
Just did a quick benchmark of the current setup here using the flocking example, but with the benchmarking parameters from #298.

Was my rule of thumb understanding correct, when we should look to put 1 agent per cell? Here we have a 100*100 grid with 300 agents. `spacing = 30` would make sense then yeah? Turns out that takes 5 seconds on this run.

```julia
for sp in 0.08:0.04:1
    a = @benchmark step!(model, agent_step!, model_step!, 100) setup = (
        (model, agent_step!, model_step!) = flocking(
            n_birds = 300,
            separation = 1,
            cohere_factor = 0.03,
            separate_factor = 0.015,
            match_factor = 0.05,
            dims = (100, 100),
            spacing = $sp,
        )
    )
    push!(times, minimum(a.times) * 1e-6)
end
```
(and an extra one at `sp=2` for good measure).

Screenshot_20200929_230155

Best we can get is around `spacing = 0.36`, which is about 74ms. Compared to 343.782 ms using `ContinuousSpace`, that's a great improvement! However I think we'll need to do some more investigation into grid size optimisation.

Datseris:
@kavir1698 I am struggling to understand something that should in principle be quite simple. How can the user increase the density of the compartments without `spacing` being a field of the struct...?

You said that it doesn't have to be a fiend because in this definition
```julia
function CompartmentSpace(d::NTuple{D,Real}, spacing;
    update_vel! = defvel2, periodic = true) where {D}
    s = GridSpace(floor.(Int, d ./ spacing), periodic=periodic, metric=:euclidean)
    return CompartmentSpace(s, update_vel!, size(s))
end
```
the spacing is "taken into account". Yet, for some reason, in the `pos2cell` function you do:
```julia
pos2cell(pos::Tuple) = ceil.(Int, pos)
cell_center(pos::Tuple) = getindex.(modf.(pos), 2) .+ 0.5
```
First of all, oddly enough, the function `pos2cell` is not even used in `cell_center`. The more weird part is in `nearby_ids` where I see:
```julia
function nearby_ids(pos::ValidPos, model::ABM{<:AbstractAgent,<:CompartmentSpace{D}}, r=1; exact=false) where {D}
    center = cell_center(pos)
```

so the actual real position of the agent, which is in continuous space, is used for the "cell_center". I wonder now, the following scenario: a user having a real space that has extend (1.0, 1.0). The user then makes two spaces, one with "spacing" 0.1, and one with spacing 0.01.

These two spaces have more grid points in their GridSpace. Yet, both of them do exactly the same operations to find the "current cell" or "nearest cell", specifically the `ceil` function. This is just so weird, how can it possibly be? _Even more weird_ is the fact that in both cases the `ceil` function will return exactly `(1.0, 1.0)` no matter where the agent is located, as this is the only integer allowed by the space's extend. 

Notice, I am asking because I already have done this thing years ago in ChaosTools.jl, to calculate entropies. There, I must do `loor.(Int, (point/ε))`, i.e. I must divide the position in real space by the spacing `ε`. So how can it be that you don't do this here?

Datseris:
@Libbum 

> Was my rule of thumb understanding correct, when we should look to put 1 agent per cell? 

yes, and I see from the graph this is indeed quite accurate and something easy to calculate from the user's perspective. The numbers you get should in principle also fluctuate because of the random distribution of agents.

Libbum:
If I have a `CompartmentSpace((10, 10), 0.02)`, this means that my space should extend from 0&ndash;10 in `x` and 0&ndash;10 in `y` correct? How therefore, can I get 

```julia
julia> model[20].pos
(165.21981672618037, 443.3392364355354)
```

kavir1698:
Here, space dimensions are not absolute, but relative to the spacing. A space in (1, 2) with spacing=0.1 is equal to (10, 20) with spacing =1.
This saves us an expensive division every time we need the cell center or cell number.

Libbum:
I don't think a user will appreciate having to do that calculation every time they want to get the position of an agent though..

kavir1698:
> 
> 
> I don't think a user will appreciate having to do that calculation every time they want to get the position of an agent though..

When a space is created, it tells the actual space dimensions, in the previous example (10, 20). The user doesn't need to do any calculations.

Libbum:
That makes little sense. Why ask them to provide an extent of their space, a spacing of their space, then return some completely different value of space for them to use? Why not obfuscate this from the user entirely? 

At the moment, I thought we were suggesting to the user "Choose a spacing that provides (on average) one agent per cell". If a user wants a space of `dims=(10, 10)` and they have `n_agents=800`, then they calculate `spacing = sqrt(prod(dims))/n_agents ≈
0.2`.

Instead, you're suggesting "Space extents should provide approximately one agent per cell". Which means you may as well always state `dims=(500,500), spacing = 1`, But then the user would have to convert back to their wanted extent of `(10,10)`...

Datseris:
@kavir1698 you have to rewrite the source code so that `spacing` (or `x`) is a parameter of the space and, as I originally instructed in #313 you do the operation `ceil(pos ./ space.x)` instead. Please stick to the original declaration in #313, we are losing so much time here constantly discussing things that I've thought through several times, with pen and paper, like e.g. the dimensional scaling or which cells should be included in the search.

kavir1698:
> 
> 
> That makes little sense. Why ask them to provide an extent of their space, a spacing of their space, then return some completely different value of space for them to use? Why not obfuscate this from the user entirely?
> 
> At the moment, I thought we were suggesting to the user "Choose a spacing that provides (on average) one agent per cell". If a user wants a space of `dims=(10, 10)` and they have `n_agents=800`, then they calculate `spacing = sqrt(prod(dims))/n_agents ≈ 0.2`.
> 
> Instead, you're suggesting "Space extents should provide approximately one agent per cell". Which means you may as well always state `dims=(500,500), spacing = 1`, But then the user would have to convert back to their wanted extent of `(10,10)`...

Your confusion is perhaps because of your different expectation given the previous implementation of continuous space. Previously, we asked the user to provide a space _extend_, but here, we ask the user to provide the _relative dimensions_ of the space. What matters in the end is the ratio of different dimensions to one another, not their actual range. As I said, this implementation is faster. But I only have one vote. The change is not so much work.

Datseris:
> What matters in the end is the ratio of different dimensions to one another, not their actual range.

No, I don't agree with this, not conceptually at least. What you say is true for the numeric implementation, but not for usage. Continuous space is the only _truly_ real space we have. It is the only space where the absolute location of agents matters, because this location is the only one that can be mapped in a true, existing in reality, space. That's why if you have a town square that is 20 by 10 meters (in _reality_), you want your real space to be 10 by 20 meters. It just makes sense.

Besides, you did not say anywhere in the documentation string that "agent positions will be considered in a modified space that has values `dimensions ./ spacing`". Besides besides, what happens with velocity in your space? This would also have to be adjusted from the physically relevant units of the user to modified ones. You also don't mention this in the documentation string.

Libbum:
Yep, that's the most important argument. I see *why* you'd like to solve this in the current way to make it the *most* performant, but we need to also consider usability. The cognitive overhead required by a user here is simply too much, it makes things too awkward if you want to operate in real units.

In this case we have to sacrifice some performance to keep our usability simple.

kavir1698:
> 
> 
> Yep, that's the most important argument. I see _why_ you'd like to solve this in the current way to make it the _most_ performant, but we need to also consider usability. The cognitive overhead required by a user here is simply too much, it makes things too awkward if you want to operate in real units.
> 
> In this case we have to sacrifice some performance to keep our usability simple.

Sure, a change is coming!

Libbum:
That looks much better. 

I'll take a look at `nearby_ids(;exact=true)`, see if I can't get this working a little better:
```julia
julia> @benchmark nearby_ids($aa, $model, $r; exact = false)
BenchmarkTools.Trial: 
  memory estimate:  8.41 KiB
  allocs estimate:  10
  --------------
  minimum time:     2.118 ms (0.00% GC)
  median time:      2.207 ms (0.00% GC)
  mean time:        2.217 ms (0.00% GC)
  maximum time:     3.233 ms (0.00% GC)
  --------------
  samples:          2255
  evals/sample:     1

julia> @benchmark nearby_ids($aa, $model, $r; exact = true)
BenchmarkTools.Trial: 
  memory estimate:  4.68 MiB
  allocs estimate:  3921
  --------------
  minimum time:     30.644 ms (0.00% GC)
  median time:      31.095 ms (0.00% GC)
  mean time:        31.541 ms (0.53% GC)
  maximum time:     35.784 ms (5.07% GC)
  --------------
  samples:          159
  evals/sample:     1
```

Something is off with the flocking benchmark now though. I get:
```julia
julia> a = @benchmark step!(model, agent_step!, model_step!, 100) setup = (
               (model, agent_step!, model_step!) = flocking(
                   n_birds = 300,
                   separation = 1,
                   cohere_factor = 0.03,
                   separate_factor = 0.015,
                   match_factor = 0.05,
                   dims = (100, 100),
                   spacing = 0.36,
               )
           )
ERROR: BoundsError: attempt to access 277×277 Array{Array{Int64,1},2} at index [278, 145]
```

Libbum:
Changes in f99d087 give us
```julia
julia> @benchmark nearby_ids($aa, $model, $r; exact = true)
BenchmarkTools.Trial: 
  memory estimate:  4.38 MiB
  allocs estimate:  26
  --------------
  minimum time:     26.156 ms (0.00% GC)
  median time:      26.943 ms (0.00% GC)
  mean time:        27.228 ms (0.48% GC)
  maximum time:     30.722 ms (4.80% GC)
  --------------
  samples:          184
  evals/sample:     1
```

kavir1698:
> Something is off with the flocking benchmark now though. I get:

Yes, I noticed that. Every few movements, an agent is moved out of space. I am still checking why this happens.

Libbum:
If the space is of size 277x277, (with `spacing = 0.36`, so the 'real' dims are `(100, 100)`), then `pos2cell` is incorrect for the far boundary.

```julia
julia> ceil.(Int, (0.01, 99.96399599238744) ./ 0.36)
(1, 278)
```

Libbum:
Changing the above fixes the issue with adding agents in the flocking benchmark, but there's something going on with moving agents still to do with boundary conditions. 
I added a `println(a.pos)` on line 57 in `add_agent_to_space!`. 
```julia
(69.77015879482535, 100.69579716118025)
ERROR: BoundsError: attempt to access 278×278 Array{Array{Int64,1},2} at index [194, 280]
Stacktrace:
 [1] getindex at ./array.jl:810 [inlined]
 [2] add_agent_to_space! at /mnt/turtle/repos/Agents.jl/src/spaces/compartment.jl:58 [inlined]
 [3] move_agent!(::Bird, ::Tuple{Float64,Float64}, ::AgentBasedModel{Bird,CompartmentSpace{2,true,typeof(Agents.defvel2)},typeof(random_activation),Nothing}) at /mnt/turtle/repos/Agents.jl/src/spaces/compartment.jl:76
 [4] move_agent! at /mnt/turtle/repos/Agents.jl/src/spaces/compartment.jl:95 [inlined]
 [5] flocking_agent_step!(::Bird, ::AgentBasedModel{Bird,CompartmentSpace{2,true,typeof(Agents.defvel2)},typeof(random_activation),Nothing}) at ./REPL[7]:12
```

So that's come through `move_agent!` but the `if periodic` check did not fix this position.

kavir1698:
Thanks for narrowing down, Tim. I'll have to check it tomorrow, I'm too busy today.

Datseris:
> So why are we using ceil here, but floor in the GridSpace constructor (line 44)? That should be a ceil too right?

Yes. From what you posted the problem is that you're making the grid space one shell smaller than what it should be.

Libbum:
These changes, as Ali feared, have increased our timing by almost a factor of 2. The optimal grid spacing has also changed as a result.

Screenshot_20201001_111400

Datseris:
That comparison doesn't help much, because the previous version can't be used in practice. How does it compare with the old `ContinuousSpace` is what we care more about.

Libbum:
~121 ms here, 343ms there.

Datseris:
Can you please report timings for the unit benchmarks: time for adding, moving, removing and finding neighbors.

Libbum:
Yep, I'm working on that now.

Libbum:
It's not currently obvious to me what the ideal spacing choice is at the moment though..

kavir1698:
It is worth comparing the timing for larger models too, like with thousands of agents instead of 100.

Libbum:
```julia
cmodel = ABM(ContinuousAgent, CompartmentSpace((10.0, 10.0, 10.0), 1))
add_agent!((0, 0, 0), cmodel, (0.8, 0.7, 1.3), 6.5, false)
```
currently fails

Datseris:
i think you have to revert all `ceil`s to `floor`s. In chaostools.jl, where I do exactly the same thing, I use `floor` instead of `ceil` and everything works fine.

Libbum:
That's not a universal solution at least. Position `(0.01, 0.01)` cannot lie in cell `(0, 0)` (from `pos2cell((0.01, 0.01), model)`). But I'll hunt down a solution...

Libbum:
Well, a `pos=(0.0, 0.0)` is a true edge case regardless of the use of `ceil` or `floor` here. It's going to be crazy if we have to check that position every single time we need a cell returned. @Datseris, was there a solution for this in chaostools?

Datseris:
`floor(...).+1`, yes of course, sorry I wasn't explicit about this part.

Libbum:
Choosing a `spacing` is really brittle, or at least depends on differing things.

```julia
function iterate_over_neighbors(a, model, r)
    s = 0
    for x in nearby_ids(a, model, r; exact = true)
        s += x
    end
    return s
end
 
times = Float64[]
spaces = 0.5:0.5:6.0
for sp in spaces 
    cmodel = ABM(ContinuousAgent, CompartmentSpace((10.0, 10.0, 10.0), sp))
    for x in range(0, stop = 10.0, length = 7)
        for y in range(0, stop = 10.0, length = 7)
            for z in range(0, stop = 10.0, length = 7)
                add_agent!((x, y, z), cmodel, (0.8, 0.7, 1.3), 6.5, false)
            end
        end
    end
    aa = cmodel[139]
   
    a = @benchmark iterate_over_neighbors($aa, $cmodel, 5) setup =
        (nearby_ids($aa, $cmodel, 5))
    push!(times, minimum(a.times) * 1e-6)
end
```

The test is a 3D, 10x10x10 space with 343 (7x7x7) agents, using `exact=true`.

Screenshot_20201001_151438

Here we see increased spacing is performing better, up to the chosen search radius (nothing to do with agent count or grid size?).
Changing the neighbor search radius to `1`, we can see that scaling holds:

Screenshot_20201001_152349

Now using `exact = false`, different scaling occurs:
Screenshot_20201001_152853
Blue is for search radius 1 (best ~1.5), and red for search radius 5 (best ~2.0, or perhaps 6.0??). 

Any suggestions on what to check for / a better way to test this? I'm pretty stuck here&mdash;choosing a sane starting point is vital for the benchmarks.

Libbum:
`move_agent!(agent, model, (1.2, 0.0, 0.7))` (moving an agent with a given velocity) is either not implemented or not working

Datseris:
pos is 2nd argument, not 3rd (at least in the API I declared)

Datseris:
Are the number in the plots averaged over different initial agent distributions?

Regardless, I don't think we should spend more time on finding a good `spacing`, we can just say in the docstring something like "performance depends on `spacing` and typically there is an optimal one, benchmark and find out".

> choosing a sane starting point is vital for the benchmarks.

That is true but we also want the benchmarks to be fair for the other softwares as well, and optimizing every possible parameter might not be fair. From my eyes, we should just pick the same spacing for our's and MASON's model and report the numbers. If we want to be even fairer, we can average over let's say 5 values of spacing for both software.

> Choosing a spacing is really brittle, or at least depends on differing things.

Yeap, I expected that. If I misled you to believe that choosing the optimal `spacing` would be deterministic, I apologize. I think that a good value (rule of thumb) would be 1 agent per cell on average.

Notice something important: I haven't seen any actual plots, those ones with the agents colored, which I initially pasted, that actually show that the current implementation is correct. And I think this should be checked first before anything else. I haven't checked the source.

Libbum:
> pos is 2nd argument, not 3rd (at least in the API I declared)

This is for a velocity move though, and is working for `ContinuousSpace`.

Libbum:
> we should just pick the same spacing for our's and MASON's model and report the numbers

They use `neighborhood / 1.5`, which is 3.33 and essentially our optimum for that setup. 

> I think that a good value (rule of thumb) would be 1 agent per cell on average.

That really doesn't reconcile with any testing I'm doing at the moment though. For the flocking benchmark setup for example, that's `spacing = (100*100) / 300 = 33.33`, which is **0.18x** against a spacing of 3.33.

Anyhow. I'll post up some neighbor selection graphs soon&mdash;I *think* they're fine, but will double check.

Libbum:
exact=false
Screenshot_20201002_101353
exact=true
Screenshot_20201002_101441
exact=false across PBCs
Screenshot_20201002_101539
exact=true across PBCs
Screenshot_20201002_101648

Libbum:
I think I'm miss-representing the boundary values in these plots. See how there's an extra row of cells above x=10 that are coloured?

Will fix that to make sure these are fine, but excluding that, things look good now.

Libbum:
Smaller spacing makes it more obvious it's working as expected when there's no wrapping:
Screenshot_20201002_112507
Screenshot_20201002_112601

But I think we're missing data when a wrap needs to happen.
Screenshot_20201002_112741
Data in (6,0) should be collected here yeah?

Datseris:
yeap, they should. Is the "periodic" part propagated properly to the grid space...? Could there be something wrong with the grid space? ( I highly doubt that, but I'm not always right, like with the optimal spacing)

Datseris:
have you uploaded changes in the plotting code? I'll have a look over the weekend.

Datseris:
I also have an idea for optimizing this further so that it doesn't select cells so far away by using the cell central midpoint to correspond to the underlying grid, and also using `sqrt(D)/2` instead of `sqrt(D)` for the extra dimension.

Libbum:
I'm working more on it at the moment. There's definitely a bug in the PBC wrapping. Will upload something better soon.

Libbum:
A little further digging tells me it actually may be in the gridspace:

Screenshot_20201002_134536

I should be able to see three cells (`x=[8,9,10]`) from the wrapped mirror, not just two. 
It'll just be an indexing issue somewhere, since all of the values are 1 indexed.

```julia
julia> pos
(0.9265913421511796, 7.4851981784038095)

julia> s.extent
(10.0, 10.0)

julia> s.spacing
1.0

julia> s.dims
(11, 11)

julia> collect(Agents.grid_space_neighborhood(CartesianIndex(Agents.pos2cell(pos, model)), model, 3))
29-element Array{Tuple{Int64,Int64},1}:
 (1, 5)
 (10, 6)
 (11, 6)
 (1, 6)
 (2, 6)
 (3, 6)
 (10, 7)
 (11, 7)
 (1, 7)
 (2, 7)
 (3, 7)
 (9, 8)
 (10, 8)
 (11, 8)
 (1, 8)
 (2, 8)
 (3, 8)
 (4, 8)
 (10, 9)
 (11, 9)
 (1, 9)
 (2, 9)
 (3, 9)
 (10, 10)
 (11, 10)
 (1, 10)
 (2, 10)
 (3, 10)
 (1, 11)
```

Libbum:
Alright. I got it. There's a few nuances that need to be taken care of. `floor.(x) + 1` for positions, but not for generating the actual grid space (it just wants `floor`). On top of that, there were no periodic distance checks (in fact, I don't think we satisfied this condition at all in `ContinuousSpace` either), so that's taken into account now when we use the `exact` method.

exact = false
Screenshot_20201002_152336

exact = true
Screenshot_20201002_152302

exact = false (PBC)
Screenshot_20201002_173023


exact = true (PBC)
Screenshot_20201002_172932

Datseris:
Great. I am now convinced that we can in fact reduce the search radius in grid space, and will show how over the weekend. In the meantime, what is the current benchmarks? How do we compare with (a) previous ContinuousSpace and (b) MASON?

Libbum:
Working on them soon. Just fixed the tests.

Can you comment on https://github.com/JuliaDynamics/Agents.jl/pull/322#issuecomment-702574611?

Datseris:
You mean this: 

> That really doesn't reconcile with any testing I'm doing at the moment though. For the flocking benchmark setup for example, that's spacing = (100*100) / 300 = 33.33, which is 0.18x against a spacing of 3.33. 

? 

Well if you have the numbers then I was just wrong. Numbers don't lie...

Libbum:
Yeah. 

OK, no worries. I guess what I'm referring to is that it will be quite important to give users a good rule-of-thumb in the documentation for this. 1 agent per cell is not a good one. I'm hoping we can suggest something else.

Datseris:
Does the performance of `exact=true` and `exact=false` have the _same_ scaling w.r.t. to `spacing`? I would guess they have different results. I don't know how far we can go with giving users a good rough estimate. Well, at least I failed. Perhaps we should motivate them to benchmark and find it themselves by doing this process in one of our continuous space examples. We then don't have to spend to much time on this.

Libbum:
I'll have to try those checks again to see. The previous fixes will alter the outcome I think.

Just found another issue whilst running the benchmarks, so will need to rectify that before some results.

Libbum:
Need some input on this.
First, we need to make the value of `extent` to be the end of the range, but not included in the range e.g. x = [0, 10), y = [0, 10) for `CompartmentSpace((10, 10), 1.5)`. That's fine, I've just change the docs to reflect this.

The second issue though has to do with the user's choice of spacing. As it is, we would cut up `CompartmentSpace((10, 10), 1.0)` in to 10 equidistant portions. In the case of `CompartmentSpace((10, 10), 1.5)`, the grid space ends up being 6x6, which means it only spans x = y = [0, 9), not the entire extent.

Should we confine users to only allow `spacing` to be divisible by all dimensions? Or is there a decent way we can handle this case?

Datseris:
> Should we confine users to only allow spacing to be divisible by all dimensions?

Yes, let's not complicate things.

Libbum:
Benchmarks for Flocking with parameters from #298 
This branch with `spacing = vision / 1.5 = 3.3333` (to compare with MASON): 110.222 ms (1)
With `ContinuousSpace`: 337.725 ms (3.1x)
MASON: 206.914 ms (1.9x)
These are results from an un-optimised version of flocking, so the comparison to MASON is not so relevant currently.

---

This branch:
```julia
Nearby Agents - exact = false
@btime nearby_ids($a, $model, 1);
  201.971 ns (4 allocations: 272 bytes)

@btime iterate_over_neighbors($a, $model, 2.0)
  5.023 μs (174 allocations: 5.77 KiB)

Nearby Agents - exact = true
@btime nearby_ids($a, $model, 1, exact=true);
  2.086 μs (18 allocations: 2.26 KiB)

@btime iterate_over_neighbors($a, $model, 2.0)
  15.119 μs (408 allocations: 43.71 KiB)

nearby positions
@btime nearby_positions($a.pos, $model);
  33.072 ns (0 allocations: 0 bytes)

Move agent
@btime move_agent!($a, $model);
  64.794 ns (0 allocations: 0 bytes)

Add agent
@btime add_agent!($model, (-0.9670145669689216, 0.2547210773962555), 1.0, 0.25, 4.0, 0.25, 0.01, 5.0)
  1.957 μs (41 allocations: 1.69 KiB)
```
With `ContinuousSpace`:
```julia
Nearby Agents - compare exact
@btime nearby_ids($a, $model, 1); #Note this call IGNORES periodic neighbors, so it's a little faster than it should be
  4.798 μs (32 allocations: 1.59 KiB)

@btime iterate_over_neighbors($a, $model, 2.0)
  23.941 ms (149118 allocations: 4.79 MiB)

Move agent
@btime move_agent!($a, $model);
  3.063 μs (14 allocations: 944 bytes)

Add agent
@btime add_agent!($model, (-0.9670145669689216, 0.2547210773962555), 1.0, 0.25, 4.0, 0.25, 0.01, 5.0)
  8.990 μs (90 allocations: 4.11 KiB)
```

Libbum:
8f15292 improves `add_agent!` to
```julia
@btime add_agent!($model, (-0.9670145669689216, 0.2547210773962555), 1.0, 0.25, 4.0, 0.25, 0.01, 5.0)
  142.753 ns (1 allocation: 96 bytes)
```

Datseris:
Okay so the performance over the previous continuous space is super-higher, when I start working on this over the weekend I will rename this to `ContinuousSpace` and throw away the old one.

Libbum:
I took a shot at getting some more speed out of the seldom used functions like `genocide!`, but I don't think it was worth the add:

For 1 million agents. Current implementation: 0 allocations, 958.854 μs
```julia
function genocide!(model::ABM{A,<:CompartmentSpace}) where {A}
    empty!(model.agents)                                
    for p in model.space.grid.s
       empty!(p)
    end                                                                              
    model.maxid[] = 0
end
```
This one: 0 allocations, 777.907 μs (1.2x)

> I will rename this to ContinuousSpace and throw away the old one

Great. I'll be busy for most of tomorrow, but can touch up the benchmarks & make a new flocking implementation when you're done with that.

Datseris:
Nah, I wouldn't complicate more functions that are used so rarely and also for such a small gain. The smaller the source code the better!

Datseris:
im working on this now, don't do any commits until tomorrow

Datseris:
Alright, using `grid_r = (r+δ)/model.space.spacing` in the inexact version decreases the search radious drastically, and for small `spacing` this becomes almost indentical with the true euclidean radious.

image


there is a small amount of cells that are in the exact diagonal that are missed for large spacings, but I think we can safely not care, given that by definition this version is inexact.

Datseris (approved):

Datseris:
Okay, I'm done here:

image

We require passing tests, I'll leave this to @kavir1698 .

Libbum:
I'll check over the changes tomorrow and see if I have anything to add.
Have we done some sanity checks in 3D as well?

Datseris:
I haven't no.

Libbum:
I think we should make the distance calculation an exported function and generalise it over grid and continuous spaces.
Something like this?:
```julia
function euclidean_distance(p1, p2, model::ABM{A,<:Union{GraphSpace{D,true}, ContinuousSpace{D, true}}}) where {A, D}    
    sqrt(sum(abs2.(p1 .- p2)))    
end    
function euclidean_distance(p1, p2, model::ABM{A,<:Union{GraphSpace{D,false}, ContinuousSpace{D, false}}}) where {A, D}    
    total = 0    
    for (i, (a, b)) in enumerate(zip(p1, p2))                                                
        delta = abs(b - a)                                                                   
        if delta > model.space.dims[i] - delta                                               
            delta = model.space.dims[i] - delta                                              
        end                                                                                  
        total += delta^2                                                                     
    end                                                                                      
    sqrt(total)                                                                              
end   
```

Reason being that we use a simple `distance` function or `LinearAlgebra.norm` all over the place in examples, but quite a few of them are periodic in nature&mdash;meaning that in those cases we're not actually satisfying boundary conditions.

Datseris:
Ah good that you mentioned this I wanted to say taht Distances.jl implements a different formula for periodic euclidean called `PeriodicEuclidean` and goes like `sqrt(sum(min(mod(abs(x - y), p), p - mod(abs(x - y), p)).^2))` for periodicity under `p`. Perhaps this version is faster than the larger one you've written above? In general `if` statements are much slower that standard numeric operations like multiplication

Libbum:
Yeah, we can certainly look for something better. I just wrote a patch essentially. 

Still need to update examples for this. Models are mostly done but we are missing `update_space!` for growing_bacteria.jl

src/spaces/continuous.jl
```diff
@@ -0,0 +1,442 @@
...
+    return a
+end
+
+function move_agent!(
+        a::A,
+        pos::Tuple,
+        model::ABM{A,<:ContinuousSpace{D,periodic}},
+    ) where {A<:AbstractAgent,D,periodic}
+    remove_agent_from_space!(a, model)
+    if periodic
```
Libbum, line 90: Should we dispatch explicit values of periodic here rather than using the `if`?

Datseris: No, it doesn't make a difference. The `if` gets pruned in compilation.

Libbum:
How should we best handle this scenario:

Bacterial_growth uses a continuous space (12,12), with no periodic boundaries. When the cells divide and grow, at some point they wish to move away from the centre and ultimately *try* to exist outside of the space: (13, 4) for example. 

I can modify the code to respect the boundary (that makes sense for bacteria in a petri dish), but for the API, should there be an inbounds check or a warning or something else to alert the user that this position cannot be used? At the moment we just get:
```julia
ERROR: BoundsError: attempt to access 12×12 Array{Array{Int64,1},2} at index [13, 4]
Stacktrace:
 [1] getindex at ./array.jl:810 [inlined]
 [2] add_agent_to_space! at /mnt/turtle/repos/Agents.jl/src/spaces/continuous.jl:70 [inlined]
 [3] add_agent_pos!(::SimpleCell, ::AgentBasedModel{SimpleCell,ContinuousSpace{2,true,Float64,typeof(Agents.defvel)},typeof(fastest),Dict{Symbol,Float64}}) at /mnt/turtle/repos/Agents.jl/src/core/space_interaction_API.jl:180
```

examples/schoolyard.jl
```diff
@@ -113,9 +113,7 @@ function agent_step!(student, model)
     end
 
     ## Add all forces together to assign the students next position
-    student.pos = student.pos .+ noise .+ teacher .+ network_force
-    ## Update the students position in space
-    update_space!(model, student)
+    move_agent!(student, student.pos .+ noise .+ teacher .+ network_force, model)
```
Datseris, line 116:
```suggestion
    newpos = student.pos .+ noise .+ teacher .+ network_force
    move_agent!(student, newpos, model)
```

src/models/growing_bacteria.jl
```diff
@@ -77,12 +77,11 @@ end
 
 function growing_bacteria_agent_step!(agent::SimpleCell, model::ABM)
     fsym, compression, torque = transform_forces(agent)
-    agent.pos = agent.pos .+ model.dt * model.mobility .* fsym
+    move_agent!(agent, agent.pos .+ model.dt * model.mobility .* fsym, model)
```
Datseris, line 80:
```suggestion
    newpos = agent.pos .+ model.dt * model.mobility .* fsym
    move_agent!(agent, newpos, model)
```

Datseris:
Hm, honestly I think this error is fine... We could add an additional check inside the continuous space, but this would duplicate operations... Julia does a bounds check anyway, and we would do it twice. Perhaps mention in the docs that if periodic=false you can't move agents outside the range.

Libbum:
Sure. I'll do that.
As for bacteria, this is a low priority. It's not a quick fix, so for now I'll just drop the step count down a little so the boundaries don't become an issue for the example.

Libbum:
```julia
model, agent_step!, model_step! = Models.flocking()

sample = Vector{Tuple{Models.Bird, Models.Bird}}(undef, 0)
for _ in 1:5000
push!(sample, (random_agent(model), random_agent(model)))
end

dims = model.space.extent

a = collect(Agents.euclidean_periodic(p1.pos, p2.pos, dims) for (p1, p2) in sample)
b = collect(ttt(p1.pos, p2.pos, dims) for (p1, p2) in sample)
sort(a) == sort(b) #true

# This is essentially how they implement it in Distances.jl
function ttt(p1, p2, dims)
s = 0.0
for i in eachindex(p1, p2, dims)
s+= tt(p1[i], p2[i], dims[i])
end
sqrt(s)
end
    
function tt(p1, p2, dims)
    s1 = abs(p1 - p2)
    s2 = mod(s1, dims)
    s3 = min(s2, dims - s2)
    abs2(s3)
end

julia> @btime collect(Agents.euclidean_periodic(p1.pos, p2.pos, $dims) for (p1, p2) in $sample)
  20.800 μs (2 allocations: 39.14 KiB)

julia> @btime collect(ttt(p1.pos, p2.pos, $dims) for (p1, p2) in $sample)
  86.085 μs (2 allocations: 39.14 KiB)
```

So there's no need to adjust our implementation. But are you OK with me making this an exported function for Grid & Continuous?

Datseris:
> But are you OK with me making this an exported function for Grid & Continuous?

Shouldn't we just make a single function `euclidean` where the third argument is the model itself, and the function dispatches on the `periodic` type parameter of the space?

Libbum:
Yes, exactly. That's the plan.

Libbum:
2f7c584 adds the euclidean metric. Don't think I can make the type signatures any tighter. Well, at least I couldn't find a way to generalise the separation of Grid and Continuous for the periodic case.

Libbum:
@Datseris has the alterations you mad to the `exact=false` code made the search region slightly asymmetric? Fine if it did&mdash;don't want to go hunting for some bug that isn't even there.

Screenshot_20201005_183643

Libbum:
Final things to get done on this as far as I'm concerned:

- [ ] Another benchmark to make sure any recent changes made regressions
- [ ] 3D sanity check (in progress)
- [ ] Drop some of these helper files that wont be needed (like the plotter and the compartment space flocking test). OR we can keep them but put them in another directory in `test/`?

docs/src/api.md
```diff
@@ -36,6 +36,7 @@ sample!
 ```@docs
 nearby_ids
 nearby_positions
+euclidean
```
Datseris, line 39: The reason I suggested `edistance` instead of `euclidean` is because the latter conflicts with Distances.jl. However, given that we use Tuples instead of SVectors people can't use distances with it anyways. I think its okay.

Libbum: Oh, I thought you said euclidean. Must have read that in the email, not here. Easily updated.

Datseris:
@Libbum do we need something else here before this is shippable? I'll take care of the changelog on this one. Do you remember what were the final performance gains on the flocking model?

Libbum:
For me, just a confirmation that https://github.com/JuliaDynamics/Agents.jl/pull/322#issuecomment-703748944 is by-design, and just a clean up question in https://github.com/JuliaDynamics/Agents.jl/pull/322#issuecomment-703750638.

Otherwise it's good to go.

Flocking is about 3.2x from the original ContinuousSpace. Don't bother with the MASON comparison yet, that'll be done in #298.

Datseris:
> @Datseris has the alterations you mad to the exact=false code made the search region slightly asymmetric? Fine if it did—don't want to go hunting for some bug that isn't even there.

What, no I don't think so. In that comment you paste a figure which is incorrect as well, right? So perhaps I made a mistake? Although I've created countless figures for the exact and never had seen a problem.

Datseris:
My `exact=false ` version is symmetric around the cell center, not around the coordinate of the point. It allows for smaller searrch radious this way. I think it is fine, you think it is not fine? I don't see any problem in the figure you pasted

Libbum:
No, it's fine. 

> My exact=false version is symmetric around the cell center, not around the coordinate of the point

Yeah, that makes sense. Was just confirming since I also altered the distance function...

Datseris:
> Yeah, that makes sense. Was just confirming since I also altered the distance function... 

? I thought you simply re-wrote it, but the actual numeric computations were the same?

Libbum:
Exactly But perhaps I introduced a bug unwittingly. It's not the case though. The re-write is the same (checked in tests).

The figure above, since you've done a good job shaving off the extra cells seemed not as 'aligned' as the simple euclidean + sqrt(2) plots we were doing a few days ago. So when I saw that it made me question the changes I made today. But all is well, it's fine.

Datseris:
alright, so I'll merge as soon as tests pass!

Datseris:
All seems good across the board, https://juliadynamics.github.io/Agents.jl/previews/PR322/examples/growing_bacteria/

ALL SYSTEMS GO CAPTAIN

PR #323, opened 2020-09-25
BenchmarkCI integration

Libbum:
Test case for getting benchmarks on associated PRs. 
I may need to open and close a few dummy PRs to test this and get it right.


Datseris:
So what this PR will do in the end is make any existing PR also run the CI?

Libbum:
It will add an action that will run benchmarks of the PR against master, then provide a report like [this one](https://github.com/mtsch/Ripserer.jl/pull/104#issuecomment-698654046). But only if we label it with the "run benchmark" label.

Libbum:
I think why it's failing at the moment is that most of the benchmarks we have will just not run with the resources a VM has. I'm going to drop a lot of those in the master branch for now and make things a little less resource hungry.

Datseris:
Hm, are you sure this is worth the time investment? We could be doing other things instead, like designing how to mix spaces with a graph space... We can always run benchmarks manually, this only takes a couple of minutes to do.

Libbum:
This should have taken 5 minutes! :P
If this last bit fails, I'll park it for the moment.

Datseris:
> This should have taken 5 minutes! :P

Haha if I had 5 cents everytime I said that _and it came out true_ I would be rich as fuck!

Libbum:
Ah, that worked! I'll just add in some of the less power hungry bits tomorrow and this is done.

benchmark/benchmarks.jl
```diff
@@ -4,6 +4,14 @@ const SUITE = BenchmarkGroup(["Agents"])
...
 include("agents.jl")
 
+#### SPACE CONSTRUCTION ####
+
+SUITE["space_creation"] = BenchmarkGroup(["graph", "grid", "continuous"])
+SUITE["space_creation"]["graph"] = @benchmarkable GraphSpace(complete_digraph(1000))
+SUITE["space_creation"]["grid"] = @benchmarkable GridSpace((500, 500))
+SUITE["space_creation"]["continuous"] =
+    @benchmarkable ContinuousSpace(5; extend = (100, 100, 100, 100, 100))
+
```
Datseris: this 1 time cost can be excluded to make CI run faster and also reduce the printed comparison. In general anything that happens only 1 doesn't need to be reported here I Would say.

benchmark/benchmarks.jl
```diff
@@ -43,3 +51,305 @@ SUITE["model"]["initialise_union"]["continuous"] = @benchmarkable ABM(
...
+SUITE["graph"]["add"]["agent_pos"] =
+    @benchmarkable add_agent_pos!($graph_agent, $graph_model) samples = 100
+SUITE["graph"]["add"]["agent_single"] =
+    @benchmarkable add_agent_single!($graph_agent, $graph_model) samples = 100
+SUITE["graph"]["add"]["create_pos"] =
+    @benchmarkable add_agent!(26, $graph_model, 6.5, false) samples = 100
+SUITE["graph"]["add"]["create_single"] =
+    @benchmarkable add_agent_single!($graph_model, 6.5, false) samples = 100
+SUITE["graph"]["add"]["create"] =
+    @benchmarkable add_agent!($graph_model, 6.5, false) samples = 100
```
Datseris: Notice that you benchmark all types of add in all spaces. Yet, only the `add_agent_pos` is different between different spaces, all other add types are centralized in the API.jl file. Since the other functions are in fact the same and should be timed in only one space (GraphSpace for simplicity), to reduce time and complexity of output

Libbum: Fair points on both. I'll alter things once I manage to get it running correctly.

Libbum:
This is good to go now. We can tweak it later if need be.

Libbum:
Any issues with merging this?

Datseris:
nope, you decide

Issue #324, opened 2020-09-25
Designing mixed spaces


Datseris:
As far as infrastructure goes, it is not hard to combine 2 spaces. We can define a `MixedSpace` struct, and because of our recent space source code rework it is easy to extend `move, add, kill` to simply call the functions for each of the underlying space.
In my understanding, a mixed space makes sense only if the second space is a graph...?

What needs discussion:
1. What should `move_agent_pos` do? Should it dispatch depending on the position type? i.e. if it is integer it moves the agent to node `i`, otherwise to position `i` ?
2. When we do "add agent" where do we put the agent in the graph? do we extend `add_agent_pos` to take a forth argument, the graph node?
3. How do we modify the `nearby` functions to be usable in both graph and real space?

@Libbum in #265 you said that "it is very standard" for MASON/NetLogo, https://github.com/JuliaDynamics/Agents.jl/issues/265#issuecomment-698141944. Yet, while opening this issue, I had difficulty even coming up with what are the correct questions to ask, much less so what the implementation should be.

Perhaps you can write here what _exactly_ do MASON and NetLogo mean by having "two spaces" and how they treat "finding nearest neighbors" and "where agents are"?

Libbum:
I'm going to psedo-code / translate to Agents.jl where I can here.

## MASON
```julia
numStudents = 50
yard = ContinuousSpace(2, extend = (100, 100))
model = ABM(Student, yard)
for _ in 1:numStudents
    add_agent!(model) # More specifically, they add students randomly around (50, 50) ± 0.5
end
```

A teacher sits in the center of the yard, students move randomly about the yard, but have a force vector that attracts them to the teacher and compels them to stay away from the fence. This means they mostly huddle around the teacher (depending on the strength of the teachers attractor).

Now a social network is added.
```julia
buddies = GraphSpace(path_graph(numStudents))
```
Which is added to the `Students` class, our equivalent would be something like `ABM(Student, yard, buddies)` or a space union perhaps.

The initialisation would be extended to something like this:

```julia
numStudents = 50
yard = ContinuousSpace(2, extend = (100, 100))
buddies = GraphSpace(path_graph(numStudents))
model = ABM(Student, yard, buddies)
for _ in 1:numStudents
    s = add_agent!(model, yard) # Not true syntax, just an example
    add_agent!(s, model, buddies) # to illustrate the mechanics of MASON
end
for student in 1:numStudents
[...]

PR #325, opened 2020-09-26
Schoolyard. LightGraphs integration

Libbum:
See #324


examples/schoolyard.jl
```diff
@@ -0,0 +1,138 @@
...
+        ## Add one friend and one foe to the social network
+        friend = rand(filter(s -> s != student, 1:numStudents))
+        add_edge!(model.buddies, student, friend, rand())
+        foe = rand(filter(s -> s != student, 1:numStudents))
+        add_edge!(model.buddies, student, foe, -rand())
+    end
+    model
+end
+
+# Our model contains the `buddies` property, which is our LightGraphs.jl undirected, weighted graph.
```
Datseris, line 76: Above you say that the friendship is not reciprocal, but here the graph is undirected. How can this be?

examples/schoolyard.jl
```diff
@@ -0,0 +1,138 @@
...
+function agent_step!(student, model)
+    ## place a teacher in the center of the yard, so we don’t go too far away
+    teacher = (model.space.extend .* 0.5 .- student.pos) .* model.teacher_attractor
+
+    ## add a bit of randomness
+    noise = model.noise .* (Tuple(rand(2)) .- 0.5)
+
+    ## Adhere to the social network
+    network = model.buddies.weights[student.id, :]
+    tidxs, tweights = findnz(network)
```
Datseris, line 94: perhaps we should explain what `findnz` does

examples/schoolyard.jl
```diff
@@ -0,0 +1,138 @@
...
+end
+
+# Our model contains the `buddies` property, which is our LightGraphs.jl undirected, weighted graph.
+# As we can see in the loop, we choose one `friend` and one `foe` at random for each `student` and
+# assign their relationship as a weighted edge on the graph.
+
+# ## Movement dynamics
+
+distance(pos) = sqrt(pos[1]^2 + pos[2]^2)
+scale(L, force) = (L / norm(force)) .* force
```
Datseris, line 83: you can replace `norm` with `distance`, and thus not have to use LinearAlgebra at the top.

examples/schoolyard.jl
```diff
@@ -0,0 +1,138 @@
...
+            else
+                L = model.max_force - distance(force)
+                force = scale(L, force)
+            end
+        end
+        network_force = network_force .+ force
+    end
+
+    ## Add all forces together to assign the students next position
+    student.pos = student.pos .+ noise .+ teacher .+ network_force
```
Datseris, line 117: here you should use `move_agent!(student, newpos, model)` otherwise the space structure won't be able to keep track of the change in position.

Libbum: That's what the update_space! call does, but sure&mdash;I can change to move_agent! here.

examples/schoolyard.jl
Libbum: Oversight. I work with undirected graphs far more often, so just typed that without thinking.

examples/schoolyard.jl
```diff
@@ -0,0 +1,142 @@
...
+# the [MASON](https://cs.gmu.edu/~eclab/projects/mason/) ABM framework uses as an introductory example
+# in their [documentation](https://cs.gmu.edu/~eclab/projects/mason/manual.pdf).
+
+# Rather than implementing an Agents.jl&ndash;specific graph structure, we can interface with
+# [LightGraphs.jl](https://github.com/JuliaGraphs/LightGraphs.jl): a high class library for managing
+# and implementing graphs, which can be re-used to establish social networks within existing spaces.
+
+# To begin, we load in some dependencies
+
+using Agents, LightGraphs, SimpleWeightedGraphs, SparseArrays, AgentsPlots
```
Datseris, line 20: Where in the example do you use SparseArrays?

Also, do you need to use both LightGraphs and SimpleWeightedGraphs? I think only the latter is enough

Libbum: findnz is in sparsearrays

Datseris:
https://juliadynamics.github.io/Agents.jl/previews/PR325/examples/schoolyard/

animation is not shown

Datseris:
Agents.jl&ndash;specific is pobably mistake

examples/schoolyard.jl
Libbum: add_edge! is in LightGraphs.

examples/schoolyard.jl
Libbum: No, I can't do this. Need to use update_space! here, since move_agent!(student, newpos, model) doesn't exist for continuous space. That's why the example fails.

Datseris:
should we scatterplot some red dot in the place where the teacher should be?

PR #326, opened 2020-09-26
Allow mutable graphs

### LINKED ISSUE #264: Allow mutable graph spaces (support add_vertex! and rem_vertex!)
Hi there, I used ABM with the `step!` function for simulation, but I found that even if some agents were killed in some steps, the graph in the space did not update. I verified this by calculating the vertices of the network using `nv(model.space.graph)`, and it seems that the output is corresponding to the beginning state of the simulation . Does anyone know why this happened? And I am wondering how can I get to know the properties of the network in each step if I cannot make use of `model.space.graph`? Thanks for your time in advance!

Datseris:
Closes #264

I am having trouble because I don't know anything about LightGraphs and perhaps someone should help me here.

How does the order of node integers change in LightGraphs.jl when you add another one or remove one? If I do `rem_vertex!`, can I just do `deleteat!(model.space.s, n)` and then the correspondence between node numbers in the graph, and node numbers in `model.space.s` stays the same? 

And what about adding new nodes? If I add a new node in lightgraphs, is it automatically becoming the `nv(graph)+1` node? From the documentation of `add_vertex!` it looks like it.

### DIFF
--- a/src/spaces/graph.jl
+++ b/src/spaces/graph.jl
@@ -140,3 +140,11 @@ end
 #######################################################################################
 # Mutable graph functions
 #######################################################################################
+"""
+    rem_node!(n::Int, model::ABM{A, <: GraphSpace})
+Remove node `n` from the model's graph. All agents in that node are killed.
+"""
+function rem_node!(n::Int, model::ABM{<:AbstractAgent, <: GraphSpace})
+    for a ∈ agents_in_position(n, model); kill_agent!(a, model); end
+    # LightGraphs.rem_vertex!(model.space.graph)
+end
--- a/src/spaces/graph.jl
+++ b/src/spaces/graph.jl
@@ -14,10 +14,9 @@ Create a `GraphSpace` instance that is underlined by an arbitrary graph from
 [LightGraphs.jl](https://github.com/JuliaGraphs/LightGraphs.jl).
 The position type for this space is `Int`.
 
-The nodes of the graph space can be altered with the functions [`add_node!`](@ref),
-[`rem_node!`](@ref) and [`add_edge`](@ref), assuming the used graph is mutable.
+The underlying graph can be altered using [`add_node!`](@ref) and [`rem_node!`](@ref).
 
-`GraphSpace` represents a space where each node of a graph can hold an arbitrary
+`GraphSpace` represents a space where each node (i.e. position) of a graph can hold an arbitrary
 amount of agents, and each agent can move between the nodes of the graph.
 If you want to model social networks, where each agent is equivalent with a node of
 a graph, you're better of using `nothing` (or other spaces) as the model space, and using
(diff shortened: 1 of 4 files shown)


Datseris:
The documentation of LightGraphs.jl says that:

  This operation has to be performed carefully if one keeps external data structures indexed    
  by edges or vertices in the graph, since internally the removal is performed swapping the     
  vertices v and |V|, and removing the last vertex |V| from the graph. After removal the        
  vertices in g will be indexed by 1:|V|-1.

which is too complicated for me to understand right now :D So I'll leave this PR until an expert of light graphs wants to continue it.

yanyuan-wang:
Hi, here's my understanding.

What `rem_vertex!` does is swapping the index of the last node and that of the one to be removed. Suppose there are 10 nodes in the graph, `rem_vertex!(graph, 2)` will delete the second node and then re-index the last node as 2, while keeping the index of the rest 8 nodes the unchanged. So, this "swapping" operation can make sure that the maximum index is equal to the number of nodes in the graph.

And whenever a new node is added, it will automatically be attached with the index of `nv(graph)+1`.

src/spaces/graph.jl
```diff
@@ -127,3 +138,23 @@ function nearby_positions(
...
+`n` node (and all its edges and agents) are deleted.
+"""
+function rem_node!(n::Int, model::ABM{<:AbstractAgent, <: GraphSpace})
+    for a ∈ agents_in_position(n, model); kill_agent!(a, model); end
+    s = model.space.s
+    V = nv(model)
+    s[V], s[n] = s[n], s[V]
+    pop!(s)
+    LightGraphs.rem_vertex!(model.space.graph, n)
+end
```
Datseris, line 160: @yanyuan-wang does this seem correct to you?

Datseris:
I've fixed this and merging when tests pass.

src/spaces/graph.jl
```diff
@@ -127,3 +137,39 @@ function nearby_positions(
...
+**Warning:** LightGraphs.jl (and thus Agents.jl) swaps the index of the last node with
+that of the one to be removed, while every other node remains as is. This means that
+when doing `rem_node!(n, model)` the last node becomes the `n`-th node while the previous
+`n`-th node (and all its edges and agents) are deleted.
+"""
+function rem_node!(model::ABM{<:AbstractAgent, <: GraphSpace}, n::Int)
+    for id ∈ copy(ids_in_position(n, model)); kill_agent!(model[id], model); end
+    V = nv(model)
+    success = LightGraphs.rem_vertex!(model.space.graph, n)
+    n ∉ 1:V && error("Node number exceeds amount of nodes in graph!")
```
kavir1698, line 158: Lines 156 and 158 should be the first two lines of the function.

Isn't `n > V` faster than `n ∉ 1:V`?

src/spaces/graph.jl
yanyuan-wang: Yes, that looks good!

src/spaces/graph.jl
Datseris: yes, by 1 boolean operation. `n \in range` reduces to:
```
in(x::Integer, r::AbstractUnitRange{<:Integer}) = (first(r) <= x) & (x <= last(r))
```

PR #328, opened 2020-09-29
Improve documentation of `paramscan`

### LINKED ISSUE #327: Directly providing model properties to paramscan requires workaround
### Preamble 
*Please note that this is not a bug report!* I'm just trying to document my findings for others who face the same situation as I just did. But maybe my writing from a user perspective is still helpful for the development of Agents.jl.

### My usual approach
Normally I pass my model `properties` dictionary directly to my initialisation function for use in the `ABM` function like so:

``` julia
function initialise(; properties)
    space = # define space
    model = ABM(Employee, space; properties = properties)
    # and so on
end

properties = Dict(a => true, b => false)
model = initialise(; properties)
```

(I'm aware that this approach is not quite in line with the examples in the documentation but I actually prefer it like that.)

### Applying same approach to `paramscan`
Which works perfectly fine. However, I now wanted to use a similar approach together with the provided `paramscan` function:

``` julia
properties = Dict(a => [false, true], b => false)
paramscan(properties, initialise)
```

But this throws me an error: `ERROR: LoadError: UndefKeywordError: keyword argument properties not assigned`.

How come it works in the one case but not in the other? I had a quick look at the source of `paramscan` and it seems like it deconstructs my `properties` dictionary and tries to pass those key-value-pairs one-by-one as kwargs to the `initialisation` function. Of course then there's no `properties` dictionary anymore and the above error gets thrown. This is certainly not an error and is working as intended but it was indeed a bit confusing when I did this for the first time.

### Assumed simple solution
This situation could be easy to solve:

``` julia
properties = Dict(:a => [false, true], :b => false)
parameters = Dict(:properties => properties)
paramscan(parameters, initialise)
```

But this entails another problem outlined in the comment below!

Datseris:
This PR improves the docs of `paramscan` and thus addresses #327

Closes #327

### DIFF
--- a/src/simulations/paramscan.jl
+++ b/src/simulations/paramscan.jl
@@ -3,9 +3,10 @@ export paramscan
 """
     paramscan(parameters, initialize; kwargs...) → adf, mdf
 
-Perform a parameter scan of the model by collecting simulation output from all parameter
-combinations into a single dataframe (one for the agents and one for the model),
-where the parameter values are also columns (besides the Agents.jl collected data).
+Perform a parameter scan of a ABM simulation output by collecting data from all
+parameter combinations into dataframes (one for agent data, one for model data).
+The dataframes columns are both the collected data (as in [`run!`](@ref)) but also the
+input parameter values used.
 
 `parameters` is a dictionary with key type `Symbol` which contains various parameters that
 will be scanned over (as well as other parameters that remain constant).
@@ -16,8 +17,9 @@ contains many parameters and thus is scanned. All other entries of
 
 The second argument `initialize` is a function that creates an ABM and returns it.
 It should accept keyword arguments which are the *keys* of the `parameters` dictionary.
-Since the user decides how to use input arguments to make an ABM, `parameters`
-may contain both model and agent properties.
+Since the user decides how to use input arguments to make an ABM, `parameters` can be
+used to affect model properties, space type and creation as well as agent properties,
+see the example below.
 
 ## Keywords
 The following keywords modify the `paramscan` function:
@@ -48,7 +50,7 @@ function initialize(; numagents = 320, griddims = (20, 20), min_to_be_happy = 3)
     return model
 end
 ```
-and collect data by doing:
+and do a parameter scan by doing:
 ```julia
 happyperc(moods) = count(x -> x == true, moods) / length(moods)
 adata = [(:mood, happyperc)]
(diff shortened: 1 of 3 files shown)


src/simulations/paramscan.jl
```diff
@@ -15,20 +15,18 @@ Keys of `parameters` should be of type `Symbol`.
...
-Keyword arguments such as `adata` and `mdata` are
-also propagated.
-
+## Keywords
 The following keywords modify the `paramscan` function:
 
-`include_constants::Bool=false` determines whether constant parameters should be
-included in the output `DataFrame`.
+- `include_constants::Bool=false` determines whether constant parameters should be
+  included in the output `DataFrame`.
```
fbanning, line 26: Not directly related to this PR but maybe it might be nice to be able to pass a list of parameters that should be included. This way users could pick which constants should be included and which should be left out of the df.

Datseris: You can simply not include them in `parameters` though, is this good enough?

Datseris:
@fbanning this PR address your comments. The short answer is, the approach you want to take is not possible with `paramscan`. Every parameter must be explicitly stated as keyword argument to the constructing function. This is the only way we can expand parameters into dataframe columns.

fbanning:
Thank you, I think the this PR solves a bit of the confusion I initially had when using `paramscan`. It will most certainly help others in the future as well. 👍

Datseris:
BTW if you haven't used DrWatson, I highly recommend checking it out

src/simulations/paramscan.jl
fbanning: I don't think I understand what you mean by this. If I normally provide some constant model properties like the grid size within `parameters`, then I cannot just leave them out of `parameters` because they're crucial for the model. What do you mean by saying "simply not include them in `parameters`"?

fbanning:
I had a look at its documentation and use cases as well as your JuliaCon video on it. After a bit of reading though, I assumed that it does too much for my taste (like creating project folder structure or predefining how files are saved). Instead of getting used to DrWatson, we simply created our own folder structure for our project and wrote our own saving functions where would easily define file names and directories. Maybe I will give it another shot at some point in the future and read more thouroughly through its API documentation but currently it seemed unnecessary to me to introduce it in our project.

Datseris:
Although this gets too much off topic, noone forces you to follow the predefined structure. You can still use things like `@quickactivate` and `@tagsave` and benefit from them, same goes with `dict_list` which we use here in `paramscan`.

Issue #329, opened 2020-10-03
Agents git history is 80 gb


Datseris:
we have to prune a buttload of stuff from the documentation branch.

Libbum:
Yeah, gh-pages just explodes this. I wonder if we can just set up an action to prune everything other than tagged releases, open PR previews and current dev.

Datseris:
I actually don't know how to _remove_ stuff from the GitHub history. If you know how, please do so, kill all docs hosted before version 3.0, all previews, everything. 

I know where to find the files (by simply checking out the gh-pages branch), but I don't really know how to remove them from `.git`. even if you delete the files, they still stay in `.git`, because that's the point of git...

Libbum:
I'm going to play around with git-filter-repo on a personal fork. I think I can do this, but it wont be something we put in an action. Serious consequences that need to be tested & retested before we do it here.

Libbum:
OK, here's what can be done. I'll use my fork as the example (these are exactly the steps taken)

```
# Bare, mirror clone the repo
git clone git@github.com:Libbum/Agents.jl.git --bare --mirror

# Analyse the repository (just makes things a bit faster)
git filter-repo --analyze

# Remove everything below v3 and clear out all PR previews on the gh-pages branch
git filter-repo --invert-paths --path-glob 'v0*' --path-glob 'v1*' --path-glob 'v2*' --path-glob 'previews/*' --refs gh-pages

git push
```
My tree looks [like this](https://github.com/Libbum/Agents.jl/tree/gh-pages) now. 

The [versions.js](https://github.com/Libbum/Agents.jl/blob/gh-pages/versions.js) can be modified manually to drop out the links to our now removed files.

Using [git filter-repo](https://github.com/newren/git-filter-repo).

Do you want to try that out too on a personal branch just to make sure? Otherwise I suggest I bare mirror clone JuliaDynamics/Agents.jl twice, one for backup, one for this operation. Push the altered version, we make a dummy PR to make sure docs and previews don't fail. If all goes well, that's that&mdash;we push the backup like nothing ever happened.

Datseris:
What's the size of your fork after these operations?

Sure, go ahead and duplicate the repo in JuliaDynamics, that is fine for me. Notice also that your fork is a perfectly valid backup just as well, right?

Libbum:
1GB.

Sure, just wanting to be extra cautious. I'll do this tomorrow.

Datseris:
[...]

PR #330, opened 2020-10-05
Run doc preview test

Libbum:
This is a test PR that should not be merged. It's only here to check that documentation & preview builds still work after changes due to #329.


Libbum:
https://juliadynamics.github.io/Agents.jl/previews/PR330/

Looks good!

Issue #265, opened 2020-07-31
Build ABMs with OpenStreetMapX.jl and Agents.jl


fbanning:
Hi,

I would like to use OSM map data for some agent-based models and was wondering if it is possible to utilise the functionality of [OpenStreetMapX.jl](https://github.com/pszufe/OpenStreetMapX.jl) for this. I saw that the maintainers have created a simple Zombie apocalypse model for their JuliaCon 2020 example. This has been very helpful for me to understand the general idea how to build an ABM in combination with OpenStreetMapX.jl and plotting it via folium (such a nice tool btw!).

However, I would prefer to use the Agents.jl framework for building ABMs and would like to ask whether or not it is possible to use the functionality of those packages together? I've been tinkering around with it but couldn't get the desired outcome so far.

My aim would be to integrate the MapData generated by OpenStreetMapX.jl into the space component (GraphSpace) of a model created with Agents.jl. After some agents have been added within the bounds of MapData, I would then imagine to have two separate networks - one for the agents and one for the underlying map on which the agents move. When plotting the model, both graph networks could be accessed by Agents.jl and plotted normally (respecting bounds, relative location, etc.).

Would this be at all feasible? I'm looking forward to hear your thoughts on this. Maybe it's way too complex. Maybe it's really easy and I just miss the forest for the trees.
(Sidenote: I have also asked this question in the issues section of the OpenStreetMapX.jl repo because I am unsure who to ask regarding this topic.)

Best wishes

Datseris:
Hi there,

I thought of exactly the same thing when I saw the talk, and I absolutely think we can get an interface between the two. I have already contacted @pszufe , which I hope to motivate, because I believe we will need some help on the side of OpenStreetMapX.jl (which I know nothing about).

The process to do this interfacing amounts to defining a new type of *space* for Agents.jl, which uses internally the mechanics of OpenStreetMapX.jl . Once we have defined this space, and fundamental methods like `move_agent!` and `add_agent!`, the rest of the interface of Agents.jl will "just work" for agents moving around in this open street map space.

Give me some time, and I will "define" the space API that Agents.jl requires (everything is already defined in the source code, but a high level developer documentation will greatly help).

Datseris:
@fbanning can you please cross-link the issue on the other repo?

kavir1698:
Hi,

[...]

PR #332, opened 2020-10-10
Support for OpenStreetMapXSpace

pszufe:
OK, so this is the first version of API for OpenStreetMapX spaces following the discussion in:
https://github.com/JuliaDynamics/Agents.jl/issues/265

I am not sure what other functions are mandatory for a space at Agents.jl. 
Currently the API seems slightly unfriendly to me and surely I have interpreted something wrongly where implementing it. 
So I await comments on the integration pattern, functions that I shall provide etc.

One thing that is particularly ugly is that currently unable to show `ABM{OSMXAgent{Nothing},OpenStreetMapXSpace}` using`display` function. Of course I could use brute force approach and implement display(::ABM{OSMXAgent{Nothing},OpenStreetMapXSpace}) but there surely must be a nicer way.

### DIFF
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -4,6 +4,8 @@ using Distributed
 using LightGraphs
 using DataFrames
 using Random
+using OpenStreetMapX
+using Parameters
 import Base.length
 
 # Core structures of Agents.jl
@@ -17,6 +19,7 @@ include("spaces/grid.jl")
 include("spaces/discrete.jl")
 include("spaces/continuous.jl")
 include("spaces/utilities.jl")
+include("spaces/openstreetmapx.jl")
 
 # Stepping and data collection functionality
 include("simulations/step.jl")
--- a/Project.toml
+++ b/Project.toml
@@ -8,6 +8,8 @@ DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+OpenStreetMapX = "86cd37e6-c0ff-550b-95fe-21d72c8d4fc9"
+Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
--- a/src/spaces/openstreetmapx.jl
+++ b/src/spaces/openstreetmapx.jl
@@ -1,0 +1,176 @@
+export OpenStreetMapXSpace, OSMXPos, OSMXAgent, get_ENU, get_coordinates
+
+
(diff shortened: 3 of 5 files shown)


src/Agents.jl
```diff
@@ -4,6 +4,8 @@ using Distributed
 using LightGraphs
 using DataFrames
 using Random
+using OpenStreetMapX
+using Parameters
```
Datseris, line 8: Julia provides the same macro as Parameters: `Base.@kwdef` and thus Parameters are not necessary as a dependency.

Datseris:
Yes there is. Simply extend `Base.show(io, space)` to dispatch on OpenStreetMapSpace. For example, in GridSpace we do:
```julia
function Base.show(io::IO, space::GridSpace{D, P}) where {D, P}
    s = "GridSpace with size $(size(space)), metric=$(space.metric) and periodic=$(P)"
    print(io, s)
end
```

src/Agents.jl
pszufe: I use `@with_kw` because of pretty-printing which makes life nicer when doing agent-based computing.
Consider  `Base.@kwdef struct B; x::Int; y::Vector{Int}; end` vs `@with_kw struct C; x::Int; y::Vector{Int}; end`:
```
julia> B(1, [2,3,4])
B(1, [2, 3, 4])

julia> C(1, [2,3,4])
C
  x: Int64 1
  y: Array{Int64}((3,)) [2, 3, 4]
```
I always use this for complex structures because it allows debugging in the REPL. However this is up to you. The constructors in my code can live without it altogether.

Datseris:
> I am not sure what other functions are mandatory for a space at Agents.jl.
Currently the API seems slightly unfriendly to me and surely I have interpreted something wrongly where implementing it.
So I await comments on the integration pattern, functions that I shall provide etc.

Okay, then this is a point to improve for us. The documentation on implementing a new space is described in detail here:  https://juliadynamics.github.io/Agents.jl/dev/devdocs/ Do you mind telling explicitly what is not clear?

pszufe:
> Okay, then this is a point to improve for us. The documentation on implementing a new space is described in detail here: https://juliadynamics.github.io/Agents.jl/dev/devdocs/ Do you mind telling explicitly what is not clear?

- looking at the codes I have a feeling that this list is incomplete, for an example what about `add_agent!`
- it should be clearly stated what functions do. For example should `nearby_ids` return indices of locations in the space or return indices of agents within a distance. Currently I am returning indices within the network. The method names are definitely never self-explanatory - each needs one- two- sentences of explanation.

Datseris:
> looking at the codes I have a feeling that this list is incomplete, for an example what about add_agent!

No the list is not incomplete. Because of the design of Agents.jl you only have to extend the functions listed. `add_agent!` is automatically implemented for any space that does so. `GridSpace, ContinuousSpace, GraphSpace` do not explicitly extend `add_agent!`. The method `add_agent_to_space` is all it takes.


> it should be clearly stated what functions do. For an example should nearby_ids return indices of locations in the space or return indices of agents within a distance? 

...? Every single function in that list has a documentation string. `nearby_ids` returns IDs of agents, while `nearby_positions` returns valid position types. I thought that this was very clear from the docstrings, no? 

image


image

and I am cutting the docstrings short in what I paste here, given that they have plethora of more information.

Datseris:
Closing in favor of #333

PR #333, opened 2020-10-15
An alternate OSM Space

Libbum:
Here's another suggestion for #265 to compare with #332.

I see some complications with #332, such as no ability to chose a 'true' random location on the map, everything works on intersections. Nearest neighbours are represented via this property as well, so if you take any given intersection, and collect all connecting intersections&mdash;an agent is listed as a neighbour regardless if it's 20m or 200m away.

This PR introduces a more complex search mechanism at the expense of performance. Unsure which direction we want to go with this however, so it's not completely finished. There are certainly bugs and it could be faster in places (it's not that bad at present anyway).

```julia
@benchmark step!(m, agent_step!, 100) setup=(m=initialise())
BenchmarkTools.Trial: 
  memory estimate:  5.92 MiB
  allocs estimate:  5043
  --------------
  minimum time:     19.446 ms (0.00% GC)
  median time:      46.263 ms (0.00% GC)
  mean time:        47.970 ms (5.54% GC)
  maximum time:     70.517 ms (12.28% GC)
  --------------
  samples:          15
  evals/sample:     1
```

For one, I've simplified the requirement of the agent constructor. `pos` needs to be `(Int, Int, Float64)` and a `route::Vector{Int}` is needed, that's it. I'd say we could also have an *optional* `destination` parameter that we respect, so that agents can arrive at the middle of a street, not just on an intersection. This is a simple addition that I'll do later if we go down this path.

There are a bunch of `NOTE:`s and `TODO:`s listed around the code which we can discuss before moving forward, so I wont re-iterate those here.

I've not looked into plotting at all yet, other than to say that using Plots is a bad idea&mdash;the map **must** be redrawn at every frame, and that takes a long time. Would be a good push to get us to at least investigate a Makie version in the example.

outbreak


Libbum:
There's a guy on the bottom right who just sails across what appears not to be a road. I suspect that's because the end of the road is clipped out of the map frame?

kavir1698:
Maybe we can keep both search mechanisms? sometimes performance becomes the top priority.

Libbum:
Well, it's pretty simple for this implementation not to consider destinations on roads to be an end-point. However, you sacrifice essentially all that it means to be a map space. An intersection-only search is just a `GraphSpace` search. I'd go as far to say that we wouldn't need a new `OSMSpace` in that case&mdash;just an integration example in the documentation.

That's my argument for this method vs the other, and keeping both really means two completely different outcomes.

kavir1698:
Fair enough. An integration example for the other implementation would be enough.

Datseris:
@Libbum I'll do a full review here. Can you please rebase/update to current master?

Libbum:
Haven't checked it will run yet, but the conflicts were minor.

Datseris (changes requested):
Why do we need 2 maps to test this? Toronto is 60mb, this is humongous. In fact, can we avoid adding maps entirely? Perhaps OpenStreetMapX.jl already has some maps in its tests.

Datseris:
I made some changes to the function names because I was getting a bit lost, given the size of this change. We can revert them at the end or keep them starting with `osm` for clarity. 

I made a core change: created a Type `OSMPos`. @Libbum , here you were committing the highest form of type piracy: pirating `Base` itself! You were extending a method for `::NTuple{Int, Int, Float64}`! We shouldn't do this! I added a new type for now as a workaround, but we can think about this later.

Datseris:
@Libbum wait, the "distance" along the road the agent moves is not bound to [0, 1] but instead in meters, the real meters between the nodes?

Datseris:
LOL, GOLDEN LOL, I just saw the section titles in the docs example. Hahahaha, worthy continuation of the `PoorSoul` saga from the social distancing example hahahaha

src/spaces/openstreetmap.jl
```diff
@@ -0,0 +1,402 @@
...
+# Custom functions for OSMSpace
+#######################################################################################
+
+# EXPORTED
+"""
+    osm_random_direction(model::ABM{A,OpenStreetMapSpace})
+
+Similar to `random_position`, but rather than providing only intersections, this method
+returns a location somewhere on a road heading in a random direction.
+"""
```
Datseris, line 69: It is not completely clear what this function returns. We should add an arrow `-> OSMPos`. But the function name also doesn't help. It doesn't feel liike this function returns a valid position instance for the map space, but perhaps something else. Given that I've added `osm` in front, we can maybe make it `osm_random_position` instead?

src/spaces/openstreetmap.jl
```diff
@@ -0,0 +1,402 @@
...
+
+The abbreviation `OSMSpace` may be used interchangably.
+
+Much of the functionality of this space is provided by interfacing with
+[OpenStreetMapX.jl](https://github.com/pszufe/OpenStreetMapX.jl), for example the two
+keyword arguments `use_cache = false` and `trim_to_connected_graph = true` are
+passed into the [`OpenStreetMapX.get_map_data`](@ref) function.
+For details on how to obtain an OSM file for your use case, consult the OpenStreetMapX.jl
+Readme.
+"""
```
Datseris, line 22: The docstring of `OpenStreetMapSpace` should explain how the business of of planning a route works. Furthermore, I believe that the expected `Agent` fields and their types should be documented here, in this docstring, instead of the `AbstractAgent` docstring.

src/spaces/openstreetmap.jl
```diff
@@ -0,0 +1,402 @@
...
+Returns the distance travelled between two intersections.
+"""
+osm_road_length(pos::OSMPos, model::ABM{<:AbstractAgent,<:OpenStreetMapSpace}) =
+    osm_road_length(pos[1], pos[2], model)
+osm_road_length(p1::Int, p2::Int, model::ABM{<:AbstractAgent,<:OpenStreetMapSpace}) =
+    model.space.m.w[p1, p2]
+
+#HELPERS, NOT EXPORTED
+
+get_ENU(pos::Int, model) = model.space.m.nodes[model.space.m.n[pos]]
```
Datseris, line 120: No idea what this means. ENU? Can we rename it to something clearer and more verbose? Since it is internal its length doesn't matter much.

src/spaces/openstreetmap.jl
```diff
@@ -0,0 +1,402 @@
...
+`distance` travelled in meters along an agents current route.
+"""
+function move_agent!(
+    agent::A,
+    model::ABM{A,<:OpenStreetMapSpace},
+    distance::Real,
+) where {A<:AbstractAgent}
+
+    #TODO: Assumption that an agent can only end its route on an intersection.
+    #It cannot pull up to someone's house, or park on the side of the road at present.
```
Datseris, line 165: I'm a bit confused by this comment. If the position of the agent is `(0, 1, 0.2)` and the road length between 0 and 1 is 4.0, then `move_agent!(..., 2.0)` will make the agent have position `(0, 1, 2.2)`, right? And therefore the movement indeed stopped in-between the nodes.

Datseris (commented):
Hi @Libbum ,
I'm adding a basic initial review here. I think it would be useful to add some tests as well since I think I'll close the alternative #332 in favor of this one.

src/spaces/openstreetmap.jl
Datseris: Oh, also we should discuss here that this space really represents the map as a _continuous_ entity, in the expence of performance. if one only cares about interesections they can just load to graph and make a `GraphSpace`.

src/spaces/openstreetmap.jl
Libbum: "Position" in OSMX.jl is a node on a graph - which for users would translate to intersections on a map. #332 uses `random_position` to choose one of these intersections at random, but therefore cannot provide all positions on the map. That function is still useful in many cases though, so I kept it as-is.

This new one, it takes a random *edge* of the graph (which translates to road), and chooses a random position on that road, as well as a heading (or direction of travel)). In that sense, this is not an `osm_random_position`, it's closer to `osm_random_vector`, but doesn't have a velocity associated with it, so "direction" or "heading" are better qualifiers.

src/spaces/openstreetmap.jl
Libbum: Good with all of those things. Was more focused on implementation. Will keep this as a TODO once everything else settles down here.

src/spaces/openstreetmap.jl
Libbum: It's an OSM thing. I don't know. I just needed it to convert the underlying structure. Will dig deeper into what it is....

src/spaces/openstreetmap.jl
Libbum: See the example, we have
```julia
start = osm_random_direction(model) # Somewhere on a road
finish = random_position(model) # At an intersection
```
as a contrast. We could rename `random_position` to `osm_random_intersection`, then reconsider what `osm_random_direction` should be called.

src/spaces/openstreetmap.jl
Libbum: Been a while since I wrote this, so I'm a bit rusty on the reason. What I recall is that the issue is not in `move_agent!` but in the route planner. As I said above, OSMX.jl only works with intersections as far as I can tell. So the call to `shortest_route` is between nodes not some way down an edge.

 In the example we use `finish = random_position(model) # At an intersection`, if we were to use `osm_random_direction` without extending the route planner I think we get into the situation where the route is finished, but there is still distance to travel and our stopping condition isn't met, but there's no understanding of where to move...

Libbum:
> Why do we need 2 maps to test this? 

We don't, one is the default for OSMX.jl, which I used initially, then the other I pulled in to be in-line with the example in #332. Good idea to use the one in the example path. Will switch it.

> keep them starting with `osm` for clarity.

Agreed. I think this is quite a special space and won't be used all the time. Good to separate it a bit. I've got ideas for other spaces in the future and it'll start getting hard to keep everything in check if we don't start making things somewhat modular.

> extending a method for `::NTuple{Int, Int, Float64}`

An oversight&mdash;this was copied directly from #332 to try and understand what/why the types were implemented as they were. Forgot to fix it.

> the "distance" along the road the agent moves is not bound to [0, 1] but instead in meters, the real meters between the nodes?

Yeah, I think so. The graph has everything in meters, so I kept using that. I recall that #332 uses percentage, but it was just an extra operation that wasn't needed. Probably I just copied the docstring and didn't update it there.

---

It'll take me a little to get reacquainted with this code and identify the pieces still to-do. But these suggestions are a good start. Thanks!

Not too sure if I'll have time for this today.

src/spaces/openstreetmap.jl
Datseris: We should instead have a `random_route` function instead then that optionally gets an input starting node.

Libbum: Not a bad idea. Will keep that an open possibility.

examples/zombies.jl
```diff
@@ -0,0 +1,117 @@
...
+
+function initialise(; map_path = TEST_MAP)
+    model = ABM(Zombie, OpenStreetMapSpace(map_path))
+
+    for _ in 1:100
+        start = osm_random_direction(model) # Somewhere on a road
+        finish = random_position(model) # At an intersection
+
+        ## We already have our start and finish edges, so we identify the in-between
+        path = osm_plan_route(start[2], finish[1], model)
```
Datseris, line 34: I think we can streamline this to a single function `add_random_route!`. In the end that's what we care about, that each agent has a random route to follow.

And specifically we can make it `add_random_route!(start = random)`.

Libbum: Maybe true, but it'll just be a wrapper. `osm_plan_route` is far more important since I suspect that users will want to look at traffic flows and other data-driven solutions with this map space. In addition, there's a route planner *by distance* and one *by quickest* or something. We should allow access to both.

Datseris: of course `osm_plan_route` will be epxported, my suggestion is just for wrapping these 3 lines. Does the route planning already have a "quickest" and "by distance" options? And what does "by distance" even mean by the way?

src/spaces/openstreetmap.jl
```diff
@@ -0,0 +1,408 @@
...
+"""
+    osm_road_length(pos::OSMPos, model)
+    osm_road_length(start::Int, finish::Int, model)
+
+Returns the distance travelled between two intersections.
+"""
+osm_road_length(pos::OSMPos, model::ABM{<:OpenStreetMapSpace}) =
+    osm_road_length(pos[1], pos[2], model)
+osm_road_length(p1::Int, p2::Int, model::ABM{<:OpenStreetMapSpace}) =
+    model.space.m.w[p1, p2]
```
Datseris, line 119: Hm, that's a bit unintuitive. I would expect the `osm_road_length` would instead give the existing real distance between two intersections, not where the agent is.

Datseris: Oh, I think this is what it does anyways. The docstring is wrong.

Datseris:
So what is really missing here? How far away are we from wrapping this up?

Libbum:
If you're happy with the implementation in essence, then mainly all I need to do is fix up some edge cases that I identified already and simplify / streamline some of these things you've highlighted. There's probably more to do, but the base is really there. We can update things in 4.x.

The more I'm looking at it, the more I don't like `OSMPos` though. I think we can keep it a raw `Tuple{Int,Int,Float64}`. Any objections to that?

I need perhaps a focused afternoon on it, but won't have one of those until Tuesday most likely.

Datseris:
> I think we can keep it a raw Tuple{Int,Int,Float64}

Fine by me.

Alright, then I give a final review before merging once you're done. 

We need to mention this in the paper as well, perhaps add a brief integration section?

src/spaces/openstreetmap.jl
Libbum: Yeah. "The distance the agent has travelled between two intersections" is perhaps correct. `osm_road_length` is perhaps not so correct either, maybe `osm_road_length_travelled` is better.

Libbum:
> We need to mention this in the paper as well

For sure. Once this is merged I'll focus on the paper. It's just 320 and 363 otherwise that need to be closed out now I think.

src/spaces/openstreetmap.jl
Libbum: Oh, you're right on this one. It has nothing to do with the agent.

examples/zombies.jl
Libbum: Fastest & shortest is actually the options. I've added the option for both now.

src/spaces/openstreetmap.jl
Libbum: Right, so the reason for this with the current implementation: `pos::Tuple{Int,Int,Float64}` gives us a location somewhere on a road or at an intersection&mdash;fine. OSMX.jl's route planner only uses intersections. It's not possible to plan a route to a specific location on a road. So we make `route::Vector{Int}`, since we can just take a stroll down each of the intersections until the end. To be able to allow an agent to finish somewhere other than an intersection, we need an additional `destination::Tuple{Int,Int,Float64}` property or perhaps something like `route::Tuple{Vec{Int},Tuple{Int,Int,Float64}}` which implies your destination is at `route[2]`, if you follow `route[1]...`. I'll think about the ramifications of this and take a stab at a better design.

Datseris: Is it really so important to make routes that do _not_ end at intersections? Given the complexity that it brings, I'm not sure whether it is _that_ beneficial. Such minute details may not matter for the actual scientific use of an ABM on a map, like e.g. spreading rate or other things, which anyways are calculated over several ABM realizations.

Libbum: It's not really added complexity, just an additional tuple. I'd argue it's definitely worth it if we're actually making this a space rather that an Integration Example.

If I want to measure traffic flows from my franchise of petrol stations that are sitting on highways (i.e. not on intersections) coming in and out of a city, I certainly don't want to have my estimations off by kilometres.

Datseris: Okay, you have a fair point I'm convinced.

Libbum:
I'm happy with the implementation as it stands now. This covers essentially all of OSMX.jl that we'll want to use. The google API stuff we don't expose, but if users want that they can just pull in OSMX.jl themselves and use it in tandem.

Have not updated the main docstring and tests are needed, but I'll wait for comments before doing those.

examples/zombies.jl
```diff
@@ -0,0 +1,104 @@
...
+
+        add_agent!(start, finish, model, false)
+    end
+
+    patient_zero = random_agent(model)
+    patient_zero.infected = true
+    return model
+end
+
+# In our model, agents are seemingly oblivious to the problem, they keep going about their
```
Datseris, line 39:
```suggestion
# In our model, zombies are seemingly oblivious to the problem, they keep going about their
```

src/spaces/openstreetmap.jl
```diff
@@ -0,0 +1,591 @@
...
+Current `pos`ition and `destination` tuples are represented as
+`(start intersection index, finish intersection index, distance travelled in meters)`.
+The `route` is an ordered list of intersections, providing a path to reach `destination`.
+
+Further details can be found in [`OSMAgent`](@ref).
+
+## Routing
+
+There are three ways to generate a route, depending on the situation.
+1. Using [`add_agent!`](@ref)`(start, finish, model)`, which will calculate a route automatically for a new agent.
```
Datseris, line 57: That's a bit confusing for me. You add an agent *and* plan its route? If the agent has extra properties, how are they created? is it with the syntax of extra keywords? 

Perhaps this should be a different function or split into two functions? one that adds the route to the agent? Because if an agent already has a valid route we can use `add_agent_pos!` instead, instead of adding further dispatches to `add_agent!`. It might be a bit complicated if `add_agent!` does too many things.

Datseris: Saw the docstring of add_agent now. Given that `plan_route` also accepts additional keywords, I'd vote to not overload `add_agent!` here for clarity.

src/spaces/openstreetmap.jl
```diff
@@ -0,0 +1,591 @@
...
+overwrite any current route.
+"""
+function osm_random_route!(agent, model::ABM{<:OpenStreetMapSpace})
+    agent.destination = osm_random_road_position(model)
+    agent.route = osm_plan_route(agent.pos, agent.destination, model)
+    return nothing
+end
+
+"""
+    osm_plan_route(start, finish, model::ABM{<:OpenStreetMapSpace}; by = :shortest, return_trip = false, kwargs...)
```
Datseris, line 112: docstring is too large to be rendered in the docs

Datseris: Also, what are `kwargs...`?

src/spaces/openstreetmap.jl
```diff
@@ -0,0 +1,591 @@
...
+For details on how to obtain an OSM file for your use case, consult the OpenStreetMapX.jl
+README. We provide a variable `TEST_MAP` to use as a `path` for testing.
+
+This space represents the underlying map as a *continuous* entity choosing accuracy over
+performance. If your solution can tolerate routes to and from intersections only, a
+faster implementation can be achieved by using the
+[graph representation](https://pszufe.github.io/OpenStreetMapX.jl/stable/reference/#OpenStreetMapX.MapData)
+of your map provided by OpenStreetMapX.jl. For tips on how to implement this, see our
+integration example: [Social networks with LightGraphs.jl](@ref).
+
```
Datseris, line 35: We link to an example for using the light graphs, but we should link to an example that _actually_ uses the space as well, the zombie apocalypse. In fact, I think that in all spaces, after the opening sentence of the docstring we link to an example directly.

Datseris (approved):
I only have documentation comments. I had a concern about the overloading of `add_agent!`, but I'll leave it to you to decide whether you want it this way or not.

src/spaces/openstreetmap.jl
Libbum: > Road speeds are needed for this method which can be passed in via extra keyword arguments. Consult the OpenStreetMapX documentation for more details.

src/spaces/openstreetmap.jl
Libbum: It's just the same as any add_agent! method. But sure, I've dropped that method and edited the example.

Libbum:
@Datseris, recent changes should resolve all open aspects of the review.

I'm having issues with the random portion of the tests. They run fine locally but fail on the actions. Can you pull this down and take a look on your machine? Or suggest what's the best way to cover these random functions.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## New features:
 - `GridSpace` has been re-written from scratch! It now supports **any dimensionality** and is about a **full order of magnitude faster** than the previous version!
 - `ContinuousSpace` has been re-written from scratch! It is now at least 3 times faster!
+- A new, continuous `OpenStreetMapSpace` which lets agents traverse real locations via planned routes.
```
Datseris, line 6:
```suggestion
- A new, continuous `OpenStreetMapSpace` which lets agents traverse real world locations via planned routes based on the Open Street Map initiative.
```

Datseris:
@Libbum other than the tests I don't see something more here I can do, so I leave the rest to you (tag me explicitly if you need something)

Libbum:
Sure. I'm still fighting with them - problem may be that the way that OSMX.jl loads the maps is not deterministic.

There's a bug in the `nearby_ids` I need to fix. Once tests are passing I'll merge.

Libbum:
Damnit. Node ids for each corresponding graph vertex are not reproducable in the OSMX.jl datastructure. So we can't use them as useful position data.
```julia
OSMX: Test Failed at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:76
  Expression: start_idx == map_one.n[start]
   Evaluated: 3625688657 == 3625692040
Stacktrace:
 [1] top-level scope at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:76
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
 [3] top-level scope at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:67
OSMX: Test Failed at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:77
  Expression: start_idx == map_two.n[start]
   Evaluated: 3625688657 == 3625692040
Stacktrace:
 [1] top-level scope at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:77
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
 [3] top-level scope at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:67
OSMX: Test Failed at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:78
  Expression: finish_idx == map_one.n[finish]
   Evaluated: 2985292255 == 580387481
Stacktrace:
 [1] top-level scope at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:78
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
 [3] top-level scope at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:67
OSMX: Test Failed at /home/runner/work/Agents.jl/Agents.jl/test/osm_tests.jl:79
  Expression: finish_idx == map_two.n[finish]
   Evaluated: 2985292255 == 580387481
```

I'll need to re-think some things before we can merge this.

Libbum:
@Datseris, can you take a look at the additions in a705801 and comment if you have concerns.
Essentially all indexes, ids, edges, whatever are not static in the OSMX.jl data struct, so every time you load a map on a different system / julia version etc, you get completely different numbers. This makes testing next to impossible without some additional mechanisms: now implemented; which end up being helpful anyway. Users won't always want randomly positioned agents for example, so they can use lat-lon coordinates.

I think I still have a problem to hunt down in `nearby_ids`, looking at that now.

Libbum:
Once CI finishes, and if you're happy with recent changes, this is ready.

PR #334, opened 2020-10-16
Measurement integration

Libbum:
An integration example of how to use [Measurements.jl](https://github.com/JuliaPhysics/Measurements.jl) to propagate uncertainty.


docs/make.jl
```diff
@@ -90,7 +90,8 @@ pages = [
     "Ecosystem Integration" => [
         "LightGraphs" => "examples/schoolyard.md",
         "DifferentialEquations" => "examples/diffeq.md",
-        "BlackBoxOptim" => "examples/optim.md"
+        "BlackBoxOptim" => "examples/optim.md",
+        "Measurements" => "examples/measurements.md",
```
Datseris, line 94: I wonder, should we put the ending `.jl` in all page headers?

Datseris:
Looks great to me, I give ok for merging.

docs/make.jl
Libbum: Perhaps. I guess it's essentially the convention for repos now. So will do this when I merge.

PR #337, opened 2020-11-03
Implement mixed-ABM collections

### LINKED ISSUE #335: Collecting data in mixed-ABMs is mostly unhandled
I think there's something here we haven't taken into account.

Suppose I have a mixed model of `Union{Wolf, Sheep}` and I want to get a property from `Wolf` that doesn't exist in `Sheep`. I'm not interested in aggregate data at present, but we may need to cover that example too.

Say `wolf.range`. Perhaps I could provide a helper
```julia
w_range(a) = hasfield(typeof(a), :range) ? a.range : missing
```
So my expected collection here would output a range value for each wolf and a missing value for each sheep. The problem is that if the *first* agent in the schedule is a `Sheep`, the dataframe will be constructed with a `Missing` type, not a `Union{Missing,Float64}`. Then, an error at runtime
```julia
┌ Error: Error adding value to column :w_range.
└ @ DataFrames ~/.julia/packages/DataFrames/cdZCk/src/dataframe/dataframe.jl:1301
ERROR: MethodError: convert(::Type{Union{}}, ::Float64) is ambiguous. Candidates:
  convert(::Type{Union{}}, x) in Base at essentials.jl:169
  convert(::Type{T}, x::Number) where T<:Number in Base at number.jl:7
  convert(::Type{T}, arg) where T<:VecElement in Base at baseext.jl:8
  convert(::Type{T}, x::Number) where T<:AbstractChar in Base at char.jl:179
  convert(::Type{T}, obj) where T<:FunctionWrappers.FunctionWrapper in FunctionWrappers at /home/tim/.julia/packages/FunctionWrappers/OAZAk/src/FunctionWrappers.jl:85
Possible fix, define
  convert(::Type{Union{}}, ::Number)
```

The reason why this is dependent on the first value is that we assume the result type of the collection of one agent is the same for all agents. That's true in the single agent case, but should somehow be handled differently in the mixed case.

I don't want to have to change schedulers on the model here; what if I also have a `Sheep` property that isn't in `Wolf`? The same problem would occur.

How should/can we best filter agent properties in this scenario?

Libbum:
Closes #335. 
No new API, just handling mixed-ABMs as they should have been already.

### DIFF
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -58,9 +58,20 @@ two `DataFrame`s, one for agent-level data and one for model-level data.
   If you remove agents during model run, you should modify the aggregating functions.
   *E.g.* instead of passing `mean`, pass `mymean(a) = isempty(a) ? 0.0 : mean(a)`.
 
-  For mixed-models, an additional column `agent_type` will be placed in the output
-  dataframe. In the case that data is needed for one agent type that does not exist
-  in a second agent type, `missing` values will be added to the dataframe.
+* `mdata::Vector` means "model data to collect" and works exactly like `adata`.
+  For the model, no aggregation is possible (nothing to aggregate over).
+
+By default both keywords are `nothing`, i.e. nothing is collected/aggregated.
+
+## Mixed-Models
+
+  For mixed-models, the `adata` keyword has some additional options & properties.
+  An additional column `agent_type` will be placed in the output
+  dataframe.
+
+  In the case that data is needed for one agent type that does not exist
+  in a second agent type, `missing` values will be added to the dataframe. *Note:
+  there is a performance hit when chosing this option.*
 
   Aggregate functions will fail if `missing` values are not handled explicitly.
   If `a1.weight` but `a2` (type: Agent2) has no `weight`, use
@@ -71,11 +82,6 @@ two `DataFrame`s, one for agent-level data and one for model-level data.
   and `a2.weight::Int64` with `adata = [(:weight, mean)]`. To achive this, it's best to
   make a promote function: `fweight(a) = Float64(a.weight)`, then use
   `adata = [(fweight, mean)]`.
-
-* `mdata::Vector` means "model data to collect" and works exactly like `adata`.
-  For the model, no aggregation is possible (nothing to aggregate over).
-
-By default both keywords are `nothing`, i.e. nothing is collected/aggregated.
 
 ### Other keywords
 * `when=true` : at which steps `s` to perform the data collection and processing.
@@ -215,6 +221,7 @@ function init_agent_dataframe(model::ABM{A}, properties::AbstractArray) where {A
             end
             unique!(current_types)
             if length(current_types) == 1
+                current_types[1] <: Missing && error("$(k) does not yield a valid agent property.")
                 types[i+std_headers] = current_types[1][]
             else
                 types[i+std_headers] = Union{current_types...}[]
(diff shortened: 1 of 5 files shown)


src/simulations/collect.jl
```diff
@@ -8,6 +8,16 @@ export run!, collect_agent_data!, collect_model_data!,
...
 get_data(a, f::Function, obtainer::Function) = obtainer(f(a))
 
+get_data_missing(a, s::Symbol, obtainer::Function) =
+    hasproperty(a, s) ? obtainer(getproperty(a, s)) : missing
+function get_data_missing(a, f::Function, obtainer::Function)
+    try
+        obtainer(f(a))
+    catch
+        missing
+    end
```
Datseris, line 18: I didn't have enough time to review this PR normally, but here I need to leave the comment that `try` blocks lead to horrible performance and should not be used in performance-critical code. 

The question is two fold. Is this performance critical code, and is there a non-try way to do it?

Libbum: 1. It's for collections on mixed models that already provide a performance warning, so that's already covered.
2. This is for checking that arbitrary user functions that operate over an arbitrary collection of possible agent types doesn't spit an error. If this isn't the place for a try/catch, then there is no place :smile:

Libbum: In addition: its only ever called in the blocks with `T>:Missing`, so it's essentially only for times that users are fast and loose with sanitation of their data. The docs suggest not doing this.

Datseris:
@Libbum can you run the benchmark suite here please?

Datseris:
Seems good to me, however I would place the documentation of this at the end of the docstring. It is too large and may interrupt the flow.

Libbum:
We took the collection tests out so this isn't really telling us much. I'll have to make something more targeted.

>  I would place the documentation of this at the end of the docstring

This is a discussion for `adata` though. The flow would be interrupted if we had a discussion of adata, then mdata, then keywords, then adata again. It would need to be renamed or something&mdash;any suggestions there?

Libbum:
```julia
function initialise()
  model = ABM(Union{Agent3, Agent4}, GridSpace((10,10)); warn = false)
  fill_space!(Agent3, model, rand())
  add_agent!(Agent4, model, 3)

  agent_step!(agent::Agent3, model) = (agent.weight = rand())
  agent_step!(agent::Agent4, model) = dummystep

  return model, agent_step!
end

function initialise_std()
  model = ABM(Agent3, GridSpace((10,10)))
  fill_space!(Agent3, model, rand())

  agent_step!(agent, model) = (agent.weight = rand())

  return model, agent_step!
end

@benchmark run!(model, agent_step!, 100; adata = [:weight]) setup=((model, agent_step!) = initialise())
# 10.344 ms

@benchmark run!(model, agent_step!, 100; adata = [:weight]) setup=((model, agent_step!) = initialise_std())
# 994.875 μs (master: 962.066 μs)

wpos(a) = a.pos[1] + a.weight

@benchmark run!(model, agent_step!, 100; adata = [wpos]) setup=((model, agent_step!) = initialise())
# 20.756 ms

@benchmark run!(model, agent_step!, 100; adata = [wpos]) setup=((model, agent_step!) = initialise_std()) 
# 868.173 μs (master: 836.269 μs)
```

As we'd expect, this does give us a big hit. Removing `missing`s from the data collection was where we managed to get most of our performance gains in the refactor we did a few months back.

I've changed the notice to a warning and put in these benchmark values so that people aware of the ramifications. 
As you can see though, there's next to no change against master for the single agent type case.

If you're happy with the changes, I don't think there's much more I can do on this. It's as good as it will get I think.

Datseris:
>  As you can see though, there's next to no change against master for the single agent type case.

Great, that is what I was really interested in. Please give me a couple of days to have a closer look, just to be on the safe side and then we merge this!

src/simulations/collect.jl
```diff
@@ -53,6 +63,29 @@ two `DataFrame`s, one for agent-level data and one for model-level data.
...
+  For mixed-models, the `adata` keyword has some additional options & properties.
+  An additional column `agent_type` will be placed in the output
+  dataframe.
+
+  In the case that data is needed for one agent type that does not exist
+  in a second agent type, `missing` values will be added to the dataframe.
+
+  **Warning:** Since this option is inherently type unstable, try to avoid this
+  in a performance critical situation. Performance will depend on specifics, but benchmarks
+  see a 10x slowdown when using `Symbol` matching and 26x for `Function` properties.
```
Datseris, line 77: My only comment actually is that I'd try to keep the docstring as lean as possible. "Performance will depend on specifics, but benchmarks see a 10x slowdown when using `Symbol` matching and 26x for `Function` properties." is something I'd remove.

src/simulations/collect.jl
```diff
@@ -53,6 +63,29 @@ two `DataFrame`s, one for agent-level data and one for model-level data.
...
+  Aggregate functions will fail if `missing` values are not handled explicitly.
+  If `a1.weight` but `a2` (type: Agent2) has no `weight`, use
+  `a2(a) = a isa Agent2; adata = [(:weight, sum, a2)]` to filter out the missing results.
+
+  Data collection will fail if two agent types a property with the same name but different
+  subtypes, and are expected to aggregate with each other. For example `a1.weight::Float64`
+  and `a2.weight::Int64` with `adata = [(:weight, mean)]`. To achive this, it's best to
+  make a promote function: `fweight(a) = Float64(a.weight)`, then use
+  `adata = [(fweight, mean)]`.
+
```
Datseris, line 88: I would remove this entire thing. It is not Agents.jl related but super basic Julia knowledge.

src/simulations/collect.jl
```diff
@@ -145,59 +178,158 @@ is not known.
 """
 collect_agent_data!(df, model, properties::Nothing, step::Int=0; kwargs...) = df
 
-function init_agent_dataframe(model::ABM, properties::AbstractArray)
-    nagents(model) < 1 && throw(ArgumentError("Model must have at least one agent to "*
-    "initialize data collection"))
+function init_agent_dataframe(model::ABM{A}, properties::AbstractArray) where {A<:AbstractAgent}
+    nagents(model) < 1 && throw(ArgumentError(
+        "Model must have at least one agent to initialize data collection",
+    ))
```
Datseris, line 184: For clarity, could you please split up this function into two? Essentially, simply split the `if std_headers` statement to just call one of two lower-level functions.

Datseris:
I'm fine with this and my only comments are about code clarity. If you'd like to address them then merge afterwards.

src/simulations/collect.jl
Libbum: Done for the aggregate call as well.

Libbum:
There are some documentation issues that need to be fixed because of this. I'll fix them tomorrow and merge then.

Issue #336, opened 2020-10-31
Comparison to DynamicGrids.jl


rafaqz:
For fixed grid models like forest fire, DynamicGrids.jl was around two orders of magnitude faster than Agents.jl the last time I checked.

It also has a good range of visualisation tools - REPL, GTK and web outputs. Also live parameter manipulation. It might be nice to mention that in the docs where you are comparing the forest fire model with mesa.

Libbum:
A valid point. We had a comparison before, but removed it once the only cellular automata we had was the game of life. Now that forest fire is back to a static system again, I can re-run some numbers to that end.

Datseris:
@rafaqz have you seen this: https://juliadynamics.github.io/InteractiveChaos.jl/dev/agents/  ? I don't see something of similar complexity or interactivity in the docs. In your readme I did read

> [DynamicGrids...] provides better visualization tools. 

however. What does this mean, exactly?

> Also live parameter manipulation.

Is there a video of this in the docs? I just went through them but I didn't find one.

Datseris:
> DynamicGrids.jl was around two orders of magnitude faster than Agents.jl

Yeap, that is true, I also remember it. Notice however that in version 4.0 we have re-written our `GridSpace` from scarttch, and it now has a design more similar to DynamicGrids.jl. It is still very much clear to me that DynamicGrids.jl will be faster, but by how much we have to check again.

rafaqz:
I only mention the visualisation because you mention it as a positive for mesa in the docs. It's good to see you have some live interfaces working too!

But there's a simple REPLOutput built in, and DynamicGridsGtk.jl, DynamicGridsInteract.jl are image-based live interfaces. You can use ColorSchemes.jl to color the simulation, do multi-grid layouts, and set fonts etc for labels/time counter if you need it. A GifOutput is built-in too. See the gifs in the readme https://github.com/cesaraustralia/DynamicGrids.jl for examples of the output. The interfaces just show the same thing, but as the simulation runs. It can do 100fps pretty easily on a simple simulation.

The InteractOutput has control widgets and start/stop etc like your example. There is a video of it in my JuliaCon talk about half way through, but yeah it should be in the DynamicGridsInteract.jl docs.

[...]

PR #338, opened 2020-11-03
DynamicGrids benchmark

Libbum:
WIP for #336 

@rafaqz, I've built a [forest fire benchmark](https://github.com/JuliaDynamics/Agents.jl/compare/dyngrids?expand=1#diff-082ce23bbeaaf882ab86d9ca1468c9abfd88b9f43d7b4863461ee618cc54e4e8) that should be a direct comparison to our current benchmark implementation (100 steps of [this model](https://github.com/JuliaDynamics/Agents.jl/blob/a8e53693dfc046506894d56937ddd51e9553bcd9/src/models/forestfire.jl)). Can you verify that this is how you would approach this setup?

If so, our runs are pretty close now:

| Agents | DynamicGrids |
|---|---|
|24.802 ms|22.271 ms|

Since our performance improvements have come a long way, we may need to discuss which other aspects of DynamicGrids a user might prefer over what Agents has to offer. There's a lot of points you mention in #336, could you expand on a few of those properties and how they would be useful in the cellular automata cases we're looking at here?

### DIFF
--- a/benchmark/compare/DynamicGrids/Project.toml
+++ b/benchmark/compare/DynamicGrids/Project.toml
@@ -1,0 +1,2 @@
+[deps]
+DynamicGrids = "a5dba43e-3abc-5203-bfc5-584ca68d3f5b"
--- a/benchmark/compare/DynamicGrids/src/DynamicGrids.jl
+++ b/benchmark/compare/DynamicGrids/src/DynamicGrids.jl
@@ -1,0 +1,43 @@
+using DynamicGrids
+using BenchmarkTools
+
+function initialise(; density = 0.7)
+    DEAD, ALIVE, BURNING = 1, 2, 3
+
+    forest = let
+        Neighbors(VonNeumann(1)) do neighborhood, cell
+            if cell == ALIVE
+                if BURNING in neighborhood
+                    BURNING
+                else
+                    cell
+                end
+            elseif cell == BURNING
+                DEAD
+            else
+                cell
+            end
+        end
+    end
+    # Set up the init array and output
+    init = fill(DEAD, 100, 100)
+    for position in CartesianIndices(init)
+        if rand() < density
+            state = position[2] == 1 ? BURNING : ALIVE
+            init[position] = state
+        end
+    end
+
+    return init, forest
+end
+
+function run_forest(init, forest)
+    output = ArrayOutput(init; tspan = 1:100)
+
+    # Run the simulation
+    sim!(output, forest)
+end
+
+a = @benchmark run_forest(init, forest) setup = ((init, forest) = initialise())
+println("DynamicGrids.jl ForestFire (ms): ", minimum(a.times) * 1e-6)
+
--- a/src/models/forestfire.jl
+++ b/src/models/forestfire.jl
@@ -7,7 +7,7 @@ end
 """
 ``` julia
 forest_fire(;
-    density = 0.8,
+    density = 0.7,
     griddims = (100, 100)
 )
 ```


Datseris:
That's **insane**. I guess using TimeseriesPrediction.jl as the codebase was a great choice after all!

rafaqz:
That's some great optimisation work you have done! Does this actually use TimeseriesPredictions.jl? I'm interested in that now.

Probably the output should be defined in setup, instead of the init array. But it only shave a little. If you really want to get competetive you could define the SimData object externally and reuse it, but it wont save that much time.

re context:

We don't actually use DynaicGrids.jl for cellular automata really, it's just the easiest demo. We use it for spatial ecological modelling and scenario planning which is often much messier. 

Sometime we have 12 rules chained together using a few grids with feedbacks between them, like the main gif in the readme. The visualisation is required to check if the parameters and the dynamics actually make sense in together before you use an optimiser, and just manipulating it live is the easiest way. Because the models can get complicated and we swap components around a lot, writing visualisation scripts for every model is painful, automating it makes it a no brainer to just play with model combinations, instead of having to code them all. 

Mostly in practice our real simulations are sparse (not actual sparse arrays just lots of zeros) so `SparseOpt` gives another order of magnitude improvement by just not running things (which you can't do with forest fire). Most of the work at this stage is syntax and making the models cleaner for people to write and compose, and running models of with complex structs in multiple grids for modelling spatial population genetics and species interactions.

The main things we value in DynamicGrids are 
- rule composability and sequecing
- multi-grid models
- grids of arbitrary types (se.g. structs and staticarrays)
- automatic interface and optimisation setup
- optimisation for sparse simulations
- optimisation for combining models in chains without array writes between them
- inclusion of arbitrary external data sources into models in a structured way, so you can e.g. switch spatial locations but use the same model object. Things like climatic data.



We also need to compose rules that do arbitrary array writes to any cell with neighborhood models like forest fire.

rafaqz:
By the way I'm not sure these implementations are what I expected, there are usually randomisations in each step where the forest randomly grows or burns? That will be much slower in both cases - and probably show some of the real differences as Agents.jl will have to allocate a lot, and DynamicGrds.jl wont.

Which also means this isn't a stochastic model... I've got a GPU optimisation for deterministic models in that might smoke both of these ;)

rafaqz:
It would be good to benchmark models with this kind of complexity:

```julia
using Dispesal.jl, DynamicGrids.jl, GeoData.jl

# ... loading lots of data, a MWE is hard...
carrycap = floattype(1e9)
growth = ExactLogisticGrowthMap{:population,:population}(
    layerkey=Val(:growthrates),
    carrycap=carrycap,
    timestep=Day(1),
    nsteps=floattype(1.0),
);
growth

### Local dispersal
λ = floattype(0.0125)
radius = 1

@time hood = DispersalKernel{radius}(
    formulation=ExponentialKernel(λ),
    distancemethod=AreaToArea(30),
    cellsize=floattype(1.0)
)
localdisp = InwardsPopulationDispersal{:population,:population}(hood)
log.(hood.kernel) |> heatmap
savefig("output/dispersal_kernel.png")


### Allee effects
allee = AlleeExtinction{:population,:population}(
    minfounders=floattype(22.0)
);

### Human Dispersal
humandisp = HumanDispersal(; human_pop=parent(aus_gdp); 
    scale = 10
    read=:population,
    scale=scale,
    human_exponent=floattype(1.075),
    dist_exponent=floattype(1.429),
    dispersalperpop=floattype(8.703e-9),
    max_dispersers=floattype(3.264e4),
    nshortlisted=100
)

# Define ruleset and output
ruleset = Ruleset(
    aus_humandisp, Chain(localdisp, allee, growth);
    timestep=timestep,
    opt=SparseOpt(),
);
output = ArrayOutput((population=aus_populationgrid,);
    mask=boolmask,
    aux=(; growthrates=growthrates,),
    tspan=DateTime(2020, 1):timestep:DateTime(2026, 1)
)
```

rafaqz:
This is a visualisation and gif output of the model in the benchmark:

```julia
using DynamicGridsGtk
init, forest = initialise()
output = GtkOutput(init; tspan=1:100, minval=0, maxval=2)
sim!(output, forest)
savegif("deterministic_forest_fire.gif", output)
```

Mostly there would be no writes happening in your IBM model as there are few cells changing. That's are large part of why this version is fast for Agents.jl. DynamicGrids.jl could do a faster implementaion of this specific model using `ManualRule`, to reduce writes.

deterministic_forest_fire

Libbum:
Thanks for all the info @rafaqz. I agree, it'd be great to highlight a lot of these inter-comparisons of the Julia ecosystem. We've almost finished a comparison paper of Agents.jl with a number of leading ABM frameworks&mdash;all of which use this simple mechanistic forest fire implementation rather than the real one ([previous version](https://github.com/JuliaDynamics/Agents.jl/commit/c025f2fe8b0b20375b6c55c425558ac6c1d8f9ad#diff-0e608539e63b21e473cfdf4cd1bf013d3de2be658c0a94da0ee1c8ac7c9f62b0)). I'll take a look at these suggestions soon and come back to you with better benchmarks.

rafaqz:
Ok, it's interesting that everyone benchmarks that. Fitting out models is only performance limited when there is stochasticity. But I'll put some time into implementing it if its a standard. There are a few ways to make that faster in DynamicGrids.jl, two that reduce how much work is being done, and brute force with a GPU.

Libbum:
I wouldn't say it's standard. In fact I'm not really happy keeping our defaults like this, so don't take the effort. Now that we have clear comparisons with those frameworks I'm thinking of reverting our model back. [The original model is pretty clear](https://en.wikipedia.org/wiki/Forest-fire_model), and I struggle to see why this simple case is the default for Netlogo and Mesa.

rafaqz:
I'm actually keen to write it up now in DynamicGrids.jl because of how much faster it could run than this benchmark. There are a few minor optimisations that should be easy, but aren't currently. It will force me to make them easy to do.

Datseris:
@rafaqz 

> Does this actually use TimeseriesPredictions.jl?

Not exactly, but the code is based on that code. The gist is that we create pre-initialized sturctures that know what indices they should be searching for given a radius, you can see this here:

https://github.com/JuliaDynamics/Agents.jl/blob/master/src/spaces/grid.jl#L84-L162

rafaqz:
Ok some benchmarks on and observations on this. There were a lot of other opdates for DG before I could merge the the generalised `SparseOpt` method.

1. DG Forest fire is an order of magnitude faster using either `SetNeighbors` or regular `Neighbors` using `SparseOpt`
to avoid running the parts that have no burning cells
2. The general time-constant `Neighbors` method used before is now 2.5x faster as well 
3. Forest fire is only active in < 1% of the grid. The time constant method here is 2.5x faster than Agents.jl but its running every neighborood in the whole grid! With dense models it will take the same amount of time. So, we really should benchmark something with more going on, or where we compare with varying amounts of activity.  Thats pretty much why DynamicGrids.jl exists. I included 3-4 life as an example below.
4. Turns out DynamicGrid.jl is ~2x faster than DSP.jl for raw small-kernel convolutions - so anything like that, like a dispersal kernel model, will probably be many orders of magnitude faster in DynamicGrids.jl.   https://github.com/JuliaDSP/DSP.jl/issues/355#issuecomment-725257524

```julia
# Forest Fire ##################################################
using DynamicGrids, BenchmarkTools
const DEAD, ALIVE, BURNING = 1, 2, 3

function setup(rule, opt, density, size) 
    init = fill(Int8(DEAD), size...)
    for position in CartesianIndices(init)
        if rand() < density
            state = position[2] == 5 ? BURNING : ALIVE
            init[position] = state
        end
    end
    output = ResultOutput(init; tspan=1:100)
    rs = Ruleset(rule; opt=opt, padval=DEAD)
    simdata = DynamicGrids.SimData(DynamicGrids.extent(output), rs)
    return output, rs, simdata
end

function forest_setneighbors(; size=(100, 100), density=0.7, opt=NoOpt())
    forest = SetNeighbors(VonNeumann(1)) do grids, neighborhood, I, state
        if state == BURNING
            for P in positions(neighborhood, I)
                @inbounds grids[P...] = (DEAD, BURNING, BURNING)[grids[P...]]
            end
            @inbounds grids[I...] = DEAD
        end
        return nothing
    end
    setup(forest, opt, density, size)
end

function forest_neighbors(; size=(100, 100), density=0.7, opt=NoOpt())
    forest = Neighbors(VonNeumann(1)) do neighborhood, cell
        if cell == ALIVE
            BURNING in neighborhood ? BURNING : cell
        else
            DEAD
        end
    end 
    setup(forest, opt, density, size)
end

# Timings on a lenovo T450 laptop, similar performance to initial results.

### Outwards/neighbor setting model ###
# Write-to-neighbors approach similar to what Agents.jl does. Not sure why I didn't 
# use this the firest time! except the nice SetNeighbors syntax wasn't available back then.
# It's fast in this case, but the approach doesn't scale well to higher activity models
o, rule, sd = forest_setneighbors()
@btime sim!($o, $rule; simdata=$sd);
#  3.559 ms (798 allocations: 59.31 KiB)

### Inwards neighborhood model with SparseOpt that ignores DEAD ad ALIVE regions ###
# Had to generalise SparseOpt for this, it could only ignore zeros previously
o, rule, sd = forest_neighbors(; opt=SparseOpt(in((DEAD,ALIVE))))
@btime sim!($o, $rule; simdata=$sd);
# 1.838 ms (798 allocations: 76.45 KiB)

### Inwards/neighbor getting model ###
# This is the same as before but faster after some performance work
o, rule, sd = forest_neighbors()
@btime sim!($o, $rule; simdata=$sd);
#  9.798 ms (798 allocations: 59.31 KiB)

# Visualise these
using DynamicGridsGtk
output = GtkOutput(DynamicGrids.init(o); tspan=1:100, fps=50, minval=DEAD, maxval=BURNING)
sim!(output, rule);
```

I don't have your actual code to compare this to (could you post it?) but I got very similar timings to you for the first model you posted ~(23ms) so this is probaby comparable.

Can Agents.jl run game of life? 3-4 life remains active so may be a good test of situations where you have to check most or all of the neighboroods all of the time. Of course you can optimise GOL to be 1000x faster that this, but this is just the generic `NeighborhoodRule/Moore` buffer used in the last forest fire model, not some hashed cache thing.

```julia
# 3-4 Life
init = rand(Bool, 100, 100)
life_34 = Ruleset(Life(born=(3, 4), survive=(3, 4)))
output = ResultOutput(init; tspan=1:100)
@time sim!(output, life_34);

@btime sim!($output, $life_34);
#  7.457 ms (581 allocations: 88.89 KiB)

# You can just view the simulations in the terminal with this:
reploutput = REPLOutput(init; tspan=1:500, fps=500, color=:blue, style=Braile())
sim!(reploutput, life_34)

⠸⢤⢯⠿⡿⣍⢀⢠⠤⡬⡞⣇⠓⡿⣄⡄⢦⣈⡔⠴⠟⠤⠢⠒⢿⡰⣯⣧⠬⢢⣠⣶⠍⠀⠉⠺⠶⠁⠠⠁⠀⣠⡴⠚⠓⣫⠋⠀⣄⠆
⠀⡮⣗⡓⡹⢿⣆⣷⢉⡫⠘⠺⣉⠦⣦⡀⠀⡴⢩⡹⣟⣌⠧⡀⣨⡁⠀⠛⠶⣤⠈⠙⠁⢀⣼⠧⢱⣯⢾⡠⡠⠀⣀⢤⣲⢱⢶⠒⠀⢘
⣲⣟⣉⢸⣼⣰⡟⢮⣭⣅⡤⠾⠲⣀⣰⢖⢶⢿⠁⠀⣺⡉⠙⢡⣄⡄⠄⠀⠲⢕⣗⢘⢏⣿⡲⢤⡈⣤⠀⠋⣦⠤⣽⠔⣶⠒⣤⠭⣄⡾
⢟⣆⡭⢃⡄⣖⣲⠂⠰⢛⠈⠘⣼⡋⣵⢶⡸⡧⠊⢜⡉⢼⠃⠭⠽⢯⠤⠷⠅⢣⡄⢇⡀⠀⠡⠅⢖⡲⡕⠶⣶⠃⢊⣲⠩⣢⢞⠅⢃⠃
⡘⣽⠅⡬⡢⡐⢗⣶⣒⢤⠷⢰⡋⢉⠍⠾⣪⡶⠆⡤⠮⢾⣋⠈⣱⡶⡁⠺⣥⠌⡜⠀⠈⠑⡶⣷⡊⢽⢩⠀⠘⠼⣤⣄⡊⢯⢲⠁⠀⠼
⠷⡰⢮⣁⣹⢷⣌⣨⣀⡀⠀⢿⢘⡫⢭⡿⢽⣯⣰⠔⡔⡆⢩⠂⠉⢪⡀⠀⠹⠼⠇⠀⢤⣴⡒⣢⣬⠹⠑⣂⡌⢤⠴⡽⡿⠍⡾⣒⣯⡠
⢨⠈⠘⡇⣀⣤⣜⢼⡞⡀⠞⣾⣦⠤⢁⢻⢻⠑⠱⠀⢀⡗⠛⡇⡀⠑⢛⣃⡄⣈⡭⠦⠽⢟⠀⠉⢀⢄⢄⠰⠸⡞⢨⢺⠟⠰⢡⣁⠰⣼
⡆⢠⠒⠛⢗⣚⢊⣴⠓⠹⢰⡴⡍⠈⠹⣾⡴⠓⢢⣐⢺⣷⡖⣈⠰⠆⢆⣿⣰⢌⡭⢷⣤⡝⡄⣀⣧⠔⠛⣤⣌⣢⠸⣾⢴⡇⡄⢬⠁⡇
⣰⣋⣷⡲⢿⡷⣀⡀⠱⠀⠈⢇⣹⢢⡄⠛⢣⡱⣖⡀⠧⣿⡓⠐⠀⠀⢀⠥⡄⠁⢞⢨⠊⢄⠃⠳⣅⡤⣲⣁⣐⣼⣶⢯⡷⢇⣁⠷⠗⡗
⣚⣛⡸⠃⢺⢷⠀⠃⣠⣼⠂⡈⣈⣬⢀⠉⠃⠈⣏⠁⠀⢯⠙⡂⣄⢔⠍⠉⠝⢢⣜⣫⣆⢿⣶⣠⠶⠺⣤⡸⡻⠟⣛⣸⠵⢾⢟⠒⢛⢲
⠊⣅⢽⢉⠒⠄⣄⣴⠹⢐⢶⢢⢛⢝⢄⠈⠀⢠⢀⢢⣴⡃⣮⣸⣣⠉⠹⡀⢸⡉⠼⢁⢄⠴⢬⡹⣠⡄⠀⠘⣾⠁⠉⢝⢁⡸⣾⡴⠘⢸
⠀⡇⡈⠯⠇⡆⢖⣂⢠⡷⠌⡞⠘⠟⠁⠭⠃⣒⢤⡥⢤⣭⣀⠈⣿⣀⣀⣈⠉⠈⠙⣁⢸⠋⠀⠝⠄⡍⠙⠱⠒⣦⡊⢻⠬⣵⢿⣡⠑⡇
⠀⣌⠝⢁⠤⠤⢂⠈⠯⡃⣈⡧⡅⠄⢺⡵⠁⠾⢸⣎⡥⠴⣝⣂⠈⠐⢘⠋⠃⡆⡒⠓⠚⠀⢀⢀⠾⠗⠙⠉⡶⢊⠑⢂⡛⠑⣕⠎⠠⡏
⠎⣵⡆⣼⠀⢀⢈⡀⢐⣯⢉⣁⣣⣸⠘⠑⠚⠻⢿⣾⢿⣸⢣⠤⣆⢼⠮⢤⠯⠬⢤⡤⠒⠶⣪⠟⠘⡳⠦⣄⡅⠈⣦⡀⣵⣬⡅⡤⠶⠞
⣨⠽⠄⢃⢰⢮⡏⠙⠷⡂⢈⠝⠋⣩⣄⠔⠀⢠⡉⠀⠸⣿⢬⠳⢧⠴⢠⣪⢍⠠⠌⣇⠤⠨⠘⢃⣷⣾⠀⠁⡅⠀⢸⡌⢐⣏⡘⠇⠃⣂
⡇⠧⣤⣶⣉⡬⡇⠠⠞⠰⣔⡇⠵⠶⣏⣚⢛⡆⢊⣛⠄⠹⣤⠴⢰⢃⢒⡌⠁⠒⠊⣹⢷⠫⢍⣄⣛⢧⠄⢖⢷⢣⣧⡪⠇⠢⢦⠄⠬⣭
⡕⡅⠝⡮⠅⠘⠣⢶⣔⠀⠀⡏⡜⡏⣱⡂⡺⢕⡍⠀⣤⣠⣢⡂⡬⡛⡋⠿⢥⠖⠉⠡⠂⢠⡴⣮⠉⡰⠶⡸⢸⠒⡀⠊⠀⠦⡣⢿⡤⣷
⣣⡅⣤⣅⢀⡄⡀⠈⠘⠘⠃⣅⢁⡀⡟⣒⠀⠈⡟⢠⠈⠉⣉⠿⠡⡍⣶⠚⡳⢼⣋⡍⠉⢰⣇⡁⠃⠁⢀⠜⠘⡖⢹⢠⢀⢡⣟⡨⠂⠹
⠀⡹⠃⡿⠛⣭⣽⠃⢪⢊⠂⣓⠨⠈⠁⠹⡷⡂⠉⣚⡗⣛⣖⠗⠊⣆⢀⢀⠛⠹⡘⡕⠆⠀⡜⢀⡶⠆⠶⠁⠀⠨⣐⡸⢰⣝⠾⠲⣲⡼
⠀⢿⢊⡾⢙⡗⡀⠙⡃⡁⠩⣍⣱⣭⣄⣼⢀⡇⠀⢠⠓⠋⠷⠝⣋⠫⢓⣣⣲⣯⠬⠅⠀⡜⢋⣓⠲⢚⠗⣀⠀⡘⠀⡇⢈⣕⣽⣛⣁⠁
⠀⠈⣛⢆⡤⡃⣽⠈⡴⣽⡘⣕⣲⠿⠙⡉⠨⢫⡴⡼⣉⢤⢀⡈⣱⡆⡶⡹⠋⠡⢨⠀⢚⣍⣉⡰⣣⡈⡀⠺⣫⠀⠘⠁⠾⡤⡭⡭⣒⠆
⠀⢠⡌⢱⢩⡴⣤⣾⢢⠹⢄⣮⢹⣀⣐⣴⡬⠈⠘⢒⣯⠛⠀⢨⡭⣤⡼⡗⠋⢨⣿⠒⠀⢹⡀⣀⠜⢩⢛⠤⣀⢚⢲⣢⣉⠙⢚⡪⡽⠁

**Comment by rafaqz** [comment]
You can also define 34-life like this so the benchmark isn't obscured by the package `Life` method. It also 
nearly twice as fast because the states are `const`! (but I'll make them types in the package to match it)

```julia
using DynamicGrids, BenchmarkTools
init = rand(Bool, 100, 100)
const sum_states = (false, false, false, true, true, false, false, false, false), 
                   (false, false, false, true, true, false, false, false, false)
life = Neighbors(Moore(1)) do hood, state
    sum_states[state + 1][sum(hood) + 1]
end
output = ResultOutput(init; tspan=1:100)
@btime sim!($output, $life);
#  4.429 ms (581 allocations: 70.09 KiB)
```

That is literally all the code to define and run GOL

Libbum:
Thanks @rafaqz. I've been off work so haven't been able to chase this up. Unsure how quick I'll be able to get back to you to be honest. That being said: 34-life should be fine to benchmark if you think it provides a better showcase of DynamicalGrids. I'll look into that once I'm up to it.

rafaqz:
No worries. I can put it together too if you don't have time, I noticed you had a life example already written up in the docs after I posted that. I have to write some benchmarks for a paper anyway.

DymicGrids has a parallel/GPU version on the way now - this is the original brute force version or forest fire 
with a 1000 * 1000 grid, my naive first attempt on a Gtx 1080:

```julia
o3, gpu_rule, sd3 = forest_neighbors(; opt=GPUopt{32}(), size=(1000, 1000))
@btime sim!($o3, $gpu_rule; simdata=$sd3)
# 27.634 ms (23552 allocations: 1.09 MiB)
```

Similar time as before, 100x larger grid! It wont do as much better on 100x100 due to setup costs, at 1.7ms

Libbum:
Closing this in favor of the solutions discussed in #336

PR #341, opened 2020-11-08
Random agent with condition

Datseris:
This was asked in the Julia slack and it's simple to add.

However, there are two ways to do it: first, you allocate a random permutation of the agent IDs, and iterate through them until you find condition. This way, you can guarantee that you check all agents once, and if no agent satisfies condition, you return `nothing`.

The other way is what I've written. It doesn't allocate, and thus is faster, but if no agent in the model satisfies the condition, then the function will never terminate.

Which one is best?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Performance increase of finding neighbors in GraphSpace with r > 1.
 - New wrapping function `nearby_agents` that returns an iterable of neighboring agents.
 - New public `schedule` function for writing custom loops.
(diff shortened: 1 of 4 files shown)


Libbum:
Hard choice. Maybe both should be implemented, and users can chose the unsafe method if they know they will always satisfy a condition at least once. Perhaps put in the docstring that a quick one-time check for this (assuming no further mutation) is `any(condition, allagents(model)) == true`.

Datseris:
I've decided to go with the safe option. If a user wants a performant version they should probably code it themselves, it is 4 lines only after all.

Issue #346, opened 2020-11-25
Selecting specific neighbours (biased random walks)


mbruna:
I am trying to set up a simple exclusion process on a 2D grid space where the jumps to neighboring cells are only allowed when unoccupied and the rates are asymmetric (right, up, left, and down moves do not have the same probability and depend on an internal variable "angle" of the agent).
At the moment I am having trouble identifying the appropriate neighbor node from the node_neighbors list, in particular when one of the neighbors falls on the other side of the domain (due to periodicity). 

I am attaching my agent_step! function.

Thanks so much.

ps: I am starting with Julia and Agents.jl, so please excuse any silly mistakes I made in the attached code (I'd also be very keen to hear of any improvements to make the code more efficient).

[agent_step.txt](https://github.com/JuliaDynamics/Agents.jl/files/5597392/agent_step.txt)

Libbum:
Hi @mbruna!

Yes, this is a problem we've been working on previously (#294, #295) and I think has fallen by the wayside a little since we've been doing some major changes for an upcoming 4.0 release.

I'm pretty sure it's only due to the periodicity problem and *may* be fixed with the new `GridSpace` in the master branch, but I don't remember checking explicitly.

Can I ask what version of Agents you're using? 3.7.1 or the current master branch?

Libbum:
(contents of agent_step.txt)
```julia
function agent_step!(agent, model)

    # Compute probabilities of 4 spatial moves (1- right, 2- up, 3-left, 4-down)
    πj = [exp(cos(agent.theta)), exp(sin(agent.theta)), exp(-cos(agent.theta)), exp(-sin(agent.theta))]
    # Wiener process for the angles
    agent.theta += sqrt(2*model.DR*model.dt)*randn()

    # Move agents in space

    # check if agent will move (generate random number between 0 and 1):
    r = rand()
    # since a no-move is always the most likely, let's start with that
    if (r >= sum(πj))
        # no move
        return
    end

    # if I made it here I will move (provided the direction I pick is empty)
    neighbor_cells = node_neighbors(agent, model)

    # For unbiased diffusion we'd do:
    # target = rand(neighbor_cells)

    # For the biased diffusion I need to be able to select in an efficient way specific neighbors (also accounting for periodic boundary conditions)
    if r < πj[1]
        # move to right
        target = # this should be the neighbour to the right of agent
    elseif r>= πj[1] && r< πj[1] + πj[2]
        # move up
[...]

PR #347, opened 2020-11-25
[WIP] Standarise neighbor directions

Libbum:
A few failing tests to confirm #346 is still an issue in the new `GridSpace`.

@Datseris: I'm wondering if we should look at ordering the neighbors when we do the periodic wrap (which will require a performance hit), or should we write a helper function (or overload to nearby_positions) which returns particular values?

For example we could do `nearby_positions(agent, model, directions = [:N, :NW])` or a new helper `nearby_direction(:E, agent, model)`.

### DIFF
--- a/test/space_test.jl
+++ b/test/space_test.jl
@@ -160,6 +160,20 @@ end
     @test sort!(collect(nearby_ids((2, 2), gridspace))) == [1, 2]
     @test collect(nearby_ids(a, gridspace)) == [2]
 
+    # Make sure directions always make sense when periodic
+    model = ABM(Agent3, GridSpace((3, 3); metric = :euclidean))
+    centre = add_agent!((2,2), model, rand())
+    offset = add_agent!((1,1), model, rand())
+    # S, W, N, E
+    @test collect(nearby_positions(centre, model)) == [(2, 1), (1, 2), (3, 2), (2, 3)]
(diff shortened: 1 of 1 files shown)


Datseris:
I must admit I haven't understood the problem :( so I'll take a step back.

First, is the `nearby_pos` function wrong in the case of periodicity? or does it still find correctly the nearby locations even near the edges?

Libbum:
Positions are correct, it's just that the order is essentially different depending on how the wrapping occurs. Therefore, it's not really that simple to know which nearby position is north of current or south east or wherever.

Tests show, for a 3x3 grid, you can put an agent in the centre (2,2) and the positions will return in the order South, West, North, East.
Put an agent at (1, 1): the correct (periodically wrapped) cells are returned, but the order is South, West, East, North.

Datseris:
Yes that is true the code i've written does no ordering. Ordering only makes sense for 1-radious neighborhood, for higher it doesn't. The code I've written is generic for any `r` and thus doesn't do this ordering. You'd have to make an explicit clause for `r==1` to get that...

(but I think this is not a priority now)

Libbum:
A fair point. It will be straightforward to make a stop gap set of functions for @mbruna's case for the moment. I'll do that now and worry about the general case later.

Libbum:
Closing, since these cases are out of scope for `nearby_positions`. I'll work on a solution for #346 in another way.

PR #349, opened 2020-11-28
Walk utility

Libbum:
This is a solution to the discussion around #346.
Users coming from Netlogo will be used to [Turtles](https://ccl.northwestern.edu/netlogo/docs/programming.html#procedures), which have been historically used in a number of other places to represent movement on a grid. 

I've simplified the concept a bit here: rather than having a heading and a direction (which works fine for us right now in `ContinuousSpace`, and doesn't require any extra help from the framework), the `walk!` function moves agents on 2D grids, respecting boundary conditions where necessary. 

This should solve @mbruna's initial question, as well as providing a solution for future netlogo converts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - New public `schedule` function for writing custom loops.
 - Mixed models are supported in data collection methods.
 - `random_agent(model, condition)` allows obtaining random agents that satisfy given condition.
+- A walk! utility for 2D `GridSpace`s, providing turtle-like agent movement.
 
 ## Breaking changes
 Most changes in this section (besides changes to default values) are deprecated and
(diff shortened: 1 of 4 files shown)


Datseris:
my complaint here is that the code is too complex and repetitive. this should be in principle easy to do with a couple of functions.

Why not translate all directions to vectors that are added to position e.g. north bcomes `(0, 1)` and southeast becomes (1, -1)` and then have one single function that just adds this vector to the position and takes the modulo if it is periodic?

Datseris:
you can still have `North` `South` etc. but you have a function like `tovector(::North) = (0, 1)`.

Libbum:
Not a bad idea. I'll adjust accordingly.

Datseris (approved):

Datseris:
awesome, simple code makes me happy

PR #350, opened 2020-11-29
Change step! and run! function order

### LINKED ISSUE #331: Multiple, successively executed `agent_step!` functions during each step
How could one add multiple `agent_step!` functions for each step, so that the order of execution would be `agent_step_1! -> agent_step_2! -> model_step!`? The idea behind this is simply to have all scheduled agents act once, then have all scheduled agents act a second time again, then let the model step happen once.

I can of course think of workarounds for this, e.g. wrapping `agent_step_1!` and `agent_step_2!` within the `model_step!` function and call them from there (providing some custom built list of agents). However, I'm wondering whether there's an option for this built into the API of Agents.jl?

Libbum:
Fixes #331, assuming all of the docs work and tests pass.

Things to do or at least open for discussion:

- `interactive_abm` should be updated to the new ordering as well when this is released
- Schelling is our goto first model, this change kind of makes the `step!` look complicated, since we must provide the `dummystep`. Can we think of a way to either discuss this, or change something here to make the first model approachable? Stepping in a number of our example models are no longer as 'clean' it seems.
- An example using a more complex model structure would be nice. @fbanning, perhaps you could suggest something here?
- I did not want to introduce the agent_step separation as a legacy utility in the tutorial (which is essentially the only place we talk about step!). There needs to be some second, more advanced place to discuss this matter.

I'd suggest we add something like an UPGRADING.md file, which will give a run through of the changes needed to go from a 3.7 model, to a 4.0 one, but I'll leave that for a different PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ therefore are not "truly breaking".
 - Default arguments for `GridSpace` are now `periodc = false, metric = :chebyshev`.
 - Internal structure of the fundamental types like `ABM, GraphSpace`, etc. is now explicitly not part of the public API, and the provided functions like `getindex` and `getproperty` have to be used. This will allow performance updates in the future that may change internals but not lead to breaking changes.
 - `vertex2coord, coord2vertex` do not exist anymore because they are unnecessary in the new design.
+- **`step!` and `run!` argument order has changed**. Now call `step!(model, model_step!, agent_step!, 1)`.
 - API simplification and renaming:
   - `space_neighbors` -> `nearby_ids`
   - `node_neighbors` -> `nearby_positions`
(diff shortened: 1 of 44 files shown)


Datseris:
Tim, I think you made things unecessarily breaking. Why make such a massive change where you change the order of agent_step and model_step, where the only change that us truly necessary is in the case where only 1 stepping function it is assumed that this function is the model_step. 

There is no reason to change the order of the agent & model step functions. The only thing necessary is to make the following call signature:

```
step!(model, [agent_step! ,] model_step!, n)
```
whose only difference from the existing one is where the optional brackets are located:
```
step!(model, agent_step!, [model_step!, ] n)
```

This breaks no code that was already using both agent_ and model_ step.

Datseris:
This PR changes 43 files I think it will be simpler to start a new one... The only files that need change are the `run` and `step` definitions and example models that have been using only `agent_step`.

Libbum:
Yep, makes sense. Sorry&mdash;still got a lot of brain fog going on.

PR #351, opened 2020-11-29
Swap single function stepping method to model_step as default

### LINKED ISSUE #331: Multiple, successively executed `agent_step!` functions during each step

Libbum:
A simpler (saner) implementation of #350, that fixes #331.

Still to consider:
-  An example using a more complex model structure would be nice. @fbanning, perhaps you could suggest something here?
- I did not want to introduce the agent_step separation as a legacy utility in the tutorial (which is essentially the only place we talk about step!). There needs to be some second, more advanced place to discuss this matter. (Ref: [this comment](https://github.com/JuliaDynamics/Agents.jl/issues/331#issuecomment-734314620))

I'd suggest we add something like an UPGRADING.md file, which will give a run through of the changes needed to go from a 3.7 model, to a 4.0 one, but I'll leave that for a different PR.

### DIFF
--- a/examples/wealth_distribution.jl
+++ b/examples/wealth_distribution.jl
@@ -54,7 +54,7 @@ N = 5
 M = 2000
 adata = [:wealth]
 model = wealth_model(numagents = M)
-data, _ = run!(model, agent_step!, N; adata = adata)
+data, _ = run!(model, agent_step!, dummystep, N; adata = adata)
 data[(end - 20):end, :]
 
 # What we mostly care about is the distribution of wealth,
@@ -117,7 +117,7 @@ Random.seed!(5) # hide
 init_wealth = 4
 model = wealth_model_2D(; wealth = init_wealth)
 adata = [:wealth, :pos]
-data, _ = run!(model, agent_step!, 10; adata = adata, when = [1, 5, 9])
+data, _ = run!(model, agent_step_2d!, dummystep, 10; adata = adata, when = [1, 5, 9])
 data[(end - 20):end, :]
 
 # Okay, now we want to get the 2D spatial wealth distribution of the model.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ therefore are not "truly breaking".
 - Default arguments for `GridSpace` are now `periodc = false, metric = :chebyshev`.
 - Internal structure of the fundamental types like `ABM, GraphSpace`, etc. is now explicitly not part of the public API, and the provided functions like `getindex` and `getproperty` have to be used. This will allow performance updates in the future that may change internals but not lead to breaking changes.
 - `vertex2coord, coord2vertex` do not exist anymore because they are unnecessary in the new design.
-- **Default stepping function has changed.** `run!` and `step!` now expect at least `model_step!`, with `agent_step!` as an optional argument.
+- **Default stepping function has changed.** `run!` and `step!` now expects (at least) `model_step!`, with `agent_step!` as an optional argument.
 - API simplification and renaming:
   - `space_neighbors` -> `nearby_ids`
   - `node_neighbors` -> `nearby_positions`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ therefore are not "truly breaking".
 - Default arguments for `GridSpace` are now `periodc = false, metric = :chebyshev`.
 - Internal structure of the fundamental types like `ABM, GraphSpace`, etc. is now explicitly not part of the public API, and the provided functions like `getindex` and `getproperty` have to be used. This will allow performance updates in the future that may change internals but not lead to breaking changes.
 - `vertex2coord, coord2vertex` do not exist anymore because they are unnecessary in the new design.
+- **Default stepping function has changed.** `run!` and `step!` now expect at least `model_step!`, with `agent_step!` as an optional argument.
 - API simplification and renaming:
   - `space_neighbors` -> `nearby_ids`
   - `node_neighbors` -> `nearby_positions`
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -52,14 +52,16 @@ For more functions visit the [API](@ref) page.
 ## 4. Evolving the model
 
 Any ABM model should have at least one and at most two step functions.
-An _agent step function_ is required by default.
-Such an agent step function defines what happens to an agent when it activates.
-Sometimes we also need a function that changes all agents at once, or changes a model property. In such cases, we can also provide a _model step function_.
-
-An agent step function should only accept two arguments: first, an agent object, and second, a model object.
-
+A _model step function_ is required by default.
 The model step function should accept only one argument, that is the model object.
-To use only a model step function, users can use the built-in [`dummystep`](@ref) as the agent step function.
+
+Optionally, we can also provide an _agent step function_, which defines what happens to
+an agent when it activates. An agent step function should only accept two arguments:
+first, an agent object, and second, a model object.
+
+Sometimes models are simple enough that changing agent properties is all that's required.
+In these cases, users can use the built-in [`dummystep`](@ref) as the model step
+function, to ignore any dynamics here.
 
 After you have defined these two functions, you evolve your model with `step!`:
 ```@docs
--- a/examples/diffeq.jl
+++ b/examples/diffeq.jl
@@ -343,7 +343,7 @@ import DiffEqCallbacks
 function fish!(integrator, model)
     integrator.p[2] = integrator.u[1] > model.min_threshold ?
         sum(a.yearly_catch for a in allagents(model)) : 0.0
-    Agents.step!(model, agent_cb_step!, 1)
+    Agents.step!(model, agent_cb_step!, dummystep, 1)
 end
 
 function fish_stock!(ds, s, p, t)
--- a/examples/flock.jl
+++ b/examples/flock.jl
@@ -126,7 +126,7 @@ cd(@__DIR__) #src
 model = initialize_model()
 e = model.space.extent
 anim = @animate for i in 0:100
-    i > 0 && step!(model, agent_step!, 1)
+    i > 0 && step!(model, agent_step!, dummystep, 1)
     p1 = plotabm(
         model;
         am = bird_triangle,
--- a/examples/forest_fire.jl
+++ b/examples/forest_fire.jl
@@ -71,12 +71,12 @@ nothing # hide
 
 # ## Running the model
 
-step!(forest, tree_step!, 1)
+step!(forest, tree_step!, dummystep, 1)
 count(t->t.status == :burnt, allagents(forest))
 
 #
 
-step!(forest, tree_step!, 10)
+step!(forest, tree_step!, dummystep, 10)
 count(t->t.status == :burnt, allagents(forest))
 
 # Now we can do some data collection as well using an aggregate function `percentage`:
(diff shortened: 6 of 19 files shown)


Datseris:
I'll take some time here to re-write parts of the tutorial.

examples/wealth_distribution.jl
```diff
@@ -117,7 +117,7 @@ Random.seed!(5) # hide
 init_wealth = 4
 model = wealth_model_2D(; wealth = init_wealth)
 adata = [:wealth, :pos]
-data, _ = run!(model, agent_step!, 10; adata = adata, when = [1, 5, 9])
+data, _ = run!(model, agent_step_2d!, dummystep, 10; adata = adata, when = [1, 5, 9])
```
fbanning, line 120: Oh good catch

fbanning:
>     * An example using a more complex model structure would be nice. @fbanning, perhaps you could suggest something here?

I think the easiest way to display and explain a more advanced model structure would be a model with substeps as I expect this to be a quite regular usecase.
I first thought about expanding an existing example (like Predator-Prey) to have a day-night cycle as substeps during which agents act differently (sheep eat at day, wolf eat at night). However, that might be a suboptimal choice because the currently existing examples all serve a specific illustration purpose, e.g. Predator-Prey shows how a model with multiple agent types is handled.
Instead we might introduce a new example that stands on its own and only serves to explain how to get more fine-grained control over what happens at each step. Which option do you think would be better?

Datseris:
> Instead we might introduce a new example that stands on its own and only serves to explain how to get more fine-grained control over what happens at each step. Which option do you think would be better?

I agree with this but I would appreciate it if you can contribute an example. I'm not sure what is most usable or takes most advantage of this approach.

examples/wealth_distribution.jl
Libbum: Yeah! I'm sure I've looked at this example 1000 times and didn't notice it.

Libbum:
Would be happy to field suggestions from the Netlogo examples library for example. If you don't have time to actually implement one, but can find something that showcases what you're looking for, I can take a stab at writing it.

fbanning:
I just had a look through the NetLogo model library. The ["Voting" model](http://ccl.northwestern.edu/netlogo/models/Voting) seems to be a perfect fit to explain what I've described in the [abstract example in my earlier post](https://github.com/JuliaDynamics/Agents.jl/issues/331#issuecomment-704974655). The underlying problem is even mentioned in the comments inside the NetLogo code

>   ;; use two ask patches blocks so all patches compute "total"
>   ;; before any patches change their votes

The implementation in Agents.jl would be very simple, so that no other concepts would have to be explained to the readers (big plus imho).
Do you think this is a good fit to illustrate how substeps inside `model_step!` work?

Datseris:
yes. Can you put together a small example and post it here as a gist?

Libbum:
Yeah, I think that looks good. I can't see how we would do this without two iterations over all agents. You may solve it with the `total` being collected in `agent_step!`, then a call again to `allagents(model)` in `model_step!`. Of course this is where these changes make the most sense.

Libbum:
```netlogo
patches-own
[
  vote   ;; my vote (0 or 1)
  total  ;; sum of votes around me
]

to setup
  clear-all
  ask patches
    [ set vote random 2
      recolor-patch ]
  reset-ticks
end

to go
  ;; keep track of whether any patch has changed their vote
  let any-votes-changed? false
  ask patches
    [ set total (sum [vote] of neighbors) ]
  ;; use two ask patches blocks so all patches compute "total"
  ;; before any patches change their votes
  ask patches
    [ let previous-vote vote
      if total > 5 [ set vote 1 ]
      if total < 3 [ set vote 0 ]
      if total = 4
        [ if change-vote-if-tied?
          [ set vote (1 - vote) ] ]
      if total = 5
        [ ifelse award-close-calls-to-loser?
          [ set vote 0 ]
          [ set vote 1 ] ]
      if total = 3
        [ ifelse award-close-calls-to-loser?
          [ set vote 1 ]
          [ set vote 0 ] ]
      if vote != previous-vote
        [ set any-votes-changed? true ]
      recolor-patch ]
  ;; if the votes have stabilized, we stop the simulation
  if not any-votes-changed? [ stop ]
  tick
end

to recolor-patch  ;; patch procedure
  ifelse vote = 0
    [ set pcolor green ]
    [ set pcolor blue ]
end
```

Libbum:
```julia
using Agents             
                                   
mutable struct Voter <: AbstractAgent
    id::Int                                     
    pos::Tuple{Int,Int}                                              
    vote::Bool                                                                          
    total::Int                                                                          
end                                                                                     
                                                                                        
model = ABM(                                                                            
    Voter,                                                                              
    GridSpace((10, 10));                                                                
    properties = Dict(                                                                  
        :tick => 0,                                                                     
        :any_votes_changed => true,                                                     
        :change_vote_if_tied => true,
        :award_close_calls_to_loser => true,
    ),
)

for _ in 1:prod(size(model.space))
    add_agent_single!(model, rand(Bool), 0)
end 

function model_step!(model)
    votes_changed = false
    for a in allagents(model)    
        a.total = sum(v.vote for v in nearby_agents(a, model))
    end
    for a in allagents(model)
        previous_vote = a.vote
        if a.total > 5
            a.vote = true
        elseif a.total < 3
            a.vote = false
        elseif a.total == 4 && model.change_vote_if_tied
            a.vote = !a.vote
        elseif a.total == 5
            a.vote = model.award_close_calls_to_loser ? false : true
        elseif a.total == 3
            a.vote = model.award_close_calls_to_loser ? true : false    
        end
    
        if a.vote != previous_vote
            votes_changed = true
        end
    end
    model.any_votes_changed = votes_changed
    model.tick += 1
end

when(model, s) = !model.any_votes_changed
    
run!(model, dummystep, model_step!, when)
```

It's still somewhat straightforward. We need to call too "Agent Loops", although we don't need to do any special scheduling really...

Datseris:
this seems suspiciously similar to the HK model which features synchronous agent update...

Libbum:
Agreed. Perhaps something more complex will be needed.

Datseris:
I will return here, I want some more time to think about it. The point I want to raise is:

* What do we gain by making such a breaking change?

Nothing changes. Like before, the convenience option of `agent_step!` exists. Like before, any complex model that requires `model_step!` more, exists. Nothing changes, yet we're making this breaking change. Is there a reason?

Shouldn't we just document things better, add a couple of more sophisticated examples, and point out in the tutorial that `agent_step!` is just a convnience? I'll anyways add an entry in the tutorial explicitly writing down the core code of `step!` so that it becomes clear what happens. (i.e. when the stepping functions are called)

fbanning:
> I will return here, I want some more time to think about it.

Sure, do whatever you must. There's no rush in any way.

> the convenience option of agent_step! exists.

In my humble opinion, `agent_step!` is not so much convenient but rather redundant. I've stated this earlier and am still convinced that this is true. I would remove it completely but we have agreed to keep it as an optional/legacy parameter in #331.

> Is there a reason?

From my perspective, I've pointed out the reasons why this change would make sense in #331.

To sum it up again:
1. `model_step!` alone can do anything that `agent_step!` does. Besides a bit of apparent convenience for simpler models, there's nothing to be gained from having `agent_step!` hard-coded into the framework. 
2. The user saves absolutely no lines of code when using `agent_step!` for simpler models. Instead of `function agent_step! ... end` they will write `for agent in scheduler ... end` with the same functional content in between.
3. The transition from using a `agent_step!` with the model's scheduler towards writing own scheduler loops inside `model_step!` is an unnecessary step in the learning curve of Agents.jl.

Certainly those points are debatable as can be seen in the related issue. After all I would be happy to read why these arguments don't apply to the case at hand. I love to be proven wrong and learn new things!
I certainly don't think that the reasons I brought forward are unassailable proof that `agent_step!` should be dropped or that we should apply the changes proposed in this PR. But they sure seem very valid to me.

> What do we gain by making such a breaking change?

First and foremost: clarity and consistency.

> I'll anyways add an entry in the tutorial explicitly writing down the core code of step! so that it becomes clear what happens. (i.e. when the stepping functions are called)

I think this is a very good idea.

Libbum:
I'm still really in two minds about this. I agree with much of what @fbanning states above, then at the same time I think "why bother". 

One thing is for sure: this half baked approach (which I've argued for up to this point) is *not* the solution we should take. It's clear that if we choose to drop `agent_step!`, and break things, we should just up and do it&mdash;rip the band-aid off.

I think it's pretty ugly having `dummystep` around, and this PR makes it more prevalent.

With all that being said, I'm struggling to find an example in the Netlogo library that even comes close to needing such a change... 

Thus, it's a matter of taste. Documentation is good start, and I'd prefer to streamline things. Removing `dummystep` would be good, which means dropping the step separation. To that end, I think points 2 and 3 of @fbanning above make a good deal of sense. For the moment the scheduler takes a back seat, and we really don't expose users to its potential, instead we focus on how you actually separate workload between agent and model land. Flipping that on its head causes no more complexity (point 2).

We're making breaking changes already, so it's now when we have an increasing user base, or down the line when we have an established one..

This is a ramble in two directions, but it's where my head is at on this issue. If I had to swing either way right now, I think I'd vote for killing off `agent_step!` and write up some more extensive documentation.

Datseris:
Just a quick comment before I read the remaining text in more detail:

> The user saves absolutely no lines of code when using `agent_step!` for simpler models. Instead of `function agent_step! ... end `they will write `for agent in scheduler ... end` with the same functional content in between.

This is incorrect though. What you do is either:
```julia
function agent_step!(a, model)
# code
end

step!(model, agent_step!, dummystep, 10) # old version
```
or
```julia
function model_step!(a, model)
    for a in schedule(model) # or other scheduler
         some_agent_step_function!(a, model)
    end
end

step!(model, model_step!, 10) # new version (this PR)
```
The second version is clearly larger, and requires knowledge of scheduling.

Datseris:
Here is my conclusion: this **should not be merged**.

As I said: Nothing changes. Like before, the convenience option of agent_step! exists. Like before, any complex model that requires model_step! more, exists. Nothing changes, yet we're making this breaking change. This simply makes no sense from a development point of view.

Instead, we should have an additional entry in the Tutorial, called "Advanced stepping", that discusses complicated models and introduces a complex stepping function as an example, like so:
```julia
function complex_step!(model)
    for a in scheduler1(model)
        agent_step1!(a, model)
    end
    intermediate_model_action!(model)
    for a in scheduler2(model)
        agent_step2!(a, model)
    end
    final_model_action!(model)
end
```

The argument:

> In my humble opinion, agent_step! is not so much convenient but rather redundant. 

is false, since for every example model we have in our documentation `agent_step!` is rather convenient and not at all redundant.

The largest concern for me is that even though we've been talking about this for months, and even though we have explicitly asked for an actual, real example that we can put in the docs, nothing has been presented so far that justifies doing this change. Even Tim's example turned out to be something already in the docs that works no problem with agent+model steps. So, this by itself, makes me think that the arguments of @fbanning represent a *minority* of use cases, not the majority.

---

What's important here isn't the order argument in the function, is documenting clearly what is possible to do.

Libbum:
I'm fine with accepting that point of view. 

Indeed, we must put in some advance tutorial documentation: I can use one of my personal models in a cut down fashion I think. I'll work on a new PR with updated docs that satisfy the discussion here. 

@fbanning, perhaps once I've put up that documentation, I'll ping you and we can see if that methodology satisfies your concerns.

PR #353, opened 2020-12-01
Simplify, and extend walk!

### LINKED ISSUE #352: Challenging the design and complexity of `walk!`
I've been thinking about the new `walk!` function, and I don't think its design is good. It suffers from the (somewhat arbitrary) complexity of introducing 8 new types and 1 new abstract type, while making selecting a random direction difficult, since Julia is a compiled language. Choosing a type North, East etc. at random doesn't play well with type deductions.

I propose to completely remove all this North, East stuff and instead allow `walk` to take a 2-tuple as second argument:

```
walk!(agent, direction::Tuple{Int, Int}, model)
```
this has the benefit that we can define easily the dispatch method
```
walk!(agent, ::typeof(rand), model) = walk!(agent, (rand(-1:1), rand(-1:1)), model
```

Libbum:
Closes #352.

Gives us ND grid and continuous walks for both PBC and non-PBC spaces. The `ifempty` keyword will only move an agent to a new position if it is empty (`GridSpace` only).

Two simple random walk helpers are also included. I was considering making these more complex and allowing arbitrary random distributions, but that would make us require Distributions.jl. Don't think it's really needed, since users can extend this method quite easily themselves.

### DIFF
--- a/src/spaces/utilities.jl
+++ b/src/spaces/utilities.jl
@@ -1,5 +1,4 @@
-export edistance,
-    walk!, North, South, East, West, NorthEast, NorthWest, SouthEast, SouthWest
+export edistance, walk!
 
 #######################################################################################
 # %% (Mostly) space agnostic helper functions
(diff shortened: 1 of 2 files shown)


src/spaces/utilities.jl
```diff
@@ -58,58 +57,86 @@ function edistance(
...
-walk to, but not exceed the boundary value.
+The dimensionality of `direction` must be the same as the space. `GridSpace` asks for
+`Int`, and `ContinuousSpace` for `Float64` vectors, describing the walk distance in
+each direction. Velocity is ignored for this opreation in `ContinuousSpace`.
 
-Possible directions are `North`, `South`, `East`, `West`, as well as `NorthEast`,
-`SouthEast`, `SouthWest` and `NorthWest`.
+## Keywords
+- `ifempty` will check that the target position is unnocupied and only move if that's
+true. Available only on `GridSpace`.
```
Datseris, line 73: i would add an example direction here for clarity

src/spaces/utilities.jl
```diff
@@ -1,5 +1,4 @@
-export edistance,
-    walk!, North, South, East, West, NorthEast, NorthWest, SouthEast, SouthWest
+export edistance, walk!
```
Datseris, line 1: Why do we export `edistance`?

Libbum: Since it respects the PBCs. Users don't need to care about them when calculating distance. It's used in `examples/flock.jl` as an example.

Datseris: oh okay sorry didn't remember it

PR #354, opened 2020-12-03
Tuple for search direction on GridSpaces

### LINKED ISSUE #314: Allow search radius in GridSpace to be a Tuple (different value for each dimension)
It is quite straightforward(*) to change the neighbor finding code in grid space to allow for `r` to be a tuple as well, so that across each dimension a different radius is used. This will make the higher dimensionality of the space truly useful. Imagine you had some 3D problem, and you want the 4th dimension to actually mean a property of whatever: color, class, wealth, I don't know. Then you can find agents not only by proximity in real space, but also in the 4th dimension, that property.

(*) this is trivial for chebyshev norm by simply broadcasting the code. For euclidean space it is not possible.

Libbum:
Closes #314 by allowing `r` to be a tuple. We need to store a separate dictionary for this, but in general there is not too much overhead.

### DIFF
--- a/src/spaces/grid.jl
+++ b/src/spaces/grid.jl
@@ -20,6 +20,7 @@ struct GridSpace{D,P} <: DiscreteSpace
     s::Array{Vector{Int},D}
     metric::Symbol
     hoods::Dict{Float64,Hood{D}}
+    hoods_tuple::Dict{NTuple{D,Float64},Hood{D}}
 end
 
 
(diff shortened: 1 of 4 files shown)


src/spaces/grid.jl
```diff
@@ -40,11 +40,11 @@ cartesian indices have Euclidean distance `≤ r` from the cartesian index of th
...
-        d::NTuple{D,Int};
-        periodic::Bool = true,
-        metric::Symbol = :chebyshev,
-        moore = nothing
-    ) where {D}
+    d::NTuple{D,Int};
+    periodic::Bool = true,
+    metric::Symbol = :chebyshev,
+    moore = nothing,
+) where {D}
```
Datseris, line 47: Haha Tim, why do you keep doing this? I find it untuitive. The extra intentation level is better to read, because the next lines don't start with an extra identation out of nowhere

Datseris (approved):

src/spaces/grid.jl
Libbum: It's just the linter. Do you have some way of setting JuliaFormatter.jl not collapse the indentation?

Libbum: https://github.com/invenia/BlueStyle#method-definitions

Datseris: I don't agree with this:
```julia
function foobar(
    df::DataFrame,
    id::Symbol,
    variable::Symbol,
    value::AbstractString,
    prefix::AbstractString="",
)
    # code
end
```
because the `# code` starts with improper identation to the previous statement. It seems unnatural. For example you wouldn't do:
```julia
if a
    2
else
    3
end
    # code
```
but instead the code would have the same identation as the preceding line.

Libbum: You're not indenting the code though, you're ending the function definition.

In general I don't really have a strong opinion, but it's nice to have things standardised. I'd be happy to change linter defaults, I just can't see a way to override this specific lint easily.

PR #357, opened 2020-12-23
Abbreviate keyword argument values

rikhuijzer:
Hi! I'm working on a linter, and saw that this repository is set to Julia 1.5 and has [26 occurrences](https://lint.huijzer.xyz/github/juliadynamics/agents.jl/#abbreviate_keyword_argument_values_26_hits) of using 
```
f(a, b; c = c)
```
which [from Julia 1.5 onwards](https://julialang.org/blog/2020/08/julia-1.5-highlights) can be abbreviated to
```
f(a, b; c)
```
This PR applies this abbreviation. Feel free to close this PR immediately if you're not interested in this small refactoring.

Note that currently, I'm not matching cases like `f(a, b; c = c, d = e)` or `f(a, b; d = e, c = c)`. If you want to include this in one big PR, then let me know, and I'll add the patterns and update this PR :smile:

PR #358, opened 2020-12-23
Avoid findall(x -> x == false, Y)

rikhuijzer:
This is an extension of PR https://github.com/JuliaDynamics/Agents.jl/pull/357.

Instead of writing
```
julia> findall(x -> x == false, [true, false])
1-element Array{Int64,1}:
 2
```
it is possible to write
```
julia> findall(!, [true, false])
1-element Array{Int64,1}:
 2
```
or even
```
findall(.!Y)
1-element Array{Int64,1}:
 2
```
I guess the second variant is the clearest to readers, so this PR suggest to rewrite to occurrences to that.

Feel free to press close immediately if you don't find this a good suggestion! :smile:

### DIFF
--- a/examples/opinion_spread.jl
+++ b/examples/opinion_spread.jl
@@ -53,7 +53,7 @@ function adopt!(agent, model)
     matches = model[neighbor].opinion .== agent.opinion
     nmatches = count(matches)
     if nmatches < model.nopinions && rand() < nmatches / model.nopinions
-        switchId = rand(findall(x -> x == false, matches))
+        switchId = rand(findall(!, matches))
         agent.opinion[switchId] = model[neighbor].opinion[switchId]
     end
 end
--- a/src/models/opinion.jl
+++ b/src/models/opinion.jl
@@ -40,7 +40,7 @@ function adopt!(agent, model)
     nmatches = count(matches)
     # Adopt a different opinion w/ calculated probability
     if nmatches < model.nopinions && rand() < nmatches / model.nopinions
-        switchId = rand(findall(x -> x == false, matches))
+        switchId = rand(findall(!, matches))
         agent.opinion[switchId] = model[neighbor].opinion[switchId]
     end
 end


Datseris:
Hm, I must admit, I'm not super sold that the `findall(!, [true, false])` is the easiest to read version... Kind of surprised me, even though it makes perfect sense! @Libbum I leave this to your decision!

Libbum:
For the moment we only have this in two places, so not a major deal. What I find problematic from a new-user point of view is two-fold:
1. They take a look at [`Base.!`](https://docs.julialang.org/en/v1/base/math/#Base.:!) and see that `!missing` should work, and would then find it confusing when `findall(!, [true, false, missing])` fails (since findall expects `if missing` to be a bool).
2. We use the *boolean not* in this manner far more often than its counterparts. So whilst this looks ... OK; it doesn't feel that natural to do `findall(&, [true, false])` or `findall(|, [true, false])` to save a few characters on a line.

Thanks for running these lints though @rikhuijzer, appreciated!

rikhuijzer:
> For the moment we only have this in two places, so not a major deal. What I find problematic from a new-user point of view is two-fold:

I completely understand. Thank you for the feedback  :+1: :smile: 
 
>     1. They take a look at [`Base.!`](https://docs.julialang.org/en/v1/base/math/#Base.:!) and see that `!missing` should work, and would then find it confusing when `findall(!, [true, false, missing])` fails (since findall expects `if missing` to be a bool).

Just to check: is this right? They seem to yield the same output
```
julia> findall(x -> !x, [true, false, missing])
ERROR: TypeError: non-boolean (Missing) used in boolean context

julia> findall(!, [true, false, missing])
ERROR: TypeError: non-boolean (Missing) used in boolean context

julia> findall(.![true, false, missing])
ERROR: TypeError: non-boolean (Missing) used in boolean context
```
>     2. We use the _boolean not_ in this manner far more often than its counterparts. So whilst this looks ... OK; it doesn't feel that natural to do `findall(&, [true, false])` or `findall(|, [true, false])` to save a few characters on a line.
> 
> 
> Thanks for running these lints though @rikhuijzer, appreciated!

You're welcome; I'll make sure that the suggestion of this PR is not popping up again :)

Datseris:
> findall(&, [true, false]) or findall(|, [true, false]) to save a few characters on a line.

this doesn't make sense because `&, |` are *binary* operators while `!` is unary, which is what `findall` expects.

Libbum:
No, [it's trinary](https://docs.julialang.org/en/v1/base/math/#Base.:!):
```julia
julia> !true
false

julia> !false
true

julia> !missing
missing
```

Datseris:
That's not what the term means. It implements three-value logic, but this doesn't say anything about the number of input arguments. `!` takes one input argument that's why it is unary. (The operation `1 ! 2` doesn't make sense). On the other hand `&, |` take two input arguments, hence they are binary. `findall` expects a unary function as first argument. Orthogonal to the internal logic of the function.

Libbum:
Sure. That follows.

I've taken a look at @rikhuijzer's [other lints](https://lint.huijzer.xyz/patterns/) and it looks like there will be some more suggestions coming concerning maps and other boolean comparisons. To that end, maybe this change is not so bad.

Looks like the plan for the linter is to have it running constantly over all registered packages. Additionally, rikhuijzer/ShowLint#17 stems from our comments here&mdash;implying the pattern will be turned off for the organisation if we don't want it. After thinking on it a bit more, are you still of the impression this is not such a clear change?

rikhuijzer:
> Turns out that would just be `findall([true, false])`.

Yes, exactly! And that it possible because `findall` happens to also accept a single argument, that is, accepts `findall(Y)` as well as `findall(f, Y)`. That is also why `findall(.!Y)` works. 
```
julia> Y = [true, false]
2-element Array{Bool,1}:
 1
 0

julia> .!Y
2-element BitArray{1}:
 0
 1
```
(I only now notice that this [broadcasting](https://docs.julialang.org/en/v1/base/arrays/#Broadcast-and-vectorization) operation does change the type, so it's not completely equivalent :grey_exclamation:). Other broadcasting syntax examples are
```
julia> broadcast(!, Y)
2-element BitArray{1}:
 0
 1
```
and 
```
julia> (!).(Y)
2-element BitArray{1}:
 0
 1
```
or
```
julia> (x -> !x).(Y)
2-element BitArray{1}:
 0
 1
```

> I've taken a look at @rikhuijzer's other lints and it looks like there will be some more suggestions coming concerning maps and other boolean comparisons. To that end, maybe this change is not so bad.

I planned to stop after the two PRs I did now. The site contains many not so good suggestions. It is always a bit difficult to find a balance between what is truly an important change and what are not so important changes. It's all a bit experimental still, and I'm trying to figure out what people find an useful rewrite and what they find not useful.

Issue #356, opened 2020-12-17
Dimension aware utilities


Libbum:
I'm writing up an example to showcase some of the new features we've been implementing (Tuples in search dimensions of neighbors, the `walk!` function, `edistance` etc). We've spoken about using the higher dimensions of `GridSpace` to seek out agents in categorical rather than spatial distance&mdash;that's the plan here. 

However, it's only the neighbor utilities are aware of this concept at present, meaning that there's a bunch of workarounds needed to get to a state that 'normal' use of our utilities can achieve. 

Here's the idea (not completely fleshed out, it's a work in progress), with a start on an implementation and a bunch of pain points:

```julia
# This example illustrates how to leverage higher dimensions of a `GridSpace` to identify
# the distance from neighbors not just spatially, but also categorically. We'll also use
# the [`walk!`](@ref) function extensively.

# ## Rules of Engagement
# Agents wander around the map looking for opponents. When a grid space is occupied by                                                                                    
# two or more agents there will be a battle. With experience gained from the fight, the
# victor searches for more opponents to crush and losers scurry away defeated or possibly
# even die. This process repeats until there is a single, definitive winner.
#
# For this battle ground to exist, the following rules must be followed:
# - Agents have an experience level, starting at level 1 up to a maximum of 15.
# - Agents will walk randomly to search for opponents when none are nearby.
# - Once opponents are found, agents will (in this explicit order)
#   1. Approach the nearest worthy opponent (one with equal or ±1 experience level)
#   2. Chase down a weak opponent (with experience level -2 or fewer)
#   3. Run from a tougher opponent (with experience level +2 or higher)
# - Running agents call for help. If heard, another agent may choose to temporarily team
# up and fight the tougher opponent together.
# - Battles are one by weighted chance - a higher level gives an agent a larger chance of
# winning, but does not guarantee it. When a victor is chosen
#   - The difference in experience between opponents is swapped.
#   - If an agents experience is below 1, they die.
#   - Opponents leave in a daze an will walk at random for 3 time steps.
#
# ## Model Setup

cd(@__DIR__) #src
using Agents
using Plots
gr() # hide

mutable struct Fighter <: AbstractAgent
    id::Int
    pos::Tuple{Int,Int,Int}
end

model = ABM(Fighter, GridSpace((100, 100, 15)))

n = 0
while n != 10
    pos = (rand(1:100, 2)..., 1)
    if isempty(pos, model)                                               
[...]

PR #359, opened 2020-12-23
Battle Royale example

Libbum:
In an attempt to investigate the discussion in #314, I've come up with this example. It runs, but I'm not completely happy with the output. (I'll need to add in a colour and marker legend before merging too, but that's less relevant than below)

I'll use *space* and *category* to separate the different dimensions of our `GridSpace`.

In no particular order:
- Cannot use AgentsPlots. It thinks we're a 3D plot. There would need to be some way to send it the spatial dimensions only.
- Periodic boundaries make no sense in a category, but would be nice to have in space.
    I wanted to have PBC's in this example initially, but when I have a level 2 agent and are looking for 'weaker' opponents, I'd like to return agents of level 1, not level 1, 10, 9, 8.
- Cannot decouple space and category.
    If I want to search for any agent with my level, I have to search ALL space, plus my current level. (100, 100, 1). This functionality really only works efficiently when we can completely decouple the categories.
- Cannot search for "larger than" or "less than" etc.
    All we have is neighbors in a radius, so in this specific example when I want to find "weak" or "strong", I have to find "weak" AND "strong", then filter one out.
- Lots of excess filtering in general. You can see through the example. Would be nice to consider if these are considered fine, or if  there is a way to make this better.
- Is there a strong motivation to do this vs agent properties?
    For really large grids, with the above things fixed, then yeah - maybe. But for the moment we're running through 100x100 grid spaces to find worthy agents, vs 50 total agents which could have a `level` property. 

As I suggested in #356, making `GridSpace` aware of categorical dimensions might be a way forward. But, take a look at the code here, and see if you can't find any better solutions, and we can adjust the example accordingly.


Libbum:
battle

Libbum:
The recent updates to the example and addition of #360 leaves my original list of issues with only one open point: cannot use AgentsPlots. I think this is completely fine, at least for the moment though. This is simply an example, and we do not have full support for such things, nor know if users will truly embrace the concept.

Datseris:
My plan for AgentsPlots is to _kill it_ and use `InteractiveChaos` directly. I've changed interactive chaos and it now has only abstract plotting as a dependency. It already has all the code for the plots. The only thing remaining is to add a function for static plots, that doesn't do all the fancy animating.

Libbum:
Sounds like a plan!

I'm having some issues getting the docs to build, but this will be ready once they pass.

Datseris:
go ahead with merging from my perspective

PR #360, opened 2021-01-12
Specified dimension nearest-neighbor search

Libbum:
This PR satisfies a number of open problems addressed in #359 by adding one additional search pattern into the neighbour search: `r::Vector{Tuple{Int,UnitRange{Int}}}`. This makes searching over specific dimensions with precise distance possible.

Hoods were not the best way of going about this in the end. I attempted to cut down the dimensionality of the hood (in an unpublished branch), but converting back to a set of true `CartesianIndex`s turned out to be impractical.

Searching over 'all' space takes a good deal of time comparatively when looking at `views` of the space. Here's a test case:

```julia
using Agents
using BenchmarkTools                                                                                                                                                      
using Random                                                                                                                                                              
Random.seed!(6548)                                                                                                                                                        
                                                                                                                                                                          
mutable struct Fighter <: AbstractAgent                                                                                                                                   
    id::Int                                                                                                                                                               
    pos::Tuple{Int,Int,Int}                                                                                                                                               
    has_prisoner::Bool                                                                                                                                                    
    capture_time::Int                                                                                                                                                     
    shape::Symbol # For plotting                                                                                                                                          
end                                                                                                                                                                       
                                                                                                                                                                          
model =                                                                                                                                                                   
    ABM(Fighter, GridSpace((100, 100, 10); periodic = false); scheduler = random_activation)                                                                              
                                                                                                                                                                          
n = 0                                                                                                                                                                     
while n != 50                                                                                                                                                             
    pos = (rand(1:100, 2)..., rand(1:10))                                                                                                                                 
    if isempty(pos, model)                                                                                                                                                
        add_agent!(pos, model, false, 0, :diamond)                                                                                                                        
        n += 1                                                                                                                                                            
    end                                                                                                                                                                   
end                                                                                                                                                                       
                                                                                                                                                                          
agent = model[23]                                                                                                                                                         
α = CartesianIndex(agent.pos)                                                                                                                                             
space = model.space                                                                                                                                                       
r = (100, 100, 1)

# level is 6. Find -1:1 -> 5:7                                                                                                                                            

function search_view(space)                                                                                                                                              
    s = view(space.s, :, :, 5:7)                                                                                                                                          
    return [model[id] for id in Iterators.flatten(s)]                                                                                                                     
end                                                                                                                                                                       
                                                                                                                                                                          
function search_index(agent, model, r)                                                                                                                                    
    nn = nearby_ids(agent.pos, model, r)                                                                                                                                  
    return [model[id] for id in nn]                                                                                                                                       
end                                                                                                                                                                       

@assert search_view(space) == search_index(agent, model, r) 

@benchmark search_view($space)

#BenchmarkTools.Trial: 
#  memory estimate:  544 bytes
#  allocs estimate:  7
#  --------------
#  minimum time:     73.800 μs (0.00% GC)
#  median time:      74.401 μs (0.00% GC)
#  mean time:        74.978 μs (0.00% GC)
#  maximum time:     246.422 μs (0.00% GC)
#  --------------
#  samples:          10000
#  evals/sample:     1
 
@benchmark search_index($agent, $model, $r) setup = (Agents.initialize_neighborhood!($space, $r))

#BenchmarkTools.Trial: 
#  memory estimate:  432 bytes
#  allocs estimate:  5
#  --------------
#  minimum time:     317.694 μs (0.00% GC)
#  median time:      338.685 μs (0.00% GC)
#  mean time:        340.662 μs (0.00% GC)
#  maximum time:     479.480 μs (0.00% GC)
#  --------------
#  samples:          6636
#  evals/sample:     1

function search_view_lazy(space)
    s = view(space.s, :, :, 5:7)
    return (model[id] for id in Iterators.flatten(s))
end

function search_index_lazy(agent, model, r)
    nn = nearby_ids(agent.pos, model, r)
    return (model[id] for id in nn)
end

@benchmark search_view_lazy($space)

#BenchmarkTools.Trial: 
#  memory estimate:  0 bytes
#  allocs estimate:  0
#  --------------
#  minimum time:     3.776 ns (0.00% GC)
#  median time:      3.799 ns (0.00% GC)
#  mean time:        3.891 ns (0.00% GC)
#  maximum time:     16.770 ns (0.00% GC)
#  --------------
#  samples:          10000
#  evals/sample:     1000
                                  
@benchmark search_index_lazy($agent, $model, $r) setup = (Agents.initialize_neighborhood!($space, $r))

#BenchmarkTools.Trial: 
#  memory estimate:  0 bytes
#  allocs estimate:  0
#  --------------
#  minimum time:     44.074 ns (0.00% GC)
#  median time:      44.487 ns (0.00% GC)
#  mean time:        45.276 ns (0.00% GC)
#  maximum time:     81.474 ns (0.00% GC)
#  --------------
#  samples:          8509
#  evals/sample:     990
```

I don't *think* there needs to be much more done to satisfy my issues in the example if this is merged. But I'll go over it again soon and see if I find anything.

TODO: Tests, changelog.

### DIFF
--- a/src/spaces/grid.jl
+++ b/src/spaces/grid.jl
@@ -212,6 +212,8 @@ Return an iterator of ids over specified dimensions of `space` with fine grained
 of distances from `pos` using each value of `r` via the (dimension, range) pattern.
 
 Example, with a `GridSpace((100, 100, 10))`: `r = [(1, -5:5), (3, 0:2)]`.
+
+**Note:** Only available for use with non-periodic grids.
 """
 function nearby_ids(
     pos::ValidPos,
@@ -222,11 +224,16 @@ function nearby_ids(
     vidx = []
     for d in 1:ndims(model.space.s)
         idx = findall(dim -> dim == d, dims)
-        dim_range = isempty(idx) ? Colon() : pos[d] .+ last(r[only(idx)])
+        dim_range = isempty(idx) ? Colon() :
+            bound_range(pos[d] .+ last(r[only(idx)]), d, model.space)
         push!(vidx, dim_range)
     end
     s = view(model.space.s, vidx...)
     Iterators.flatten(s)
+end
+
+function bound_range(unbound, d, space::GridSpace{D,false}) where {D}
+    return range(max(unbound.start, 1), stop = min(unbound.stop, size(space)[d]))
 end
 
 function nearby_positions(pos::ValidPos, model::ABM{<:AbstractAgent,<:GridSpace}, r = 1)
--- a/src/spaces/grid.jl
+++ b/src/spaces/grid.jl
@@ -213,7 +213,7 @@ of distances from `pos` using each value of `r` via the (dimension, range) patte
 
 Example, with a `GridSpace((100, 100, 10))`: `r = [(1, -5:5), (3, 0:2)]`.
 
-**Note:** Only available for use with non-periodic grids.
+**Note:** Only available for use with non-periodic chebyshev grids.
 """
 function nearby_ids(
     pos::ValidPos,
--- a/src/spaces/grid.jl
+++ b/src/spaces/grid.jl
@@ -150,11 +150,7 @@ reducing the amount of computations necessary (i.e. we don't "find" new indices,
 we only add a pre-determined amount of indices to `α`).
 """
 function grid_space_neighborhood(α::CartesianIndex, space::GridSpace, r::Real)
-    hood = if haskey(space.hoods, r)
-        space.hoods[r]
-    else
-        initialize_neighborhood!(space, r)
-    end
+    hood = get(_ -> initialize_neighborhood!(space, r), space.hoods, r)
     _grid_space_neighborhood(α, space, hood)
 end
 
@@ -170,11 +166,7 @@ function grid_space_neighborhood(
     space::GridSpace{D},
     r::NTuple{D,Real},
 ) where {D}
-    hood = if haskey(space.hoods_tuple, r)
-        space.hoods_tuple[r]
-    else
-        initialize_neighborhood!(space, r)
-    end
+    hood = get(_ -> initialize_neighborhood!(space, r), space.hoods_tuple, r)
     _grid_space_neighborhood(α, space, hood)
 end
 
--- a/src/spaces/grid.jl
+++ b/src/spaces/grid.jl
@@ -205,6 +205,18 @@ function nearby_ids(pos::ValidPos, model::ABM{<:AbstractAgent,<:GridSpace}, r = 
     Iterators.flatten((s[i...] for i in nn))
 end
 
+function nearby_ids(pos::ValidPos, model::ABM{<:AbstractAgent,<:GridSpace}, r::Vector{Tuple{Int,UnitRange{Int}}})
+    dims = first.(r)
+    vidx = []
+    for d in 1:ndims(model.space.s)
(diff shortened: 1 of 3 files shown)


Libbum:
Running an updated battleroyale from #359:

```diff
80,83c80
<             strong_ids = collect(Iterators.filter(
<                 id -> level(agent) + 1 < level(model[id]),
<                 nearby_ids(agent, model, (5, 5, 4)),
<             ))
---
>             strong_ids = collect(nearby_ids(agent, model, [(1, -5:5), (2, -5:5), (3, 2:4)]))
90,91c87
<                 worthy_ids =
<                     collect(nearby_ids(agent, model, (size(model.space)[1:2]..., 1)))
---
>                 worthy_ids = collect(nearby_ids(agent, model, [(3, -1:1)]))
106,108c102,105
<                     weak_ids = collect(Iterators.filter(
<                         id -> level(agent) - 1 > level(model[id]),
<                         nearby_ids(agent, model, (10, 10, 4)),
---
>                     weak_ids = collect(nearby_ids(
>                         agent,
>                         model,
>                         [(1, -10:10), (2, -10:10), (3, -4:-2)],
218c215
```

Benchmark for 225 steps, arriving at the same solution:
Previous (ms): 653.615
This PR (ms): 183.393

I've kept two tuple searches in the example:
```julia
close_ids = collect(nearby_ids(agent, model, (0, 0, 10)))
occupied_ids = collect(Iterators.filter(
            id -> model[id].has_prisoner,
            nearby_ids(agent, model, (7, 7, 10)),
        ))
```
These could be simplified since we don't really need to search 'all' of the levels, just the space. For example: `close_ids = collect(nearby_ids(agent, model, [(1, 0:0), (2, 0:0)]))`

Doing that drops the time to 164.22 ms.

So when you get the chance @Datseris, could you comment on your choice of this `Hood` mechanism vs `views`? I guess for the Euclidean case it may be an issue to obtain a decent `Hood`, but the chebyshev case really seems like we'd be better off using views overall. Perhaps I've missed something though..

I don't know why that continuous test failed there, I'll make some test soon...

src/spaces/grid.jl
```diff
@@ -205,6 +197,37 @@ function nearby_ids(pos::ValidPos, model::ABM{<:AbstractAgent,<:GridSpace}, r =
     Iterators.flatten((s[i...] for i in nn))
 end
 
+"""
+    nearby_ids(pos, model::ABM{<:GridSpace}, r::Vector{Tuple{Int,UnitRange{Int}}})
+
+Return an iterator of ids over specified dimensions of `space` with fine grained control
+of distances from `pos` using each value of `r` via the (dimension, range) pattern.
+
+Example, with a `GridSpace((100, 100, 10))`: `r = [(1, -5:5), (3, 0:2)]`.
```
Datseris, line 206: I'm not sure this is as clear as it could be. So in principle here you would search the 3rd dimension, from current level up to at most +2 levels, right? Perhaps we should discuss this more in the docstring and also crossref to the battle royale example?

Datseris (approved):
Using views is totally fine since it is more performant! I guess only docs can be improved here, feel free to merge after.

src/spaces/grid.jl
Libbum: Certainly. Will do so tomorrow.

Libbum: This will be easier to extend once the battle example is updated, so I'll fix this in #359.

PR #366, opened 2021-01-15
Changing type order in `ABM`

### LINKED ISSUE #364: Change Type parameterization order for ::ABM, make `Space` first
It is almost exclusively the case that we dispatch on the _second_ type parameter of `ABM`, the space type. Perhaps it makes sense to make this type first in the type paraterization, so we no longer have to write `::ABM{A, GridSpace} where {A<:AbstractAgent}`, but simply `::ABM{GridSpace}` ?

Besides the fact that we have to do this change throughout the source, is there any downside? The simplification upside is huge I believe.

Libbum:
Closes #364

First pass at this. Tests run, although there may be some questions as to how daisyworld runs - will see how documentation builds.

Think I covered all of the needed changes. Unsure how much of an improvement this makes are though...

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -10,6 +10,7 @@ using Agents
 println("Plots...")
 using Plots
 println("AgentsPlots...")
+Pkg.add("AgentsPlots#master")
 using AgentsPlots
 println("Literate...")
 import Literate
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -10,7 +10,7 @@ using Agents
 println("Plots...")
 using Plots
 println("AgentsPlots...")
-Pkg.add("AgentsPlots#master")
+Pkg.add(PackageSpec(name="AgentsPlots", rev="master"))
 using AgentsPlots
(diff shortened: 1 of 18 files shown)


Libbum:
Right.. We'll need to patch AgentsPlots for this to work...

Datseris:
Yeah haven't really realized how often we were using `A` inside functions. Thankfully this definitely simplifies many docstrings, where now the `{A, ` is no longer necessary.

For agents plots we can simply make the docs use master branch again by doing `Pkg.add("AgentsPlots#master")` in the `make.jl`, and make a quick update on the plots repo

Datseris:
i will spend some time trying out cairo makie anyways for our docs

Libbum:
Examples all look good!

Datseris:
So this is good to merge right? Anything missing? If not, go ahead.

Libbum:
Yep. Good for someone else to take a glance at it, just in case.

Issue #365, opened 2021-01-14
[FR] Inheritance from pre-defined agent types


Datseris:
c.f discussion here: https://discourse.julialang.org/t/composition-and-inheritance-the-julian-way/11231/4

I've been thinking this for a long time now, because we kind of force the user to make a type with given fields at given order, we could simplify with making a 
```julia
mutable type ContinuousAgent{D}
id::Int
pos::NTuple{D, Float64}
end
```
and the user could do
```julia
@inherit Agent ContinuousAgent{2}
w::Float64
end
```

and the new open street map space made me rethink about it. The post above provides a simple macro to do this, all in all I think it is not a bad idea? I definitely think this is more reasonable than enforcing extension of a `pos` function the standard Julian way.

Libbum:
Would take some design consideration, but in general this would indeed be nicer. I've used something like this before in [DICE.jl](https://github.com/Libbum/DICE.jl/blob/master/src/Abstractions.jl), which was quite helpful.

Libbum:
The good thing about this is that each space will have an agent type, so we may not even need to worry about #366, since we can dispatch on `GridAgent` when needed, and `GridSpace` otherwise.

Datseris:
I still think the dispatch should happen on the space nevertheless, because the suggestion here is that using this subtyping system is _optional_ (because it reduces code clarity after all).

I'll put together a draft once I'm doing reviewing the rest of the PRs.

Libbum:
Oh, but I've just done it.

PR #367, opened 2021-01-15
Initial pre-defined types

Libbum:
WIP for #365.

Types and docstrings added, but have not filtered them through examples or anything. No tests done yet, but the results do work out as the docstrings suggest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - `random_agent(model, condition)` allows obtaining random agents that satisfy given condition.
 - A walk! utility for `GridSpace` and `ContinuousSpace`s, providing turtle-like agent movement and random walks.
 - The Battle Royal example explores using categorical neighbor searching in a high dimensional `GridSpace`.
+- An `@agent` macro provides a quick way of creating agent structs for any space.
 
 ## Breaking changes
 Most changes in this section (besides changes to default values) are deprecated and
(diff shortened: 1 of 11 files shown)


Datseris:
ah you did it already. okay I'll review tonight

Datseris:
my suggestion is that agent abstract types go to a new file in the `core` folder: `agents.jl`

Libbum:
At the moment this generates `<: AbstractAgent`s only. If we want to dispatch on `<: GraphAgent` then we'll need to create subtypes of AbstractAgent for each space, then modify structs and the macro accordingly. From your comment on the other thread, this may not be that important though...

Datseris:
I've done some doc updates and this is good to go from my end. Let's only add one basic test to be on the clear side regarding codecov.

Libbum:
I was going to also update some / all examples to use the macro - does that makes sense to you too?

Looks like you've made `vel` a mandatory variable in the `ContinuousSpace`. We only have it as [optional](https://github.com/JuliaDynamics/Agents.jl/blob/5df08e22d10d5fc4721760b66a12a0cd12ac69ed/src/core/model.jl#L277) in the `agent_validator` however. Should I promote it to mandatory there?

Datseris:
Sorry, forgot to mention the velocity change I did. Yes I made that because since it is anyway automated, we might as well add it for convenience. No reason to add it to the validator though, it is still not mandatory.

Datseris:
> I was going to also update some / all examples to use the macro - does that makes sense to you too?

I'm not sure about this. I was thinking that we'd leave the examples as-is for clarity. Macros tend to be unclear unless you really know what they do. But on the other hand, it is worth it to change some of the examples? Because this new interface is more convenient? Not sure! I'd say we leave it for now, its a really recent change after all.

Libbum:
I saw you added a discussion in Schelling. I think this enough for now, we'll see how people use it. I'll add a changelog entry and merge this then.

PR #368, opened 2021-01-16
Documentation and test updates

Libbum:
This will close out a number of documentation issues when complete. I'm having some issues building docs locally though, so pushing this up to get actions to test the back-referencing. 

Also done some tidying up of never-used small functions and closed some coverage gaps in the tests.

### DIFF
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -183,7 +183,7 @@ end
 Return the number of agents in the `model`.
 
 Example usage in:
-- [Wright-Fisher model of evolution](@ref): [sample!](@ref neutral)
+- [Wright-Fisher model of evolution](@ref): [A neutral model](@ref)
 """
 nagents(model::ABM) = length(model.agents)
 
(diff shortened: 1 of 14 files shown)


src/core/model.jl
```diff
@@ -181,6 +181,9 @@ end
 """
     nagents(model::ABM)
 Return the number of agents in the `model`.
+
+Example usage in:
+- [Wright-Fisher model of evolution](@ref): [sample!](@id neutral)
```
Datseris, line 186: I'm not sure I agree with this change. It essentially makes the docstring of a very simple function twice as large... Perhaps for `sample!` it does make sense to cross-ref ton an example, but for `nagents`? Not sure.

Libbum: I'll perhaps do all the complex ones, yeah. This is essentially just a test to make sure things work out as I suspect they should. But I'll dump it later - sure.

Libbum:
If the doc build passes after that last commit, I think this is good to merge if there are no comments. I didn't go too wild with the example trackbacks - just some to the more complex functions. 

All of this can be extended in the future of course. Documentation is never finished.. Let me know if you have some specifics that I've missed that you think are worthwhile though.

Datseris:
Nah go ahead, good time to wrap things up

PR #370, opened 2021-01-19
Rework Agents.jl entirely on Makie + InteractiveChaos

Datseris:
This works locally for me but needs https://github.com/JuliaDynamics/InteractiveChaos.jl/pull/30 , which will be merged soon.


Libbum:
- I wonder if InteractiveChaos needs a new name now that there's more than just the orbits, cobwebs etc.
- Pretty sure the doc failures here are coming from `workflows/ci.yml`, we can just drop the PyCall stuff there...

Datseris:
> I wonder if InteractiveChaos needs a new name now that there's more than just the orbits, cobwebs etc.

Sure, good idea. Any suggestions?

Libbum:
InteractiveDynamics is general enough to cover everything that's there now and it fits with the organisation name. You wrote InteractivePlotting before, perhaps a slip of the fingers, but its apt too.

Datseris:
Yeap InteractiveDynamics seems like the way to go.

AayushSabharwal:
The current commit only shows Schelling model changed to the new plotting system. Should the other examples be updated as well? I can do that if necessary.

Additionally, I wanted to ask why the backend in the updated example is Cairo? Makie seems to recommend GL.

Datseris:
> The current commit only shows Schelling model changed to the new plotting system. Should the other examples be updated as well? I can do that if necessary.

Yes, I've updated only Schelling model for now because I wanted to test whether it works. Everything has to be updated, and if you want to help that would be great. But first, have you reproduced this PR locally? Did you manage to create the schelling video?

> Additionally, I wanted to ask why the backend in the updated example is Cairo? Makie seems to recommend GL.

For the simple reason that GLMakie is very hard to get working on CI (so that the docs get plots automatically), while CairoMakie works out of the box. On my work I also use GLMakie exclusively.

AayushSabharwal:
I'll try replicate this PR locally and test it out.

I was asking about the backend since I thought someone might assume from the examples that Cairo is the recommended backend for using with Agents, especially if they've never explicitly used Makie before.

Datseris:
Yes, good idea to add a note in the docs then

AayushSabharwal:
Could you give me a quick pointer on how you set up this PR? I'm not too experienced with managing Julia packages, and I can't seem to get my clone of the `cairomakie` branch to use the clone of `new_agents` branch of InteractiveChaos. It doesn't recognise the `abm_plot` function.

Datseris:
You need to do:
```
add Agents#cairomakie
add InteractiveChaos#new_agents
```
so that julia checks out specific branches instead of stable versions

AayushSabharwal:
Thanks for the help. I've managed to replicate the schelling video. It works out of the box, as expected. I'll get to work on updating the rest of the examples.

Datseris:
InteractiveChaos 0.12 is now tagged. You just need this version, no reason to follow master branch at the moment.

Libbum:
Regarding the Schelling example: output from `abm_plot` looks super clean - great! However, looks like we have issues with `abm_video`. In firefox I get "No video with supported format and MIME type found", and in Chrome I get a small video window, but there is no actual video there.

Datseris:
Hm, unfortunate. It worked locally for me. Does it work locally for you with the video?

Libbum:
I'm just checking that now. Looks like there's a 404 error to the video, so I'm investigating that.

Libbum:
Ah, it's [one folder down the tree](https://github.com/JuliaDynamics/Agents.jl/tree/gh-pages/previews/PR370/examples)...

Datseris:
What's the fix?

Libbum:
In a meeting, I'll look later. Either just use "../schelling.mp4" or fix the `@__DIR__` directory in this instance.

Datseris:
Ugh, I remember there was a comment about that on the Docs of Documenter.jl, whereby hosting changes some paths...

AayushSabharwal:
I've updated daisyworld, it generates the following video on my system.

https://user-images.githubusercontent.com/23384717/105180211-4353aa80-5b50-11eb-9f53-27b629ea956c.mp4

Libbum:
It's not clear to me how to solve the path question yet. [Animations in AbstractPlotting](https://raw.githubusercontent.com/JuliaPlots/AbstractPlotting.jl/master/docs/src/animation.md) for example use no `cd` command and have no special `workdir` command in make.jl.

We write to `examples/name/`, but `cd(@__DIR__)` seems to be saving to `examples/`, so I don't know what the deal is with that. I'm going to try a few things here...

Libbum:
So that works. And by that, I mean just putting ".." in the path of the raw html blocks. It shouldn't be an issue so long as we name the animations uniquely. It'd be nice to push them into the correct path though...

Datseris:
@AayushSabharwal in the InteractiveChaos app I use some flowers as markers, see here: https://github.com/JuliaDynamics/InteractiveChaos.jl/blob/master/examples/agents_daisyworld.jl

Datseris:
Also, now that the videos make for easily adding a nice title, all videos should have a title.

Libbum:
Flocking has a section talking about AgentsPlots and its custom shape capacity. We use it in Bacterial Growth as well.
```julia
function bird_triangle(b::Bird)    
    φ = atan(b.vel[2], b.vel[1])    
    xs = [(i ∈ (0, 3) ? 2 : 1) * cos(i * 2π / 3 + φ) for i in 0:3]    
    ys = [(i ∈ (0, 3) ? 2 : 1) * sin(i * 2π / 3 + φ) for i in 0:3]    
    Shape(xs, ys)    
end
```

I'm not sure where to start with converting, this but I guess you did something to this end with billiards @Datseris?

Datseris:
Yes, just use `poly!` to plot a polygon. Here is how I plot disks: 
```julia
    θ = 0:0.01:2π
    p = [Point2f0(cos(t)*o.r + o.c[1], sin(t)*o.r + o.c[2]) for t in θ]
    poly!(ax, p; color = obfill(o), scale_plot=false, kwargs...)
```

Libbum:
So the problem that I see here is that we can't really use `poly` as a marker type for `scatter` (`to_spritemarker` only accepts string/symbol/char). If that's the case, then we need a way to identify the need for custom markers and swap out the scatter call to a bunch of poly calls. 

Perhaps you see a different angle?

Datseris:
its possible to use arbitrary polygons for markers, I'm almost certain about it. We have to ask in the makie slack for it

AayushSabharwal:
I've added flower markers and a title for the daisyworld video. I've also tested the `abm_play` function and used it to replace the old `interactive_abm`. There doesn't seem to be a way to add parameter sliders to the interactive player, or have I missed something?

https://user-images.githubusercontent.com/23384717/105284594-cddfec80-5bd8-11eb-9600-2d9a2079266a.mp4

AayushSabharwal:
I additionally updated HK opinion dynamics to use Makie. It doesn't use InteractiveChaos, but I thought it appropriate to shift all plotting to Makie in the examples. Below is the resultant plot
hk

AayushSabharwal:
I've updated predator-prey too. I'm not attaching the images, since it feels like it's getting a bit spammy, but I can upload them when necessary

Libbum:
- [x] Schelling's segregation model
- [ ] Sugarscape
- [ ] SIR model for the spread of COVID-19
- [ ] Continuous space social distancing for COVID-19
- [ ] Wealth distribution
- [x] Forest fire
- [x] Conway's game of life
- [ ] Wright-Fisher model of evolution
- [x] Hegselmann-Krause opinion dynamics
- [ ] Flocking (partial)
- [x] Daisyworld
- [x] Predator-Prey
- [ ] Bacteria Growth
- [x] Opinion spread
- [ ] Battle Royale
- [ ] BlackBoxOptim.jl
- [ ] DifferentialEquations.jl
- [ ] LightGraphs.jl
- [ ] Measurements.jl

AayushSabharwal:
Should I send a PR to the `cairomakie` branch?

Libbum:
Yes - thanks. I just sent you an email outlining the steps.

AayushSabharwal:
I got the mail, thanks for the help! The PR is at https://github.com/JuliaDynamics/Agents.jl/pull/373

AayushSabharwal:
It looks like Wealth Distribution didn't have a random seed set before running it for the first time (to generate the unicode plot), and I'm not able to reproduce the exact plot. I can get fairly close though, and it still obeys power law

Libbum:
That's fine.

AayushSabharwal:
Wright-Fisher doesn't have any plots. Should that just be ticked off the list?

AayushSabharwal:
In sugarscape, the second plot that consists of an animated gif of two plots (a heatmap and a `plotabm`) doesn't seem like it can be replicated with the current setup, since any call to `abm_plot` generates a new plot, and `abm_video` doesn't allow for manual modification of the underlying figure. My proposed solutions are as follows:

1. Split it into two animations
2. Expose the `!` functions (maybe just `abm_plot!`) in InteractiveChaos to allow for the figure returned from `abm_plot` to be modified

I'll be happy to implement any solution deemed appropriate.

Libbum:
Let's try do do this as simple as possible now, and build fixes for things we don't like later. I propose splitting the figure into two now, and opening an issue about this on InteractiveChaos so we can figure out the best solution later.

Datseris:
Yeah we definitely need some `abm_plot!` but the code will need a lot of work for this to happen because at the moment the current stepping function has 20 input arguments... I'll have to simplify this by defining a lot of `on` statements.

Datseris:
I know *exactly* how to fix this and make it is easy to use. Unfortunately, I have to give a very important presentation on Monday and therefore can't spend much time on other projects.

Libbum:
LightGraphs.jl will need the `abm_plot!` adjustment too, and I'm not sure what to do with Battle Royale, probably another custom implementation for the moment...

The outliers now are the SIR model and the plots that need paths on scatters.

Datseris:
@Libbum Makie does not support custom polygon plotting at the moment, which makes bacterial and flocking impossible to plot. How do we proceed? Should we proceed with the paper with the old AgentsPlots.jl and leave this PR for after Makie supports what we need?

AayushSabharwal:
Could `meshscatter` be of some potential use?

Datseris:
No:
```
ulia> markers = [:circle, :rect, :circle]
3-element Array{Symbol,1}:
 :circle
 :rect
 :circle

julia> meshscatter(rand(3), rand(3); marker = markers)
Error showing value of type AbstractPlotting.FigureAxisPlot:
ERROR: MethodError: no method matching _default(::Tuple{Observable{Array{Symbol,1}},Observable{Array{Point{2,Float32},1}}}, ::GLMakie.GLAbstraction.Style{:default}, ::Dict{Symbol,Any})
Closest candidates are:
  _default(::Union{Observable{M}, M}, ::GLMakie.GLAbstraction.Style, ::Dict) where M<:GeometryBasics.Mesh at C:\Users\datse\.julia\packages\GLMakie\mjJL8\src\GLVisualize\visualize\mesh.jl:2
  _default(::Union{GLMakie.GLAbstraction.GPUArray{T,2}, Union{Observable{X}, X} where X<:AbstractArray{T,2}}, ::GLMakie.GLAbstraction.Style, ::Dict) where T<:ColorTypes.Colorant at C:\Users\datse\.julia\packages\GLMakie\mjJL8\src\GLVisualize\visualize\image_like.jl:7
```

Libbum:
I'm a bit torn about that. These new introductions are much better than Plots.jl implementations in so many ways, but there's a lot of work for us individually if we want to get the lot transitioned and we can't hope that these issues will be a priority for Makie devs.

What about hybrid approach? Since v4.0 is not necessarily a 'truly breaking' release (we deprecate all old function names and such); can we introduce this new system and put up depreciation notices for AgentsPlots?

What I see as a solution:
- Introduce InteractiveChaos as the new plotting mechanism we are transitioning to (should we do the InteractiveDynamics rename now, so we don't have to change it in the docs in the near future?).
- Archive AgentsPlots, don't support it anymore, but still have the few examples that need it (SIR, flock, bacteria, zombies) building with it.
- Put `!!! note` or `!!! warning` block in the SIR, flock, bacteria, zombies examples that these solutions will change in the near future.

Datseris:
sounds good

Libbum:
Cool. For the moment I'm going to write something custom for Battle Royale, and we can look at the complexities of that example later. (Doing that now)

Would be nice to be able to have this `abm_plot!` solution to enable the LightGraphs.jl and Sugarscape examples to work as they used to.

I'll reset the other examples back to AgentsPlots and re-enable them here too.

Datseris:
Okay I'm done reworking the internal code. See here: https://github.com/JuliaDynamics/InteractiveChaos.jl/pull/32

As explained in the docstring of `abm_plot`, the function returns `fig, abmstepper`. You can update the fig interactively in a loop by doing:
```
Agents.step!(abmstepper, model, agent_step!, model_step!, n)

```

See for example the source code of `abm_video`:


```julia
    # add some title stuff
    s = Observable(0) # counter of current step
    if title ≠ "" && showstep
        t = lift(x -> title*", step = "*string(x), s)
    elseif showstep
        t = lift(x -> "step = "*string(x), s)
    else
        t = title
    end

    fig = Figure(; resolution, backgroundcolor = DEFAULT_BG)
    ax = fig[1,1] = Axis(fig; title = t, titlealign = :left)
    abmstepper = abm_init_stepper_and_plot!(ax, model; kwargs...)

    record(fig, file, 1:frames; framerate) do j
        Agents.step!(abmstepper, model, agent_step!, model_step!, spf)
        s[] += spf; s[] = s[]
        # (title ≠ "" || showstep) && (abmax.title = titlef(s))
    end
    return nothing
```

I haven't updated the slider app yet, but that will be fast. However, I have an extremely important presentation for Monday and I will spend the rest of my weekend exclusively on that. I'll return here on Monday afternoon and see what remains to be done.

Libbum:
Cool. I'll need some help on making Battle Royale work nicely. I've made a decent static plot but don't want to reinvent the wheel for the video.

Hopefully these changes help with that. But sure&mdash;good luck, we'll pick it up again Monday.

Libbum:
To be completed before merge
- [ ] Rename the Interactive Application page to the Plotting page and extend discussion around the new methods
- [ ] Update the LightGraphs.jl and Sugarscape examples with the new extensions of InteractiveChaos
- [ ] Add a depreciation warning about `AgentsPlots` in the examples that still use it (SIR, flocking, bacteria, zombie)
- [ ] Rename InteractiveChaos to InteractiveDynamics and adjust the wording here accordingly
- [ ] Fix, then merge InteractiveChaos#32 and tag a new release - drop the custom branch in `make.jl`
- [ ] Do a once-over of the final output and make sure all text now references the new system not the old
- [ ] CHANGELOG update, additional wording in the intro/tutorial section perhaps

examples/flock.jl
```diff
@@ -123,6 +123,13 @@ using AgentsPlots
 gr() # hide
 Random.seed!(23182) # hide
 cd(@__DIR__) #src
+
+!!! note
+    [AgentsPlots.jl](https://github.com/JuliaDynamics/AgentsPlots.jl) is now depreciated
```
Datseris, line 128:
```suggestion
    [AgentsPlots.jl](https://github.com/JuliaDynamics/AgentsPlots.jl) is now deprecated
```

AayushSabharwal:
Should I make another PR for sugarscape?

Libbum:
Of you have a solution, sure.. I only have a half functioning one.

AayushSabharwal:
There's still the issue that providing constant values for `color` etc. of `abm_plot` throws an error

Libbum:
This should be fixed in the new branch, try `] up` to update your local copy to the current commit.

Libbum:
Could do with some eyes on the `schoolyard` example:
```julia
model = schoolyard()

s = Observable(0) # counter of current step
t = lift(x -> "Playgound Dynamics, step = "*string(x), s)
f, abmstepper = abm_plot(model, as = 15)
ax = contents(f[1, 1])[1]
## ax.title = t[]
scatter!([50 50]; color = :red) # Show position of teacher
limits!(ax, 0, 100, 0, 100)
hidedecorations!(ax)
record(f, "schoolyard.mp4", 1:40; framerate = 15) do i
    Agents.step!(abmstepper, model, agent_step!, dummystep, 1)
    s[] = i; s[] = s[]
    ## ax.title = t[]
end
```
Finding it quite difficult to get the title working in a nice way. The commented out sections 'work', but the text jumps all over the place.

I'm also just looking to add that one scatter point on top of the abm plot, and it seems like a lot of boiler plate to get that to happen. Perhaps I've not quite followed how best to use the `abmstepper` here...

AayushSabharwal:
Made PR https://github.com/JuliaDynamics/Agents.jl/pull/375 for sugarscape

Libbum:
More things to fix/look into:
- [ ] The social distancing videos look wrong. Like they only show 10% of the model. I think there's an issue with the limit finder there.
- [ ] There's some Fontconfig error spitting out in the CI now. Not sure why they're appearing, but we get this kind of output in places:
image
- [ ] Output from `abm_plot` is not being displayed, even with a `display(f)` call.

Datseris:
With https://github.com/JuliaDynamics/InteractiveChaos.jl/pull/32 , what remains to be done here (besides the fact that models that rely on plotting arbitrary polygons are impossible)?

Libbum:
Checklists [here](https://github.com/JuliaDynamics/Agents.jl/pull/370#issuecomment-766085068) and [here](https://github.com/JuliaDynamics/Agents.jl/pull/370#issuecomment-766372419) still have things to check off.

Datseris:
Not sure what's the problem with the social distancing. The function that finds the limits is:
```
function modellims(model)
    if model.space isa Agents.ContinuousSpace
        e = model.space.extent
        o = zero.(e) .+ 0.5
    elseif model.space isa Agents.DiscreteSpace
        e = size(model.space.s) .+ 1
        o = zero.(e)
    end
    return o, e
end
```
so that seems correct to me...?

Libbum:
> so that seems correct to me...?

It is, but look at the output: 
https://juliadynamics.github.io/Agents.jl/previews/PR370/examples/socialdist5.mp4

Datseris:
What's wrong with it? Seems to get the periodicity correct? Where do the compartments start *precicely*? Is the +0.5 I have incorrect?

Libbum:
> What's wrong with it?
- number of agents are dramatically smaller
- only one agent gets infected (none get cured)
- there elastic behaviour is quite 'far' away in comparison to the master example, so it seems zoomed in to me

> Seems to get the periodicity correct?

I dont think so. Take a look at bottom left guy, he's bouncing off of nothing on the right when he goes through the wall.

Compartments shouldn't matter in a plotting context right? +0.5 is definitely incorrect, the system is 0.0 - 1.0.

Libbum:
`o = zero.(e) .+ 0.5`, Oh, no this is definitely wrong. It's always from zero.

Datseris:
i'm checking this out locally now

Datseris:
Works locally for me without a problem with InteractiveChaos 10.3


https://user-images.githubusercontent.com/19669089/105643707-aff0e100-5e91-11eb-98f8-c6ea5b5f7d25.mp4

Libbum:
That looks better! I think this run goes out to 1000 in master though.

(also, check slack)

AayushSabharwal:
https://juliadynamics.github.io/JuliaDynamics/
Wouldn't the Flocking example here also require updating when this is merged?

Datseris:
Yes, that is easy to do tho' don't worry :)

AayushSabharwal:
Awesome. I stumbled across it and thought I'd point it out in the off chance it might not get changed.

PR #372, opened 2021-01-20
Preview cleanup action

Libbum:
Adds a new action that should keep the previews in `gh-branches` a little cleaner than historically. Taken from  CliMA/TimeMachine.jl and Documenter.jl docs.


Datseris:
So did this work? Can I also use it in DynamicalSystems.jl?

Libbum:
We need to close/merge an open PR before we'll see it in action. I'll do a manual prune of the others later on (before we tag v4).

Libbum:
Now we can see if 374 gets removed by the action.

PR #383, opened 2021-01-26
Fix typo in GridSpace periodic default

fbanning:
See https://github.com/JuliaDynamics/Agents.jl/blob/88af0ec548ade7ec5d805b7aa7c85e49dadbb931/src/spaces/grid.jl#L26-L65 for the correct defaults.


Libbum:
Thanks!

Issue #155, opened 2020-03-20
Allowing agents to have size


kavir1698:
One way is to Extend the SQLite database So that it has one more column, agent's diameter. This will be used to determine when agents interact with one another. Currently, only if the center of an agent is in interaction radius of another one's, they interact.

Datseris:
Although this is an okay idea, I am not sure whether we should enforce agents to have a field `diameter`. It is not always the case that agents in continuous space have size.

Datseris:
To the best of my understanding, the search happens only in this function: https://github.com/JuliaDynamics/Agents.jl/blob/master/src/core/continuous_space.jl#L217  .

One can extend the `left` and `right` limiters by an agent diameter. But I want to propose an alternative than adding a field `diameter` to the agent, as this requires to add necessary strictness.

I propose to define a function
```julia
agentsize(a::AbstractAgent) = ntuple(x -> 0.0, length(agent.pos))
```

and invite users to extend this function via multiple dispatch for their agent. This is very Julia and will also allow scenarios where agents are not spheres.

Datseris:
@lhupe the bacteria of your simulations have physical size, _and_ are not spheres. Would you mind telling us how you guys tackle this?

lhupe:
> How do you find nearest bacteria to a given bacterium? 

We don't. Since the interactions between two bacteria are completely described by a continuous force, we just find those that are closeish (by binning the bacteria spatially) and compute that force term (which is zero for all bacteria pairs that do not intersect).
In order to simulate "hard" bacteria, we just use a very large force. 

> Do you take into account its orientation and shape?

Kind of. For the actual force computation, we do need a complete representation of the bacteria. For choosing the size of the bins that we use for finding closeish bacteria, we use a worst case approximation for the interaction radius.

Datseris:
A follow-up question then: Do you find all pairs of bacteria in one-go? Or you go bacterium by bacterium, and find for a given bacterium all of its pairs?

lhupe:
We loop over all bins, compute the forces of all pairs of bacteria within the bin, then loop over nearby bins and compute the forces of all pairs of bacteria from the one bin and bacteria from the other bin.

Since I don't really understand the previous sentence myself, have some pseudocode!
```
for firstbin in bins
    for (i, b1) in enumerate(firstbin)
        for b2 in firstbin[i:end]
            do_stuff(b1, b2)
    
    for secondbin in nearby_bins(firstbin)
[...]

PR #384, opened 2021-01-27
Fractal Growth example

AayushSabharwal:
I just have preliminary work on the example. It currently runs, and is configurable. I'm not fully satisfied with marker size, which will require some trial end error. I'm also looking for better implementations for how velocity is calculated. If there's any other feedback, I'd love to hear it


Libbum:
Thanks @AayushSabharwal, I wont have time to look into this until tomorrow, but at first glance it looks nice.

Libbum:
This looks absolutely amazing!

examples/diffusion_limited_aggregation.jl
```diff
@@ -0,0 +1,123 @@
+using Agents, Random, LinearAlgebra
```
Libbum, line 1: So the next step will be to write up some documentation for this. You've seen the `examples` section already, have go at doing something similar, I'll add some comments about flow here.

examples/diffusion_limited_aggregation.jl
```diff
@@ -0,0 +1,123 @@
...
+) = Particle(
+    id,
+    pos,
+    (0.0, 0.0),
+    radius,
+    is_stuck,
+    [0.0, 0.0, twist_clockwise ? -1.0 : 1.0],
+)
+
+properties = Dict(
```
Libbum, line 25: This is best to go in the initialization function, so that we can also add this to the `models` section.

examples/diffusion_limited_aggregation.jl
```diff
@@ -0,0 +1,123 @@
...
+    :spawn_count => 0,
+    :min_radius => 1.0,
+    :max_radius => 2.0,
+)
+
+rand_circle() = (θ = rand(0.0:0.1:359.9); (cos(θ), sin(θ)))
+particle_radius(min_radius::Float64, max_radius::Float64) = min_radius <= max_radius ?
+rand(min_radius:0.01:max_radius) : min_radius
+function agent_step!(agent::Particle, model)
+    if agent.is_stuck
```
Libbum, line 40: Can simplify this to `agents.is_stuck && return`

examples/diffusion_limited_aggregation.jl
```diff
@@ -0,0 +1,123 @@
...
+        if model[id].is_stuck
+            agent.is_stuck = true
+            model.spawn_count += 1
+            return
+        end
+    end
+    radial = model.space.extent ./ 2.0 .- agent.pos
+    radial = radial ./ norm(radial)
+    tangent = Tuple(cross([radial..., 0.0], agent.twist_axis)[1:2])
+    agent.vel =
```
Libbum, line 53: Yep, I think this is better than what you had before and makes the most sense for the moment. Maybe this can be improved upon in future versions, but fine for now.

examples/diffusion_limited_aggregation.jl
```diff
@@ -0,0 +1,123 @@
...
+    :attraction => 0.0:0.01:2.0,
+    :speed => 0.0:0.01:2.0,
+    :wiggle => 0.0:0.01:2.0,
+    :twist => 0.0:0.01:2.0,
+    :clockwise_fraction => 0.0:0.01:1.0,
+    :min_radius => 0.5:0.01:3.,
+    :max_radius => 0.5:0.01:3.,
+)
+
+interactive_abm(
```
Libbum, line 115: Lets wait until we can use `abm_data_exploration` with InteractiveDynamics before merging, but it looks really great here.

examples/diffusion_limited_aggregation.jl
```diff
@@ -0,0 +1,123 @@
...
+            particle_radius(model.min_radius, model.max_radius),
+            rand() < model.clockwise_fraction;
+            pos = (rand_circle() .+ 1.0) .* model.space.extent .* 0.49,
+        )
+        add_agent_pos!(particle, model)
+        model.spawn_count -= 1
+    end
+end
+
+function initialize_model(;
```
Libbum, line 72: Better in the example to start with this, and add the dynamics (`_step!` functions) later.

examples/diffusion_limited_aggregation.jl
Libbum: We can make this change now!

AayushSabharwal:
Thoughts on renaming "Diffusion-limited aggregation" to "Fractal growth"? The current name is a bit of a mouthful, and doesn't really convey what it's about to someone who has no background in it's specific area of study.

Libbum:
Yep, sounds good! We can certainly state in the introduction that this is a diffusion-limited aggregation simulation.

AayushSabharwal:
How can I see how the documentation looks? Also, what kind of videos should be in there? I could record the interactive session, or maybe `abm_video` it running with default parameters?

Libbum:
> How can I see how the documentation looks?

We have previews in our CI, but cant use them for your current PR since it's not connected to our build system. If you go into the docs directory and instantiate the packages locally then you can run `make.jl` to build a local copy. 

Oh, but I'll have to push some changes for that to make sense. I'll do that now. I'm not sure if I'll be able to look over the tonight, but tomorrow probably.

> Also, what kind of videos should be in there? I could record the interactive session, or maybe `abm_video` it running with default parameters?

Since you've put in the work to get the interactive session running, it'd be best to record that I think. @Datseris, for the examples in InteractiveDynamics, what are you doing there? Just using a screen recorder?

Libbum:
Exceptional work @AayushSabharwal, it looks super great. I've just played with it for 15 minutes getting it to generate different fractal patterns.

Have updated a few bits of the text, very minor changes. Changed some of the `variables` to document links: [`variables`](@ref), which will make the documentation hyperlink to the relevant place.

I'll have a go at making a video for it tomorrow.

I like your solution to agent sizes, at leas for simple circular interactions. We can use some of this as a start for tackling #155.

Not all the time, but generally we put a distraction free version of the example into the `src/models` section, so perhaps that could be done here.
Lastly, we should reference the original page the example was shown in an appropriate manner.

Datseris:
> Since you've put in the work to get the interactive session running, it'd be best to record that I think. @Datseris, for the examples in InteractiveDynamics, what are you doing there? Just using a screen recorder?

Hi, thanks this indeed looks awesome and fractals are extremely dear to my heart. One of the main things that brought me where I stand now, as a researcher of nonlinear dynamics, was doing fractal art on programs like apophysis and the like!

To my understanding we can use `abm_video` directly here, right? The output can be directly put in the picture, just like in the open PR we have #370. This PR works totally fine last time I checked; the reason we haven't merged it is simply because we cannot have custom polygons at the moment. However I Think it is a good idea to merge this PR, and simply continue using the old "AgentsPlots" style for things that cannot be done with Makie yet. Because `abm_video` is just super convenient.

For recording interactions that also show my mouse I use `ScreenToGif` : https://www.screentogif.com/

Libbum:
Yep. We can definitely use abm_video, it's nice to play around with the settings interactively though!

Thanks. I'll take a look at screentogif tomorrow.

AayushSabharwal:
I'm trying to build the docs locally. Currently it's getting stuck on  `┌ Debug: Running ExpanderPipeline on examples\battle.md`. Some window keeps periodically opening for a fraction of a second then closing again. I'm not sure if this is supposed to happen. I'll wait a while and see what happens, I've noticed the checks on PRs tend to take a while on the docs part too.

AayushSabharwal:
I've added this to Models, in the pattern of the rest of the examples. Should I commit the changes? Is there a specific reason none of the others use the `@agent` macro?

Libbum:
Sure, commit them!

`@agents` is very new and is optional. It's good we have a blend of both the normal struct and ones built with the macro though, so keep it like this.

Doc builds are at least 35 minutes now. Mostly because of zombies. Plotting the OSM is extremely slow and needs to be addressed soon. 
I've looked at them myself and other than the latest changes you've just made they were good. Will take a look at the new stuff soon. I think it's ready though.

Datseris:
Why not pre-run the zombies example and upload the mp4 to the juliadynamics website repo, which also host the Agents.jl animation logo? Then we can skip running most of the Map example...

Datseris:
Also, can you please post the video here?

AayushSabharwal:
This is the video for fractal growth
https://user-images.githubusercontent.com/23384717/106379730-03a08480-63d4-11eb-97d2-7d6202eb7f4c.mp4

Libbum:
Re: zombie video, great idea. Will do that in another PR.

Datseris:
That looks ridiculously cool! You can increase the `spf` (steps per frame) and decrease the framerate to make an overall smaller sized video

Libbum:
- Just added some changes that should let the CI build the `abm_play` call (GLMakie cannot be used there without a huge effort).
- Made the function calls in the models section unique (@AayushSabharwal, since we just return functions to the user here, we cannot hope that multiple dispatch sorts out name ambiguities, so we just make them unique)
- Recorded a little interaction video, will wait for JuliaDynamics/JuliaDynamics#27 is merged before we do the same to this

Since we can't preview this one, I'm happy to merge it and do any final adjustments after the fact. I'll monitor it.

Libbum (approved):

AayushSabharwal:
Got it. My bad, I forgot to make those changes in `Models`. I'll check out the changed video too in a while.

Libbum:
Oh, it's no problem! Don't expect you to know all of the quirks of our system.

Libbum:
https://juliadynamics.github.io/Agents.jl/dev/examples/fractal_growth/

Looking good! A few small things I'll adjust, but otherwise a great example.

AayushSabharwal:
I'll add the `nothing #hide` lines to suppress unwanted output.

Libbum:
I've pushed some changes to master already that suppress most important things (like the Random call). The Makie cache spill is a small issue which I tried to fix in `docs/make.jl`. It didn't work, but it's not that big of a deal. Otherwise the output that's there is fine.

We had a bit of a discussion about whether to hide essentially irrelevant output or not a while back, and it was decided that having these small verifications along the way that a user following along can compare against might be helpful.

So there's no need to suppress anything else.

PR #388, opened 2021-01-28
Mesa benchmark: Remove data collection code

rht:
To make the comparison of the speed and LOC more fair. How do I update the benchmark table at https://juliadynamics.github.io/Agents.jl/stable/comparison/#An-in-depth-comparison-of-four-frameworks ?

### DIFF
--- a/benchmark/compare/Mesa/ForestFire/model.py
+++ b/benchmark/compare/Mesa/ForestFire/model.py
@@ -1,6 +1,5 @@
 from mesa import Model
 from mesa import Agent
-from mesa.datacollection import DataCollector
 from mesa.space import Grid
 from mesa.time import RandomActivation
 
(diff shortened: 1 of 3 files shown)


rht:
E.g.:
current:
```
Mesa ForestFire (ms): 1050.3123858943582
```

with data collection removed:
```
Mesa ForestFire (ms): 760.3026318829507
```

Libbum:
Great, thanks! Do you see any other issues in the Mesa benchmarks?

rht:
I'm still reading. I think the data collection one is the low hanging fruit. There could be further optimization like data class to remove boilerplate line like `self.height = height` (to make it Julia-like), but this practice is not commonly used in Mesa I think.

Libbum:
Sure. It would be good to get them as close to 'Mesa standard' as possible, but as you say - reflective of what the benchmark is actually doing. These were done by another contributor, I thought it was already done :/

rht:
There are some further adjustments, e.g. `self.running` in forest fire is no longer relevant because it is used only when `run_model()` is called, which keeps doing `step()` until `self.running` is False. But this is outside the scope of this PR, which purpose is to remove data collection.

Libbum:
Thanks. I'll take a look into that. Appreciate you bringing these things to my attention. If you notice anything else let me know!

rht:
- In Mesa's WolfSheep/model.py, all the code associated with `verbose` can be removed.
- In Mesa's WolfSheep/model.py, `run_model()` can be removed, and the explicit number of steps can be specified in benchmark.py.

PR #394, opened 2021-01-29
Update Mesa benchmark code

rht:
WIP until I can no longer find unused code.

### DIFF
--- a/benchmark/compare/Mesa/Flocking/boid.py
+++ b/benchmark/compare/Mesa/Flocking/boid.py
@@ -56,51 +56,26 @@ class Boid(Agent):
         self.separate_factor = separate
         self.match_factor = match
 
-    def cohere(self, neighbors):
-        """
-        Return the vector toward the center of mass of the local neighbors.
-        """
-        cohere = np.zeros(2)
-        if neighbors:
-            for neighbor in neighbors:
-                cohere += self.model.space.get_heading(self.pos, neighbor.pos)
-            cohere /= len(neighbors)
-        return cohere
-
-    def separate(self, neighbors):
-        """
-        Return a vector away from any neighbors closer than separation dist.
-        """
-        me = self.pos
-        them = (n.pos for n in neighbors)
-        separation_vector = np.zeros(2)
-        for other in them:
-            if self.model.space.get_distance(me, other) < self.separation:
-                separation_vector -= self.model.space.get_heading(me, other)
-        return separation_vector
-
-    def match_heading(self, neighbors):
-        """
-        Return a vector of the neighbors' average heading.
-        """
-        match_vector = np.zeros(2)
-        if neighbors:
-            for neighbor in neighbors:
-                match_vector += neighbor.velocity
-            match_vector /= len(neighbors)
-        return match_vector
-
     def step(self):
         """
         Get the Boid's neighbors, compute the new vector, and move accordingly.
         """
-
         neighbors = self.model.space.get_neighbors(self.pos, self.vision, False)
-        self.velocity += (
-            self.cohere(neighbors) * self.cohere_factor
-            + self.separate(neighbors) * self.separate_factor
-            + self.match_heading(neighbors) * self.match_factor
-        ) / 2
+        N = 0
+        match_vector, separation_vector, cohere = np.zeros((3, 2))
+        for neighbor in neighbors:
+            N += 1
+            heading = self.model.space.get_heading(self.pos, neighbor.pos)
+            cohere += heading
+            if self.model.space.get_distance(self.pos, neighbor.pos) < self.separation:
+                separation_vector -= heading
+            match_vector += neighbor.velocity
+        N = max(N, 1)
+        cohere = cohere / N * self.cohere_factor
+        separation_vector = separation_vector * self.separate_factor
+        match_vector = match_vector / N * self.match_factor
+        #self.velocity = (self.velocity + cohere + separation_vector + match_vector) / 2
+        self.velocity += (cohere + separation_vector + match_vector) / 2
         self.velocity /= np.linalg.norm(self.velocity)
         new_pos = self.pos + self.velocity * self.speed
         self.model.space.move_agent(self, new_pos)
--- a/benchmark/compare/Mesa/Flocking/model.py
+++ b/benchmark/compare/Mesa/Flocking/model.py
@@ -51,7 +51,6 @@ class BoidFlockers(Model):
         self.space = ContinuousSpace(width, height, True)
         self.factors = dict(cohere=cohere, separate=separate, match=match)
         self.make_agents()
-        self.running = True
 
     def make_agents(self):
         """
--- a/benchmark/compare/Mesa/ForestFire/model.py
+++ b/benchmark/compare/Mesa/ForestFire/model.py
@@ -1,7 +1,7 @@
 from mesa import Model
 from mesa import Agent
 from mesa.space import Grid
-from mesa.time import RandomActivation
+from mesa.time import BaseScheduler
 
 class TreeCell(Agent):
     """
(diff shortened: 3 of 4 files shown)


Libbum:
In the docs:
> For LOC, we use the following convention: code is formatted using standard practices & linting for the associated language. Documentation strings and in-line comments (residing on lines of their own) are discarded, as well as any benchmark infrastructure. NetLogo is assigned two values since its files have a code base section and an encoding of the GUI. Since many parameters live in the GUI, we must take this into account. Thus 375 (785) in a NetLogo count means 375 lines in the code section, 785 lines total in the file.

The discrepancies you have with the count is probably 'white space between function calls'. Since MASON code is distributed in binary format and the NetLogo files have GUI code embedded, it was difficult to settle on a decent comparison. So using python / julia linters on the code from Mesa and Agents seemed like the best option - making the style guides be part of the process.

We didn't want to make this a race to the bottom. For example
```julia
function flocking(;
    n_birds = 100,
    speed = 1.0,
    cohere_factor = 0.25,
    separation = 4.0,
    separate_factor = 0.25,
    match_factor = 0.01,
    visual_distance = 5.0,
    extent = (100, 100),
)
```
11 lines just for this function header, since Julia has a 90 character line limit.
We could drop that to two being sneaky
```julia
more_kws = (sf = 0.25, mf = 0.01, vd = 5.0, extent = (100, 100), spacing = vd / 1.5)    
function flocking(; n_birds = 100, speed = 1.0, cf = 0.25, separation = 4.0, more_kws...)
```
or just ignore the linter and make it one line...

rht:
Worry not. The cleanup I have done so far are real cruft, and have already been applied to the Mesa repo (https://github.com/projectmesa/mesa/pull/981). The code in github.com/projectmesa/mesa has been formatted with Black, which will do the same multiline arguments for flocking.

The concern with the whitespace is real though. People can't tell that the Python version has more lines just because of some whitespace, they'd instead infer that Mesa has more boilerplate code.

Corvince:
Hi, I am one of the mesa maintainers. Great discussion so far and really great and motivating benchmark.

I found another major bottleneck for mesa: You were using `RandomActivation` as a scheduler, whereas Julia uses "fastest", which mesas closest equivalent (esp. in terms of performance) is `BaseScheduler`. This gives me about a 3x increase in mesas performance for forestfire. This would then be in line with the 7x speed difference between Julias predator/prey and mesas WolfSheep, which both use RandomActivation. I haven't looked into the Flockers example yet, but will in the soonish future.

Since I can't contribute to this PR directly, [I opened a PR with @rht s base branch](https://github.com/rht/Agents.jl/pull/1). If @rht merges this PR, I think it should show up here.

rht:
I have merged @Corvince 's PR.

rht:
For reference, here is the benchmark output with `BaseScheduler`
```
Mesa ForestFire (ms): 367.46351909823716
```

The same, but instead with PyPy3:
```
Mesa ForestFire (ms): 65.83644193597138
```
(2x slower)

rht:
The Agents.jl version on my laptop
```
Agents.jl ForestFire (ms): 32.785682
```

rht:
Why not use RandomActivation equivalent on Agents.jl for ForestFire?

Libbum:
The lines-of-whitespace question isn't too big of a deal from my side. Once you guys are happy with the representation of the Mesa runs, let me know a count and why, we can work off that. Julia and Java don't have the same whitespace concerns of Python, so I can adjust the numbers there.

> Why not use RandomActivation equivalent on Agents.jl for ForestFire?

Essentially since MASON doesn't use it and I seriously do not wish to work with Java and more than I need to! :man: :gun:

rht:
Also, just a reminder, this line in docs/src/index.md needs to be updated
> Agents.jl was originally inspired by the [Mesa](https://github.com/projectmesa/mesa) framework for Python, but has since then departed in design, leading to a dramatically simpler and cleaner API, besides having obvious performance benefits (more than 30 times better performance than Mesa in some cases, see our [Agents.jl Performance and Complexity Comparison](@ref)).

10x instead of 30x?

rht:
I found 2 discrepancies in the Flocking example.
1. The `separate` vector is divided by N in Agents.jl (https://github.com/JuliaDynamics/Agents.jl/blob/0e472cc6e7c2d310e649f85b1a78e1569d9ef583/src/models/flocking.jl#L77), but it is not in Mesa (https://github.com/projectmesa/mesa/blob/e8757cfe7c84b1d5ad980a5b7daab9880813b1f8/examples/boid_flockers/boid_flockers/boid.py#L80)
2. The formula for velocity update is different: https://github.com/projectmesa/mesa/blob/e8757cfe7c84b1d5ad980a5b7daab9880813b1f8/examples/boid_flockers/boid_flockers/boid.py#L99-L103 vs https://github.com/JuliaDynamics/Agents.jl/blob/0e472cc6e7c2d310e649f85b1a78e1569d9ef583/src/models/flocking.jl#L79 (they differ by `self.velocity / 2`)

I have looked at the NetLogo and MASON version, but they are too cryptic. Which version is correct?

Libbum:
For separate:
[Mesa normalises](https://github.com/JuliaDynamics/Agents.jl/blob/0e472cc6e7c2d310e649f85b1a78e1569d9ef583/benchmark/compare/Mason/Flocking/Flocker.java#L130). 
Since NetLogo is doing things based on turtle movement, we are not comparing things explicitly there. It takes the average of the coherence angles. I've taken a single NetLogo bird-turtle and tracked its movement. Can confirm that Agents.jl birds were following its path.

For velocity:
That's most likely a mistake in the Agents code, nice catch! Will take a closer look.

rht:
For separate: you meant MASON; it looks like the Mesa version is wrong because of the lack of normalization.

Libbum:
Sorry, yes I did! I've been doing that constantly throughout this comparison process.

Libbum:
Think there's some more investigation to be done with the Flockers. Visually, these examples are not matching up with NetLogo when I introduce the alteration to Agents. Also, I suspect that when Mesa adds normalisation it'll behave quite differently.
Can you give me a quick way of plotting the Mesa example to compare on-screen?

rht:
@Libbum you can visualize the animation on a web browser for Mesa if you do `python run.py` in the `examples/boid_flockers` of the Mesa repo.

rht:
From the wikipedia page of boids (https://en.wikipedia.org/wiki/Boids), there is an external link on the pseudocode: http://www.vergenet.net/~conrad/boids/pseudocode.html.

The takeaway is that:
1. `separate` is not normalized by division by N, see rule 2 of the pseudocode
2. the velocity update formula is `b.velocity = b.velocity + v1 + v2 + v3` (there is no division by 2 at all; in fact, I couldn't find any division by 2 in the MASON version)

This is probably not authoritative enough. Better go to the original paper.

Libbum:
@rht, #474 is going to move our benchmarking to a dedicated repo. I think we still have some open questions here before we can successfully resolve the flocking code. 

Are you happy with the code reduction so far? I'm thinking the best short term solution would be to merge those changes, then open a new issue on [ABM_Framework_Comparisons](https://github.com/JuliaDynamics/ABM_Framework_Comparisons) concerning the flockers example.

Another solution would just be to re-target this PR over on the new repo and we can continue discussions without merging current work first.

rht:
Yeah, let's merge this PR first. The flockers example problem is only 1 line of code, so wouldn't affect the line count. Not sure about the benchmark time though. I have gone through the 4 examples (Flocking, ForestFire, Schelling, WolfSheep) and couldn't find much to be made equivalent in terms of implementation detail for now.

Though my last question is that, in the Mesa WolfSheep code, there is `RandomActivationByBreed` that is not yet part of Mesa standard library -- is this code include in the line count of Mesa's WolfSheep? Though one could argue that it is a feature of Agents.jl to implement `by_type` scheduler.

Libbum:
Ok great, thanks!

We're doing a redesign of our wolf sheep too, so I'll keep that question in mind and adjust accordingly. It'll be better to keep it close to standard libraries.

rht:
Since the benchmark repo has already existed (https://github.com/JuliaDynamics/ABM_Framework_Comparisons), I think it is better if I redo the commits into that repo? Just need your confirmation before doing this.

Libbum:
Sure, go ahead!

rht:
Superseded by https://github.com/JuliaDynamics/ABM_Framework_Comparisons/pull/4.

PR #400, opened 2021-01-31
Attempt to fix fontconfig

Libbum:
Something's wrong with fontconfig in the CI image. It's been a problem since we started using CairoMakie, but I don't think its in that package. 

Most online discussions state that there's a version missmatch somewhere, but since we spool up a new container, that seems odd..

Trying some solutions in this branch. May end up pretty noisy.


Libbum:
Problem: Fontconfig_jll, which is needed by GR and Cairo build against fontconfig 2.13, whereas the ubuntu container is at 2.12.6.

Issue #277, opened 2020-08-13
Obstacles


dunefox:
I need to simulate agents moving around between points in a certain environment to generate log files according to their contacts (under, say, 1.5 meters generates a log entry with the agent ids): similar to this https://jaantollander.github.io/crowddynamics/multiagent.html#field.

Is this possible currently?

Datseris:
Hi, no this is not possible "out of the box". 

---

Of course you can do everything manually yourself there isn't any inherent limitation from Agents.jl.

You can e.g. create an array that is as large as your grid and has 0s if there is no obstacle and 1s if there is an obstacle.  Then it is possible to manually check the distance of Agents from obstacles. Notice however that your description is not adequate for me to understand **exactly** what you want to do:

> to generate log files according to their contacts (under, say, 1.5 meters generates a log entry with the agent ids): 

I just cannot imagine what this means and need more information.

dunefox:
I should have mentioned that I need a continuous simulation, not a grid.

Basically I want to check if two or more agents are closer than 1.5 meters to each other (this should be easy with ```interacting_pairs(model, 0.015, :all)```, I think). If this happens I want to log the agent ids and the duration. 

Ideally, I would like to have spawn points, target points, and obstacles as well having agents move between the points. Imagine a corona simulation with an event log if people come closer to each other than 1.5 meters. I hope it makes more sense now.

Datseris:
I see. From what you have said so far, there is only one thing not possible: Automatically avoiding obstacles. The rest you can already do with the API of Agents.jl. As you said `interacting_pairs(model, 0.015, :all)` gives you what you want, while you can simply orient the agent's velocity to their "target" and then `move_agent!`. 

So yeah, entities that are "obstacles" and "automatically calculating trajectories that avoid them" would be super cool to have, but I fear that it will take tremendous development effort (and also the skills to do it, I think this path findingalgorithms like [A*](https://en.wikipedia.org/wiki/A*_search_algorithm)) .

Datseris:
Relevant links from the crowddynamics repo:

https://github.com/jaantollander/crowddynamics/tree/master  (seems not active anymore)

https://jaantollander.com/post/how-to-implement-continuous-time-multi-agent-crowd-simulation/ (some blog with guidance/discussion on how to do crowd dynamics)

Libbum:
Two libraries that may be of assistance here:
[...]

PR #405, opened 2021-02-02
Pathfinding in GridSpace

AayushSabharwal:
`GridSpace` is arguably the easiest target for pathfinding. This has been implemented through the A* algorithm, extended to arbitrary dimensions.
The pathfinding data is managed by the `AStar` struct. It stores:
- Agent paths through a `Dict` mapping agent ID to a linked list containing grid positions along the path that agent will follow
- Dimensions of the grid. This has been kept separate since custom solutions might need it, for example `BattleRoyale`.
- `D`-dimensional neighborhood offsets, either Moore or VonNeumann depending on the `M` type parameter
- admissibility, which allows approximating the path for a substantial increase in performance
- The cost metric used to evaluate the distance between two points. This is how heightmaps are handled.
- Periodic nature of the grid in the `P` type parameter

The API is as follows
- A new dummy `Pathfinder` struct holds parameters for creating `AStar` and serves as a "gateway" of sorts to add pathfinding to `GridSpace`
- `GridSpace` now has a `pathfinder` field, which is `Union{Pathfinder,Nothing}`. A model with pathfinding is created by specifying the pathfinder through the kwarg of `GridSpace`. It defaults to `nothing`, indicating no pathfinding. A pathfinder can be created by passing in a `Pathfinder`, with the appropriate fields (keyword arguments) overriden. For example, `GridSpace((10, 10); pathfinder=Pathfinder(admissibility=0.3))` creates a pathfinder with `admissibility=0.3` and the rest of the fields default. To use all default values, the default constructor `Pathfinder()` is used.
- The `set_target!` function calculates and stores the path from the agent's current position to the specified one
- The `move_agent!` function now has a new method, that moves the agent along its precalulated path. If the path doesn't exist or is empty, the agent doesn't move
- `kill_agent!` has a new method, which additionally removed path data (if any) from the `AStar` struct
- A couple utility functions like `heightmap`, `walkmap`, `is_stationary` to query pathfinder data.

Apart from these, there are some details exposed that allow for custom solutions to deal with more complicated cases
- Custom cost metrics can be defined with the `CostMetric{D}` abstract type. Each metric requires a custom `delta_cost` method, so that is exposed as well. `DirectDistance` has two methods depending on whether the `AStar` struct has a Moore or VonNeumann neighborhood
- `find_path!` is exposed, in cases where the user wants to manually manage multiple pathfinders or the pathfinding space doesn't directly map to the `GridSpace`

[...]

PR #412, opened 2021-02-06
Add FMP example

rmcsqrd:
As discussed in #277 I added in an implementation example of the [Force Based Motion Planning (FMP) Paper](https://arxiv.org/pdf/1909.05415.pdf). 

I do not regularly commit to large projects so please let me know if I am missing any best practices. The two items that I think might be a sticking point are:
- the `using` dependencies. I don't think LinearAlgebra, Colors or ProgressMeter are standard within the Agents.jl PROJECT.toml. Is it okay to leave them in?
- I have not used Documenter.jl before. I tried to match the existing example files formatting. I checked the generated markdown file as noted via CONTRIBUTING.md but it was not rendering the gifs. 

Other than that any feedback is welcome/much appreciated.


Libbum:
This is an awesome example! Really cool! It will take me a while to read into it and understand how best to move forward. There are some ways that we can make it more idiomatic Agents.jl code, and then the question of if it makes sense to be a new addition to the API.

@AayushSabharwal is doing a good job building up A* for `GridSpace` in #405, which may be a little complicated to extend to a Continuous example, so it may mean that this algorithm is definitely something we can consider for the path-finding suite we want to build up. You may be right though, it could be better in its own package and Agents just calls on it.

Great job!

Libbum:
Oh, dependencies don't matter. We can use however many we want when building the documentation. It's how many we choose to add into Agents itself that we try to keep small.

Building docs is a bit of a pain to get right locally, don't worry too much about that&mdash;I can make sure they look as they should.

rmcsqrd:
Thanks for the kind words. Any feedback is welcome. Happy to explain anything that is unclear while you're looking into it.

rmcsqrd:
@Libbum Have you had a chance to review this/consider if you'd like to incorporate this into core Agents.jl functionality? As previously mentioned I wrote this for my MS thesis and am starting to work on the reinforcement learning piece of that project. I'm trying to keep things relatively modular and am planning to either fork Agents.jl and roll it into that or modularize FMP into its own package depending on your decision. 

Please don't interpret this as me trying to rush you - I'm just at a point where having a decision from you on this would give me a clearer path forward with how I architect my thesis code; that said it wouldn't be big deal to integrate down the road if you need more time to make a decision. Thanks.

Libbum:
Heya, sorry no - not yet. I'm in the process of moving jobs so haven't had extra time and there's a backlog of tasks here to get to!

I started looking into #405 yesterday, with considerations on how we should go about designing pathfinding as a general concept for Agents. Should it be something modular or tightly bound to space is one aspect to really consider.

Should have some time this evening and tomorrow to look at this PR. Is it OK for me to do some edits directly or would you prefer I just comment on the branch?

Libbum:
So I've had a read through and it's definitely something we can include. I can do some work to update bits and pieces, separate things out into functional vs example code and see what works out. 

From the looks of it though, `AgentPositionInit` is essentially a couple of pre-defined experiments and if we don't include the plotting helpers there's essentially a few functions left that could be made into a library as you say.

What I might do is make a new PR that would integrate these algorithmic sections into the structure of Agents.jl, provide a few exposed functions that we could then re-write the examples around. From that point we could make the decision on whether or not to continue down that path or make a library that can become a dependency for Agents. Does that sound like a plan?

rmcsqrd:
`AgentPositionInit` is indeed a pre-defined experiments chunk of code. The example in this PR is mostly just a combined form of the code in [this repo](https://github.com/rmcsqrd/antfarm/tree/master/src). In general I think the code of interest is found in [/fmp/fmp.jl](https://github.com/rmcsqrd/antfarm/blob/master/src/fmp/fmp.jl) and a lot of the other stuff is auxiliary. 

Feel free to edit as you see fit - that being said I am also happy to make any edits per your direction. If you'd like I am happy to take a stab at making the PR that integrates the FMP algorithm into Agents.jl + reworking this example around the added code. If you give me a set of criteria to work within (aka what part of Agents.jl to add this stuff, what sort of exposed functions you are envisioning, etc) I can fork, make edits, and submit a new PR. 

Would that work for you? I'd love an opportunity to contribute to Agents.jl substantively but also understand if there is something on your end that might preclude that.

Libbum:
Certainly, if you'd like to refactor a bit to integrate the important functions into Agents that would be great!

The open question I have is where best to put it. For the moment it looks like the algorithm works best on Continuous spaces, but would it also work on a grid space? If so, maybe we target inclusion at the descrete space level. 

However, following discussions in #405, we're uncertain if the pathfinding should be part of space, part of model or something on its own.

Best case for us at the moment is to have one API to use whichever algorithm we like, or if we must use FMP only on Continuous space and A* in GridSpace, then the API should automatically choose the algorithm for us. 

I think it's going to take some iteration. So for a start, have a go at placing the methods where you can and we'll work from there.

rmcsqrd:
To refresh, the algorithm works as follows:
1) the algorithm computes the resultant force from all inter-agent/obstacle repulsive forces and navigational feedback force
2) These forces are treated as an acceleration (Newtons 2nd law with unit mass). The acceleration is multiplied by `model.dt` which yields a velocity. 
3) the resultant velocity is passed to `move_agent()`

I think the "resultant force" approach fits much more naturally into a continuous space and trying to apply it into a discrete space may be a little hacky. 

I took a look at the discussion going on in #405 and will model my implementation of the FMP algorithm in a similar way to that.

Libbum:
Great. We're on the home stretch there and will have an `AbstractPathfinder` type that `FMP` can be a subtype of. I'm hoping we'll get it merged tomorrow. The design there should be able to sort us out a pretty streamlined implementation for your algorithm too. 

If FMP  is best to stay in `ContinuousSpace`, that's no issue at all.

Datseris:
I don't have the time to give a review here, but what's the difference between this and the force-based approach of `ContinuousSpace`? Where you evolve agents based on Euler scheme, and a force field tunes the agent velocities?

Libbum:
Do you mean [elastic_collision!](https://juliadynamics.github.io/Agents.jl/stable/api/#Agents.elastic_collision!) etc? My understanding is that it's essentially trying to avoid collisions if possible based on motion planning. So rather than force fields and bouncing, the motion here has the potential to be more non-linear and 'natural' looking.

That sound about right Rio?

rmcsqrd:
> Where you evolve agents based on Euler scheme, and a force field tunes the agent velocities?

@Datseris That is a pretty concise description of what the FMP algorithm is doing. That said, I am unclear on where this is implemented in `ContinuousSpace`. I see a reference to `update_vel!` being related to spatial force fields but am having trouble finding where that is implemented. If you point me towards that I can give a better comparison. TBH I did not realize that this functionality existed in Agents. 

@Libbum I actually think that a "force field" is actually a decent characterization of what is going on under the hood - I liken agents to magnets when I explain this algorithm to people. As agents get within an interactive proximity of one another they experience repulsive forces. 

This diagram is pretty illustrative:

Screen Shot 2021-02-22 at 3 06 49 PM

`u_i` is the resultant velocity of the agent at position `p_i` that is the vector sum of 
- `f_i^obs` (obstacle repulsive force)
- `f_i^R` (repulsive force from the agent at position `p_j`)
- `f_i^\gamma` (attractive force to goal at position `T_i`)

Like I mentioned above, I am unclear on how the existing `update_vel!` handles forces. If `elastic_collision!` is what @Datseris is referring to then the main differences are:
- Agents don't need to collide to interact. Agents experience a repulsive force that is inversely proportional to inter-agent proximity.
- A resultant force is computed for each agents local environment meaning that all inter-agent forces are considered during each time step. `elastic_collision` looks like it only considers the collision of two agents per time step (I think). 
- The FMP algorithm has a "goal" parameter for each agent that exerts an attractive force.

Datseris:
> I see a reference to update_vel! being related to spatial force fields but am having trouble finding where that is implemented. If you point me towards that I can give a better comparison. TBH I did not realize that this functionality existed in Agents.

There is no implementation because `update_vel!` is a user provided function. For example, this function would get the agent position, and calculate a magnetic field force vector, or electric field or whatever, which in turn modefiies the agent velocity. The update happens here: https://github.com/JuliaDynamics/Agents.jl/blob/master/src/spaces/continuous.jl#L113-L116 

This is not related to `elastic_collision!` in any way. But given your plot and the description of the forces, this seems to me like something possible via the `update_vel!` interface. Do you think you could give it a try? this could simplify the total amount of code, if possible.

Datseris:
one thing is how efficient things are. because in the current set up `update_vel!` goes through each agent one by one. And thus for each agent it calculates the forces. Does your implementation do any optimization on this part? We could use `Agents.interactive_pairs` as well.

rmcsqrd:
I use `interacting_pairs` using a user-specified interaction radius to figure out which agents are interacting. The algorithm is `O(nk)` where `n` is the number of agents in the state space and `k` is the number of agents for each agent. If you assume that all agents have a homogenous radius, you can "non-rigorously show" (aka I haven't sat down to formally prove it but I am pretty sure I am correct) that the complexity reduces to `O(6n)=O(n)` because it is a [packing problem](https://en.wikipedia.org/wiki/Packing_problems#Hexagonal_packing_of_circles) and agents arrange themselves hexagonally.

I'm getting some slightly mixed directions from you and @Libbum - originally I was planning to put it into `src/pathfinding` and emulate what is being done with the A* algorithm in #405 . Technically this is a motion planning algorithm but I also agree this fits pretty naturally into the `update_vel!` realm. I'm fine with either approach.

Datseris:
You're the one most familiar with the method, so you need to decide what approach fits best. As far as I can tell, the figure you showed is exactly what `update_vel!` aims to do. I guess there are other ways to look at the same problem, like e.g. path finding. 

I think it is worth it to try the `update_vel!` approach as well. We can't know which approach suits best without actually trying both. However we might need to do some updates in the `update_vel!` approach,  to allow using `interacting_pairs` for performance.

rmcsqrd:
I agree with all that - I'll implement it as an alternative to `defvel` in `src/spaces/continuous.jl` as the current path moving forward.

rmcsqrd:
Okay, these recent commits are my preliminary implementation of the FMP algorithm as an `update_vel` method. My "git-fu" is lacking so only pay attention to the modifications in `src/spaces/continuous.jl` and `examples/FMP.jl`. The FMP example code should demonstrate usage although there are still some bugs I need to iron out.

Libbum:
Thanks Rio, yes&mdash;sorry for the push-pull from us. Something like this that's new to us all takes a bit of iteration before knowing the best way to go about it. Reading the recent discussion, I'd also agree that looking at the `update_vel!` approach would be better situated for FMP, so thanks for these changes. I'll be able to take a deeper look in a day or so.

Project.toml
```diff
@@ -4,15 +4,21 @@ authors = ["Tim DuBois", "George Datseris", "Ali Vahdati"]
...
 GraphRecipes = "bd48cda9-67a9-57be-86fa-5b3c104eda73"
 LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 OpenStreetMapX = "86cd37e6-c0ff-550b-95fe-21d72c8d4fc9"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
Datseris, line 21: What happened here? None of these should be added as Agents dependencies. Also AgentsPlots is deprecated.

Datseris (changes requested):

Project.toml
AayushSabharwal: `DataStructures` is an added dependency for A* Pathfinding in #405. It appears as if the rest were added in [this commit](https://github.com/JuliaDynamics/Agents.jl/pull/412/commits/fe7a516ee3c2cbc761c35139670a3f09ad2f8120)

Datseris: Okay, the rest should be removed. If they are dependencies for building documentation then they can be put in the `docs/Project.toml` instead. Although I doubt revise and progress meters are needed to build the docs.

Libbum:
Don't worry about these merge issues Rio. We're doing a lot of major things elsewhere. I'll clean it up.

rmcsqrd:
Okay - thanks. It is probably becoming clear that I am not good with git haha. Basically the two files I've made modifications to are:
- `src/spaces/continuous.jl`
- `examples/FMP.jl`
Everything else can be ignored. I manually recloned the current master branch of Agents.jl and copied in the offending files to try and isolate my PR to the two files mentioned above but was unsuccessful. Sorry about that.

rmcsqrd:
Alright I just pushed a commit with a fleshed out example  (`examples/FMP.jl`) and better docstrings for the FMP implementation (`src/spaces/continuous.jl`). Please ignore any other file changes. 

There is a small bug with the velocity stuff but in general I think that this is ready for review when you have a chance.

rmcsqrd:
I figured out the bug I was having an issue with. One thing that I would like to note is that I am not sure that the FMP algorithm works with `update_vel!` in its current form. The FMP algorithm computes the resultant velocities of agents at each iteration based on all agent positions at that snapshot in time. `update_vel!` updates each agents position when `move_agent!` is called. This means that some agents update their velocity based on agents that have already moved. This technically yields some "undesired behavior" (which is totally subjective; I am just trying my best to make a true implementation of the algorithm). Checkout the examples:

**FMP computes resultant velocities of all agents before updating agent positions**
circle_swap_pretty

**FMP computes resultant velocity and updates position each time `update_vel!` is called**
circle_swap_update_vel

(That gridlock condition is actually expected behavior)

Currently I make a function call before `step!` that updates all agent velocities and then step the model conventionally (no `update_vel` argument in `ContinuousSpace`). One option I was considering is another `move_agent!` definition that updates all agent velocities before updating their positions. I was curious if anyone had any other ideas? I think `update_vel!` is a good place to put this but I need a way to compute all agent velocities before changing their positions.

rmcsqrd:
Looks like `model_step!` per the advanced stepping documentation should accomplish this - I will try that

Datseris:
Sorry for being away too long. I can now review. There are two example files. Why is that, and which one should I review?

examples/FMP_simple_temp.jl
```diff
@@ -0,0 +1,252 @@
...
+# - :A = agent (has velocity, subject to repulsive forces)
+# - :O = object (has no velocity, cannot move, agents must go around)
+# - :T = target (no velocity, just a projection of agent target zone)
+
+using Agents, Random, Plots, LinearAlgebra, Colors, ProgressMeter
+
+ENV["GKSwstype"]="nul"
+gr()
+cd(@__DIR__)
+
```
Datseris, line 31: Almost all of this is unecessary. Plotting now has moved to `InteractiveDynamics`. See for example the updated docs of the Flocking model, or the Social Distancing model. No reason for `ProgressMeter` either.

examples/FMP_simple_temp.jl
```diff
@@ -0,0 +1,252 @@
...
+function FMP_Model()
+    properties = Dict(:FMP_params=>FMP_Parameter_Init(),
+                      :dt => 0.01,
+                      :num_agents=>100,
+                      :num_steps=>1500,
+                      :step_inc=>2,
+                     )
+
+    #space2d = ContinuousSpace((1,1); periodic = true, update_vel! = FMP_Update_Vel)  #BONE commented this out
```
Datseris, line 64: Be sure that the final examples for merging do not have forgotten comments like these

examples/FMP_simple_temp.jl
```diff
@@ -0,0 +1,252 @@
...
+"""
+This function is a utility function that takes an agent and overwrites its color if it is of type :T (target) and gives it a grey color for display purposes.
+"""
+function PlotABM_ColorUtil(a::AbstractAgent)
+    if a.type == :A || a.type == :O
+        return a.color
+    else
+        return "#ffffff"
+    end
+end
```
Datseris, line 103: all of this function can be made as `f(a) = a.type in (:A, :B) ? a.color : "#ffffff"`

Datseris:
@rmcsqrd some of the name choices you do, do not abide per Julia's style guide. For example `AgentInitColor`. CamelCaseSyntax is used for Types only, not functions. For functions please use exclusively lower case and `_` if necessary. Also have a look at : https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/

rmcsqrd:
Sorry for the confusion and thank you for the comments - I've been struggling with git a lot on this. At some point I just recloned from `Agents.jl/master` to get a clean start and started a new branch at: https://github.com/rmcsqrd/Agents.jl/tree/FMP

That has one example `examples/FMP.jl` (reflecting your comments above except for the Julia style guide stuff) and changes to `src/spaces/continuous.jl`. 

It seems like closing this PR and opening a new one for my new branch is probably the best course of action unless you have a better idea? Thanks.

Datseris:
> It seems like closing this PR and opening a new one for my new branch is probably the best course of action unless you have a better idea? Thanks.

Yeap, go ahead with that, while address my comments as well. Don't worry, when I started contributing to open source git was a true nightmare for me. Did something wrong all the time.

rmcsqrd:
Continued in #456

Issue #408, opened 2021-02-03
Data collection interface inspired by DataFrames.jl


flipgthb:
Hi, 
First, I would like to say that I appreciate the work and that I'm way more comfortable with Agents.jl than with Mesa or other approaches I've tried for agent modeling, so thanks a lot for the great job.
One thing, however, that I think could be improved is the data collection interface. Not that it is bad right now, just that think it could be even more convenient.
One approach I really appreciate regarding data manipulation is the DataFrames.jl package interface to pass functions to apply to dataframes columns, like:
```julia
using DataFrames
df = DataFrame(a=collect(1:6),b=rand([:x,:y],6),c=randn(6))
    select(df,[:a,:b]=>ByRow((a,b)->"$(b)_$(a)")=>:new_col_from_a_and_b)
```
Two features of this approach that, in my opinion, Agents.jl approach lacks are:
   - The possibility to use selectors to specify how to use the attributes (in the above example the `ByRow`).
   - The ability to use anonymous functions and give it a name on the fly, which helps to clean the global namespace from functions that will be used only once.

I propose for Agents.jl to adapt its data collection interface to follow the approach of DataFrames.jl.
Sure, `Agents.AgentBasedModel` does not follow the Tables.jl interface, so the "selectors" should be different.

The idea looks somewhat like this:
```julia
run!(
    fake_model_instance,
    fake_agent_step!,
    n_steps;
    report=[
    [:agent_property,:agent_other_property]=>ByAgent((x,y)->x+y)=>:agent_complicated_property,
    :agent_property=>Neighbors(((a1prop,a2prop)->max(a1prop,a2prop)),r=1)=>:pairwise_property,
    ],
    report_when=[:agent_complicated_property=>x->mean(x)<1.0]
)
``` 
 Forgive me if what I said is to much nonsense (not a programmer and never requested a feature before).

Libbum:
Hi Felippe!

An interesting proposal, although I'm not convinced yet that it makes the most sense for our collection model. Granted, there are a lot of 'simple' examples in our documentation, so we have yet to really considered scenarios that may require greater flexibility. If you have one, it would be helpful to take a look at it.

Some things we need to decide upon:
- Does this change keep the interface simple when needed but give power users the capacity they need? At the moment I cannot see how I just store `:agent_property` directly for example.
[...]

PR #417, opened 2021-02-09
High-order iterators and maps

Libbum:
A start on #408. These methods handle any order you need / can fit in memory. Initially I thought it would only be required to export the map functions, but in the case of the filtered versions of the map it's not clear which ids correspond to the output.

So, in our examples for these methods, we'll need to state clearly that you can get indexes for your results in a few ways:
- No filter, you just need one row of indices: `by_id(model)`
- No filter, you want the lot: collect the `group_id_iter`. The results are correct for the agent version as well.
- Filter: use the filter to filter the `group_id_iter`. For example: `Iterators.filter(allunique, group_id_iter(2,model))`.

Feel free to suggest better names here. Still needs tests, and the examples we discussed.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -89,8 +89,8 @@ provide an interface for such calculation.
 
 ```@docs
 iter_agent_groups
-iter_id_groups
 map_agent_groups
+index_groups_filtered
 ```
 
 ## Discrete space exclusives
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -81,6 +81,19 @@ a = random_agent(model)
 sort!(collect(nearby_agents(a, model)))
 ```
 
+## Higher-order interactions
+
+There may be times when pair-wise, triplet-wise or higher interactions need to be
+accounted for across most or all of the model's agent population. The following methods
+provide an interface for such calculation.
+
+```@docs
+group_id_iter
+group_agent_iter
+group_id_map
+group_agent_map
+```
+
 ## Discrete space exclusives
 ```@docs
 positions
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -88,10 +88,9 @@ accounted for across most or all of the model's agent population. The following 
 provide an interface for such calculation.
 
 ```@docs
-group_id_iter
-group_agent_iter
-group_id_map
-group_agent_map
+iter_agent_groups
+iter_id_groups
+map_agent_groups
 ```
 
 ## Discrete space exclusives
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -203,63 +203,43 @@ allids(model) = keys(model.agents)
 #######################################################################################
 # %% Higher order collections
 #######################################################################################
-export group_agent_iter, group_id_iter, group_agent_map, group_id_map
+export iter_agent_groups, iter_id_groups, map_agent_groups
 
 """
-    group_id_iter(order::Int, model::ABM)
-
-Return an iterator over all agent IDs of the model, group by order. When `order = 2`, the
-iterator returns agent pairs, and when `order = 3`: agent triples.
-
-See also [`group_agent_iter`](@ref).
+    iter_id_groups(order::Int, model::ABM)
+Return an iterator over all agents of the model, grouped by order. When `order = 2`, the
+iterator returns ID pairs, e.g `(1, 2)` and when `order = 3`: ID triples, e.g.
+`(1, 7, 8)`.
 """
-group_id_iter(order::Int, model::ABM) =
+iter_id_groups(order::Int, model::ABM) =
     Iterators.product((by_id(model) for _ in 1:order)...)
 
 """
-    group_agent_iter(order::Int, model::ABM)
+    iter_agent_groups(order::Int, model::ABM)
 
-Return an iterator over all agents of the model, group by order. When `order = 2`, the
-iterator returns ID pairs, e.g `(1, 2)` and when `order = 3`: ID triples, e.g.
-`(1, 7, 8)`.
-
-See also [`group_id_iter`](@ref).
+Return an iterator over all agents of the model, grouped by order. When `order = 2`, the
+iterator returns agent pairs, e.g `(agent1, agent2)` and when `order = 3`: agent triples,
+e.g. `(agent1, agent7, agent8)`. `order` must be larger than `1` but has no upper bound.
 """
-group_agent_iter(order::Int, model::ABM) =
+iter_agent_groups(order::Int, model::ABM) =
     Iterators.product((map(i -> model[i], by_id(model)) for _ in 1:order)...)
 
 """
-    group_id_map(order::Int, f::Function, model::ABM)
-    group_id_map(order::Int, f::Function, model::ABM, filter::Function)
+    map_agent_groups(order::Int, f::Function, model::ABM)
+    map_agent_groups(order::Int, f::Function, model::ABM, filter::Function)
 
-Applies function `f` to all grouped agent IDs of a [`group_id_iter`](@ref). `f` must take
-the form `f(model, a, b)`, where the number of arguments (after `model`) is equal to
-`order`.
-Optionally, a `filter` function that accepts an iterable and returns a `Bool` can be
-applied to remove unwanted matches from the results.
-
-See also [`group_agent_map`](@ref).
-"""
-group_id_map(order::Int, f::Function, model::ABM) =
-    (f(model, idx...) for idx in group_id_iter(order, model))
-group_id_map(order::Int, f::Function, model::ABM, filter::Function) =
-    (f(model, idx...) for idx in group_id_iter(order, model) if filter(idx))
-
-"""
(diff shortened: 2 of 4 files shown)


fbanning (approved):
Seems solid to me.

src/core/model.jl
```diff
@@ -190,13 +194,73 @@ Return an iterator over all agents of the model.
...
 
+#######################################################################################
+# %% Higher order collections
+#######################################################################################
+export group_agent_iter, group_id_iter, group_agent_map, group_id_map
+
+"""
+    group_id_iter(order::Int, model::ABM)
+
+Return an iterator over all agent IDs of the model, group by order. When `order = 2`, the
```
Datseris, line 211: group -> grouped

src/core/model.jl
```diff
@@ -190,13 +194,73 @@ Return an iterator over all agents of the model.
...
+
+"""
+    group_agent_iter(order::Int, model::ABM)
+
+Return an iterator over all agents of the model, group by order. When `order = 2`, the
+iterator returns ID pairs, e.g `(1, 2)` and when `order = 3`: ID triples, e.g.
+`(1, 7, 8)`.
+
+See also [`group_id_iter`](@ref).
+"""
```
Datseris, line 227: The example given is not what this function does, but what the function `group_id_iter` does.

Datseris: We should move the example to the `group_id_iter` and in this docstring only cross-ref` group_id_iter` and say "this function does the same but with agents. This is how we deal with `nearby_ids` and `nearby_agents`.

src/core/model.jl
```diff
@@ -190,13 +194,73 @@ Return an iterator over all agents of the model.
...
+group_agent_iter(order::Int, model::ABM) =
+    Iterators.product((map(i -> model[i], by_id(model)) for _ in 1:order)...)
+
+"""
+    group_id_map(order::Int, f::Function, model::ABM)
+    group_id_map(order::Int, f::Function, model::ABM, filter::Function)
+
+Applies function `f` to all grouped agent IDs of a [`group_id_iter`](@ref). `f` must take
+the form `f(model, a, b)`, where the number of arguments (after `model`) is equal to
+`order`.
```
Datseris, line 237: The syntax `f(model, a, b)` is not clear to me, when followed by "where the number of arguments..". Does this refer to `a, b`? I.e. if order was 3 you would need `a, b, c`? Then it is better to define the call as `f(model, args...)` with `args...` as many as `order`.

Datseris (commented):
I left some documentation comments. One thing that is not yet clear to me is whether we really need both `group_id_iter` and `group_agent_iter`. I would vote to only keep the `agent` version. In functions like `nearby_ids`, there is some use in having both, because the `nearby_agents` version skips the given agent. 

So yeah, my vote is to keep only `group_agent_iter` and `group_agent_map`. 

I would also like to challenge the names. Specifically the `group_agent_map`, I feel like it does not convey exactly what the function does. Perhaps `map_agent_groups` is better? Not sure!

src/core/model.jl
Libbum: Yep, that's correct. It's essentially a curried function.

Libbum:
Changes look fine, I'll get to them soon.

The id version is much faster, but I guess since this is really only going to operate on agent properties, then sure: let's drop the id runners.

Names were hard for this one. `map_agent_groups` is as good as any! And perhaps a bit more standard.

src/core/model.jl
```diff
@@ -190,13 +194,73 @@ Return an iterator over all agents of the model.
...
+the form `f(a, b)`, where the number of arguments is equal to `order`.
+Optionally, a `filter` function that accepts an iterable and returns a `Bool` can be
+applied to remove unwanted matches from the results.
+
+See also [`group_agent_map`](@ref).
+"""
+group_agent_map(order::Int, f::Function, model::ABM) =
+    (f(idx...) for idx in group_agent_iter(order, model))
+group_agent_map(order::Int, f::Function, model::ABM, filter::Function) =
+    (f(idx...) for idx in group_agent_iter(order, model) if filter(idx))
```
Datseris, line 262: the mapping function `f` and the filtering function `filter` accept the arguments in a different way, which I find unintuitive. One requires `f(a, b, c)` while the other `f((a, b, c))`. They should have same input syntax.

Libbum:
Two open points on this one (well, excluding tests & examples):
 1. I've left what is now called `iter_id_groups` in, but removed the map counterpart. Reason for this is when users want to filter their result, the output is no longer in matrix form. You cannot use `by_id(model)` as an index to your data. Therefore a set of indexes should also be collected. 
     - with `iter_id_groups`: `Iterators.filter(filter_function, iter_id_groups(3,model))`
     - without: `map(agents->map(a->a.id, agents), Iterators.filter(filter_function, iter_agent_groups(3,model)))`
    
    In fact, I'd perhaps not export `iter_id_groups`, but make something else available like `index_groups_filtered(order, filter) = Iterators.filter(filter, iter_id_groups(order,model))`
2. f(a, b, c) while the other f((a, b, c)): we're better off using the iterable interface ` f((a, b, c))` so we can use base functions like `allunique`. 
    The question I'd raise here is how much understanding a user needs to get this to function. These are 'advanced' features in our framework: sure. Take @flipgthb's `balance` function as an example.
    ```julia
    balance(a_i::A,a_j::A,a_k::A) where {A<:NNAgent} = coherence(a_i,a_j) + coherence(a_j,a_k) + coherence(a_i,a_k)
    ```
   *must* take the form
    ```julia
   balance(a::NTuple{3,<:NNAgent}) = coherence(a[1],a[2]) + coherence(a[2],a[3]) + coherence(a[1],a[3])
    ```
    I think I'm just speaking out loud. Maybe it's fine to expect advanced users to know about `NTuples`...

flipgthb:
>        The question I'd raise here is how much understanding a user needs to get this to function. These are 'advanced' features in our framework: sure. Take @flipgthb's `balance` function as an example.
>        ```julia
>        balance(a_i::A,a_j::A,a_k::A) where {A<:NNAgent} = coherence(a_i,a_j) + coherence(a_j,a_k) + coherence(a_i,a_k)
>        ```
>        
>        
>        _must_ take the form
>        ```julia
>        balance(a::NTuple{3,<:NNAgent}) = coherence(a[1],a[2]) + coherence(a[2],a[3]) + coherence(a[1],a[3])
>        ```
>        
>        
>        I think I'm just speaking out loud. Maybe it's fine to expect advanced users to know about `NTuples`...

In my opinion, it is fine to define the function in terms of `NTuples` except, maybe, for models with multiple types of agents.
For instance, if there is a `NNAgent{Type1}` and `NNAgent{Type2}`, It is possible to specify the function as
```julia
balance(a::Tuple{NNAgent{Type1},NNAgent{Type2},NNAgent{Type1}}) = ...
```
but not as a function of an `NTuple`. In my particular case, this kind of thing is not a problem, though

flipgthb:
Nah, what I said is not correct, one can always do something like
```julia
balance(a::NTuple{3,Union{NNAgent{Type1},NNAgent{Type2}}) = ...
```

Libbum:
@flipgthb, once these tests pass I think this one is ready to merge. We don't have an example of how to use it yet, but it seemed like you'd be interested in making something for it. OK if I make a new issue for that and assign you to it?

flipgthb:
@Libbum its ok

src/core/model.jl
```diff
@@ -190,13 +194,56 @@ Return an iterator over all agents of the model.
...
+export iter_agent_groups, map_agent_groups, index_groups_filtered
+
+"""
+    iter_agent_groups(order::Int, model::ABM)
+
+Return an iterator over all agents of the model, grouped by order. When `order = 2`, the
+iterator returns agent pairs, e.g `(agent1, agent2)` and when `order = 3`: agent triples,
+e.g. `(agent1, agent7, agent8)`. `order` must be larger than `1` but has no upper bound.
+
+Index order is provided by the [`by_id`](@ref) scheduler.
```
Datseris, line 215: I mean, here it should be trivial to allow any scheduler instead of forcing the `by_id`, right? Not that it matters much

docs/src/api.md
```diff
@@ -81,6 +81,18 @@ a = random_agent(model)
...
+## Higher-order interactions
+
+There may be times when pair-wise, triplet-wise or higher interactions need to be
+accounted for across most or all of the model's agent population. The following methods
+provide an interface for such calculation.
+
+```@docs
+iter_agent_groups
+map_agent_groups
+index_groups_filtered
```
Datseris, line 93: I don't think the `_filtered` part at the end should be here. I propose to rename this function to `index_mapped_groups`. I would be using this function even without hte `filter` argument, to get a convenience vector that maps results of `map_agent_groups` to agents used in the mapping.

Libbum:
Sure, I'll change these things soon. Won't tag 4.1 until the RNG pool is sorted, so no problem. 

-----Original Message-----

Issue #415, opened 2021-02-08
Seed consideration in parallel_replicates


Libbum:
At the moment, `parallel_replicates` uses a [deepcopy of the model](https://github.com/JuliaDynamics/Agents.jl/blob/0c660872f983b5a981671bc8f1b5c901c8799a93/src/simulations/collect.jl#L454). If users are depending on seeds inside their model to produce replicable random results, they will want replicates to modify such a seed. 

We need to allow users to provide a function which appropriately modifies each replicate to achieve this.

Datseris:
True, but the seeding is a bigger problem, because functions that choose e.g. random positions are not seeded properly by the model. Therefore, I think the issue of using a global RNG from the ABM _everywhere_ precedes solving the current issue.

Libbum:
Ah, yes. That may be more important. I'll think on that.

Datseris:
We could define `get_rng(model::ABM) = get(model.properties, :rng, Random.GLOBAL_RNG)` and use the output of `get_rng` in every `random` call. What I'm a bit uncertain is whether this is type stable.

Libbum:
I've been thinking about Bogumił's [blog about seeds](https://bkamins.github.io/julialang/2020/11/20/rand.html) too, I think something like what you've suggested above is a good place to start.

Snowed under at the moment with other projects, but it'll be good to chat about this in a few weeks.

Datseris:
Very interesting. I knew almost nothing the post said. It only makes it clear that the seed business is more important than we thought.

Datseris:
I think we should simply allow an extra keyword `seeds = [rand(Int) for i in 1:replicates]` by default, which is used to seed each replicate. That seems to be the simplest scenario. Not really sure what to do with the `RandomDevice()` version though.

Libbum:
Something like that would be a solution, yeah.

Concerning `RandomDevice()`: I've got a hardware entropy RNG thingy plugged into this machine. No point not using it!

Datseris:
Have you checked that `deepcopy`-ing a model with `RandomDevice()` instance produces same numbers, even if run at different points in time?

Libbum:
At least with a modified parallel replicates implementation its true. Hence #436 

https://github.com/Libbum/LimnoSES.jl/blob/656a5c13d274a23ab6e59773e25f091d5173346b/src/replicates.jl#L5

[...]

PR #419, opened 2021-02-18
Enable 1.6rc testing on CI

Libbum:
I think we'll have seed failures here, as we did previously in #271, where we 'solved' the issue by dropping 1.0 support. We don't have that luxury this time round.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,8 @@ jobs:
       matrix:
         version:
           - '1'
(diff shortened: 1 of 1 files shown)


Libbum:
Relevant discussion in #415, which may need to be solved to sort out this problem. As with #271, StableRNGs is not a solution for us.

Libbum:
So, this is working in the sense that it enables testing. I've got a new PR coming to fix tests on 1.6.

PR #420, opened 2021-02-18
Provide an in-model RNG pool

Libbum:
WIP to get seeds working correctly.

Things look fine on the surface. Models run, types seem alright.

Still issues with tests on different versions though (at least locally). So continuing discussion from #419 here.

### DIFF
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -13,12 +13,12 @@ abstract type DiscreteSpace <: AbstractSpace end
 ValidPos =
     Union{Int,NTuple{N,Int},NTuple{M,<:AbstractFloat},Tuple{Int,Int,Float64}} where {N,M}
 
-struct AgentBasedModel{S<:SpaceType,A<:AbstractAgent,F,P}
+struct AgentBasedModel{S<:SpaceType,A<:AbstractAgent,F,P,R<:AbstractRNG}
     agents::Dict{Int,A}
     space::S
     scheduler::F
     properties::P
-    rng::AbstractRNG
+    rng::R
     maxid::Base.RefValue{Int64}
 end
 
@@ -79,13 +79,13 @@ function AgentBasedModel(
     space::S = nothing;
     scheduler::F = fastest,
     properties::P = nothing,
-    rng = Random.default_rng(),
+    rng::R = Random.default_rng(),
     warn = true,
-) where {A<:AbstractAgent,S<:SpaceType,F,P}
+) where {A<:AbstractAgent,S<:SpaceType,F,P,R<:AbstractRNG}
     agent_validator(A, space, warn)
 
     agents = Dict{Int,A}()
-    return ABM{S,A,F,P}(agents, space, scheduler, properties, rng, Ref(0))
+    return ABM{S,A,F,P,R}(agents, space, scheduler, properties, rng, Ref(0))
 end
 
 function AgentBasedModel(agent::AbstractAgent, args...; kwargs...)
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -18,7 +18,7 @@ struct AgentBasedModel{S<:SpaceType,A<:AbstractAgent,F,P,R<:AbstractRNG}
     space::S
     scheduler::F
     properties::P
-    rng::R
+    rng::Base.RefValue{R}
     maxid::Base.RefValue{Int64}
 end
 
@@ -66,10 +66,7 @@ which are the fields of `AgentBasedModel`.
 (see e.g. [`by_id`](@ref) and the scheduler API).
 
 `rng = Random.default_rng()` provides random number generation to the model.
-Passing, for example `MersenneTwister(1234)` will initialise with a repeatable random
-seed, and `RandomDevice()` will use the system's entropy source (coupled with hardware
-like [TrueRNG](https://ubld.it/truerng_v3) will invoke a true random source, rather
-than pseudo-random methods like `MersenneTwister`).
+Accepts any subtype of `AbstractRNG`.
 
 Type tests for `AgentType` are done, and by default
 warnings are thrown when appropriate. Use keyword `warn=false` to suppress that.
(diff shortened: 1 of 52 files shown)


Libbum:
Tests that fail at present is a very odd subset. Using StableRNG seed on the first one that's failing didn't help. There's something else deeper happening perhaps.

src/core/model.jl
```diff
@@ -5,24 +5,26 @@ export ABM, AgentBasedModel
...
+ValidPos =
+    Union{Int,NTuple{N,Int},NTuple{M,<:AbstractFloat},Tuple{Int,Int,Float64}} where {N,M}
 
-struct AgentBasedModel{S<:SpaceType, A<:AbstractAgent, F, P}
+struct AgentBasedModel{S<:SpaceType,A<:AbstractAgent,F,P}
     agents::Dict{Int,A}
     space::S
     scheduler::F
     properties::P
+    rng::AbstractRNG
```
Datseris, line 21: `AbstractRNG` is an abstract type (says so in the name)! You need to parameterize this like the other fields.

Datseris:
I had scheduled to do this over a long train ride I will have on Saturday, but you got on it sooner. I will still commit here during my train ride.

Libbum:
Ah! Well, I could really use some help with the failing tests when you get the time

For the schedulers, it looks like maybe the dict order has changed from 1.5->1.6? That's all I can see.

For the walks: I have no idea why they're different yet.

Datseris:
Is it possible to re-seed RNGs in-place? Because `ABM` is immutable, you can't do `model.rng = MersenneTwister(21581)` to set a new seed.

src/core/model.jl
```diff
@@ -64,17 +65,27 @@ which are the fields of `AgentBasedModel`.
 `scheduler = fastest` decides the order with which agents are activated
 (see e.g. [`by_id`](@ref) and the scheduler API).
 
+`rng = Random.default_rng()` provides random number generation to the model.
+Passing, for example `MersenneTwister(1234)` will initialise with a repeatable random
+seed, and `RandomDevice()` will use the system's entropy source (coupled with hardware
+like [TrueRNG](https://ubld.it/truerng_v3) will invoke a true random source, rather
+than pseudo-random methods like `MersenneTwister`).
```
Datseris, line 72: I think that is not necessary information. We can simply say that it must be a subtype of `AbstractRNG`. This information can be in the Tutorial.md documentation page instead. 

(I believe docstrings should be as lean as humanly possible)

Libbum:
Ah, hadn't looked at that yet. Was too focused on getting 1.6 to pass tests.

```julia
Random.seed!(54)
model = ABM(Agent3, GridSpace((5, 5)))
add_agent!(model, rand(model.rng))

model[1].weight # 0.5330409816783452
model[1].weight = rand(model.rng) # 0.6328988203983412
```

```julia
Random.seed!(54)
model = ABM(Agent3, GridSpace((5, 5)))
add_agent!(model, rand(model.rng))

model[1].weight # 0.5330409816783452
Random.seed!(64)
model[1].weight = rand(model.rng) # 0.9626830433141891
```

So it works in the global case, since we default to `Random.default_rng()`.

```julia
model = ABM(Agent3, GridSpace((5, 5)); rng = MersenneTwister(54))
add_agent!(model, rand(model.rng))

model[1].weight # 0.5330409816783452
Random.seed!(64) # Should not change since this changes the global seed
model[1].weight = rand(model.rng)  # 0.6328988203983412 (expected)
```

So yes, we'll need to write an update method for the local seed (or use Ref like we do for maxid).

Datseris:
I'm currently asking other developers whether we should make `ABM` a mutable type. I'm not sure, or to be frank I don't remember, what are the benefits of having it immutable.

Libbum:
[Speed for one](https://bkamins.github.io/julialang/2020/10/16/gctime.html). I've been watching [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) for a while now. In the future it'd be better if we made even Agent structs immutable rather than going the other way and making the model mutable.

Datseris:
how come this Bogumil guy never contributed here or said anything about Agents.jl. Would have been reaaaally use full to have him around :P

Datseris:
I'm really confused by this blog post though. It seems to me that mutable objects have no reason to exist then...?

Libbum:
It's a stack vs heap thing I _think_. I've not understood these lens things, how they works or whatever. But I think oxinabox talks about [some changes in v1.5 that assist here](https://www.oxinabox.net/2021/02/13/Julia-1.6-what-has-changed-since-1.0.html#references-to-the-heap-from-the-stack-performance).

Libbum:
Solved the tests now. Problems were:
- **Scheduler failures** Dict order is not the same across versions. So even with a StableRNG seed, `collect(keys(model.agents))` is not comparable. When shuffling, the seed is the same, but since the original order is not, we return different values.
    - I've altered the `random_activation` test to compare against the `fastest` order to prove there's a shuffle
    - The `partial_activation` does not test for explicit order now but makes sure the activation is around 10% (as requested by the test)
- **`walk!`** problems were essentially twofold.
    - `rand(rng, -1:1, 2)` is fine to use across both versions so long as `StableRNGs` are being tested (There must be algorithmic alterations happening for `rand(2)`, but not `rand()`).
    - `Tuple(rand(model.rng, -1:1) for _ in 1:2)` has one fewer allocation in 1.5.3 than `Tuple(rand(rng, -1:1, 2))`, but it seems that order of operations has changed in 1.6rc1. Even with `StableRNGs` we cannot use the former. On 1.6 the number of allocations is the same, so using the latter is not a problem.

I've added tests and seed changes etc for the in-model pool too. I think we keep the question of immutable vs mutable in #421 and do some benchmarking there. For now I'm happy with this and think it's ready to merge, but @Datseris perhaps you can give it another once-over.

Datseris:
@Libbum what was the reason to use a `Ref` instead of the direct field? At the moment I've changed it to non-ref to see if things are okay, but given the API you have defined, I don't see a reason to use `Ref`. MersenneTwister is mutable by itself for example.

Libbum:
Yeah, those changes look good actually! Cool.

Datseris:
Your reason to use `Ref` was to be able to "re-seed" `RandomDevice` generators. Hwoever, since those are anyways not reproducible, no reason to "seed" them.

Libbum:
Yep. The new methods work fine without the need for `Ref`.

PR #425, opened 2021-02-22
New intro docs

Datseris:
This is an attempt to make the documentation of Agents.jl more welcoming to new users. 

- Remove comparison from homepage
- Add some words about the design of agents in homepage
- Declutter tutorial by removing non-crucial information. (obviously the spaces are crucial information however listing there the docs of all spaces takes a lot of space, and is not necessary at the moment)

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -1,6 +1,7 @@
 # API
 
 The API of Agents.jl is defined on top of the fundamental structures  [`AgentBasedModel`](@ref), [Space](@ref Space), [`AbstractAgent`](@ref) which are described in the [Tutorial](@ref) page.
+In this page we list the remaining API functions, which constitute the bulk of Agents.jl functionality.
 
 ## `@agents` macro
 The [`@agents`](@ref) macro makes defining agent types within Agents.jl simple.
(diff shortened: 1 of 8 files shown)


docs/src/api.md
```diff
@@ -37,9 +52,21 @@ genocide!
 sample!
 ```
 
+### `@agents` macro
```
Libbum, line 55: Should this section be moved to above "Adding agents"? Since this is ultimately "creating agents".

Datseris: good point

docs/src/performance.md
```diff
@@ -0,0 +1,4 @@
+# Performance Tips
```
Libbum, line 1: This isn't included anywhere. 

I guess we could move the 'a note on iteration' from the API page here though?

Libbum:
These are good additions / changes. Whilst we're looking into this, one thing that always bothered me was that methods like `run!` were only on the tutorial page and not listed in the API.

Datseris:
> Whilst we're looking into this, one thing that always bothered me was that methods like run! were only on the tutorial page and not listed in the API.

Unfortunately documenter doesn't allow expanding a documentation string in two places :(

Libbum:
Oh really! That's... odd. 

Sure, then having a tutorial is more important.

docs/src/performance.md
Libbum: This file is still here, but not linked anywhere...

Issue #426, opened 2021-02-22
Moving OSM Agents by time, rather than by distance


SteffenMuehle:
## Feature request:
I would like to move agents on an OSM map by *time* rather than by *distance*, taking into account local road speeds.
For intuition, I've attached a .gif I've made where it's clear that agents have different speeds on different roads.
![bus_time_test](https://user-images.githubusercontent.com/28302575/108683656-2fee8300-74f2-11eb-8967-b6dcd6485d82.gif)

## Solution wanted
The solution (=function) I'd like would look like
> time=5.0
> move_agents_by_time(model,time)

or as a kwarg
> move_agents(model;time=5.0)

This function should take into account the local road's (=graph edge) speed limit.
This is already given in the model's map:
> model.space.m.class
with speed limits in km/h
> OpenStreetMapX.SPEED_ROADS_URBAN

## My custom solution
I've written two functions, which together solve this issue. Both are attached. Two further little helper functions are also attached:
[get_travel_times.txt](https://github.com/JuliaDynamics/Agents.jl/files/6020559/get_travel_times.txt)
[move_agent_by_time!.txt](https://github.com/JuliaDynamics/Agents.jl/files/6020560/move_agent_by_time.txt)
[two_little_helper_functions.txt](https://github.com/JuliaDynamics/Agents.jl/files/6020561/two_little_helper_functions.txt)

#### Explanation
1) "get_travel_times": Obtains the map's edge's weights as travel times.
This function is exactly taken from the OpenStreetMapX.jl package (see here https://github.com/pszufe/OpenStreetMapX.jl/blob/b74a8d432efd61683a9eaea3282381feceaa8b44/src/routing.jl#L63), but I've altered it such that the output is a sparse matrix in the same form as the map's original weights (travel distances) that are given in model.space.m.w.

2) "move_agents_by_time"
- Converts agent's position' third element (=distance between two nodes in meters) into a time (in seconds)
- exchanges model's graph weights with travel times rather than distances
- performs "time step" by using given Agents.jl functions for distance-steps.
- turns agent.pos[3] into distance again
- retrieves distance-weights

#### Suggestion
My solution works but needs to pass the travel_times as an explicit argument. It would be more convenient if this sparse matrix would just be a field of the model's map model.space.m. This map is a struct of data type "MapData" as given by the OpenStreetMapX.jl package. So far, it has a single weight matrix model.space.m.w for distances between nodes.

## Minimal example
[...]

PR #429, opened 2021-02-22
added 4 functions to src/spaces/openstreetmap.jl that allow moving agents in time, not space

SteffenMuehle:
This pull requests refers to issue 426:
https://github.com/JuliaDynamics/Agents.jl/issues/426

I've made a new branch move_by_time starting off from the master branch.
Only change: I added te 4 functions described in the issue to src/spaces/openstreetmap.jl

### DIFF
--- a/src/spaces/openstreetmap.jl
+++ b/src/spaces/openstreetmap.jl
@@ -662,3 +662,69 @@ function nearby_positions(
     end
     neighborfn(model.space.m.g, position)
 end
+
+
+
+#returns sparse matrix with travel times between connected nodes
+function get_travel_times(m::MapData, class_speeds::Dict{Int,Float64} = OpenStreetMapX.SPEED_ROADS_URBAN)::SparseMatrixCSC{Float64,Int64}
+    sparse_times=copy(m.w)
+    @assert length(m.e) == length(m.w.nzval)
+    indices = [(m.v[i],m.v[j]) for (i,j) in m.e]
+    for i = 1:length(m.e)
+        sparse_times[indices[i]...] = 3.6 * (m.w[indices[i]]/class_speeds[m.class[i]])  # t=s/v. 3.6 factor because of physical unit change from km/h to m/s
+    end
+
+    return sparse_times
+end
+
+
+#translate progress between nodes (agent.pos[3]) from distance into time
+function distance_to_time!(agent,sparse_distances,sparse_times) 
+    distance_since_node=agent.pos[3]
+    edge_length=sparse_distances[agent.pos[1:2]]
+    edge_time=sparse_times[agent.pos[1:2]]
+    time_since_node= edge_length != 0 ? distance_since_node*edge_time/edge_length : 0    
+    agent.pos=(agent.pos[1:2]...,time_since_node)
+    
+    return nothing
+end
+
+
+#translate progress between nodes (agent.pos[3]) from time into distance
+function time_to_distance!(agent,sparse_times,sparse_distances) 
+    time_since_node=agent.pos[3]
+    edge_length=sparse_distances[agent.pos[1:2]]
+    edge_time=sparse_times[agent.pos[1:2]]
+    distance_since_node= edge_time != 0 ? time_since_node*edge_length/edge_time : 0
+    agent.pos=(agent.pos[1:2]...,distance_since_node)
+    
+    return nothing
+end
+
+
+#moves a single agent in time, not space
+function move_agent_by_time!(
+    agent::A,
+    model::ABM{<:OpenStreetMapSpace,A},
+    time::Real,
+    sparse_times::SparseMatrixCSC{Float64,Int64}
+) where {A<:AbstractAgent}
+    
+    # make identical model, only difference: node weights by time, not distance
+    sparse_distances=model.space.m.w
+    model.space.m.w=sparse_times
+    
+    # turn pos[3] into time increment rather than distance increment
+    distance_to_time!(agent,sparse_distances,sparse_times) 
+    
+    # perform move with time-weights instead of distance-weights between nodes
+    move_agent!(agent,model,time)
+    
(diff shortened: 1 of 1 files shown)


Libbum:
Thanks Steffen! I'll have to think a little about how we name this function and consider if we can incorporate it into `move_agent!` directly. We try not to make too many custom move functions if possible.

Leave it with me for a while and I'll come back to it soon.

SteffenMuehle:
I got two emails saying that a bunch of stuff failed and I don't really know what it means. Pls let me know if I'm supposed to do something ;)
Also, no stress from my end!

Libbum:
SparseMatrixCSC is not something we expose in Agents.jl, so the tests cannot find it and fails.

Don't worry about it at all&mdash;we first figure out what we should do in terms of design, then make sure our tests are passing.

src/spaces/openstreetmap.jl
```diff
@@ -662,3 +662,69 @@ function nearby_positions(
     end
     neighborfn(model.space.m.g, position)
 end
+
+
+
+#returns sparse matrix with travel times between connected nodes
+function get_travel_times(m::MapData, class_speeds::Dict{Int,Float64} = OpenStreetMapX.SPEED_ROADS_URBAN)::SparseMatrixCSC{Float64,Int64}
```
Datseris, line 669: There is no gain here in having type anotations for both `class_speeds` or the return value of the function (Sparse matrix)

Datseris:
So, here, is "max speed" dictated only by the street speed limit? If so, we should allow also the possibility of considering agent allowed speed, or agent possible speed. So I can imagine an interface that works as follows:

* `move_agent!` has a keyword `by` with values `:minutes` or `:meters`. It dictates whether the agent moves meters or by time.
* If by time, a "speed" is produced: First, the road allowed max speed is used. Then, we check if agent has a field `speed`, or something similar, or even better providing `speed` as a keyword to `move_agent!`. The agent then moves with the given speed, as long as it does not exceed the street speed limit.

This keeps the interface to a single function `move_agent` (which of course will have an _independent documentation string_ specifically for OSM space).

SteffenMuehle:
@Datseris that sounds like a good idea, having a maximum speed for each bus.
With the current code I pushed, this could be easily implemented:
Each graph edge has a class that is given in the .osm file. The Dict `class_speeds` dictates the used velocity for each edge, depending on its class. Each value `x` in `class_speeds` could simply be replaced by `maximum([x,agent_max_speed])`.

Side note: The OSM package provides two class_speed dicts: `OpenStreetMapX.SPEED_ROADS_RURAL` and `OpenStreetMapX.SPEED_ROADS_URBAN` with the latter being used by default. I'm not sure why that is so, however, as the road class should in principle already tell us whether its an urban or a rural road.. just wanted to mention it.

Datseris:
(unrelated: don't do this: `maximum([x,agent_max_speed])`, it allocates a vector with these two values. use instead `max(x, agent_max_speed)`.

SteffenMuehle:
> (unrelated: don't do this: `maximum([x,agent_max_speed])`, it allocates a vector with these two values. use instead `max(x, agent_max_speed)`.

Right, makes sense, thanks!

Datseris:
@SteffenMuehle I think you should go ahead in this PR with the design I proposed above, set up some basic tests for it, so that we can then merge it as a new feature.

Datseris:
@SteffenMuehle bump

SteffenMuehle:
Yep, don't worry I'll do it, it's still on my agenda, I've just been a bit busy this week, sorry :)

Datseris:
No worries, a bump isn't supposed to put pressure or anything, just in case you forgot, as it is quite hard to keep track of notifications on GitHub (at least if you are like me)

SteffenMuehle:
@Datseris 
Alright, I have a functioning version that allows me to move agents by kwarg meters, kilometers, miles, seconds, minutes and hours, all with the single function 'move_agents!(..., :minutes) for example. A maximum speed can be chosen which the agents cannot exceed, even if the road would allow it.
Like before in my custom solution, it works by (only once) precalculating the graph weights (=sparse matrix) for time-moves and for distance-moves. I'm storing both of these matrices as model properties.

It works but may not be the most elegant way of doing it. One certain downside is that the maximum speed is a global model property and cannot differ from agent to agent: That would require calculating the graph weights "live" during the simulation.

Can you remind me where to put the code in the PR?
I currently have put the functions into src/spaces/openstreetmap.jl and use a Jupyter notebook for testing it.

Datseris:
> Like before in my custom solution, it works by (only once) precalculating the graph weights (=sparse matrix) for time-moves and for distance-moves. I'm storing both of these matrices as model properties.

This should be stored in the space instance, not the model properties, and should be done once at space creation.

> One certain downside is that the maximum speed is a global model property and cannot differ from agent to agent: That would require calculating the graph weights "live" during the simulation.

Sure, the agent speed is what matters more as the actual velocity used will be `minimum(agent_speed, road_speed)`


> Can you remind me where to put the code in the PR? I currently have put the functions into src/spaces/openstreetmap.jl and use a Jupyter notebook for testing it.

Yes put them there in this PR for now and we'll comment accordingly.

src/spaces/openstreetmap.jl
```diff
@@ -664,67 +620,192 @@ function nearby_positions(
...
-    end
+export initiate,
+    move_agent!
 
-    return sparse_times
+mutable struct Dummy <: AbstractAgent
+    id::Int                              #agent id
+    pos::Tuple{Int,Int,Float64}          #current position
+    route::Array{Int,1}                  #route of currently active job.
+    destination::Tuple{Int,Int,Float64}  #current destination (of active job)
```
Datseris, line 642: ????

AayushSabharwal:
Can we close this? Moving by time is supported now thanks to LightOSM.jl

PR #430, opened 2021-02-22
fixed bug in src/spaces/openstreetmap.jl's function 'travel'

### LINKED ISSUE #427: OSM move_agents not working properly for long distances
# Bug
Given an OSM agent "a" with some position a.pos, destination a.destination and route a.route can be moved with
> move_agent!(a, model, distance)

I believe this function (and in particular subfunction "travel!") silently assume that the distance steps "distance" are small compared to the OSM map's typical edge lengths:
A sequence of consecutive steps leads the agent to its destination.
**One very long distance step does not lead the agent to its destination**.
Instead, the agent seems to rest at its route's last node instead.

# Reproduce bug:
Take the Zombie outbreak example, then do
> model = initialise()

check agent's position, destination and route
> model[1]

perform large number of small steps (**good**)
> for _ in 1:10^4
move_agents!(model[1],model,10)
end

and compare the outcome with that of one large step (**bad**)
> move_agent(model[1],model,10^5)

In the latter case, you should see that the agents position and destination do not coincide as they should.

# Proposed solution
I believe the function "travel!" ([line 446](https://github.com/JuliaDynamics/Agents.jl/blob/8e51d6926b65eb31d5fb49ce8cd6cdc0cfc09652/src/spaces/openstreetmap.jl#L446)) does not behave correctly in the case of a huge step length. Line 446 should instead have the same logic as the function "park!" ([line 453](https://github.com/JuliaDynamics/Agents.jl/blob/8e51d6926b65eb31d5fb49ce8cd6cdc0cfc09652/src/spaces/openstreetmap.jl#L453)), but with the *virtual agent* (inside the travel! function) having already moved so to say.
I've fixed this and it now seems to work properly for me:
[my_travel.txt](https://github.com/JuliaDynamics/Agents.jl/files/6021316/my_travel.txt)

I've commented the respective code section I've changed in the attached file.

### Explanation
I think the logical error the function "travel!" makes is that it did not expect the agent's route to become empty while running with plenty of distance left to travel. Fortunately, this special case is handled in the function "park!" which I've inserted manually.

# Version
Status `~/.julia/environments/v1.5/Project.toml`
  [46ada45e] Agents v4.0.1

SteffenMuehle:
This closes #427

Starting from the master branch, I've updated the function "travel!" in src/spaces/openstreetmap.jl:
Only the commented part is new, and it is essentially a copy+pasted version of the "park!" function (same file).

### DIFF
--- a/src/spaces/openstreetmap.jl
+++ b/src/spaces/openstreetmap.jl
@@ -444,6 +444,7 @@ function travel!(start, finish, distance, agent, model)
             )
         else
             #######################################
+            # TODO: The code here can be simplified by using the existing `park`.
             # alright, so here we're in a situation where the agent is imagined to be at 'start' with 'distance left to travel.
             # the route is empty, but (start,finish) does not equal agent.destination[1:2]
             # what follows is the srouce code of the function "park", but the 'virtual' agent in it has position
--- a/src/spaces/openstreetmap.jl
+++ b/src/spaces/openstreetmap.jl
@@ -443,7 +443,35 @@ function travel!(start, finish, distance, agent, model)
                 model,
             )
         else
-            return (finish, finish, 0.0)
+            #######################################
+            # alright, so here we're in a situation where the agent is imagined to be at 'start' with 'distance left to travel.
+            # the route is empty, but (start,finish) does not equal agent.destination[1:2]
+            # what follows is the srouce code of the function "park", but the 'virtual' agent in it has position
+            # pos=(start,finish,distance-edge_distance)
+            # so I replaced that and nothing else.
+            distance-=edge_distance
+            if finish != agent.destination[1]
+                # At the end of the route, we must travel
+                last_distance = osm_road_length(finish, agent.destination[1], model)
+                if distance >= last_distance + agent.destination[3]
+                    # We reach the destination
+                    return agent.destination
+                elseif distance >= last_distance
+                    # We reach the final road, but not the destination
+                    return (agent.destination[1:2]..., distance - last_distance)
+                else
+                    # We travel the final leg
+                    return (finish, agent.destination[1], distance)
+                end
+            else
+                # Reached final road
+                if distance >= agent.destination[3]
+                    return agent.destination
+                else
+                    return (agent.destination[1:2]..., distance)
+                end
+            end
+            #######################################
         end
     else
         return (start, finish, distance)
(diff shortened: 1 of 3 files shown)


Datseris:
@SteffenMuehle thanks, great. 

What is good practice once a new feature (or improvement/fix of an existing one is added) is to test it. Do you think you can write a quick test for this? You can re-use the zombie model. Initialize a single agent, set a route for it, and then step it. Is the agent in the correct position?

You can put this test in the `test` folder of the repository. Then you can use the `@test` macro, like `@test agent.pos == expected_pos`. Put this in the file `osm_test.jl`. In fact, you should re-use the model used in these tests.

SteffenMuehle:
Hi George, yep, I've tested it and it ends up at the expected position.
I have a minimal example running here for the Zombie model in my Jupyter notebook.
Therein, the @test macro doesn't work though.. I could upload my notebook, or my notebook in .jl form?
Or paste all my code into that osm_test.jl file?

Libbum:
`test/osm_tests.jl` is the right place, yep!

Datseris:
@SteffenMuehle 

> I could upload my notebook, or my notebook in .jl form?

You can upload Jupyter notebooks as Gists: https://gist.github.com/  and GitHub will render them properly as well BTW.

SteffenMuehle:
I've now added the code from my minimal notebook in test/osm_test.jl.
I'm not entirely sure if that's what you wanted or if the @test macro is used correctly.. so let me know :)

Datseris:
Great, yeah that looks good to me! Here are the steps that remain:

1. Merge current master into this PR. Do you know how to do it? (I don't use `git`, but only GitKraken, and there it is just a drag and drop)
2. Increment patch version in the `Project.toml` file to `4.1.1`.

and that's it. Because this is a bug fix I (personally) do not think it is worth adding in the changelog.

Libbum:
I've made the test case use the existing version of OSMSpace (since it takes a bit to load). 

There's probably a way we can re-use most of `park` rather than copying a lot of it, but I think it's a bit too late in the evening for me to see how :sleeping:

Datseris:
If we can simplify source, then we should wait and do it before merging. Clean source is the top priority and is what allowed us to have a simple interface and easy contributions from other users!

src/spaces/openstreetmap.jl
```diff
@@ -443,7 +443,35 @@ function travel!(start, finish, distance, agent, model)
                 model,
             )
         else
-            return (finish, finish, 0.0)
+            #######################################
```
Datseris, line 446:
```suggestion
            #######################################
            # TODO: The code here can be simplified by using the existing `park`.
```

Issue #433, opened 2021-02-22
Plotting Aspect Ratio


rmcsqrd:
**Describe the bug**
When using the `plotabm` function the aspect ratio of the plots is funky. I encountered this bug working on the FMP algorithm implementation (#412). I think that what is happening is that the underlying aspect ratio in the `plotabm` function is not 1:1 which results in an inaccurate output gif during the FMP simulations. Passing in the argument `aspect_ratio=:equal` into the `plotabm()` function call appears to fix this. See the below simulation results which are identical sans the `aspect_ratio` argument:

`aspect_ratio=:equal`
![equal_aspect_ratio](https://user-images.githubusercontent.com/48159564/108757301-e4929f80-7506-11eb-9a7e-5c8d00dc56f6.gif)

`aspect_ratio` not set
![_moving_line_output](https://user-images.githubusercontent.com/48159564/108757293-e0668200-7506-11eb-80e4-a776431b462c.gif)

**Minimal Working Example**
[This section](https://github.com/JuliaDynamics/Agents.jl/pull/412/commits/7d18ea6a8162164fa4f7183d28a7a6690f1661fd#diff-38ed7ef460188bd80803f3f8237d07bb1df6e04b2b5578bbfd355eef950e02eaR615) of my FMP example in #412 sets the option.

My proposed fix would be to hardcode this option into `src/visualization/plot-recipes.jl` and I am happy to make the modification in my FMP implementation (#412 ) but wanted to file an issue to see if there is a reason not to do this. 

**Agents.jl version**
v4.04

AayushSabharwal:
`plotabm` is from the old `InteractiveChaos`. It's now [`InteractiveDynamics`](https://github.com/JuliaDynamics/InteractiveDynamics.jl), which should use `abm_video`/`abm_play`. Maybe trying that fixes it?

Libbum:
Whilst it's true that we'll move to InteractiveDynamics once it's completely ready, the `plotabm` recipes will still be available. 

I'm not really sure if there's any example that requires the current ratio, so it may be more representative to make `aspect_ratio=:equal` be the default for the recipe. 

I'll do that now and let the docs run, see what it looks like.

Datseris:
I don't think we should be spending much time on `plotabm` and instead focus on `abm_plot` as well as porting the available examples to it already in the docs using PR#370

Libbum:
[...]

PR #434, opened 2021-02-22
Equal aspect ratio by default

Libbum:
Quick test to see if it makes sense to add a default aspect_ratio to the plot recipe.


Libbum:
Not worth perusing. See discussion in #433

PR #435, opened 2021-02-22
Move plotting to InteractiveDynamics + Makie, part 1

Datseris:
I strongly recommend that we do this.  Only graph space and open street map space cannot be ported.

Remaining to be ported to use CairoMakie / InteractiveDynamics

- [x] Sugarscape
- [x] Weath distribution: No reason to use unicode plots, we should make a barplot (since later down the page we do a heatmap anyways)
- [x] battle royale
- [x] fractal growth


Libbum:
Was thinking the same thing really. 100% agree. Doesn't look like there's too much push from the Makie crowd, which I don't quite understand.

Datseris:
Apparently makie does support some kind of recipes, so maybe we could port the graph stuff, but I don't know how and at the moment don't have the time :D After we merge this we can keep track of remaining things to be ported in an open issue.

Libbum:
I've looked at the Plots.jl graph recipes. They're super involved. We don't have the time for that one for sure.

Libbum:
(latest commit added an .mp4)

Datseris:
https://user-images.githubusercontent.com/19669089/108860329-627da600-75ee-11eb-9c3a-d337f346227f.mp4

Datseris:
I've updated flocking and bacteria, let's see if they show up...

Datseris:
Only 4 more to go before merge. I've updated original post.

AayushSabharwal:
I'll update sugarscape. It's already on `InteractiveDynamics`, but the heatmap is absurdly large

Datseris:
What do you mean by `large`? In filesize? Notice that to update the sugar scape is much simpler in the new InteractiveDynamics interface because you can do:

```julia
fig, abmstepper = abm_plot(...)
ax_abm = content(fig[1,1])
ax_heatmap = fig[1, 2] = Axis(fig)
heatmap!(ax_heatmap)
for i in 1:steps
step!(abmstepper, model, ...) # automatically updates agents plot
update_heatmap_plot # somehow
end
```

AayushSabharwal:
I meant large in terms of visual size. It occupies basically the entire screen in the documenter/deploy preview

Datseris:
> I meant large in terms of visual size. It occupies basically the entire screen in the documenter/deploy preview

MakieLayout does automatic sizing. If you use the approach I pasted above, it should take exactly as much space as the agent plot.

AayushSabharwal:
I should've been clearer. The one in the video is fine, but the static one above it showing sugar capacities is way too large
image

Datseris:
Can you share your code in a Gist?

AayushSabharwal:
> ```julia
> fig, abmstepper = abm_plot(...)
> ax_abm = content(fig[1,1])
> ax_heatmap = fig[1, 2] = Axis(fig)
> heatmap!(ax_heatmap)
> for i in 1:steps
> step!(abmstepper, model, ...) # automatically updates agents plot
> update_heatmap_plot # somehow
> end
> ```

Would this method automatically animate in the documenter build? Also, it becomes a little more annoying because of the Colorbar beside the heatmap, and labels for titles

AayushSabharwal:
The code generating the oversized heatmap is just a couple lines
```jl
fig = Figure()
ax = fig[1, 1] = Axis(fig)
hmap = heatmap!(ax, model.sugar_capacities; colormap=cgrad(:thermal))
fig[1, 2] = Colorbar(fig, hmap; width=30)
fig
```

Datseris:
Your approach is incorrect, because it plots the heatmap and the agents on the same axis instance. I'll have a look in a couple of hours. The code I pasted is the correct way to go. It was a sketch. To make a video you wrap the `for` loop in a `record` call.

AayushSabharwal:
Alright. I don't quite get it, so I'll wait. Not quite sure on how to use `abmstepper`

Datseris:
I've updated sugarscape. @Libbum tomorrow I'm presenting Agents.jl, so I'll merge this tonight even if we don't manage to port the other two minor things

Libbum:
Battle Royale was already done. Where did it go?

Datseris:
oh yeah, you are right. I didn't think of that. https://juliadynamics.github.io/Agents.jl/previews/PR435/examples/battle/

Libbum:
OK. I can take a look over things and do some minor mods.

Something I just saw: doesn't look like the wolfsheep figure is showing up.

Datseris:
Actually I'm confused now. Where does matplotlib and conda even come from? I thought open street map was Plots.jl...?

src/visualization/plot-recipes.jl
```diff
@@ -115,3 +115,20 @@ end
     markerstrokewidth --> 1.5
     model.space.graph
 end
+
+"""
+    abm_plot_on_graph(model::ABM{<: GraphSpace}; ac, as, am, kwargs...)
```
Libbum, line 120: What does this do? plotabm already handles graphs?

Datseris: `plotabm` has no reason to exist anymore for gridspace and continuousspace, and the name was too close to `abm_plot`. I thought having an explicit name for graph plots is better for the user.

Datseris: If you see the new docs on visualization, `plotabm` is not expanded anywhere

Libbum: I added it in to the plot discussions (the bottom of the docs were not finished). Change the recipe names, sure, but just because we don't want to use Plots.jl anymore doesn't mean everyone wants to.

Datseris:
Turns out, we can't remove matplotlib :(

Datseris:
@Tim, if you can record and upload the video of the openstreet map as a saved animation to JuliaDynamics documentation repo, and crosslink it here, that's probably a massively better idea....

Libbum:
What's the dependency. Just Plots.jl?

Libbum:
It's sitting in the docs/Project.toml, I just forgot to remove it there. Perhaps that's all that needs to go?

Libbum:
There's a failure in diffeq which I've now fixed locally. I'll wait to push it with another attempt at getting rid of pyplot after this round finishes so we can check the preview again.

Libbum (approved):
If this build goes through, I think I'm finished with changes for now.

Libbum:
Preview looks pretty good. Sugarscape plots are perhaps a bit wide, but that's a minor issue.

There are sometimes problems with labels (see last plot in diffeq), but this happens only occasionally. It's probably a cairomakie bug.

Libbum:
Ah. Yes, merge is fine.

Datseris:
ops, diff eq was the one page i didn't check. But yeah, minor for now

Issue #437, opened 2021-02-23
Creating a Makie-based version of open street map plotting


Datseris:
This issue is probably better to open in OpenStreetMapXPlot.jl, but we care about it as well.

In Agents.jl we are currently moving largely towards plotting only with Makie.jl. It is too powerful to not do it!

The problem with the current open street map space is that it is extremely slow to plot. If we manage the Makie port, we will have huge benefits regarding plotting time.

Porting open street map to Makie is doable when we find some spare time. From here https://github.com/pszufe/OpenStreetMapXPlot.jl/blob/master/src/plot.jl  it looks daunting and large, but I think it is because of design. Much code can be made simpler by removing repetition of functions and propagating arguments. We could have a Makie version in 200 lines or even less. In principle we only plot `lines!` and `poly!`. For example, our Zombie outbreak example only plots `lines!` with different color. 

The core plotting is this loop:
```julia
# Draw all buildings
    if !isa(buildings,Nothing)
        OpenStreetMapXPlot.draw_buildings!(p,nodes, buildings, buildingStyle, km)
    end
    # Draw all roadways
    if !isa(roadways,Nothing)
        OpenStreetMapXPlot.draw_roadways!(p,nodes, roadways, roadwayStyle, km)
    end
    # Draw all walkways
    if !isa(walkways,Nothing)
        OpenStreetMapXPlot.draw_walkways!(p,nodes, walkways, walkwayStyle, km)
    end
    # Draw all cycleways
    if !isa(cycleways,Nothing)
        OpenStreetMapXPlot.draw_cycleways!(p,nodes, cycleways, cyclewayStyle, km)
    end
    #Draw all features
    if !isa(features,Nothing)
        OpenStreetMapXPlot.draw_features!(p,nodes, features, featureStyle, km)
    end
```
and for example all functions that plot roads should be a single function but with different style: color, width, and transparency. This will make the source code simpler.

@pszufe I am tagging you as well here, you would benefit greatly from having faster plotting of maps.

pszufe:
Yes - this would be quite easy to program.

Actually it is worth knowing that `gr()` is very fast once you correctly set the environment variable:
```
ENV["GKSwstype"]="nul"
```
But most people never set it and than observe poor performance.

For discussion see:
https://stackoverflow.com/questions/64371677/plot-window-showing-up-when-run-from-terminal-in-plots-jl

Datseris:
[...]

PR #438, opened 2021-02-23
GR nul test

Libbum:
Ref #437. Our build times are around 35-40 minutes on docs at the moment.. Lets see how this fairs out of curiosity.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,7 +1,7 @@
 cd(@__DIR__)
 using Pkg; Pkg.activate(@__DIR__)
 const CI = get(ENV, "CI", nothing) == "true"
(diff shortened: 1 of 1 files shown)


Datseris:
Just installing Conda and matplotlib takes 10 minutes... Is this because of open street map?

Datseris:
Actually I'm confused now. Where does matplotlib and conda even come from? I thought open street map was Plots.jl...?

Libbum:
Yes, its OSM. `pyplot` backend.

Libbum:
Oh, and we also needed it for things like Wolf Sheep transparent markers. That's no longer an issue though, so we can probably drop pyplot completely in #435.

Libbum:
That made exactly zero change.

PR #440, opened 2021-02-24
Pyplot removal attempt

Libbum:
Think I found the problem. Hopefully this is the fix.


test/collisions_tests.jl
```diff
@@ -71,20 +71,4 @@ K1, p1 = kinetic(model)
...
-# anim = @animate for i ∈ 1:200
-#   xs = [a.pos[1] for a in values(model.agents)];
-#   ys = [a.pos[2] for a in values(model.agents)];
-#   p1 = scatter(xs, ys, label="", marker_z=colors, xlims=[0,1], ylims=[0, 1], colorbar_title = "Agent ID")
-#   title!(p1, "step $(i)")
-#   step!(model, agent_step!, model_step!, round(Int, 1/dt))
-#   println("step $i")
-# end
-# gif(anim, "movement.gif", fps = 45);
-
```
Datseris, line 89: This was useful during testing, and may be needed in the future for debugging. You can leave it in commented and just remove the `pyplot()` line.

Libbum: Alright.

PR #441, opened 2021-02-24
Output svgs to file first

Libbum:
Following on from [discussions about DOM poisioning](https://discourse.julialang.org/t/cairomakie-and-fontconfig/55931?u=libbum), this PR will try to figure out the best way to solve that issue.

### DIFF
--- a/examples/diffeq.jl
+++ b/examples/diffeq.jl
@@ -415,7 +415,7 @@ ax =
     )
 lines!(ax, discrete, linewidth = 2, color = :blue)
 save("final.svg", f) # hide
-![](final.svg)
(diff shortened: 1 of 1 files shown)


Datseris:
why do you save svg instead of `png`?

Libbum:
Doesn't really matter. We're fixing an svg issue, so..

If this 'fixes' the issue we can use either. Julius has opened an issue on Documenter to fix the problem we have at present.

Libbum:
I think this has been recently solved in the latest Documenter.jl. Will need to test.

Libbum:
Closing, as #467 solves this problem.

PR #447, opened 2021-03-04
Pathfinding in GridSpace

AayushSabharwal:
`GridSpace` is arguably the easiest target for pathfinding. This has been implemented through the A* algorithm, extended to arbitrary dimensions.
The pathfinding data is managed by the `AStar` struct. It stores:
- Agent paths through a `Dict` mapping agent ID to a linked list containing grid positions along the path that agent will follow
- Dimensions of the grid. This has been kept separate since custom solutions might need it, for example `BattleRoyale`.
- `D`-dimensional neighborhood offsets, either Moore or VonNeumann depending on the `M` type parameter
- admissibility, which allows approximating the path for a substantial increase in performance
- The cost metric used to evaluate the distance between two points. This is how heightmaps are handled.
- Periodic nature of the grid in the `P` type parameter

The API is as follows
- A new dummy `Pathfinder` struct holds parameters for creating `AStar` and serves as a "gateway" of sorts to add pathfinding to `GridSpace`
- `GridSpace` now has a `pathfinder` field, which is `Union{Pathfinder,Nothing}`. A model with pathfinding is created by specifying the pathfinder through the kwarg of `GridSpace`. It defaults to `nothing`, indicating no pathfinding. A pathfinder can be created by passing in a `Pathfinder`, with the appropriate fields (keyword arguments) overriden. For example, `GridSpace((10, 10); pathfinder=Pathfinder(admissibility=0.3))` creates a pathfinder with `admissibility=0.3` and the rest of the fields default. To use all default values, the default constructor `Pathfinder()` is used.
- The `set_target!` function calculates and stores the path from the agent's current position to the specified one
- The `move_agent!` function now has a new method, that moves the agent along its precalulated path. If the path doesn't exist or is empty, the agent doesn't move
- `kill_agent!` has a new method, which additionally removed path data (if any) from the `AStar` struct
- A couple utility functions like `heightmap`, `walkmap`, `is_stationary` to query pathfinder data.

Apart from these, there are some details exposed that allow for custom solutions to deal with more complicated cases
- Custom cost metrics can be defined with the `CostMetric{D}` abstract type. Each metric requires a custom `delta_cost` method, so that is exposed as well. `DirectDistance` has two methods depending on whether the `AStar` struct has a Moore or VonNeumann neighborhood
- `find_path!` is exposed, in cases where the user wants to manually manage multiple pathfinders or the pathfinding space doesn't directly map to the `GridSpace`

The goal is to make the common case of a single pathfinder easy to use, with basically only one specialised function (`set_target!`) that might need to be called. Enough of the underlying code is exposed to allow for more complicated solutions to be implemented without having to either fight against the API or reimplement too much. (This will be further addressed in another PR)

The implementation is as fast as it could be made, and allows for significant speedups if approximate paths are allowed. In a benchmark with 100 agents pathing across a random maze on a 100x100 grid back and forth for 10000 steps (see https://github.com/JuliaDynamics/Agents.jl/pull/405#issuecomment-782820129) the results are as follows:
- 0 admissibility (optimal paths): `13.897 s (49249059 allocations: 10.42 GiB)`
- 0.3 admissbiiliy (paths are atmost 1.3x the optimal length): `2.491 s (10197280 allocations: 2.40 GiB)`

`DataStructures.jl` is added as a dependency for its `MutableBinaryMinHeap` (used in the A* algorithm) and `MutableLinkedList` (used to store the path, allows for efficient deletion of traversed vertices).

The runners example illustrates usage of `HeightMap` and `MaxDistance` metrics, and the maze example is a demo for using the `walkable` matrix.

https://juliadynamics.github.io/Agents.jl/previews/PR447/


AayushSabharwal:
I've actually just written up some basic docs for `Pathfinder`. If you haven't started yet @Libbum , I can commit it.

Libbum:
Sure, go ahead.

The open question for George was to consider `Chebyshev` as the default for hightmap. We should copy across the comparison runs here.

AayushSabharwal:
I would have added `(see [`CostMetric`](@ref)` but it appears that isn't exported or part of the documentation, so I left it out for now. Shouldn't it be mentioned, since it's mentioned with all the metrics?

AayushSabharwal:
Is there any work we're waiting on? Any changes that I can do?

Libbum:
There's a shader failure in our examples, so they're not plotting. Something to d with FileIO. but I don't know exactly what it is.
If you can find that, it'd be great.

Basically it just now needs some time for George and I to look it over. He's moving apartment and I'm moving jobs, so just a little unavailable at the moment!

Libbum:
Oh, CostMetric: it's in the developer docs, so you could link it, but I don't think it's needed.

AayushSabharwal:
Oh, sure. I'll see what I can do about the plotting. I'm sorry for prompting, I hope things stabilise soon.

Libbum:
I think the FileIO problem is JuliaPlots/Makie.jl#864. Patches are on the way, so probably nothing to worry about just yet. Otherwise we may need to pin FileIO 1.5.

Datseris:
The latest version of InteractiveDynamics.jl allows the keyword `static_preplot!`. They value of this keyword is a function `f(ax, model)` that plots some static elements over the plot *before* the agents are plotted. At some point in the future Makie might fix this lucky bug so we shouldn't count on the conveniencde that at the moment heatmaps are always plotted behind scatter plots.

Can you pklease update the plots so that they take advantage of this new feature? Maybe if we try to build the docs again, the plots will work.

AayushSabharwal:
I've updated the examples. They also needed to be changed to use `move_along_path!` instead of the old `move_agent!` method.

Additionally, I've left the `Colorbar` out of `runners.jl` for now, since I'm not entirely sure how to do it with only the axis available in `static_preplot!`. I'll look up Makie docs and see if I can find something

AayushSabharwal:
I resolved a conflict to follow the pattern in the rest of the modified example files in this PR.

AayushSabharwal:
It worked! The docs build, and the videos render properly (bar the absence of a Colorbar)

AayushSabharwal:
Another merge conflict

Datseris:
I finally have a functioning desk!!! Expect a proper review until the end of the weekend!

Datseris:
I'm working on this now and will change somethings. For example `diagonal_neighbors` does not make sense. This has nothing to do with neighbors of other agents. It feels to me that often in the ABM lingo the space and the agent are used interchangeably, as if the only type of ABMs that exist are Cellular Automata. There aren't any neighbors, pathfinding can happen with only 1 existing neighbor. We are talking about **neighboring positions**, something fundamentally different.

Datseris (approved):
Allright @AayushSabharwal this looks great to me! I'm happy to merge, once the following comments are addressed. They are mostly about harmonious package design and documentation.

- The functions `heatmap` and `walkable` MUST mention that if these objects are mutated, a new path needs to be planned again via `set_target!`.
- All files should be at most 96 characters length at every line
- The documentation of Pathfinding needs to be exp[lained by an initial paragraph at Path-finding docs, that expains, in English not in code, how the process works, and points out to the maze and runner examples. Alternatively, this high level description can be the opening message of the docstring of `Pathfinder`. But it must be somewhere.
- The HeightMap metric must cite the mountain runners excample
- The `walkable` keyword of `Pathfinder` must cite the maze runner exsample
- Ensure everywhere that only instances of `CostMetric` are used, not types. E.g. `HeightMap(heightmap, MaxDistance)` becomes `HeightMap(heightmap, MaxDistance())`. See for example the Distances.jl package, where such sintax of using `Euclidean` instead of `Euclidean()` is never allowed.
- (and as a result fix the bugs I introduced by not allowing `Pathfinder` to get types of metrics instead of instances)

Libbum:
The last two points are due to me. What's the reason to not use types directly here? Always using instances will simplify the interface, sure; but I don't understand why using naked types is a problem.

Datseris:
You have to write more code to support it, while they give you no real benefit. As I said in the review somewhere, the argument of "skipping writting an extra `()` " is not enough for me.

If any, this approach gives you the downside of wrongly interchanging an instance of a type with its type. Teaches wrong way of thinking, given how important the type system hierarchy is for julia. In other scenarios, using a type directly makes sense. For example `round(Type, x)` makes sense, because it makes `x` to be _of type `Type`_. Here, we use the instance of the type directly. The fact that `DirectDistance` has no actual input arguments is not relevant here.

Libbum:
Cool. Good arguments&mdash;thanks!

Datseris:
@Libbum it is also explicitly stated in the Julia Style Guide: https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-confusion-about-whether-something-is-an-instance-or-a-type

AayushSabharwal:
I have a hectic couple of days. I can address some of these issues now, and I'll get to the rest as soon as I have a little time.

> * The functions `heatmap` and `walkable` MUST mention that if these objects are mutated, a new path needs to be planned again via `set_target!`.

Currently, this doesn't happen. The only thing that invalidates paths is setting a new path for that agent though `set_target!`. or `set_best_target!`. I agree that this should happen though. I'll take a look around to see what can be done to enforce this, since currently there's nothing differentiating using the functions to view the matrices or mutate them.

> * All files should be at most 96 characters length at every line

I'll run `JuliaFormatter` on the necessary files.

AayushSabharwal:
I'm not sure how to implement automatically invalidating paths. Modifying the array elements doesn't call `Base.setproperty!` on the struct, since the struct still stores a reference to the same array. There are three other solutions I can think of:
- mention in the docs that some `invalidate_paths!` function must be called if the values are modified
- The functions return a copy. Add two new `heightmap!` and `walkmap!` functions that return the reference, assume that the value was modified and invalidate paths
- Add a boolean flag ot the functions for whether they should invalidate paths or not. Debatable whether this should default to `true` or `false`

Datseris:
> Currently, this doesn't happen. The only thing that invalidates paths  ...
You got confused. You only need to state, in plan English, in the documentation strings that a user shoulds set a new path. We don't have to do anything automatically here, the user has to.

AayushSabharwal:
Should an `invalidate_paths!` function be provided then?

Datseris:
No, why? The user just calls again `set_target!

AayushSabharwal:
Modifying either matrix invalidates paths for all agents. Rather than forcing the user to propogate this change through all agents, all paths could be easily invalidated.

Datseris:
> Modifying either matrix invalidates paths for all agents. Rather than forcing the user to propogate this change through all agents, all paths could be easily invalidated. Sorry if I'm missing something here

Yes, true, and you can write the trivial function `reset_paths!(model) = for a in allagents(model); set_target!(a, WHAT GOES HERE???); end`. There are too many problems:

1. Knowing when a user has modified an aribtrary mutable object is **extremely hard** and often impossible.
2. Are all agents of the model using path finding? Not necessarily. 
3. Updating the matrices from the user perspective might also imply that they would want to set new targets. Why not let the user decide? Otherwise how can you possibly know what target the user wants? See above `WHAT GOES HERE`.

In general my advice against too much automation is: Just make a good and simple API, but not more. The users of Agents.jl definitely have the brain to use it. Too much automation takes away power from the user, which I definitely don't like!

Datseris:
> Modifying either matrix invalidates paths for all agents

That's only conditionaly true. Take the Maze example, and imagine an agent halfway through the maze. I modify the starting half of the maze. The agent in the middle is unafected. The user should have the intelligence and information to decide which agents should be `set_target!` again and which not.

AayushSabharwal:
I see your point. It's not necessary to invalidate all paths. However, the `invalidate_paths!` function I imagined simply removes key-value pairs from the `agent_paths` field of  `AStar`. It's not necessary for the user to call it, but it simplifies invalidating everything.

Datseris:
I don't find it useful enough at the moment, and it will increase exported names. So at the moment we let users call `set_target!` as they see fit instead. A function that only removes all paths is not useful enough by itself. If it would replace wrong paths with correct ones then sure, but I already outlined above why it is impossible for us, the developers, to know this.

AayushSabharwal:
Sure.

Additionally, the only overflowing lines are a couple of `@test` macros in `grid_pathfinder_tests.jl`, and a couple `@assert`s in `grid_pathfinder.jl`. There are a couple other places with an overflowing `@assert`. Is it necessary to fix all of them? I didn't run `JuliaFormatter` since it seems to make a bunch of fairly unnecessary changes

Datseris:
Okay, let's leave the overflow for now, but keep it in mind for any upcoming PR!

AayushSabharwal:
> * The documentation of Pathfinding needs to be exp[lained by an initial paragraph at Path-finding docs, that expains, in English not in code, how the process works, and points out to the maze and runner examples

To clarify, by this you mean a plain English explanation of how the process of using the API works, mentioning the relevant functions?

Datseris:
yeap

AayushSabharwal:
Additionally, I noticed the `Base.show` methods are now outdated, since they're never used. Need to fix that

AayushSabharwal:
Looks like GLMakie is acting up

docs/Project.toml
```diff
@@ -10,6 +10,9 @@ Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterTools = "35a29f4d-8980-5a13-9543-d66fff28ecb8"
 DrWatson = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
+FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
```
Datseris, line 13: Why do we need `FIleIO` here?

Datseris:
Another change should be done here: there is no reason to load pathfinding-related things before loading spaces. The field `pathfinedr` of Discrete Space should be fully parameterized on some type `W`, and be `nothing` by default. `PathfinderType` has no reason to exist.

docs/Project.toml
AayushSabharwal: It's required to load the images in Maze Solver and Runners

AayushSabharwal:
> Another change should be done here: there is no reason to load pathfinding-related things before loading spaces. The field `pathfinedr` of Discrete Space should be fully parameterized on some type `W`, instead of `W<:PathfinderType`, and be `nothing` by default. `PathfinderType` has no reason to exist.

I'll take a look. That simplifies things a fair bit. I guess we could probably then do away with `pathfinding_options.jl` altogether, and have just one file.


> Can you please remind me why we need GLMakie?

I think it was something to do with how Mountain Runners works, but I'm not entirely sure. It'd be best if @Libbum could clarify this. Either way, I'll see if I can remove it and the related changes and build docs locally

Libbum:
GLMakie is essentially in testing here, since the more complex graphics (i.e. heatmaps) look pretty bad in CairoMakie. Plus we were using it to get around what `static_preplot` fixes. 

My assumption is that the CI tool has recently swapped `ubuntu-latest` to actually be the latest version, not a 2 year old one. If that's the case&mdash;we'll need to check how Makie.jl are building their docs now and update the CI. OR: we drop GLMakie and see how Cairo handles the new maps...

Libbum:
Yep, the [runners we use for CI are migrating](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). I've pinned the docs to run on 18, since AbstractPlotting.jl has yet to update how it builds GLFW on ubuntu 20 yet.

Libbum:
Something's wrong with the CI now...

Libbum:
Oh. 
>  GitHub Actions is now experiencing degraded performance. We are continuing to investigate.
> Mar 15, 22:05 UTC

I'll revert back to 18 and wait this out then

Libbum:
Sorry I'm behind on this (expecting approval today). Did that last change go back to using `Pathfinder` instead of `AStar`? If so, there are still many references to `AStar` throughout. A few docstrings seem to be missing as well.

AayushSabharwal:
That change removed the `pathfinding_options.jl` file. The only reason it was there was because `GridSpace` needed `PathfinderType`. With that removed, `GridSpace` is now simply parametrised on `W`. So `pathfinding_options.jl` was merged into `grid_pathfinding.jl`

Datseris:
@Libbum , I have been contemplating over this, and the FMP pull request as well, and the exported names we have are becoming quite a lot... Should we consider submodules? `Pathfinding.names`, `FMP.names`, `OSM.names` ? Because I am thinking that many names become hard to understand without context, which is e.g. why we append `osm_` to all names from open street map. If we specify by module, this appending to names is not necessary anymore.

Libbum:
Yeah, that's conversation we need to have. It would help with loading times too, since we could put OSM specific requirements (or whatever) behind a requires block.

Exactly what gets separated would need to be considered though. Does OSM get its own space, but Graph/Grid/Continuous stay in core? etc.

AayushSabharwal:
There's only one occurence of `AStar` outside of a function definition. Should the ones in the function definition stay? It's not mentioned anywhere what `AStar` is, so it could be confusing. However, changing it to `Pathfinder` isn't really true to how it works, and could lead to confusion if someone tries to look at the source.

Libbum:
We're interested in making sure the user documentation is solid, and we as developers know what's going on explicitly. The latest fix does that, so I wouldn't worry any further over that.

AayushSabharwal:
So the only thing being waited on is a review of the paragraph I wrote in the beginning of Path-finding docs and encapsulating this in a module? Is there anything else I can do?

Libbum:
```
┌ Warning: failed to run `@example` block in src/examples/schelling.md:34-36
│ ```@example schelling
│ space = GridSpace((10, 10); periodic = false)
│ ```
│   c.value = UndefVarError: GridSpace not defined
└ @ Documenter.Expanders ~/.julia/packages/Documenter/bFHi4/src/Expanders.jl:563
┌ Warning: failed to run `@example` block in src/examples/schelling.md:72-75
│ ```@example schelling
│ properties = Dict(:min_to_be_happy => 3)
│ schelling = ABM(SchellingAgent, space; properties)
│ ```
│   c.value = UndefVarError: space not defined
└ @ Documenter.Expanders ~/.julia/packages/Documenter/bFHi4/src/Expanders.jl:563
┌ Warning: failed to run `@example` block in src/examples/schelling.md:82-89
│ ```@example schelling
│ schelling2 = ABM(
│     SchellingAgent,
│     space;
│     properties = properties,
│     scheduler = property_activation(:group),
│ )
│ ```
│   c.value = UndefVarError: space not defined

┌ Warning: invalid local link: unresolved path in api.md
│   link.text =
│    1-element Array{Any,1}:
│     Markdown.Code("", "OSMSpace")
│   link.url = "@ref"
└ @ Documenter.Writers.HTMLWriter ~/.julia/packages/Documenter/bFHi4/src/Writers/HTMLWriter.jl:1823
┌ Warning: invalid local link: unresolved path in api.md
│   link.text =
│    1-element Array{Any,1}:
│     Markdown.Code("", "move_along_route")
│   link.url = "@ref"
└ @ Documenter.Writers.HTMLWriter ~/.julia/packages/Documenter/bFHi4/src/Writers/HTMLWriter.jl:1823

┌ Warning: invalid local link: unresolved path in examples/maze.md
│   link.text =
│    1-element Array{Any,1}:
│     Markdown.Code("", "static_preplot!")
│   link.url = "@ref"
└ @ Documenter.Writers.HTMLWriter ~/.julia/packages/Documenter/bFHi4/src/Writers/HTMLWriter.jl:1823
```

AayushSabharwal:
The schelling issues look like because the code should be in comments, since it's used as an illustration of the API and not as part of the actual example

AayushSabharwal:
It's mentioned in the docs that OSMSpace should also have a `move_along_route!` function. It doesn't look like that function exists for OSMSpace, which is confusing. Should the separate section (Movement with paths) be removed till the function is implemented?

Libbum:
>  It doesn't look like that function exists for OSMSpace

It's actually the standard `move_agent!` now, but the discussion was that we make this explicit. We can do it in another PR, so leave the language for now.

AayushSabharwal:
Once this builds, all those warnings should be resolved

Libbum (approved):
This looks good to go I think. Two outstanding issues that are not in scope for the PR:
1. @Datseris, the `static_preplot!` for the maze example has an off-by-one error I think. Take a look at the top and right margins.
2. I suggest we solve #458 before tagging the next release. In that way we can put this behind a module before we make any breaking changes.

Libbum:
Oh. I looked at the code for `static_preplot!` and there's nothing to it. That off by one issue is something we'll need to fix in this PR. I suspect just setting `limits!` will do.

AayushSabharwal:
I reran the benchmarks, and the new values are in the first comment. `BitArrays` seem to have made a noticeable difference

Libbum:
Yeah. I'm really not sure why `fill(true, x)` is not an alias for `trues(x)`, since they can in essence be used interchangeably, but initialize twice as fast.

AayushSabharwal:
What if `Pathfinder` constructor just took a boolean array instead of a BitArray, and converted it when initializing `AStar`

Datseris:
Mountain runners work but maze solver does not build: https://juliadynamics.github.io/Agents.jl/previews/PR447/examples/maze/ There is also some error regarding headings:

image

AayushSabharwal:
I fixed the not building in the last commit, I have the heading fix ready to go but I'm waiting for docs to build first

Datseris:
Why are some examples build with CairoMakie and some with GLMakie?

Libbum:
> Why are some examples build with CairoMakie and some with GLMakie?

No reason other than the rest are working with Cairo. It's just the two with `heatmap`s here that *need* GL. If it makes sense, we can transition everything to GL for example... There was another issue around somewhere where we discussed this, can't seem to see it at the moment though.

Datseris:
Well, provided that GLMakie *works*, then sure. What I remember is that the reason I initially wrote the docs with Cairo during the transition to InteractiveDynamics, is because CairoMakie was working great with CI. Do we use GLMakie in anyother example in this PR outside the image-loading ones?

Libbum:
Not at the moment, no. In principle most of them should be able to transition quite easily if we want to.

I know that @AayushSabharwal had to deal with size differences between Cairo and GL in the fractal model though, so it'll take some finesse and checks. So probably worth an additional PR not to pollute this one.

Datseris:
At the moment, every single example file is pulluted by having:
```julia
using CairoMakie
CairoMakie.activate!() # hide
```
so we could de-pollute it by replacing this with just
```julia
using GLMakie
```

AayushSabharwal:
Switching to GLMakie could have other consequences. Every example would need to be verified if it renders correctly, and there could be a fair chance that they don't. Either way, isn't the diff for both cases one line?

Datseris:
okay, let's do this in a different PR then. I'm convinced. I'll have a final look at the docs and everything and then merge. As Tim said, no new version will come out until we do #458 anyways , so we will have another PR that moves everything to GLMakie.

AayushSabharwal:
image
I guess since comments don't link, the `@ref` should be removed?

Libbum:
Yes, definitely. I'm also wondering why the maze has the boundary correct on the left/bottom but not on the right/top.

AayushSabharwal:
I'm not sure why that happens either. I'll look into it

Additionally, should comments still contain \` as in \`DirectDistance\` even though it doesn't affect formatting? Some pre-existing examples do, and some don't.

Datseris:
> Additionally, should comments still contain ` as in `DirectDistance` even though it doesn't affect formatting?

Yes.

AayushSabharwal:
It looks like `InteractiveDynamics` leaves that border on the top and right even without `static_preplot!` so I'm guessing it's some issue there

AayushSabharwal:
Are there any other changes to be made? I can push all of them at the same time so the final docs build once

Datseris:
Let me have a review and I will merge myself and do any minor changes necessary.

Datseris:
Thanks for the contribution @AayushSabharwal !

AayushSabharwal:
I'm delighted I could help. I plan on using this library in the future, so I want to make it as amazing as I can.

PR #450, opened 2021-03-05
Fix non-unique column names in model DataFrame when using anonymous accessor functions

ulido:
This PR changes `init_model_dataframe` to pass `makeunique=true` to the DataFrame constructor. This allows one to use dynamically generated accessor functions in `mdata` with the same internal identifier.

For example, to split the elements of an array in the model properties separate DataFrame columns one can write:
```julia
mdata = [(model) -> model.value_array[i] for i=1:N]
init_model_dataframe(model, mdata)
```
However, this fails because the resulting column names (generated from Julia's identifiers of the anonymous functions `(model) -> model.value_array[i]`) are non-unique:
```
LoadError: ArgumentError: Duplicate variable names: :#543. Pass makeunique=true to make them unique using a suffix automatically.
```

The simple fix implemented here is to instruct DataFrame to make the column names unique.


Libbum:
We have tried to disincentivize the use of anonymous functions for data collection in general. I don't see a straightforward way to provide a solution for your example without `makeunique=true`, but also fail to see what true benefit it provides. Post-processing the data from an array to cells is ultimately what DataFrames is designed to do.

Some additional questions to consider: 
- How would this example work when  `df = init_model_dataframe(model, nothing)` is called, then `collect_agent_data!(df, model, mdata)` is called later? My expectation is that this version would actually break, even with `makeunique=true` involved, so we'd need this tested and verified.
- What is the behavior for this change on the agent and aggregate agent dataframes? We would need the change to be across the board. These methods function differently to that of the model, so more verification is needed there too.

If these things can be answered and verified, with all the current machinery in place in the data collection already, such a change can be added in principle to make something like `mdata = [(model) -> model.value_array[i] for i=1:N]` easier to deal with. So we'd consider the change if these concerns can be addressed and performance does not suffer.

ulido:
Thanks for your comments, @Libbum !

>We have tried to disincentivize the use of anonymous functions for data collection in general. I don't see a straightforward way to provide a solution for your example without makeunique=true, but also fail to see what true benefit it provides. Post-processing the data from an array to cells is ultimately what DataFrames is designed to do.

The benefit (from my perspective) is that we can save a (possibly memory-intensive) post-processing step. I can collect the whole array into a single cell each time step, but then I would need to iterate over the whole column of the DataFrame and split it into different columns. Depending on the amount of collected data, this can become quite costly. If I can do the splitting (and selection) of array entries at collection time, this isn't an issue.

> How would this example work when df = init_model_dataframe(model, nothing) is called, then collect_agent_data!(df, model, mdata) is called later? My expectation is that this version would actually break, even with makeunique=true involved, so we'd need this tested and verified.

I assume you meant `collect_model_data!` instead of `collect_agent_data!`? Wouldn't this fail anyway, because there is no `:step` column in the empty DataFrame but `collect_model_data!` expects it to be there?

> What is the behavior for this change on the agent and aggregate agent dataframes? We would need the change to be across the board. These methods function differently to that of the model, so more verification is needed there too.

I've changed the approach in my latest commit, dropping the `makeunique=true` in the DataFrame initializer. I introduced a `colname` function similar to the `aggname` function for aggregators, and integrated it into the existing functionality in `collect.jl`. This also includes the agent and aggregate functionality. I've also written some tests for it in `collect_tests.jl` and made sure that all tests passed.

The `colname` function takes the function symbol name and appends any properties and their values (as strings of course), producing a unique column name even if the same function is used. This can still fail, because the conversion to string is lossy and there are edge cases when the same function is used twice with the same properties, etc.

Please let me know what you think.

ulido:
> We have tried to disincentivize the use of anonymous functions for data collection in general.

I forgot to say that this isn't just an issue for anonymous functions. You also don't get unique column names with the current code if you write it with a named function like this:
```julia
props = Vector{Function}(undef,2)
for i=1:2
    function func(m)
        m.example_array[i]
    end
    props[i] = func
end
```
This translates into both columns being called `func`. The `colname` approach gives the column names `["func_i=1", "func_i=2"]`.

Datseris:
So now why do both `colname` and `aggname` exist? Seems to me only one is necessary: whatever will access the DataFrame column.

Datseris:
Does this break anything? Are the previously created names remaining completely unafaccted?

ulido:
Thank you! That's ok, no rush.

> So now why do both `colname` and `aggname` exist? Seems to me only one is necessary: whatever will access the DataFrame column.

I thought they might be unified, yes. But I did not have time yet to really understand the aggregation code well enough to be sure nothing would break. I'll look into it.

ulido:
> Does this break anything? Are the previously created names remaining completely unafaccted?

The only case where the behavior should be different than before is when you give a function with properties - while before the column name was simply the function name, now it will have the function name plus the property names and their values attached. If the function has no properties, then the column name is the same as before.

Julia automatically assigns outer scope variables and their values as function type properties. But I don't know if there are any other cases when Julia sets function type properties.

Datseris:
> I thought they might be unified, yes. 

Yeap, let's keep things clean and simple. Here you should simply improve the existing `aggname`.

ulido:
> Yeap, let's keep things clean and simple. Here you should simply improve the existing aggname.

In the latest commit I've now merged `colname` into `aggname`. Something to consider would be to change the name of `aggname` because it's use is more general now than just aggregation, but changing user-visible APIs is likely a bad idea.

> Of course, cases without anonymous functions should be completely unaffected by this PR, and thus there shouldn't be anything breaking. Given that we so-far instructed to not use anonymous functions, this should be non-breaking if we do it right.

I agree, the only breakage possible is when someone explicitly relied on the column name generated from an anonymous function.

If you're happy with it, I can squash the commits and update the commit description.

Datseris:
Great, that's fine, I approve the changes! Thanks a lot!

> If you're happy with it, I can squash the commits and update the commit description.

Please do.

> Something to consider would be to change the name of aggname because it's use is more general now than just aggregation, but changing user-visible APIs is likely a bad idea.

Yes that is a good idea. It is not a problem to do it, we can deprecate : `@deprecate aggname new_name`. Can you suggest a better name? While `colname` sounds good, are we sure DataFrames.jl doesn't already export something like that? Perhaps `dataname`?

Datseris (approved):

Datseris:
Ah also please update Project.toml version accordingly: if we change name of `aggname` then this should be a version in the CHANGELOG as well, and increment minor version number.

ulido:
> Great, that's fine, I approve the changes! Thanks a lot!

Thank you!

> Yes that is a good idea. It is not a problem to do it, we can deprecate : `@deprecate aggname new_name`. Can you suggest a better name? While `colname` sounds good, are we sure DataFrames.jl doesn't already export something like that? Perhaps `dataname`?

I will go with `dataname`, I think that's a good name. `colname` isn't exported by DataFrames.jl either. And I'll @deprecate `aggname`.

> Ah also please update Project.toml version accordingly: if we change name of aggname then this should be a version in the CHANGELOG as well, and increment minor version number.

When you say minor version number, you mean to bump the version to 4.2? Not the patchlevel to 4.1.6?
I'll make an entry in CHANGELOG.

Datseris:
Yes. Initially this would have been a patch increment, but as we are introducing a new API name `dataname` it has to be a minor increase.

ulido:
Ok, I've updated version and changelog (also adding lines for the previous patchlevels) and squashed the commits.

Sorry for the delay, I was swamped with other stuff for the last few days.

src/simulations/collect.jl
```diff
@@ -343,19 +345,29 @@ function multi_agent_agg_types!(types::Vector{Vector{T} where T}, utypes::Tuple,
...
+@deprecate (aggname(x::Tuple{K,A}) where {K,A}) dataname(x)
+dataname(x::Tuple{K,A,C}) where {K,A,C} = dataname(x[1], x[2], x[3])
+@deprecate (aggname(x::Tuple{K,A,C}) where {K,A,C}) dataname(x)
+dataname(x::Union{Symbol,String}) = string(x)
+@deprecate aggname(x::Union{Symbol,String}) dataname(x)
+# This takes care to include fieldnames and values in the column name to make column names unique
+# if the same function is used with different values of outer scope variables.
+dataname(x::Function) = join(
+    vcat([string(x)], ["$(prop)=$(getproperty(x, prop))" for prop in propertynames(x)]), "_")
+@deprecate aggname(x::Function) dataname(x)
```
Datseris, line 370: A single `@deprecate aggname dataname` is enough, no reason to repeat it everywhere.

ulido: Ok, thanks! I thought it needed mapping for every method of `dataname`/`aggname`. I'm new to Julia's multiple dispatch infrastructure.

Datseris: No problem. multiple dispatch is the best thing in the programming universe in my opinion :D

ulido: I've replaced the multiple `@deprecate` lines with a single `@deprecate aggname dataname` in the latest push.

ulido: > No problem. multiple dispatch is the best thing in the programming universe in my opinion :D

Ah, I agree!! I fell in love with that concept once I grasped it!

src/simulations/collect.jl
```diff
@@ -343,19 +345,24 @@ function multi_agent_agg_types!(types::Vector{Vector{T} where T}, utypes::Tuple,
...
+dataname(k, agg) = string(agg)*"_"*string(k)
+dataname(k, agg, condition) = string(agg)*"_"*string(k)*"_"*string(condition)
+dataname(x::Tuple{K,A}) where {K,A} = dataname(x[1], x[2])
+dataname(x::Tuple{K,A,C}) where {K,A,C} = dataname(x[1], x[2], x[3])
+dataname(x::Union{Symbol,String}) = string(x)
+# This takes care to include fieldnames and values in the column name to make column names unique
+# if the same function is used with different values of outer scope variables.
+dataname(x::Function) = join(
+    vcat([string(x)], ["$(prop)=$(getproperty(x, prop))" for prop in propertynames(x)]), "_")
+@deprecate aggname dataname
```
Datseris, line 365: I'm not conviced about this part. Is there a reason to make a function `dataname` that accepts both tuples, but also expanded tuples? 

I think `dataname(k)` should be the only valid call signature, and if `k` is a tuple we know what to do internally. It should not be part of the public APi the version `dataname(k, agg, condition)`.

ulido: I agree, it would make more sense to leave the two and three argument methods out in favor of the tuple version. However, this was what the old `aggname` function did, so I left it in. I'm happy to deprecate the two and three argument `aggname` methods.

For this, do you know how to write a method that accepts an NTuple of length 2 or 3? Right now I have
```julia
dataname(x::NTuple{Union{Symbol,String,Function},2}) = join([dataname(s) for s in x], "_")
dataname(x::NTuple{Union{Symbol,String,Function},3}) = join([dataname(s) for s in x], "_")
```
but it would be nice to merge these two lines.

On the other hand, it might be more readable to keep it as
```julia
dataname(x::Tuple{K,A}) where {K,A} = dataname(x[1])*"_"*dataname(x[2])
dataname(x::Tuple{K,A,C}) where {K,A,C} = dataname(x[1])*"_"*dataname(x[2])*"_"*dataname(x[3])
```
i.e. similar to what the `aggname` code does.

Datseris: Just write `dataname(x::Tuple) = join([dataname(s) for s in x], "_")` ? No need to specialize further on the tuple.

ulido: Ok, that makes sense.

Datseris: > I'm happy to deprecate the two and three argument aggname methods.

Sorry, missed that, just don't mention them in the docstring, that's enough.

ulido: Hmmmm... there is an issue with the call semantics. `aggname` was written such that if you have a symbol `:weight` and you aggregate with `mean`, calling `aggname(:weight, mean)` you'll get the column name "mean_weight". With the new calling scheme `dataname((:weight, mean))` you'll get "weight_mean". Hence, it breaks the assumption of argument order in the tests, so they now fail. The tests can be changed, but I wonder if it would make sense to swap the first and second entry like this
```julia
dataname(x::Tuple) = join(vcat([dataname(x[2]), dataname(x[1])], [dataname(s) for s in x[3:end]]), "_")
```
?

ulido: > Sorry, missed that, just don't mention them in the docstring, that's enough.

Ok!

Datseris: Oh yea sorry forgot about that, yes do what you have written. This must absolutely be non-breaking! 

> The tests can be changed

No no, we have to keep it non breaking!

ulido: Makes sense! However, removing the two- and three-argument versions of `dataname` breaks tests too (not to speak of the rename `aggname` -> `dataname`). Is it ok to alter the tests in those cases?

Libbum: Don't alter the tests at the moment. It'll be best to get the solution 'right', then we can revisit to see if that makes the most sense. Once everyone is agreed, we can alter tests as a final step: only if necessary.

ulido:
I've now implemented the changes in the latest push, but left the tests alone (hence the CI failures).

src/simulations/collect.jl
ulido: Ok, I've implemented now it as discussed (leaving out any alterations to tests), and rebased against latest master.

Datseris:
I am confused. Why would the test fail now? The change you were supposed to do was only documentation: Do not mention explicitly in the docs that a version `dataname(x, y, z)` is possible. In the soruce code this version must remain, otherwise it is breaking.

ulido:
I see, I'm sorry, I misunderstood. I've restored the two- and three-argument versions now, with the tests all passing (with the appropriate deprecation warnings).

But I did not introduce `dataname(k, agg)` and `dataname(k, agg, cond)` without documenting them. I thought it would be clearer to explicitly deprecate the methods `aggname(k, agg)` and `aggname(k, agg, cond)` like this
```julia
@deprecate aggname(k, agg) dataname((k, agg))
@deprecate aggname(k, agg, condition) dataname((k, agg, condition))
```
This displays a very clear warning "Warning: `aggname(k, agg)` is deprecated, use `dataname((k, agg))` instead."

Let me know what you think, I'm happy to do it the other way too (i.e. defining but not documenting `dataname(k, agg)` and `dataname(k, agg, cond)`).

Datseris:
Sure, your suggestion is valid. There should still be `@deprecate aggname(k) dataname(k)` nevertheless.

ulido:
> Sure, your suggestion is valid. There should still be `@deprecate aggname(k) dataname(k)` nevertheless.

I have left the generic `@deprecate aggname dataname`.

I also propose to add a few tests to collect_tests.jl for the new way `dataname` deals with functions and their properties. 

* Something to test for the explicit column name, like this:
```julia
funcs = Vector{Function}(undef, 2)
for i=1:length(funcs)
    inline_func(x) = i*x
    funcs[i] = inline_func
end
@test dataname(funcs[2]) == "inline_func_i=2"
```
* A test that check if there are no issues with duplicate columns, not explicitly calling `dataname`, like this:
```julia
 _, model_data = run!(
      model,
      agent_step!,
      model_step!,
      365 * 5;
      when_model = [365 * 5],
      when = false,
      mdata = [(m) -> (m.deep.data[i]) for i=1:length(model.deep.data)],
  )
  @test Array{Float64,1}(model_data[1, 2:end]) == model.deep.data
```
* And to test that anonymous functions work well:
```julia
props = [(a)->(a.pos[i] + a.weight) for i=1:2]
df = init_agent_dataframe(model, props)
collect_agent_data!(df, model, props)
@test size(df) == (2, 5)
@test df[1, dataname(props[1])] == model[1].pos[1] + model[1].weight
@test ismissing(df[2, dataname(props[2])])
```

All of these at appropriate places in collect_tests.jl of course.

Datseris:
Yes, go ahead!

ulido:
Ok, thanks! All implemented in the latest commit.

Libbum:
@Datseris, some comments on this.
- I've dropped the version down to 4.2, we can release it as a minor bump with pathfinding
- I don't like the direction you've suggested to deprecate `aggname(a, b)` in favour of `aggname((a, b))`, since it isn't really consistent. Tests pass by using the deprecation path, but when we update them to using `dataname` the will go from
    ```julia
    @testset "aggname" begin
        @test aggname(:weight) == "weight"
        @test aggname(:weight, mean) == "mean_weight"
    # to
   @testset "dataname" begin
        @test dataname(:weight) == "weight"
        @test dataname((:weight, mean)) == "mean_weight"
   ```
   Why do I have the possibility of a single argument, then have to use a tuple in the second instance? We created this convenience for a reason. I don't see why it needs to be dropped.

@ulido: [This string](https://github.com/JuliaDynamics/Agents.jl/blob/a9a2fdae4c26120650db2d2193d28fb6952d12d8/src/simulations/collect.jl#L55-L56) in the documentation is no longer relevant. Can you suggest a better wording here that corresponds to your changes?

ulido:
> @ulido: [This string](https://github.com/JuliaDynamics/Agents.jl/blob/a9a2fdae4c26120650db2d2193d28fb6952d12d8/src/simulations/collect.jl#L55-L56) in the documentation is no longer relevant. Can you suggest a better wording here that corresponds to your changes?

Interesting you should say that. It was this sentence that made me think, ok I can use anonymous functions as accessors in a loop and then simply rename (which obviously didn't work...).
How about:

> The resulting column names are generated from the function [`dataname`](@ref). They create something like `:mean_weight` or `:maximum_f_x_pos`. In particular, you can use anonymous functions in a list comprehension to assign elements of an array owned by the agents to different columns:
> ```julia
> adata = [(a)->(a.interesting_array[i]) for i=1:N]
> ```
> For anonymous functions the column names are not predictable, but you can simply use `DataFrames.rename!` to change the returned dataframe's column names.

Libbum:
That looks good! I'll update this and fix the conflict in CHANGELOG soon. 

The test files should ultimately rename `aggname` to `dataname` before merging as well, but I'd like to resolve this question about deprecating too much before that's done.

ulido:
Ok, sounds good!

Datseris:
@Libbum my thinking is as follows:

`dataname` takes as an input `adata[i]`. It doesn't matter what `adata` is, or what `i` is. Then `dataname` will give you the name of the DataFrame column. Now, depending on the data collection the user chose, `adata[i]` is either a a symbol, a function, a 2-tuple, or a 3-tuple. In any case, all of these things is 1 argument even if they 'contain' more stuff. That is why I feel that only a single-argument version of `dataname` should be part of the public API.

Datseris:
P.s.: technical thing: this should be part of 4.2, but no reason to rush 4.2 because we want to consider the possibility of submodules. We should discuss in #458 which functionalities can be part of submodules.

Libbum:
> `dataname` takes as an input `adata[i]`

That's a decent reason in context, although current examples don't reflect this often. Most of the time we write out what we're interested in plotting. It's not so educational to say `plot(dataname(adata[17]))`, but it is to say `plot(dataname(:weight, sum))`.

It's a small point though. Will update the tests etc later today.

Datseris:
The previous docstring was actually wrong, not just suboptimal. It stated:

image

But clearly, in the 2nd and 3rd version `k` is something fundamentally different than what it is in the first version. In the first version it is valid that `k=adata[i]`, but in the remaining versions it must be that `k = adata[i][1]`. 

Allowing multi-argument input is fine, but one has to be clear about it.

> It's not so educational to say plot(dataname(adata[17])), but it is to say plot(dataname(:weight, sum)).

Yes, but you can simply do `dataname((:weight, sum))`, you don't have to do `dataname(adata[17]))`. its about giving it a tuple or not a tuple, not how you express the tuple.

Libbum:
I'm running the documentation locally to look it over, but in principle this should be good to go now. Will merge once the docs are verified. Thanks for your work on this @ulido! It will come out in 4.2, which shouldn't be too far away&mdash;there are two issues we need to resolve first.

Libbum:
Optim example:
```julia
function cost_multi(x)
    model = model_initiation(;
        Ns = [500, 500, 500],
        migration_rate = x[1],
        death_rate = x[2],
        β_det = x[3],
        β_und = x[4],
        infection_period = x[5],
        reinfection_probability = x[6],
        detection_time = x[7],
    )

    initial_size = nagents(model)

    infected_fraction(model) =
        count(a.status == :I for a in allagents(model)) / nagents(model)
    n_fraction(model) = -1.0 * nagents(model) / initial_size

    _, data = run!(
        model,
        agent_step!,
        50;
        mdata = [infected_fraction, n_fraction],
        when_model = [50],
        replicates = 10,
    )

    return mean(data.infected_fraction), mean(data.n_fraction)
end
```
```julia
julia> cost_multi(x0)
ERROR: ArgumentError: column name :n_fraction not found in the data frame
Stacktrace:
 [1] lookupname
   @ ~/.julia/packages/DataFrames/oQ5c7/src/other/index.jl:288 [inlined]
 [2] getindex
   @ ~/.julia/packages/DataFrames/oQ5c7/src/other/index.jl:297 [inlined]
 [3] getindex(df::DataFrame, #unused#::typeof(!), col_ind::Symbol)
   @ DataFrames ~/.julia/packages/DataFrames/oQ5c7/src/dataframe/dataframe.jl:440
 [4] getproperty
   @ ~/.julia/packages/DataFrames/oQ5c7/src/abstractdataframe/abstractdataframe.jl:348 [inlined]
 [5] cost_multi(x::Vector{Float64})
   @ Main ./REPL[28]:28
 [6] top-level scope
   @ REPL[39]:1
```

Pretty sure this will have to do with changing the column names to strings rather than symbols. I'll take a look later today.

Datseris:
is `data.n_fraction` valid dataframes syntax? @ulido why did you change all column names from Symbol to String?

Libbum:
> is `data.n_fraction` valid dataframes syntax?

Yes, we use it all the time.

Oddly enough, taking the method out of the function works without error.

I've noticed we use Symbols as headers for agent and model frames, but String for the aggregate agent frame&mdash;so we're not consistent anyway. I'll investigate further.

ulido:
I can‘t run it myself at the moment, but I suspect it has nothing to do with changing to strings. The difference between `infected_fraction` and `n_fraction` is that the latter references `initial_size` from the outer scope. This means that `initial_size` as stored as a property of `n_fraction`. `dataname` in turn takes this and appends it to the column name. So the resulting column name is “n_fraction_initial_size=...“ with whatever the value of `initial_size` was.

This is obviously not ideal and needs to be fixed. Perhaps a flag as part of the tuple in `mdata` or `adata` could be set if we want `dataname` to include properties, defaulting to not including them?

ulido:
> is `data.n_fraction` valid dataframes syntax? @ulido why did you change all column names from Symbol to String?

I had to change it, otherwise I could not have appended the property name and value pairs.

ulido:
Or we could only allow `dataname` to append properties when the function is truly anonymous and not names? Then nobody would try reference the column name.

ulido:
Instead of what we have now:
```julia
dataname(x::Function) = join(
    vcat([string(x)], ["$(prop)=$(getproperty(x, prop))" for prop in propertynames(x)]), "_")
```
we could do an extra check like this
```julia
function dataname(x::Function)
    name = string(x)
    if startswith(name, "#")    # Check if `x` is an anonymous (unnamed) function
        for prop in propertynames(x)
            name = name*"_"*"$(prop)=$(getproperty(x, prop))"
        end
    end
    return name
end
```

This would avoid the issue above and produces unique names for the comprehension construction of `mdata` and `adata`. However, it would NOT produce unique names in this case:
```julia
mdata = []
for i=1:2
    function accessor(m)
        return m.interesting_array[i]
    end
    push!(mdata, accessor)
end
```

Libbum:
Thanks @ulido. Haven't had the chance to get back to this yet. I'm not certain we even need to go that direction, but good that you've got a working example if needed.

Datseris:
What's holding us back from wrapping and merging this up? We know why the optim example cited is not working?

Libbum:
Essentially making sure it doesn't break, that's all. This was ultimately my initial concern with such a change. It's close, but still needs some attention. The solution above only gets us half way across the line, so it's a stop gap really. Would be better to go back to Symbols in headers as we had previously, but keep the new `dataname` methods. I just need some free time to do that.

Libbum:
So the ultimate issue here is that the optim example is poorly parameterized, and as such the `n_fraction` function is curried. Thus the 'true' column name for `mdata[2]` in this case should rightly be `n_fraction_initial_size=1500`, as suggested by `dataname`.

Rather than messing about with anything further, I've altered the optim example to use `dataname` rather than attempting to use `data.n_fraction`. 

Once docs are ran (they wont pass due to the fact we're on a fork), this should be ready to merge provided there are no errors.

PR #463, opened 2021-03-23
Move to GLMakie for all plots

Libbum:
A straight removal of Cairo. Will wait for the docs to build, check timing and visual differences, then consider if we should proceed down this route or not.


Libbum:
- [ ] Size of agents in Social Distancing is too big
- [ ] Same with Game of Life
- [ ] HK final plot needs to be enlarged vertically
- [ ] Daisyworld static plots need to be resized, they are clipping the legends
- [ ] Same issue for Wolf-Sheep
- [ ] Agent plot in Wolf-Sheep needs some work. Probably stroke width, alpha channel alterations and size
- [ ] Maze Runner still off-by-one
- [ ] Can we do something about aliasing? Flockers, bacteria etc are very pixelated
- [ ] Aspect ratio of bacteria is wrong
- [ ] Quality of plots in general are quite poor. Should we increase the default resolution? See Population Stability plot in Opinion spread for example.
- [ ] Legend cutoff in battle royale
- [ ] Size in fractal growth way too big
- [ ] General quality issues in DiffEq  and Measurements

Overall, I'm not that happy with the quality of the plots vs the nice SVGs we had. Now that documenter has a fix for DOM poisoning, I don't know if its worth the change over. 

Using the cache, build times are similar.

Thoughts?

Datseris:
How can build time not be affected, because both Cairo and GL take some time to build..,.? For me most things look fine btw, besides some size issues. But I think simply going with CairoMakie and not caring that the mountain climbers example doesn't look that good is also totally fine. The aliasing in bacteria model / flocking is just too bad here.

Libbum:
We are now caching the artifacts directory, so builds are better (unless there's a new release of GL/Cairo)

I interpret your sentence in multiple ways here sorry. We have three options
1. Fix these issues if possible and transition to GLMakie.
2. Drop GLMakie, use Cairo all the time, including heatmaps.
3. Stay as we are and use both.

What are you leaning towards? At the moment I'd advocate for 3, although we need to do some small fixes to make that decent (Documenter.jl#master basically).

Datseris:
Sorry, I wasn't clear! I'm definitely in favor of using a single backend, and more so towards 2, CairoMakie. The aliasing of `Polygon`-based plotting is really not great with GLMakie, and I don't think we can do something about it here on our end. The heatmap problems were not that bad with CairoMakie. 

What does Documenter#master offer?

Libbum:
It makes #441 redundant, fixing [this issue](https://discourse.julialang.org/t/cairomakie-and-fontconfig/55931?u=libbum).

Alright. Let's do that then. I'll work on a reverse version of this moving to Cairo completely.

PR #467, opened 2021-03-25
Complete CairoMakie solution

Libbum:
Need to verify everything runs, but should close out a number of plotting concerns we have open at present. May need to move to Documenter.jl#master, but yet to be confirmed.

Drops Plots, GLMakie and AbstractPlotting from the docs dependencies.

TODO: Drop all the GLMakie-specific dependencies in the CI if all this works.

docs: https://juliadynamics.github.io/Agents.jl/previews/PR467/


Datseris:
Why do we need plot recipes again?

Libbum:
I thought it was related to this branch, but it's not.

Our documentation is *sometimes* spooling forever. See the latest master commit - killed after 5.5 hours. 

Something has changed on the runners, but it's very difficult to pin down.

Libbum:
Issues to clear up before merging
- [ ] Colours don't seem to work now on fractal_growth
- [ ] No colourbar on runners
- [ ] Daisyworld looks awful (game of life is not so good either - removing stroke color for these examples is the fix I think)
- [ ] Measurements has something off with it. The black values may not be plotting?

Datseris:
Runners and daisyworld have been revamped in my branch `performance_tips` and they look gorgeus. I'm waiting for you to merge this in so that I rebase my pr on this.

Libbum:
I have no idea what this colours issue is. Turn on Cairo, everything is black, turn on GL, everything is fine.

Datseris:
It might be the marker that you use, try one of the default from http://makie.juliaplots.org/stable/abstractplotting_api.html#AbstractPlotting.available_marker_symbols-Tuple{}

Datseris:
> I have no idea what this colours issue is. Turn on Cairo, everything is black, turn on GL, everything is fine.

Sorry, I lost touch: the option of using GLMakie for every example is because it looks like crap for saving output, right?

Libbum:
Marker style was the solution!

A number of factors, but yes&mdash;output was terrible.

PR #472, opened 2021-03-31
Performance Tips & Small rework of multi-agent models

### LINKED ISSUE #454: Create performance tips documentation page
I'm starting this here to group various scattered tips until someone starts a pull request that adds the page.

- [ ] Avoid mixed type models due to inherent calls to dynamic dispatch, see #445 
- [ ] Use type-stable containers for `properties`, see #453 
- [ ] SOmething about immutable agent types, see #421

Feel free to leave comment below and I'll add it to the list.

Datseris:
Closes #454


examples/predator_prey.jl
```diff
@@ -1,4 +1,4 @@
-# # Model of predator-prey dynamics
+# # Predator-prey dynamics: simple
```
Libbum, line 1: This is listed as simple, but also `predator_prey_fast.jl` is also simple?

docs/src/performance_tips.md
```diff
@@ -0,0 +1,62 @@
...
+
+Please do read through Julia's own [Performance Tips](https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-tips) section as well, as it will help you write performant code in general.
+
+## Avoid `Union`s for multi-agent models
+Due to the way Julia's type system works, and the fact that agents are grouped in a dictionary mapping IDs to agent instances, using multiple types for different agents always creates a performance hit because it leads to type instability.
+
+To avoid this, you can have a single agent type including all properties all kinds should have. You can specify what "kind" of agent it is via including a field `type` or `kind` whose value is a symbol: `:wolf, :sheep, :grass`. Properties that should only belong to one kind of agent could be initialized with a "null" value for the other kinds.
+
+While this will increase slightly the amount of memory used by the model, as all agent instances will contain more data than necessary, the performance gain due to type stability typically makes up for it.
```
Libbum, line 16: I don't know if this is the case generally. It could lead to a *lot* of branching in certain cases.

Datseris: >  It could lead to a lot of branching in certain situations.

Only one `if` statement is necessary . `if kind == :wolf; agent_step_wolf!() elseif kind == :sheep; agent_step_sheep!()`, etc.? I should write this down explicitly to improve the docs.

Datseris:
After https://github.com/JuliaDynamics/InteractiveDynamics.jl/pull/42  I've reworked two models: daisyworld and wolfsheep to not have an agent for the space property, but rather matrices under the model properties. 

@Libbum can you please do me a favor and check two things:

1. I may have done something wrong on the wolfsheep model. Populations always die out in all cases. 
1. Do we get any performance improvement with the new wolfsheep or new daisyworld.

Libbum:
Sure, I'll hopefully get some time tomorrow.

Libbum:
Your 'faster' implementation looks good. I'm not sure where the failure is coming from on the surface. Will need to do some digging.

Libbum:
Is there a reason why we don't just return the CartesianIndices in spaces/grid.jl?
```julia
function positions(model::ABM{<:GridSpace})
    x = CartesianIndices(model.space.s)
    return (Tuple(y) for y in x)
end
```

Datseris:
@Libbum did you confirm whether the "faster" version of wolf sheep is actually faster or not? I'd like to wrap up this PR, which means replacing the old models with the new ones, provided the new ones are actually faster.

Libbum:
Yeah exactly. I spent some time on it this morning and I'm still not sure why everything dies.

Basically though, I can't confirm if it's faster or not if extinction occurs. That needs to be fixed before we can run any benchmark.

docs/src/performance_tips.md
```diff
@@ -9,6 +9,14 @@ This provides maximum freedom on creating an ABM.
...
 So, be sure that you benchmark your code, and you follow Julia's Performance Tips!
 
+## Take advantage of parallelization
+Julia provides several tools for [parallelization and distributed computing](https://docs.julialang.org/en/v1/manual/parallel-computing/).
+Notice that we cannot help you with parallelizing the _actual model evolution_ via the agent- and model-stepping functions. This is something you must do manually, as depending on the model, parallelization might not be possible at all due to e.g. the access and overwrite of the same memory location (writing on same agent in different threads or killing/creating agents).
+If your model evolution satisfies the [criteria allowing parallelism](https://docs.julialang.org/en/v1/manual/multi-threading/#Caveats), the simplest way to do it is using Julia's [`@threads` or `@spawn` macros](https://docs.julialang.org/en/v1/manual/multi-threading/#man-multithreading).
+
+The only native parallelization we can offer is parallelizing over the full model evolution and data collection loop. This is done by providing a `replicates > 0` and `parallel = true` argument to [`run!`](@ref).
```
fbanning, line 17: has to be adapted for the new `ensemblerun!`

Libbum:
A sanity check I needed to do to make sure branching was indeed a faster solution that unstable dispatch.

```julia

mutable struct GridAgentOne <: AbstractAgent
    id::Int
    pos::Dims{2}
    one::Float64
    two::Bool
end

mutable struct GridAgentTwo <: AbstractAgent
    id::Int
    pos::Dims{2}
    one::Float64
    two::Bool
end

mutable struct GridAgentThree <: AbstractAgent
    id::Int
    pos::Dims{2}
    one::Float64
    two::Bool
end

mutable struct GridAgentFour <: AbstractAgent
    id::Int
    pos::Dims{2}
    one::Float64
    two::Bool
end

mutable struct GridAgentFive <: AbstractAgent
    id::Int
    pos::Dims{2}
    one::Float64
    two::Bool
end

model = ABM(
    Union{GridAgentOne,GridAgentTwo,GridAgentThree,GridAgentFour,GridAgentFive},
    GridSpace((15, 15));
    warn = false, 
    rng = MersenneTwister(42),
    scheduler = random_activation,
)

for _ in 1:50
    a = GridAgentOne(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool))
    add_agent!(a, model)
    a = GridAgentTwo(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool))
    add_agent!(a, model)
    a = GridAgentThree(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool))
    add_agent!(a, model)
    a = GridAgentFour(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool))
    add_agent!(a, model)
    a = GridAgentFive(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool))
    add_agent!(a, model)
end

agent_step!(agent::GridAgentOne, model) = walk!(agent, rand, model)
function agent_step!(agent::GridAgentTwo, model)
    agent.one += rand(model.rng)
    agent.two = rand(model.rng, Bool)
end
function agent_step!(agent::GridAgentThree, model)
    if any(a-> a isa GridAgentTwo, nearby_agents(agent, model))
        agent.two = true
        walk!(agent, rand, model)
    end
end
function agent_step!(agent::GridAgentFour, model)
    agent.one += sum(a.one for a in nearby_agents(agent, model))
end
function agent_step!(agent::GridAgentFive, model)
    targets = filter!(a->a.one > 0.8, collect(nearby_agents(agent, model, 3)))
    idx = argmax(map(t->edistance(agent, t, model), targets))
    farthest = targets[idx]
    walk!(agent, sign.(farthest.pos .- agent.pos), model)
end

@btime step!($model, agent_step!, dummystep, 500)
900.044 ms (11500106 allocations: 700.68 MiB)

mutable struct GridAgentAll <: AbstractAgent
    id::Int
    pos::Dims{2}
    one::Float64
    two::Bool
    type::Symbol
end

model = ABM(
    GridAgentAll,
    GridSpace((15, 15));
    rng = MersenneTwister(42),
    scheduler = random_activation,
)

for _ in 1:50
    a = GridAgentAll(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool), :one)
    add_agent!(a, model)
    a = GridAgentAll(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool), :two)
    add_agent!(a, model)
    a = GridAgentAll(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool), :three)
    add_agent!(a, model)
    a = GridAgentAll(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool), :four)
    add_agent!(a, model)
    a = GridAgentAll(nextid(model), (1,1), rand(model.rng), rand(model.rng, Bool), :five)
    add_agent!(a, model)
end

    if agent.type == :one
        agent_step_one!(agent, model)
    elseif agent.type == :two
        agent_step_two!(agent, model)
    elseif agent.type == :three
        agent_step_three!(agent, model)
    elseif agent.type == :four
        agent_step_four!(agent, model)
    else
        agent_step_five!(agent, model)
    end
end
agent_step_one!(agent, model) = walk!(agent, rand, model)
function agent_step_two!(agent, model)
    agent.one += rand(model.rng)
    agent.two = rand(model.rng, Bool)
end
function agent_step_three!(agent, model)
    if any(a-> a.type == :two, nearby_agents(agent, model))
        agent.two = true
        walk!(agent, rand, model)
    end
end
function agent_step_four!(agent, model)
    agent.one += sum(a.one for a in nearby_agents(agent, model))
end
function agent_step_five!(agent, model)
    targets = filter!(a->a.one > 1.0, collect(nearby_agents(agent, model, 3)))
    if !isempty(targets)
        idx = argmax(map(t->edistance(agent, t, model), targets))
        farthest = targets[idx]
        walk!(agent, sign.(farthest.pos .- agent.pos), model)
    end
end

@btime step!($model, agent_step!, dummystep, 500)
166.813 ms (2292318 allocations: 149.54 MiB)
```

So that's 5.4x in this case.

docs/src/performance_tips.md
```diff
@@ -0,0 +1,82 @@
...
+Body::Any
+1 ─ %1 = Base.getproperty(model, :par1)::Any
+│   %2 = Base.getproperty(model, :par2)::Any
+│   %3 = (%1 * %2)::Any
+│        (a = %3)
+└──      return %3
+```
+which makes the model stepping function have type instability due to the model properties themselves being type unstable.
+
+The solution is to use a Dictionary for model properties only when all values are of the same type, or to use a custom `mutable struct` for model properties where each property is type annoted, e.g:
```
fbanning, line 54: typo: annotated

Datseris:
There isn't something much here besides confirming that the "fast" wolfsheep model is indeed "faster" than the original. Then, my plan is to eradicate all the versions of daisyworld and wolfsheep and keep only 1 version of daisy world, and the two versions of wolf sheep as examples of using 3 agents or only 1 for future reference.

Libbum:
I'm looking at the wolf sheep again now. See if I can figure out why it's failing.

Libbum:
@Datseris, that's got wolf-sheep running. I'll do a quick benchmark now and see how it fares.

examples/predator_prey_fast.jl
```diff
@@ -150,12 +148,10 @@ end
...
 
-function sheep_eat!(sheep, grass_array, model)
-    isempty(grass_array) && return
-    grass = grass_array[1]
-    if grass.fully_grown
+function sheep_eat!(sheep, model)
+    if model.fully_grown[sheep.pos...]
         sheep.energy += sheep.Δenergy
-        grass.fully_grown = false
+        model.fully_grown[sheep.pos...] = false
```
Datseris, line 154: I think splatting `...` is bad for performance and should be replaced with `pos[1], pos[2]`. Not sure though...

Libbum: Nah, there's no difference. It's fine to change if you want to put it in line with the rest though.

Libbum:
Benchmarking results in arxiv version of paper: 159.115 ms

This version: 46.799 ms

(Needs a bit of confirmation but I've got to go deal with this puppy, he's going crazy)

examples/predator_prey_fast.jl
Datseris: okay if there is no difference then I'd rather re-write other parts where I write explicitly `pos[1], pos[2]`.

docs/src/performance_tips.md
```diff
@@ -0,0 +1,82 @@
...
+struct MyAgent <: AbstractAgent
+	id::Int
+end
+properties = Dict(:par1 => 1, :par2 => 1.0, :par3 => "Test")
+model = ABM(MyAgent; properties = properties)
+model_step!(model) = begin
+	a = model.par1 * model.par2
+end
+```
+is a bad idea, because of:
```
Libbum, line 34: I'd re-frame this to 'is not performant, because of', since nearly all of our examples use dicts.

Libbum: Sorry. I didn't read the rest of this portion before commenting. Leave as-is.

docs/src/performance_tips.md
```diff
@@ -0,0 +1,82 @@
...
+## Avoid `Union`s of many different agent types (temporary!)
+Due to the way Julia's type system works, and the fact that agents are grouped in a dictionary mapping IDs to agent instances, using multiple types for different agents always creates a performance hit because it leads to type instability.
+
+Thankfully, due to some performance enhancements in Base Julia, unions of up to three different Agent types do not suffer much. You can see this by running the `test/performance/variable_agent_types_simple_dynamics.jl` file, which benchmarks the time to run a model that will do exactly the same amount of numeric operations, but each time subdividing it among an increasing number of agent types. Its output is
+
+```@example performance
+using Agents
+x = pathof(Agents)
+t = joinpath(dirname(dirname(x)), "test", "performance", "variable_agent_types_simple_dynamics.jl")
+include(t)
```
Libbum, line 75: This doesn't look finished. Nothing happens?

examples/predator_prey_fast.jl
```diff
@@ -0,0 +1,294 @@
...
+# controls how much energy is acquired after consuming a food source.
+
+# Grass is a replenishing resource that occupies every position in the grid space. Grass can be
+# consumed only if it is `fully_grown`. Once the grass has been consumed, it replenishes
+# after a delay specified by the property `regrowth_time`. The property `countdown` tracks
+# the delay between being consumed and the regrowth time.
+
+# It is also available from the `Models` module as [`Models.predator_prey`](@ref).
+
+# Notice that in this file we create the "simple" or "more intuitive" or "less lines of code"
```
Libbum, line 32: This text is not correct I don't think. It's for the 'simple' version"

Libbum:
@Datseris, I think this is close, but some of the bits and pieces you want to do to finish it off are not so clear cut from my point of view. So if you can do those changes yourself and let me know when you're happy, I'll give it a final look and then merge.

Libbum:
Oh, please also make sure that `Agents.Models` are updated to the newer versions so I can re-run the comparisons when we're ready.

Datseris:
Something very wrong is happening here: docs do not build locally for me any more: all names are not picked up by documenter... Anyone having same issue? Could it be that I just moved to Julia 1.6?

Datseris:
> Also we need Documenter#master

why?

Datseris:
I'm merging this once tests pass.

Libbum:
There are two missing referencs still, but otherwise looks fine to merge

Issue #436, opened 2021-02-23
seed! is ineffective for RandomDevice


Libbum:
I *thought* seed! was working with the current implementation, but upon further testing, it's not. Reading into the docs further: *Two such objects will always generate different streams of random numbers.*

So to truly `seed!` a random device, we need to pass a new instance to `model.rng`. We'll need to go back to the Ref implementation I think.

Datseris:
What is the purpose of "seeding" something without providing an actual reproducible seed?

Datseris:
> Two such objects will always generate different streams of random numbers

Okay, so what? It's like shuffling a well-shuffled deck.

Libbum:
Well, I'm happy to not call it `seed!` in this instance, but providing an ability to have random replicates (#415) is what I'm talking about here.

Datseris:
I see. I am personally not happy with putting Ref everywhere. It seems hacky and fishy. If this was a "good way to go", then mutable types wouldn't exist. Julia would tell you "just make everything have Ref fields, its better this way, and you only need to type an extra `[]`".

But I guess here a `Ref` is the only way to go forwards, or simply benchmarking whether making `ABM` mutable has any impact.

Libbum:
Yep. Will do some mutability checks I think.

Out of time on this problem now anyway. I'm going to switch back to MersenneTwister for the moment.

Datseris:
Just to be clear, at the moment parallel replicates do not work right? as they deep copy the model, and `rng` is part of the model...?

Issue #243, opened 2020-04-29
Parallelization tests & documentation


Libbum:
There are no real tests about parallelization nor distributed computing.

Whilst we provide the ability to run parallel parameter scans, it is only documented in a spotty manner, and no examples leverage it.

Ref #96

PR #476, opened 2021-04-05
Seeds and parallelization

Datseris:
This PR fixes the lack of seeding in parallel replicates. 

Todo:

- [x] RandomDevice behavior. @Libbum Julia docs say that two instances of `RandomDevice` will always generate a different stream of random numbers. Are you sure that `deepcopy(model)` generates identical behavior with `RandomDevice`?
- [x] Actually add tests that replicate simulations provide different outcome.
- [x] Resolve: initialization is not random.
- [x] Resolve: `paramscan` with `replicates`...
- [x] Rename `replicates` to `ensemble` (breaking)


Libbum:
Point 1 - pretty certain this gives us an issue, yes. I'll write an MWE in a sec to verify.

Datseris:
@Libbum I got a discussion point: the way `run!` works now with replicates, the _initialization_ of the model is identical across replicates. Is this the desired thing?

Libbum:
You know, it's probably not? At least not for me. I'd actually be happier if we didn't take a model as input, but rather an initialise function that obtained in instance for each replicate. That may mean altering the requirements of `run!`, but it removes the need of additional machinery we're trying to provide here, as well as the `RandomDevice` question.

@fbanning, can you comment on this? Is it `parallel_replicates` or `paramscan` you rely on (sorry, can't remember).

The only places we currently use it is in [schelling](https://github.com/JuliaDynamics/Agents.jl/blob/671285aa698657e2d7f022ce05ca968270597900/examples/schelling.jl#L297) (which literally would just run the same model 5 times and not change results), and the optim example. There, the way things are solved is to [force a new seed](https://github.com/JuliaDynamics/Agents.jl/blob/671285aa698657e2d7f022ce05ca968270597900/examples/siroptim.jl#L26-L28). So yeah, I think we could handle this differently.

Datseris:
> At least not for me. I'd actually be happier if we didn't take a model as input, but rather an initialise function that obtained an instance for each replicate. 

But still how would the function know to initialize 5 different models instead of the same model 5 times? It would still need to be a function with at least 1 input argument `seed`.

Libbum:
Not really. 

```julia
function init()
    ...
    model = ABM(AgentX, GridSpace((5,5))
    ...
    return model
end
```
defaults to using the internal `model.rng`, which is a new seed (or new `RandomDevice`) each call of `init`.

Datseris:
> defaults to using the internal model.rng

But the "default" seed is 

```julia
function AgentBasedModel(
    ::Type{A},
    space::S = nothing;
    scheduler::F = fastest,
    properties::P = nothing,
    rng::R = Random.default_rng(),
    warn = true,
```

the global RNG?

Libbum:
Ah, yeah I see where you're going. Hmm. I'll think of some alternatives. Got to get off the computer now.

src/simulations/collect.jl
```diff
@@ -454,7 +454,7 @@ end
 
 "Run replicates of the same simulation in parallel"
 function parallel_replicates(model::ABM, agent_step!, model_step!, n, replicates;
-    seeds = [rand(Int) for _ in 1:replicates], kwargs...)
+    seeds = [abs(rand(Int)) for _ in 1:replicates], kwargs...)
```
fbanning, line 457: Could be written as `seeds = rand(UInt32, replicates)`, I think

fbanning:
> the _initialization_ of the model is identical across replicates

Yes, this is also something that I have noticed last fall when I added parallelisation to our model (both in `run!` and  `paramscan`), so in the end we are relying on `parallel_replicates` for our whole model. I first noticed because agents' positions (from their initialisation) were always the same across all replicates. I assumed this behaviour was intended as the use of deepcopies explicitly transfers these initial model properties to all the replicates.

> I'd actually be happier if we didn't take a model as input, but rather an initialise function that obtained an instance for each replicate.

I would also rather have fully independent model instances for each replicate and drop the use of deepcopies altogether. Instead all model instances should be created within the `parallel_replicates` function call.

Datseris:
I'll push a different design soon, but I think I'd rather create a new function `parallel_run!` whose first input argument is not an `ABM` but a `Function` that generates `ABM`.

Libbum:
That's better choice, yes.

src/simulations/multirun.jl
```diff
@@ -0,0 +1,69 @@
+Vector_or_Tuple = Union{AbstractArray, Tuple}
+
+# version that takes as input generator
+function multirun!(
+        generator, args...;
+        replicates = 5, seeds = [abs(rand(Int)) for _ in 1:replicates], kwargs...
```
Libbum, line 6: Needs to be a UInt at least. Can't be negative.

Libbum:
This looks nicer. Give me a day or so to play around with it.

Datseris:
A consequence of this PR is that `paramscan` with replicates is now invalid. I think it is better this way: the goal of paramscan is not to run the same model the same time, but to run it with different parameters.

Datseris:
@Libbum a thing I don't like here is the name "replicates". Isn't the name "ensemble" more fitting to what we are doing here? Problem is, changing the dataframe key from `:replicate` to something else is a breaking change... :(

Libbum:
> `paramscan` with replicates is now invalid

I think this makes sense for 98% of cases. It's better to have these two capabilities separated&mdash;this wasn't obvious previously.

`ensemble` is better yeah. Hmm. Depends on how fast-and-loose we want to adhere to strict versioning requirements. I think we've previously done some breaking work on minor revisions....

src/simulations/multirun.jl
```diff
@@ -0,0 +1,87 @@
...
+Call [`run!`](@ref) for all `model ∈ models`. Each `model` should be a (different) instance
+of an [`AgentBasedModel`](@ref) but probably initialized with a different random seed or
+different initial agent distribution. All models obey the same rules `agent_step!, model_step!`.
+
+Similarly to [`run!`](@ref) this function will collect data. It will furthermore
+add one additional column to the dataframe called `:replicate`, which will have
+integer values counting the replicate simulations.
+The function returns `agent_df, model_df, models`.
+
+The keyword `parallel=false` will run the simulations in parallel using Julia's
```
Libbum, line 15: = true

Datseris: It should be `false` by default, because noone would define everything prefaced with `@everywhere` by default.

Libbum: Agreed, but then you need to change the wording of the text. Right now it says that false runs the simulations in parallel.

Libbum:
Can confirm this works with `RandomDevice`

```julia
using Agents, Random

mutable struct Agent3 <: AbstractAgent
    id::Int
    pos::Tuple{Int,Int}
    weight::Float64
end

function init(seed)
    model = ABM(Agent3, GridSpace((10, 10)); rng = RandomDevice())
    add_agent!(model, rand(model.rng))
    add_agent!(model, rand(model.rng))
    add_agent!(model, rand(model.rng))
    return model
end

    agent.weight = rand(model.rng)
end

sum_all(model) = sum(a.weight for a in allagents(model))
mdata = [sum_all]
_, data = multirun!(init, agent_step!, dummystep, 10; mdata)
 Row │ step   sum_all   replicate 
     │ Int64  Float64   Int64     
─────┼────────────────────────────
   1 │     0  1.73603           1
   2 │     1  2.26063           1
   3 │     2  2.02521           1
   4 │     3  2.42302           1
   5 │     4  1.55008           1
   6 │     5  1.23275           1
   7 │     6  1.34099           1
   8 │     7  1.88797           1
   9 │     8  1.78147           1
  10 │     9  1.50423           1
  11 │    10  1.49149           1
  12 │     0  0.970136          2
  13 │     1  2.54795           2
  14 │     2  1.11096           2
  15 │     3  1.38222           2
  16 │     4  1.66397           2
  17 │     5  1.11606           2
  18 │     6  1.85462           2
  19 │     7  2.18949           2
  20 │     8  1.72076           2
  21 │     9  1.49147           2
  22 │    10  1.19565           2
  23 │     0  1.60773           3
  24 │     1  1.73295           3
  25 │     2  1.06095           3
  26 │     3  1.1006            3
  27 │     4  1.83976           3
  28 │     5  1.08743           3
  29 │     6  0.652535          3
  30 │     7  1.84696           3
  31 │     8  1.56826           3
  32 │     9  1.59459           3
  33 │    10  1.47409           3
  34 │     0  0.744897          4
  35 │     1  1.64755           4
  36 │     2  1.86353           4
  37 │     3  1.92173           4
  38 │     4  0.651095          4
  39 │     5  1.67071           4
  40 │     6  1.60658           4
  41 │     7  0.755377          4
  42 │     8  1.42983           4
  43 │     9  2.14067           4
  44 │    10  0.481782          4
  45 │     0  1.33844           5
  46 │     1  1.88296           5
  47 │     2  2.25485           5
  48 │     3  1.13188           5
  49 │     4  1.80496           5
  50 │     5  0.556998          5
  51 │     6  1.26536           5
  52 │     7  1.48862           5
  53 │     8  1.615             5
  54 │     9  1.52321           5
  55 │    10  0.712008          5
```

Libbum:
Confirm `deepcopy` does not:

```julia
julia> model = init(654)
AgentBasedModel with 3 agents of type Agent3
 space: GridSpace with size (10, 10), metric=chebyshev, periodic=true
 scheduler: fastest

julia> model2 = deepcopy(model)
AgentBasedModel with 3 agents of type Agent3
 space: GridSpace with size (10, 10), metric=chebyshev, periodic=true
 scheduler: fastest

julia> step!(model, agent_step!, dummystep, 10)

julia> model.agents
Dict{Int64, Agent3} with 3 entries:
  2 => Agent3(2, (3, 6), 0.617409)
  3 => Agent3(3, (3, 10), 0.197784)
  1 => Agent3(1, (7, 1), 0.0882612)

julia> step!(model2, agent_step!, dummystep, 10)

julia> model.agents
Dict{Int64, Agent3} with 3 entries:
  2 => Agent3(2, (3, 6), 0.617409)
  3 => Agent3(3, (3, 10), 0.197784)
  1 => Agent3(1, (7, 1), 0.0882612)
```

src/simulations/multirun.jl
Datseris: okay I'm stupid.

src/simulations/multirun.jl
```diff
@@ -0,0 +1,87 @@
...
+    replicate_col!(df_model, 1)
+
+    for rep in 2:length(models)
+        df_agentTemp, df_modelTemp = _run!(models[rep], agent_step!, model_step!, n; kwargs...)
+        replicate_col!(df_agentTemp, rep)
+        replicate_col!(df_modelTemp, rep)
+        append!(df_agent, df_agentTemp)
+        append!(df_model, df_modelTemp)
+    end
+    return df_agent, df_model, models
```
Libbum, line 65: The return of the models list is currently undocumented.

Datseris:
Turns out that there is no need for both `:ensemble` and `:replicates` to exist as columns. The deprecated `run!` with `replicates` simply renames the columns.

I am now renaming `multirun!` to `ensemblerun!` and I am done. @Libbum  please give a final review and merge as you see fit.

src/simulations/multirun.jl
Datseris: Why, the docstring says "The function returns `agent_df, model_df, models`."

Libbum:
Great. This looks really nice. There are a few doc fixes etc that I'll get to tomorrow morning & will merge then. 

I think there's a number of open issues this addresses that can close once it's merged. #436 #415 #243

src/simulations/multirun.jl
Libbum: Ah, yes. I didn't see that. My bad.

Datseris:
huh that's odd: https://juliadynamics.github.io/Agents.jl/previews/PR476/examples/schelling/#Replicates-and-parallel-computing  even though we have changed that.

Libbum:
Then you need to refresh. I can't see it. `count(x -> x == true, moods)` I thought you changed though. Checking..

Libbum (approved):
Will merge once I verify this small doc change worked.

PR #479, opened 2021-04-07
schedule: Rename existing schedulers to all have same prefix

### LINKED ISSUE #363: Renaming existing schedulers to all have same start
`by_type` -> `schedule_by_type`, `random_activation` -> `schedule_randomly` etc.

rht:
This fixes #363.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,9 @@
 - New function `ensemblerun!` for running ensemble model simulations.
 
 ## Deprecated
-- Keyword `replicates` of `run!` is deprecated in favor of `ensemblerun!` and will be invalid in a future version.
-- `paramscan` with `replicates` is deprecated and will be invalid in a future version.
-- All the scheduler names have been deprecated in favor of the same schedulers but prefixed with `schedule_`.
-  `fastest` to `schedule_fastest`, `by_id` to `schedule_by_id`, `random_activation` to `schedule_randomly`, `partial_activation` to `schedule_partially`, `property_activation` to `schedule_by_property`, `by_type` to `schedule_by_type`.
+- Keyword `replicates` of `run!` is deprecated in favor of `ensemblerun!`.
+- `paramscan` with `replicates` is deprecated and this functionality of having _both_ replicates and parameter scanning in one function will be invalid in a future version.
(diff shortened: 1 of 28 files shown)


Datseris:
Great, thanks!

You need to properly _deprecate_ all existing names however. E.g. write down: `@deprecate by_type schedule_by_type`. This needs to be done for every name.

Furthermore the usage of `by_type` throughout the library `src` and `examples` folders needs to be replaced by its non-deprecated version.

Also, please add entries in the CHANGELOG file :)

rht:
Update:
- Entire codebase grepped and has all the occurrences of the old scheduler names renamed.
- CHANGELOG updated.
- Tests passed except for Documentation still pending, which I didn't touch.

Let me know if you agree with the new names.

Datseris:
You also need to replace the scheduler names here: https://github.com/JuliaDynamics/Agents.jl/pull/479/checks?check_run_id=2286278812#step:7:80

rht:
Ah gotcha, I seemed to only pay attention to .jl files.

rht:
I see. I seemed to do the `grep` not from the root repo dir. I was doing it inside `src/core`.

rht:
OK updated. Waiting for tests.

docs/src/api.md
```diff
@@ -227,12 +227,12 @@ Notice that schedulers can be given directly to model creation, and thus become
...
-by_id
-random_activation
-partial_activation
-property_activation
-by_type
+schedule_fastest
+schedule_by_id
+schedule_randomly
+schedule_partial
+schedule_property
```
Datseris, line 234: Could you rename these two to `schedule_partially` and `schedule_by_property`? I think these names are more sensible...

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ## Deprecated
 - Keyword `replicates` of `run!` is deprecated in favor of `ensemblerun!` and will be invalid in a future version.
 - `paramscan` with `replicates` is deprecated and will be invalid in a future version.
+- All the scheduler names have been deprecated in favor of the same schedulers but prefixed with `schedule_`.
```
Datseris, line 12: Here I think we should explicitly state all name changes.

docs/src/api.md
rht: Updated.

CHANGELOG.md
rht: Added.

Datseris:
Seems great thanks a lot. I will build the docs locally once tonight to make sure everything is well and then merge.

PR #483, opened 2021-04-09
Move some functionality into submodules

Datseris:
* Schedulers become submodule
* Pathfinding becomes submodule
* Functions `move_along_route!` and `is_stationary` become part of standard movement API, as they are already present for at least two occasions. Notice that we have to actually write the OSM space to use these two names as well.

Still to do:
- [x] OSM space
- [x] Tests pass
- [ ] Docs pass


Datseris:
@fbanning can you please review this as well? its a really difficult PR to be sure that nothing was missed.

CHANGELOG.md
```diff
@@ -1,16 +1,17 @@
...
-- `paramscan` with `replicates` is deprecated and this functionality of having _both_ replicates and parameter scanning in one function will be invalid in a future version.
-- All the scheduler names have been deprecated in favor of the same schedulers but prefixed with `schedule_` as follows: `fastest` to `schedule_fastest`, `by_id` to `schedule_by_id`, `random_activation` to `schedule_randomly`, `partial_activation` to `schedule_partially`, `property_activation` to `schedule_by_property`, `by_type` to `schedule_by_type`.
+- `paramscan` with `replicates` is deprecated. If you want to parameter scan and at the same time run multiple simulations at each parameter combination, simply use `seed` as a parameter, which tunes the model's initial random seed.
+- All the scheduler names have been deprecated in favor of the same schedulers but prefixed with `schedule_` as follows: `fastest` to `Schedulers.fastest`, `by_id` to `Schedulers.by_id`, `random_activation` to `Schedulers.randomly`, `partial_activation` to `Schedulers.partially`, `property_activation` to `Schedulers.by_property`, `by_type` to `Schedulers.by_type`.
```
Libbum, line 14: > prefixed with `schedule_` 

Needs to be updated to module language

src/submodules/schedulers.jl
```diff
@@ -1,89 +1,101 @@
...
+if you prefer to write your own `step!`-like loop.
+
+See also [Advanced scheduling](@ref) for making more advanced schedulers.
+
+Notice that schedulers can be given directly to model creation, and thus become the
+"default" scheduler a model uses, but they can just as easily be incorporated in a
+`model_step!` function as shown in [Advanced stepping](@ref).
+"""
+module Schedulers
+using Agents
```
Libbum, line 27: Is this the right way to do this? (more of a TODO myself to look into module design)

Libbum:
Will get to this on Sunday, won't be available tomorrow.

Datseris:
This is an extremely difficult PR. I'm genuinely exhausted... I've fixed all tests besides one related to pathfinding, which I don't see why it is related to my changes:
```
constructors: Test Failed at C:\Users\datse\.julia\dev\Agents\test\grid_pathfinder_tests.jl:73    
  Expression: length(model.space.pathfinder.agent_paths) == 0
   Evaluated: 1 == 0
```
I haven't even checked the docs yet.

src/spaces/openstreetmap.jl
```diff
@@ -1,32 +1,41 @@
...
+export random_road_position,
+    plan_route,
+    map_coordinates,
+    road_length,
+    random_route!,
+    latlon,
+    intersection,
+    road
+
+struct OpenStreetMapSpace <: Agents.DiscreteSpace # TODO: Why is this a discrete space?
```
Libbum, line 22: Because it's graph space under the hood.

It's continuous since we build machinery around the OSMX.jl struct to make it so, but it still needs discrete methods for the most part.

src/deprecated.jl
```diff
@@ -0,0 +1,18 @@
+@deprecate fastest Schedulers.fastest
```
Libbum, line 1: There are some methods in `src/Agents.jl` that could move here.

src/submodules/schedulers.jl
Libbum: Answer to this is that parent scope is not accessible in children. We can `import .specific_function` if we want, but I don't think it's all that different.

Datseris:
image

I guess we do not need the middle method. Should we consider also moving `Adding agents, kiling, etc.` to higher level heading starting with `##` ?

Libbum:
Sounds good. 

I just got to the point where all warnings are now gone. Haven't gone over any of the changes at all.

Datseris:
another to-do to fix this spacing:

image

src/simulations/ensemblerun.jl
```diff
@@ -19,6 +19,10 @@ docs online).
 All other keywords are propagated to [`run!`](@ref) as-is.
 
 Example usage in [Schelling's segregation model](@ref).
+
+If you want to scan parameters and at the same time run multiple simulations at each
```
Libbum, line 23: Was this added in the parallel paramscan PR already?

Datseris: This was always possible de-fact actually, as `seed` is nothing more as just another model parameter that can be "scanned" over.

Datseris: (of course the user has to actually add this seed to the model by `ABM(...; rng = MersenneTwister(seed))` but I hope they are intelligent enough to figure this out :D )

src/deprecated.jl
Libbum: This file isn't actually linked anywhere yet I don't think...

Issue #481, opened 2021-04-08
Parallel paramscan


Adriankhl:
If I understand it correctly, it is not possible to parallelize paramscan, especially after (https://github.com/JuliaDynamics/Agents.jl/pull/476).

May I ask why this is the case? It seems like it is strict forward to implement with something like parallel map, and it is really useful to speedup big run.

Libbum:
You can certainly distribute paramscan with a `pmap` if that's something you need. #476 removed the replication capability (run multiple versions of the same parameter set), since that never really made sense.

This is a new change though, and we have yet to release this change (its only on master). If you have a use case, please provide an example of what you're interested in doing and we can either guide you to a solution with the new syntax; or modify these recent changes to accommodate.

Datseris:
> If I understand it correctly, it is not possible to parallelize paramscan, especially after (#476).

Let's be a bit precise here because what was happening was not what you currently have in mind.

The old `paramscan` was doing something "weird": when `replicates > 0` it was running a `deepcopy` of the model, for the exact same parameter combinations, many times in parallel. With Agents.jl 4.1, where the random number generation became an internal property of the model, this became somewhat nonsensical due to the deep copying. Unfrotuantely we did not realize this when releasing 4.1 so `paramscan` in 4.1 is just a "bad function".

Furthermore, I _personally_ believe it does not make sense to do at the same time a parameter scanning _and_ an ensemble simulation and it is better to keep these two processes separated. I might be wrong.

Regardless, with the new design currently on master, **it is in fact indeed possible to truly parallelize `paramscan` over different parameters**. This is because `paramscan` generates a _new model_ for each parameter combination.  This is clear from these lines: https://github.com/JuliaDynamics/Agents.jl/blob/master/src/simulations/paramscan.jl#L85-L86

@Adriankhl are you perhaps interested to look in the source code here https://github.com/JuliaDynamics/Agents.jl/blob/master/src/simulations/ensemblerun.jl and make a PR that offers the same parallization for `paramscan` based on `pmap`? I believe it is an easy contribution.

Adriankhl:
> Furthermore, I _personally_ believe that parameter scanning and ensemble simulations should be separate processes. I might be wrong.

I have done such things in my project (multiple parameter scanning and multiple runs for same set of parameters) to plot relations between variables and the error bars/bands. If the random seed can also be a parameter in the scanning then it should be good.


[...]

PR #484, opened 2021-04-10
Parallelize paramscan 

Adriankhl:
Address https://github.com/JuliaDynamics/Agents.jl/issues/481.

`paramscan` takes `parallel::Bool` to decide whether `map` or `pmap` is used

I have also added `ProgressMeter.jl` (https://github.com/timholy/ProgressMeter.jl) as dependency to show the progress bar of `paramscan` since the original simple `print` doesn't look good on my jupyter notebook, and I am not sure if `print` is thread-safe, In https://discourse.julialang.org/t/print-functions-in-a-threaded-loop/12112 someone suggested to use `Core.println` but I couldn't find it in the current documentation, and `ProgressMeter.jl` takes care of `pmap`.

This takes 1:41:29 on my laptop
```
happyperc(moods) = count(moods) / length(moods)
adata = [(:mood, happyperc)]

parameters = Dict(
    :min_to_be_happy => collect(2:5), # expanded
    :numagents => collect(200:800),         # expanded
    :griddims => (30, 30),            # not Vector = not expanded
)

adf, _ = paramscan(parameters, initialize; adata, agent_step!, n = 2000)
```

This takes 0:32:58 with `addprocs(4)`, so it seems like parallelize paramscan is working.
```
@everywhere happyperc(moods) = count(moods) / length(moods)
@everywhere adata = [(:mood, happyperc)]

@everywhere parameters = Dict(
    :min_to_be_happy => collect(2:5), # expanded
    :numagents => collect(200:800),         # expanded
    :griddims => (30, 30),            # not Vector = not expanded
)

adf, _ = paramscan(parameters, initialize; adata, parallel = true, agent_step!, n = 2000)
```

Some other points

* It is tedious to write `@everywhere`, maybe `@everywhere begin end` and `@everywhere include` should be introduced in the example/tutorial.
* Should this be added somewhere in the test/benchmark


Libbum:
Great work @Adriankhl! We're not fans of adding dependencies of we don't need to, but will look into your suggestion here.

The docs you've added are a welcomed addition too. For tests: that's a different beast. We haven't got the CI set up with threads. If that's an easy thing to do, then sure, it's be great to have a test for this.

src/simulations/paramscan.jl
```diff
@@ -25,13 +25,19 @@ see the example below.
 The following keywords modify the `paramscan` function:
 
 - `include_constants::Bool=false` determines whether constant parameters should be
-  included in the output `DataFrame`.
+  included in the output `DataFrame`. If false, the varying parameters will still be
+  included in the output.
```
Datseris, line 29: I don't understand this. If `false` it is not included...?

src/simulations/paramscan.jl
```diff
@@ -62,59 +68,85 @@ parameters = Dict(
...
+@everywhere adata ...
+@everywhere parameters
+```
+then do the parameter scan by
+```
+adf, _ = paramscan(parameters, initialize; parallel = true, adata, agent_step!, n = 3)
+```
+
+To avoid `@everywhere` in everywhere, you can write the core part of the model
+in `Schelling.jl` and do `@everywhere include("Schelling.jl"`
```
Datseris, line 98: I would prefer avoiding duplicate information. This already exists in the `[Performance Tips](@ref)` section and thus should not make the docstring here bigger. Please cross-ref to this section in the docstring.

Datseris: Notice that the last two lines you've added are genuinely true information, and you should simply port them to the Performance Tips doc page.

Datseris:
THank you so miuch for this contribution! Seriously!!!

ProgressMeter's only dependencies are `Distributed` (ie base julia).

My suggestion would be to remove the `progress` keyword completely and always show progress ...?

src/simulations/paramscan.jl
Adriankhl: Yeah it is a bit confusing, if false then only the varying parameters will be included, if true than the constant parameters will be included as well. I have updated this.

src/simulations/paramscan.jl
Adriankhl: Done. 

Maybe a full working example about parallel run should be included in the documentation?

Adriankhl:
> THank you so miuch for this contribution! Seriously!!!

It is fun to work with julia, I am glad that I discovered an ABM library here :)

> My suggestion would be to remove the `progress` keyword completely and always show progress ...?

It is up to you guys, I have removed it in the latest commit. I also feel like it doesn't harm to always have a progress bar.

@Libbum At least the previous progress bar is not displayed properly in my jupyter notebook. As this is a single file dependency, I think it is possible to directly copy the necessary part of the source to this project. The problem is that a proper progress bar for multiple thread should contain lock and atomic operation, which may make the code a little bit more complicated.

Libbum:
I think you've made s good argument for its inclusion, and it looks like it's not so large of a package, so my point is mostly irrelevant now.

Thanks for the help on this!

Datseris:
@rht the tests are failing because they use the keyword `progress` still. Can you update them?

Libbum:
I've fixed the tests. Will check the docs locally and if all looks well, will merge soon.

PR #488, opened 2021-04-13
Automata forest

Libbum:
A more performant forest fire which is more Cellular Automata than Agent, as the current description has it.
v4.0 runtime: 24.3 ms, this version: 1.5 ms

### DIFF
--- a/examples/forest_fire.jl
+++ b/examples/forest_fire.jl
@@ -16,116 +16,113 @@
 # 1. A tree will burn if at least one neighbor is burning
 
 # The forest has an innate `density`, which is the proportion of trees initialized as
-# `green`, however all trees that reside at `x=1` on the grid are `burning`.
+# `green`, however all trees that reside on the left side of the grid are `burning`.
 # The model is also available from the `Models` module as [`Models.forest_fire`](@ref).
 
 # ## Defining the core structures
 
-# We start by defining the agent type
+# Cellular automata don't necessarily require an agent-like structure. Here we will
+# demonstrate how a model focused solution is possible.
 using Agents, Random
-using InteractiveDynamics
 using CairoMakie
 
-mutable struct Tree <: AbstractAgent
+mutable struct Automata <: AbstractAgent
     id::Int
-    pos::Dims{2}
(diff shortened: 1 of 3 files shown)


Datseris:
lol when we update the comparison table people will _lose it_

Libbum:
@Datseris, can I animate the heat array using abm_video?

examples/forest_fire.jl
```diff
@@ -16,116 +16,113 @@
...
+
+function tree_step!(forest)
+    nx, ny = size(forest.trees)
+    ## Find trees that are burning (coded as 2)
+    for I in findall(isequal(2), forest.trees)
+        ## Look up all euclidean neighbors
+        neighbors = Iterators.filter(
+            x -> 1 <= x[1] <= nx && 1 <= x[2] <= ny,
+            (I.I .+ n for n in forest.offsets),
+        )
```
Datseris, line 68: Hm, I'm a bit afraid with this one. Is it that much slower to use `nearby_positions`? Because someone could complain "Well, you're not really using Agents.jl anymore but writing the implementation from scratch, including even finding neighboring positions".

Datseris:
> @Datseris, can I animate the heat array using abm_video?

Yes, it will be automatically animated. That's how daisyworld works, you just give `heatarray = :tree` and will do everything (but also give your nice colormap)

examples/forest_fire.jl
Libbum: To do that we'd also add a space that does nothing. I can check.

Datseris: aaaaaaaaaaaaaaaaaaaa aaaaaaa sorry yeah you are right...

I wonder if we can somehow re-use the `Hood` stuff. Hm... but maybe not worth the development ATM.

Datseris: A no, what am I even talking about. We can add the space and just use `neaby_positions` to get the indices with which to access the `forest.tree` array. We don't have to actually develop anything as long as the space and the `trees` have the same size.

Libbum:
Including a space and using `nearby_positions` bumps us to a snails pace of 1.9 ms :scream:

Issue #396, opened 2021-01-29
Consider StructArrays for internal handling of agent dict


Libbum:
```julia
model = ABM(Agent3, GridSpace((100, 100)))
for _ in 1:500
    add_agent!(model, rand())
end

aos = collect(values(model.agents)) # 1.770 μs (1 allocation: 4.06 KiB)
soa = StructArray(values(model.agents)) # 3.966 μs (12 allocations: 16.22 KiB) 
# ^^ initialisation is slower here, but maybe not with a better collection method

julia> @btime sum(a->a.weight, $aos)
  428.955 ns (0 allocations: 0 bytes)
252.1354667521684

julia> @btime sum($soa.weight)
  57.521 ns (0 allocations: 0 bytes)
252.13546675216827
```

Libbum:
It remains to be seen how quick this would make things in general. Looking through most of the codebase, we do tend to use `model[id] for id in ids` quite a lot, meaning iteration over an array is certainly something to consider. We don't, mostly, look up `model[7]`... So the purpose of the O(1) lookup time in `Dict`s may not be as needed. The question is how this change affects the nearest_neighbor and `Hood` mechanisms. I'm not a fan of having to do some form of `findfirst(isequal(id), soa.id)`, O(n) lookup in those functions...

Libbum:
Building this with an associated `Dict{Int,Int}` (id, location in array) has been straightforward, except for one major caveat: no mixed model support. That's about half of the machinery we have here, so in general it may not be viable. Will have to think on it.

Libbum:
Actually, now that this is obviously an issue it essentially means a large portion of user-made agents wont be able to be stored in the model structure. So this is a non-starter. Performance gains are interesting, so I'll consider writing up an Integration example if an idea presents itself.

fbanning:
Ah, I didn't see this issue thread earlier and might have missed it because you opened and closed it on the same day. :D 
Generally I find the StructArray approach very promising and would like to delve a bit deeper into this if I can find the time at some point. The mixed model support is indeed a problem but there has to be a way around this. A lot more thinking is needed here but it seems too promising to dismiss entirely...

Libbum:
Yep, it could be a boon for certain models just not all. If there's a way to identify which models can work with this strategy and a good way to separate those into a SA model type that doesn't clash with the rest of the API, it'd be a really nice addition. My quick analysis was that would be a hard ask at this point&mdash;not impossible though perhaps!

[...]

PR #492, opened 2021-04-22
Implement StructArrays as core for agent data

fbanning:
# StructArrays in Agents.jl?

## The mission

Attempting to solve #396 via a complete reimplementation of agent data in the `ABM` struct by replacing the `model.agents` Dict with a StructArray. Current API should remain "as is" as far as possible.

## The caveats

Tests mostly work fine but not all of them, of course, because so much of the underlying structure of Agents.jl has to be changed to accomodate a move away from the battle-tested dictionary approach.

Mixed models don't work right now (not the focus of this first iteration). Those tests have been commented out because the thrown errors were annoying me.

## The benchmarks

Arbitrary benchmark code:

```julia
using BenchmarkTools, Agents, Random

mutable struct SAAgent <: AbstractAgent
    id::Int
    money::Float64
end

function init(n)
    Random.seed!(42)
    p = Dict(
        :abs_growth => 0.0,
    )
    m = ABM(SAAgent, properties = p)
    for i in 1:n
        add_agent!(SAAgent(i, rand(m.rng)), m)
    end
    return m
end

function m_s!(m)
    m.abs_growth = 0 
    for a in allagents(m)
        gains = a.money * rand(m.rng, [-0.005, -0.001, 0.01])
        m.abs_growth += gains
        a.money += gains
    end
end

@benchmark begin
    m = init(2_500)
    adata = [:money]
    mdata = [:abs_growth]
    run!(m, dummystep, m_s!, 2_500; adata=adata, mdata=mdata);
end
```

Dict:
```julia
BenchmarkTools.Trial:
  memory estimate:  3.58 GiB
  allocs estimate:  43936674
  --------------
  minimum time:     3.347 s (14.65% GC)
  median time:      3.535 s (12.17% GC)
  mean time:        3.535 s (12.17% GC)
  maximum time:     3.724 s (9.95% GC)
  --------------
  samples:          2
  evals/sample:     1
```

StructArrays:
```julia
BenchmarkTools.Trial:
  memory estimate:  7.12 GiB
  allocs estimate:  198792134
  --------------
  minimum time:     13.096 s (6.24% GC)
  median time:      13.096 s (6.24% GC)
  mean time:        13.096 s (6.24% GC)
  maximum time:     13.096 s (6.24% GC)
  --------------
  samples:          1
  evals/sample:     1
```

## The conclusion

Either my implementation is borked (actually highly likely because this is a major overhaul) or it's just not more performant to use StructArrays like I tried to do here.

Eager to hear your thoughts. :)

### DIFF
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -20,6 +20,7 @@ struct AgentBasedModel{S<:SpaceType,A<:AbstractAgent,F,P,R<:AbstractRNG}
     properties::P
     rng::R
     maxid::Base.RefValue{Int64}
+    idmap::Dict{Int,Int}
 end
 
 const ABM = AgentBasedModel
@@ -89,7 +90,7 @@ function AgentBasedModel(
     agent_validator(A, space, warn)
 
     agents = StructArray{A}(undef, 0)
-    return ABM{S,A,F,P,R}(agents, space, scheduler, properties, rng, Ref(0))
+    return ABM{S,A,F,P,R}(agents, space, scheduler, properties, rng, Ref(0), Dict())
 end
 
 function AgentBasedModel(agent::AbstractAgent, args...; kwargs...)
@@ -108,9 +109,10 @@ export random_agent, nagents, allagents, allids, nextid, seed!
 Return an agent given its ID.
 """
 function Base.getindex(model::ABM, id::Integer) 
-    ind = findfirst(x -> x == id, model.agents.id)
-    ind === nothing && throw(ArgumentError("Given ID could not be found in the model!"))
-    return LazyRow(model.agents, ind)
+    index = model.idmap[id] 
+    index === nothing && 
+        throw(ArgumentError("Given ID could not be found in the model!"))
+    return LazyRow(model.agents, index)
 end
 
 """
@@ -125,16 +127,33 @@ function Base.setindex!(model::ABM, agent::AbstractAgent, id::Int)
     agent.id ≠ id &&
     throw(ArgumentError("You are adding an agent to an ID not equal with the agent's ID!"))
     push!(model.agents, agent)
-    model.maxid[] < id && update_maxid(model)
+    model.maxid[] < id && update_maxid!(model)
     return agent
 end
 
 """
-    update_maxid(model::ABM)
+    update_idmap!(model::ABM)
+
+Map `id`s of currently active agents in the `model` to their indices in the `model.agents` 
+array.
+"""
+function update_idmap!(model::ABM)
+    for id in model.agents.id
+        index = findfirst(x -> x == id, model.agents.id)
+        if index === nothing
+            delete!(model.idmap, id)
+        else
+            model.idmap[id] = index
+        end
+    end
+end
+
+"""
+    update_maxid!(model::ABM)
 
 Update `model.maxid` to store the highest `id` of currently active agents in the `model`.
 """
-function update_maxid(model::ABM)
+function update_maxid!(model::ABM)
     model.maxid[] = findmax(model.agents.id)[1]
 end
 
@@ -169,6 +188,8 @@ function Base.getproperty(m::ABM{S,A,F,P,R}, s::Symbol) where {S,A,F,P,R}
         return getfield(m, :rng)
     elseif s === :maxid
         return getfield(m, :maxid)
+    elseif s === :idmap
+        return getfield(m, :idmap)
     elseif P <: Dict
         return getindex(getfield(m, :properties), s)
     else # properties is assumed to be a struct
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -14,7 +14,7 @@ ValidPos =
     Union{Int,NTuple{N,Int},NTuple{M,<:AbstractFloat},Tuple{Int,Int,Float64}} where {N,M}
 
 struct AgentBasedModel{S<:SpaceType,A<:AbstractAgent,F,P,R<:AbstractRNG}
-    agents::Dict{Int,A}
+    agents::StructArray{A}
     space::S
     scheduler::F
     properties::P
@@ -88,7 +88,7 @@ function AgentBasedModel(
 ) where {A<:AbstractAgent,S<:SpaceType,F,P,R<:AbstractRNG}
     agent_validator(A, space, warn)
 
-    agents = Dict{Int,A}()
+    agents = StructArray{A}(undef, 0)
     return ABM{S,A,F,P,R}(agents, space, scheduler, properties, rng, Ref(0))
 end
 
@@ -107,7 +107,11 @@ export random_agent, nagents, allagents, allids, nextid, seed!
 
 Return an agent given its ID.
 """
-Base.getindex(m::ABM, id::Integer) = m.agents[id]
+function Base.getindex(model::ABM, id::Integer) 
+    ind = findfirst(x -> x == id, model.agents.id)
+    ind === nothing && throw(ArgumentError("Given ID could not be found in the model!"))
+    return LazyRow(model.agents, ind)
+end
 
 """
     model[id] = agent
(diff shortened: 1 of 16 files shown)


src/core/model.jl
```diff
@@ -107,7 +107,11 @@ export random_agent, nagents, allagents, allids, nextid, seed!
 
 Return an agent given its ID.
 """
-Base.getindex(m::ABM, id::Integer) = m.agents[id]
+function Base.getindex(model::ABM, id::Integer) 
+    ind = findfirst(x -> x == id, model.agents.id)
```
Libbum, line 111: This is super slow, which is most likely causing your performance issues.

I just pushed up my old work on this [here](https://github.com/JuliaDynamics/Agents.jl/commit/9750323f0c2e030b888e0e31671c7157bc23dd04), which uses a `Dict{Int,Int}` index property. Take a look and see if it's of any assistance.

fbanning: Yeees, I know. :( But I couldn't think of a better way to find the index of the given id. I tried some other methods but findfirst was already the fastest implementation I could think of. The O(1) lookup of dicts is basically unbeatable. Sticking with a `Dict{Int,Int}` would require us to recreate the dict from scratch every time an agent gets added/deleted. Will try it out though.

Libbum: > require us to recreate the dict from scratch every time an agent gets added/deleted. Will try it out though.

Is that true? 

add_agent: would use a specific id, thus `m.index[id] = id`,
kill_agent: `delete_at!(m.index[id])`

I may have missed something there though.

fbanning: In a very simple model or right at initialisation, the statement `m.index[id] = id` would work. But we we want to map `id => index` in the dict and we know that `id == index` is not always true so we need to take care of that, I suppose? That's why I think recreating the dict is necessary after every addition/removal of an agent.

Thinking about it a bit more, once you delete an agent from the StructArray, all the rows after it will be altered as well (go up one row, i.e. have their index in the array reduced by 1). Let's assume that the rows of the StructArray are always sorted by `id` (which would be the default order without interference by users). Since we only ever delete one agent at a time, we could possibly do `m.index[id] -= 1` for the agents with higher ids. But adding is not so easy anymore afterwards because then `id == index` isn't necessarily the case.

Will need to think about this a bit more and try a few things out. (This screams "write some tests" at me...)

Libbum: Ah, yes, that's right&mdash;I remember having that though and thinking it pretty untenable. Although I'm not sure if such an operation is actually that bad knowing the alternative. We swap O(1) add/deletes for O(n), but searches stay O(1) and we get the speed up of StructArrays as the tradeoff.

How that fares overall for all our cases will remain to be seen.

fbanning:
Improvements have been made with this naive implementation of a `Dict{Int,Int}(id => index)`:

```julia
BenchmarkTools.Trial: 
  memory estimate:  7.16 GiB
  allocs estimate:  201602564
  --------------
  minimum time:     10.755 s (8.20% GC)
  median time:      10.755 s (8.20% GC)
  mean time:        10.755 s (8.20% GC)
  maximum time:     10.755 s (8.20% GC)
  --------------
  samples:          1
  evals/sample:     1
```

fbanning:
Extending on the benchmark, this one also aggregates some data across all agents and should be a better way to judge the potential benefits of directly accessing columns of the StructArray :

```julia
using BenchmarkTools, Agents, Random
using StatsBase: mean

mutable struct SAAgent <: AbstractAgent
    id::Int
    money::Float64
end

function init(n)
    Random.seed!(42)
    p = Dict(
        :abs_growth => 0.0,
        :mean_wealth => 0.0,
    )
    m = ABM(SAAgent, properties = p)
    for i in 1:n
        add_agent!(SAAgent(i, rand(m.rng)), m)
    end
    return m
end

function m_s!(m)
    m.abs_growth = 0 
    for a in allagents(m)
        gains = a.money * rand(m.rng, [-0.005, -0.001, 0.01])
        m.abs_growth += gains
        a.money += gains
    end
    m.mean_wealth = mean(a.money for a in allagents(m)) # for Dictionary
    m.mean_wealth = mean(m.agents.money) # for StructArray
end

@benchmark begin
    m = init(2_500)
    adata = [:money]
    mdata = [:abs_growth, :mean_wealth]
    run!(m, dummystep, m_s!, 2_500; adata=adata, mdata=mdata);
end
```

Dict:

```julia
BenchmarkTools.Trial: 
  memory estimate:  3.59 GiB
  allocs estimate:  43956686
  --------------
  minimum time:     3.820 s (10.55% GC)
  median time:      3.894 s (13.87% GC)
  mean time:        3.894 s (13.87% GC)
  maximum time:     3.967 s (17.07% GC)
  --------------
  samples:          2
  evals/sample:     1
```

StructArrays:

```julia
BenchmarkTools.Trial: 
  memory estimate:  7.17 GiB
  allocs estimate:  201630070
  --------------
  minimum time:     10.408 s (8.25% GC)
  median time:      10.408 s (8.25% GC)
  mean time:        10.408 s (8.25% GC)
  maximum time:     10.408 s (8.25% GC)
  --------------
  samples:          1
  evals/sample:     1
```

As you can see, aggregating data across all agents costs some time (about 10% increase) in the regular Agents.jl approach.  
The StructArray approach does not seem to have any measurable cost to it at all.  
So while it's overall ~3x as slow in this example, the expected speedup of aggregation over columns is real.

AayushSabharwal:
This piqued my interest. Currently, I'm a bit tight on time as my semester is coming to a close, so I haven't been able to explore the code properly. However, if I understand correctly, add/deleting an agent is an O(n) operation, since the subsequent elements have to be shifted back an index, and the corresponding entries in the `Dict` updated? If so, I have a couple ideas, but I'm not sure of their feasibility. Just throwing them out to see if it's valid, 

- To delete an agent, it could just be swapped with the last one and popped off the end? That way shifting of all higher indices is avoided, and only one (the former last index) needs to be updated.
- Deletes could possibly be pooled and avoid even swapping if the indices are simply invalidated. All invalid indices are stored in a `SortedSet`. Any access to these indices throws an error/does whatever is appropriate. When an agent is added, it simply takes the smallest invalid index. This would maintain the sorted order of ids in the array. However, this approach will leave a lot of invalid indices if the agent population falls drastically.

It's possible I've misunderstood something. If so, please correct me.

Libbum:
Decent ideas @AayushSabharwal. We're looking at any way it's possible to leverage this without causing major slow-downs in other regions of the codebase. It's all 'is this possible' at the moment, so any further ideas are welcomed.

fbanning:
> To delete an agent, it could just be swapped with the last one and popped off the end

Definitely feasible, I think.

Keep in mind that the current benchmark only adds 100 agents and doesn't remove any agents at all. So at least from my point of view, the process of deleting agents is of a lower priority right now.

>     All invalid indices are stored in a `SortedSet`

Possible, yes. However, it would leave a big memory footprint because we would have to keep the rows in the StructArray so that the indexing functions as before.  
If we delete an agent, this will leave an unused row in our agent data until a new agent is added there. Furthermore, constant checks are needed to always check whether a given id is valid in the current context of the model. It's basically introducing another lookup which can be avoided by an up-to-date dict in the first place.  
Or imagine an empty model where the user adds an agent with `id = 1` and another one with `id = 101` (for whatever reasons). Having static indices with a set of invalid indices would then require us to add 99 rows to the StructArray which are just placeholders to be filled later, whenever the user decides to add an agent with exactly this id. But the model still needs to store all these "placeholder agents".  
I would rather not go down that route with respect to memory efficiency.

> It's all 'is this possible' at the moment

Currently I'm inclined to say "yes, but it isn't really more performant".

Compared to our regular "id => struct" implementation, performance is bad enough as it currently is. Just lazily accessing the rows (i.e. accessing a single agent's field values via `model[id].property`) takes so much computational time, that it just doesn't seem sensible to opt for `StructArrays` instead of regular structs.  
The benefit of aggregation isn't used often enough in standard ABMs (i.e. altering microdata of heterogenous agents) that it would really allow us to leverage the power of a `StructArray`. The downside seems just too big for me at the moment. If we can somehow avoid that downside of `LazyRow` access then we might actually have a trade-off that is worthy of consideration. Right now (with my naive implementation here), this doesn't seem feasible.

> so any further ideas are welcome

Most definitely!

Datseris:
I guess we can close this? Thanks a lot Frederik for the effort, but the conclusion is that it doesn't really offer that much benefit?

fbanning:
> I guess we can close this? Thanks a lot Frederik for the effort, but the conclusion is that it doesn't really offer that much benefit?

I agree because nobody has commented on this PR since April and nobody seems to have found any better ways of implementing StructArrays. It was a nice idea but sadly didn't fulfill its promise. I'll just close this draft PR.

If at some point somebody wants to dive back into this topic, I'd be very happy to chime back in. For now we got other things to focus on. :)

Issue #494, opened 2021-05-21
`init_model_dataframe` errors when using a properties struct


fbanning:
When creating an ABM with a properties struct, we are currently facing a problem during the calll to [`init_model_dataframe`](https://github.com/JuliaDynamics/Agents.jl/blob/c9a7fa13493c0ba24aacd0a6a6de4ba19394d181/src/simulations/collect.jl#L397). We want to collect some model data (through the regular `mdata = [:fieldname]` approach) and get the following error (which I've generalised a bit to make it more easily understandable):

```julia
julia> model.properties[:scenario]
ERROR: MethodError: no method matching getindex(::ModelProperties, ::Symbol)
Stacktrace:
 [1] top-level scope
   @ REPL[7]:1
```

I'm relatively sure that [this line](https://github.com/JuliaDynamics/Agents.jl/blob/c9a7fa13493c0ba24aacd0a6a6de4ba19394d181/src/simulations/collect.jl#L407) is what causes the problem because we can't access a struct's fields like this (i.e. via the fieldnames as symbols). This approach only works for Dicts and Tuples.

Is my assumption correct that this is what is causing the problem?  
If so, maybe it's already enough to add a quick check for the type of `model.properties` and adapt accordingly.

Libbum:
Without looking too deeply, I'd say your assumption is correct. We wrote that method on the assumption that we're using a dict and haven't done thorough testing with the recent improvements when using structs. 

Yep, a type check would be best!

PR #497, opened 2021-05-25
Model property type checking in `init_model_dataframe`

LiamBlake:
Very minor fix to #494 by adding type checking of `model.property`. New to Julia so let me know if I missed anything obvious.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["Tim DuBois", "George Datseris", "Ali Vahdati"]
-version = "4.2.4"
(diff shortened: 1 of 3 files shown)


Datseris:
Don't we have a function that makes this automatic...?

I guess we should extend `get_data`

https://github.com/JuliaDynamics/Agents.jl/blob/master/src/simulations/collect.jl#L8-L9

to also dispatch on whether the first argument is dictionary or not...? 

I just think that this approach is quite sloppy in the source, and a couple of smart definitions will allow us to reduce a lot of complexity and duplication in the source code.

Datseris (commented):
Thanks, can you do the following:

1. Add appropriate tests in the `test` folder
2. Increment patch version in Project.toml

and then we can merge?

PR #499, opened 2021-05-26
Read agent data from CSV

### LINKED ISSUE #403: Read data from file directly into agents/model
We have decent data collecting utilities, but it'd be good to extend capacity to allow users to read in data from files to populate agents.

Unsure if this needs to be new methods or just an Ecosystem Integration example with CSV.jl and / or DataFrames.jl

AayushSabharwal:
closes #403


Libbum:
@AayushSabharwal, anything you're doing for GSoC, add the label to it so we can showcase your work later.

AayushSabharwal:
I had to overload `Base.parse(::Type{NTuple{N,T}}, s::String)` to get it to parse `pos` fields. I'm not sure if this is ideal or recommended. I'll try looking for another method if I can find one, otherwise this currently works. I would've written tests for this too, but I need csv file(s) to use for testing, and I am not sure where I can put those.

src/submodules/io/csv_reader.jl
```diff
@@ -12,12 +13,27 @@ to fields of the struct.
...
+    kwargs...
+) where {S, A}
+    for row in CSV.Rows(filename; kwargs...)
+        add_agent_pos!(agent_type(row...), model)
+    end
+end
+
+function Base.parse(::Type{NTuple{N,T}}, s::String) where {N,T}
+    return Tuple(map(x -> parse(T, x), split(s[2:(end-1)], ", ")))
+end
```
Datseris: TYPE PIRACY DETECTED!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Datseris: IM CALLING THE POLICE

Datseris: 🚓  👮🏻

AayushSabharwal: I didn't know what that was. It seems to be a bad idea. I'll try look for another way around this. Without it, CSV.jl complains about this method not existing

Datseris: Ooooh, then you should open an issue there to support parsing as Tuples? But I'm not sure whether this is intended for CSV. Data there in each column do not typically come as tuples. Why is it the case here? CSV format is quite barebones, don't think you can support "tuples as dcolumns".

AayushSabharwal: CSV.jl doesn't allow specifying a different function to parse a specific column. The only solutions I could think of are:
- Create a wrapper for `NTuple`, and define `Base.parse` for that type
- Keeping the Tuple fields as `String`, then once CSV delivers a row convert the required field(s) manually
- Only keeping the `populate_from_csv!` method that takes a `Function`
- Require that `agent_type` have a constructor taking keyword arguments, and have the user manually handle non-primitive types that will be handed over as `String`s

Datseris: You should open a request issue on CSV to allow custom parsing function that defaults to `Base.parse`, that seems straight-frward

AayushSabharwal: > Ooooh, then you should open an issue there to support parsing as Tuples? But I'm not sure whether this is intended for CSV. Data there in each column do not typically come as tuples. Why is it the case here? CSV format is quite barebones, don't think you can support "tuples as dcolumns".

For some reason none of your messages after the first seemed to load in until I reloaded the page, sorry for ignoring this.

Agents would require one column to be a tuple for their position. Usually non-scalar fields would span multiple columns, but that would result in multiple fields instead of one tuple.

Datseris:
@AayushSabharwal I do not think your mockup is a good idea;
```
id,pos,f
5,"(5, 10)",0.15
```
. You should make positions independent columns, `pos1, pos2, ...` etc. It makes for clunky CSV files.

Datseris:
Let's start considering this in a more realistic scenario. What input CSV files should we expect? Think of a hospital list, or a list of citizes with the locations being the houses of the people. How do you imagine such a CSV looking? I imagine a longitude and latitude columns. 

And how would you read such a file? You can try to create one yourself or find a real one for a test case.

AayushSabharwal:
> Let's start considering this in a more realistic scenario. What input CSV files should we expect? Think of a hospital list, or a list of citizes with the locations being the houses of the people. How do you imagine such a CSV looking? I imagine a longitude and latitude columns.

I see your point. I looked too much into the parsing instead of what sane data formats we would receive. What if the function merges all fields `field_1`, `field_2`... into a `Tuple`, and passed all fields as keyword instead of positional arguments?

Should I close the issue in CSV.jl then?

> The ID is an optional column btw, because we can take the row number as id.

That's a good point! I'll keep it in mind

Datseris:
> Should I close the issue in CSV.jl then?

Making CSV allow a custom parsing function is a good idea; However I'm not sure the proposal you wrote in that issue is the best way to go. Let's see what they reply first.

Datseris:
Let's discuss the interface in the next call, if you've setup a meeting time already

AayushSabharwal:
We haven't discussed a fixed time yet. I'll send a mail, so we can try and set up something

AayushSabharwal:
According to the issue on CSV.jl, the recommended method is to create a wrapper type and define `Base.parse` for that

AayushSabharwal:
Example using the current setup:

```julia
julia> function Models.HKAgent(; id = -1, old = -1, new = -1, pre = -1)
           return Models.HKAgent(id, old, new, pre)
       end
julia> model, s1, s2 = Models.hk()
julia> genocide!(model)
julia> populate_from_csv!("test.csv", model, Models.HKAgent, Dict(:id => 1, :old => 2, :new => 2); header = false, types = [Int, Float64])
```
"test.csv" contains:
```
1,3.4
2,4.5
3,3.6
4,6.7
```
And the result:
```julia
julia> model.agents
Dict{Int64, Agents.Models.HKAgent} with 4 entries:
  4 => HKAgent(4, 6.7, 6.7, -1.0)
  2 => HKAgent(2, 4.5, 4.5, -1.0)
  1 => HKAgent(1, 3.4, 3.4, -1.0)
  3 => HKAgent(3, 3.6, 3.6, -1.0)
```

Libbum:
Good start!

Some questions since I've not dug into the CSV internals:
- `types = [Int, Float64]` if we leave these empty, can they be inferred / cast? Will our system fail if these are not exact? I can see a situation where we want `Float64` but the data on disk is `15, 16, 17` for example. It would be fine to make this `15.0, 16.0, 17.0`.
- `header = false` if `header = true`, do we use that information? Say we had 3 columns with headers `id, old, new`, can we map those automatically with some default for `col_map`?

AayushSabharwal:
> I don't quite follow why Models.HKAgent needs to get wrapped like that for this example, can you elaborate?

`populate_from_csv!` splats keyword-value pairs into the `agent_type` constructor, so it requires the corresponding keyword arguments to be available. As a result, a custom constructor would be required for this approach. It would be possible to use positional arguments, but I think that reduces flexibility. For example, in this example not all the fields need to be specified. `pre` just uses a default value, and  `old` and `new` are mapped to the same column number. Having to use positional arguments would make it a lot more complex.

> if we leave these empty, can they be inferred / cast? 

`CSV.File` automatically detects field types, but `CSV.Rows` doesn't so they need to be explicitly provided as of now. If missing, all fields are `String`. If the `Symbol`s in `col_map` correspond to fields in `agent_type`, it might be possible to figure out the type. I'll explore this and see where I get. If `agent_type` is a function, then I don't think they can be inferred

> Will our system fail if these are not exact?

It's up to the user how to handle this. For example, the `HKAgent` constructor I made doesn't  specify types, so it could just take `Int` or `Float32` or any compatible type, and the inner constructor for the type would automatically promote it.

> if `header = true`, do we use that information? Say we had 3 columns with headers `id, old, new`, can we map those automatically with some default for `col_map`?

Currently `header=false` prevents it ignoring the first row. Assuming the header corresponds exactly to field names, it should be possible to infer `col_map` and `types` assuming they aren't provided.

AayushSabharwal:
Idea:

If the user leaves `col_map` empty, we assume the constructor doesn't take keyword arguments and simply splat the row
If `col_map` is a `Vector{Int}`, splat the columns in that order
If `col_map` is a `Dict{Symbol, Int}`, the current implementation

It might also be possible to combine the last two, so both can be used simultaneously.

Datseris:
suggestion: `id` should be an optional key for the mapping. We could just use the row number as the agent id.

src/submodules/io/csv_reader.jl
```diff
@@ -1,39 +1,22 @@
...
-    agent_type::B;
+    agent_type::B,
+    col_map::Dict{Symbol,Int};
     kwargs...
-) where {A, B<:Type{<:A},S}
-    !haskey(kwargs, :types) && (kwargs = (kwargs..., types = collect(fieldtypes(agent_type))))
+) where {A,B<:Type{<:A},S}
     for row in CSV.Rows(filename; kwargs...)
-        add_agent_pos!(agent_type(row...), model)
+        add_agent_pos!(agent_type(; (k => row[v] for (k, v) in  col_map)...), model)
```
Datseris, line 20: What do you do if the user has not created a keyword constructor for `agent_type`?

AayushSabharwal: Currently it's required. That's what I'm trying to address with the idea I mentioned

Datseris: Sorry, didn't read when I posted this! Yes on your suggestion, not sure if second idea is worth the effort.

> If col_map is a Vector{Int}, splat the columns in the order specified by the indexes in the vector. Could also support Vector{String} or Vector{Symbol}

perhaps skip it for now and only attempt to implement if requested. Time=money ;)

src/submodules/io/csv_reader.jl
```diff
@@ -0,0 +1,32 @@
+using CSV
+
+"""
+    populate_from_csv!(filename, model, agent_type, col_map; kwargs...)
```
Datseris, line 4: In Julia the typical convention to follow is that the argument that is mutated is first in the function.

AayushSabharwal:
Is this functionally complete now? Should I write tests?

Datseris:
yes please, it is critical that we test this with a real data frame ASAP. I'd suggest to use one of the existing models and save its agent list as a CSV and use this to load.

AayushSabharwal:
Is there any specific place to store the csv file, or can I just include it in the tests directory?

Datseris:
We wont store one. We generate it during the tests. Would be nice to have a function `dump_to_csv(agents)`.

src/submodules/io/csv_reader.jl
```diff
@@ -30,3 +30,12 @@ function populate_from_csv!(
         end
     end
 end
+
+function dump_to_csv(filename, agents::Vector{A}, fields = collect(fieldnames(A)); kwargs...) where {A<:AbstractAgent}
```
Datseris, line 34: the restriction to `Vector` is not good. The most common scenario would be to call this function on `agents(model)` which returns a value iterator.

I have a suggestion/advice: change your approach to writing functions. Start development on an interactive environment in VS Code where you run every single line and see what it does. Makes you understand and design code much much better.

Datseris: After you have the algorithm doing what you want, put it in a function body.

AayushSabharwal: I'll keep that in mind now on

AayushSabharwal:
`HKAgent.previous_opinion` seems to be misspelled as `HKAgent.previous_opinon`. I'll fix it in the example and model

AayushSabharwal:
Currently, `dump_to_csv` will dump non-scalar fields such as `pos` and `vel` as `"(x, y)"` strings. I'll work on splitting them into multiple columns. Should this always happen or be optional, and controlled through a keyword argument?

Datseris:
it should happen always.

Datseris:
@AayushSabharwal notice also that `IO` is a conflicting name with `Base.IO`. We need another name, maybe `AgentsIO` or something else.

AayushSabharwal:
I saw that earlier. Currently the module is called `ModelIO` but there's probably a better alternative. `Agents.AgentsIO` feels a little repetitive.

Datseris:
But `AgentsIO` is / will be an exported name. A user doesn't have to type `Agents.AgentsIO`. (But sure I'm totally up for a better name)

AayushSabharwal:
Ah, I forgot about that. Sounds great!

AayushSabharwal:
I'm unable to pinpoint why the tests refuse to run on Windows. It only happens when the following two lines are run in sequence:

```julia
Agents.ModelIO.populate_from_csv!(empty_model, "test.csv", Models.HKAgent, Dict(:id => 1, :op1 => 3, :op2 => 2))
Agents.ModelIO.dump_to_csv("test.csv", allagents(model), [:id, :old_opinion])
```

In fact, after the first line even `open("test.csv", "w")` throws the same error. (This is the code inside CSV.jl that throws the error)

I'm guessing this means splatting keyword arguments in `populate_from_csv!` is somehow causing this. I've tried deep-copying `row`, to no avail. Do you have an idea what else could cause this? I'll keep looking around for now.

Libbum:
Odd. I don't have a windows machine to test on, so can't say straight away. Will do some digging too.

AayushSabharwal:
I came across this issue @Datseris made in CSV: https://github.com/JuliaData/CSV.jl/issues/170
`use_mmap` is deprecated and didn't work, but opening the file manually as `f = open(filename, "r")` worked, and tests passed (albeit with a deprecation warning)

AayushSabharwal:
As suggested by the deprecation warning:
```
┌ Warning: `CSV.File` or `CSV.Rows` with `IOStream` object is deprecated; pass a filename, `IOBuffer`, or byte buffer directly (via `read(x)`)
│   caller = Header at header.jl:109 [inlined]
└ @ Core D:\ProgramFiles\symlinks\.julia\packages\CSV\Zl2ww\src\header.jl:109
```
I passed `read(filename)` and tests seem to pass locally

Datseris (commented):
Now that tests pass we need to add a new section at the API about documentation.

AayushSabharwal:
There are just a couple things left to be implemented first in `populate_from_csv!`:

- (suggestion) `agent_type` takes a default value using the model type. This is checked to not be a `Union` at the beginning of the function (and an appropriate error raised otherwise)
- Using row number as `id`, toggled by a keyword argument

AayushSabharwal:
Considering there's so much going on under the hood with these two functions, I've tried to be as explicit as possible, and provided examples too. I've also tried to maintain the same style as the rest of the docs. To preview, I've added a section at the end of the API docs named AgentsIO, although that can probably be improved.

src/submodules/io/csv_integration.jl
```diff
@@ -63,6 +100,33 @@ function populate_from_csv!(
     end
 end
 
+"""
+    function dump_to_csv(filename, agents, fields; kwargs...)
```
Datseris: Since these functions are not accessible in direct scope, their docstring should be prefaced with `ModelIO.`

AayushSabharwal:
A request: Once this is finalised can I be the one to merge it? That way it stays on my commit record (and I get to press a button I've never pressed before)

AayushSabharwal:
I'm not sure why the section I made isn't there in the built docs. There's no warning either, and the two functions aren't in the list of docstrings that weren't included.

AayushSabharwal:
Looks like it just took a while

Datseris (approved):
> A request: Once this is finalised can I be the one to merge it? That way it stays on my commit record (and I get to press a button I've never pressed before)

Of course you can merge, however I must say I'm confused by this commend. Where else would your commits go if I was the one that merged? maybe I have misunderstood [how github counts commits](https://docs.github.com/en/github/setting-up-and-managing-your-github-profile/managing-contribution-graphs-on-your-profile/why-are-my-contributions-not-showing-up-on-my-profile) can you explain, so that this doesn't jappen with other contributors either?

---

The module `AgentsIO` needs to have a documentation string and also expanded in the code block, just like we did with e.g. Pathfinding

AayushSabharwal:
> Where else would your commits go if I was the one that merged?

Going by the pathfinding PR the squash-and-merge counted as a single commit under your name since all commits were merged into one and you pressed the button. I got credit for the additions/deletions that I did.

AayushSabharwal:
If the docstring is satisfactory, I'll squash-and-merge once docs build.

Libbum:
> I'm not sure why the section I made isn't there in the built docs

Behind the scenes github runs some jekyll pipeline to update the static sites on github pages. Either they run them on really low priority or they don't to cache updates well. Either way, what used to be instantaneous (i.e. pages showing up on merge), now could be two-three hours late sometimes. This is what happens when Microsoft engineering gets involved.

> squash-and-merge

We found that especially for these long pathfinder PRs, having this as the default makes the most sense. You're right about the 'author' issue though as a consequence. 

Let's make this arrangement: if you get approval from one of George or myself, the other has 5 hours to object. Within that time they can also approve or comment, and you can commit each of your PRs.

Also, once it's merged, please delete the branch here if you don't need it for something temporary, otherwise it gets difficult to navigate in the long run (local branches you can clean up or keep around).

Libbum:
Oh, and when you add new stuff, don't forget to add it to the CHANGELOG.md! (I've done it for this one on master already).

Datseris:
Besides the changelog this need also to:

1. Increment minor version in project toml
2. Add a refernce to the new stuff in the special box we have at the very beginning in the documentation `index.md`
3. Tag a new release.

sorry forgot to mention those.

AayushSabharwal:
I'll address these now

Issue #493, opened 2021-04-28
Use Scratch.jl for version updates


Libbum:
> From a discussion on Slack that will be forever lost, it is better to use https://github.com/JuliaPackaging/Scratch.jl to display update messages instead of the current pipeline with `touch`. Stefan Karpinski said that
> 
> > We really don’t want people modifying package code in place cry
> 
> and also cited other related problems, while Elliot Saba further commended that the approach will be impossible in the future:
> 
> > We’re eventually going to make packages read-only.
> 
> Elliot thankfully pointed out that the replacement is easy:
> 
> > https://github.com/JuliaPackaging/Scratch.jl , Basically, instead of using the DIR macro, you would use the get_scratch macro, that's the only change. But you get a lot of benefits, including that it will work even if your package is included in PackageCompiler output. The scratch directory can be cleared, if for example the package is uninstalled then reinstalled later. The user can also manually clear it.

&mdash;JuliaDynamics/DrWatson.jl#246

fbanning:
I don't think that I understand the benefits of using Scratch.jl, could you maybe ELI5 it for me?

Libbum:
```shell
/repos/Agents.jl/src % ls -l
total 32K
-rw-r--r-- 1 tim users 2.5K Jun  6 19:52 Agents.jl
drwxr-xr-x 2 tim users 4.0K Apr 30 14:19 core/
-rw-r--r-- 1 tim users 1.1K Apr 30 14:19 deprecated.jl
drwxr-xr-x 2 tim users 4.0K Jun  5 18:28 models/
drwxr-xr-x 2 tim users 4.0K Jun  5 18:23 simulations/
drwxr-xr-x 2 tim users 4.0K Apr 30 14:19 spaces/
drwxr-xr-x 4 tim users 4.0K Jun  5 18:28 submodules/
-rw-r--r-- 1 tim users    0 Jan 25 21:07 update_v4.0
-rw-r--r-- 1 tim users    0 Feb 22 21:43 update_v4.1
-rw-r--r-- 1 tim users    0 Mar 17 09:46 update_v4.2
drwxr-xr-x 2 tim users 4.0K Apr 30 14:19 visualization/
```

See all those update files? We track their creation and show an update message in the repl when we need to showcase changes and features. The files just sit around and pollute the folder. 

[...]

PR #504, opened 2021-06-06
Incremented minor version number, added to news

AayushSabharwal:


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["Tim DuBois", "George Datseris", "Ali Vahdati"]
-version = "4.2.5"
+version = "4.2.6"
```
Datseris, line 4: Please see https://semver.org/ . You have to increment the MINOR version

docs/src/index.md
```diff
@@ -15,6 +15,7 @@ To get started, please read the [Tutorial](@ref) page.
       * New function `ensemblerun!` which replaces using `replicates` in `run!`
       * New documentation page "Performance Tips"
       * Parallelization of `paramscan`
+      * New functions `dump_to_csv` and `populate_from_csv!` to save and load agent data to and from CSV files
```
Datseris, line 18: The old news have to be removed though ;) The point of this block is to bring the users attention to something really new. It is not to keep a changelog. We have a changelog that lists all incremental changes :)

Project.toml
AayushSabharwal: Didn't know that, thanks. I'll keep it in mind

AayushSabharwal: This would become 4.3.0, right?

AayushSabharwal:
I updated the printed message in `src/Agents.jl` too

Datseris:
I wouldn't do any message printing yet. We should have a message printing once the "full" AgentsIO functionality is present, which includes the checkpoints. BTW we need to resolve #493 before we can do another update message.

AayushSabharwal:
I'll revert the message printing then, and just bump the `version_number` in that file.

Once this is merged, should I create a new release or wait for #493 to be resolved?

Datseris:
Nah we just tag a new release, #493 only needs to be resolved when we want an update message.

Datseris (approved):

src/Agents.jl
```diff
@@ -58,7 +58,7 @@ export Models
 
 # Update message:
 display_update = true
-version_number = "4.2"
+version_number = "4.3"
```
Datseris, line 61: If this is changed, the update message will trigger. If we don't want to trigger an update message, then either set `display_update = false` or leave the version unchanged.

Libbum (approved):

Datseris:
do not forget to delete the branches after merge ⬆️

PR #505, opened 2021-06-06
JLD2 Integration

### LINKED ISSUE #496: Model (De)Serialization
Points to consider:
- What should the target API look like? I think it should just be two functions `save` and `load` (names are of course up for debate)
- What format should it be saved in? JLD2, or using `Serialization` from stdlib (See the example at the bottom)? It might be worth testing if one is significantly faster than the other, or results in smaller files. I also saw `BSON.jl`, so maybe that's a possibility?
- This ties in to the previous point: Should the serialized file be able to be viewed without having to load it in as a model? Or is the sole purpose of this feature to load and save, and have the file be completely opaque? I'm in favour of the first option, although it would require some additional effort
- In some cases (GraphSpace and OSMSpace) multiple save files may be required. How should this be handled?
- Should the pathfinding extensions be implemented first, in case there's additional data to be saved or a change in the data layout?

I've gone through all the structs that would need to be saved, and mapped it out:

Properties to save:
- `ABM.agents` # Possibly as a dataframe-esque format, since the dictionary can be reconstructed from that and it may make viewing easier
- `ABM.space`
- `ABM.properties` # The structure and contents of this are unknown, so ideally it should load back to the exact same state it was
- `ABM.rng` # Each RNG struct would have it's own internal state, so it can resume from the same point in the sequence
- `ABM.maxid`
- `GridSpace.metric`
- `GridSpace.hoods`
- `GridSpace.hoods_tuple`
- `GridSpace.pathfinder`
- `ContinuousSpace.grid`
- `ContinuousSpace.dims`
- `ContinuousSpace.spacing`
- `ContinuousSpace.extent`
- `GraphSpace.graph` # `JuliaGraphs/GraphIO.jl` exists, but it saves to a separate file. I don't think there's a point in reinventing the wheel, so there's the possibility of having multiple save files to consider
- `OSMSpace.m`    # A quick glance at `pszufe/OpenStreetMapX.jl` didn't show any functions to save to a .osm file
- `AStar.agent_paths` # This is a Dict mapping to Linked List. Not difficult to implement, but worth exploring if the linked list state can be preserved or if it will have to be regenerated from an array
- `AStar.grid_dims`
- `AStar.neighborhood` # just save Pathfinder.diagonal_movement
- `AStar.admissibility`
- `AStar.walkable`
- `AStar.cost_metric` # HeightMap is recursive

Properties taken from user at deserialization:
- `ContinuousSpace.update_vel!`
- `ABM.scheduler`
What if we save a symbol, and assume the function exists in the environment when it is loaded?

Properties reconstructed from saved data:
- `GridSpace.s`
- `GraphSpace.s`
- `OSMSpace.s`
Agents store their positions, so these structures can be regenerated at deserialization

A serialization method that @Libbum found. It uses Serialization from the standard library, so there isn't any additional dependency. However, it's completely opaque. At a glance it seems to work. I'm currently testing if it saves the more intricate data listed above
```julia
using Agents
using Serialization # stdlib

(model, agent_step!, model_step!) = Models.predator_prey(   
               n_wolves = 40,   
               n_sheep = 60,   
               dims = (25, 25),   
               Δenergy_sheep = 5,   
               Δenergy_wolf = 13,   
               sheep_reproduce = 0.2,   
               wolf_reproduce = 0.1,   
               regrowth_time = 20,   
           )

serialize("model.bin", model)
# Exit julia, restart
using Agents
model = deserialize("model.bin")
```

AayushSabharwal:
#496 

Uses JLD2's custom serialization functionality (https://juliaio.github.io/JLD2.jl/dev/customserialization/)


AayushSabharwal:
Looking at https://github.com/JuliaGraphs/MetaGraphs.jl/blob/master/src/persistence.jl and from some testing I did, it appears as though `GraphSpace.graph` can be saved through JLD2.

Additionally, reimplementing all schedulers as function-like objects would make serialization significantly simpler. Currently, since `ABM` isn't mutable, `model.scheduler = ...` isn't possible. JLD2 when deserializing `SerializedABM` to `ABM` would (as of now) set the scheduler to `fastest`. The only other way would be manually converting `ABM` to `SerializedABM` (through a constructor) and give that to JLD2 to save, so that the reverse conversion function can take the scheduler as a parameter. This issue also then crops up with `ContinuousSpace.update_vel!`

Libbum:
I wonder if using accessors.jl would be good for this. I've had success with it setting model.rng to a new RandomDevice(). Will be a good discussion for tomorrow.

AayushSabharwal:
Is it an acceptable pattern to encapsulate repetitive tests in a function, or does that reduce traceability in the case that those tests fail? For example,

```julia
function check_model_properties(model, other)
        @test model.scheduler == other.scheduler
        @test model.rng == other.rng
        @test model.maxid.x == other.maxid.x
    end
```

AayushSabharwal:
There's some weird behaviour with comparing `GridSpace.hoods`, which I plan on looking into. Comparing the structs doesn't work, but comparing their fields passes

Datseris:
> I'll cover what I've written in tests before moving on to tackling OSMSpace. 

I propose to do this in a different PR. OSMSpace is a bit "special" so to speak

Datseris:
> Is it an acceptable pattern to encapsulate repetitive tests in a function, or does that reduce traceability in the case that those tests fail?

From my end this is fine, I do something similar in DynamicalSystems.jl

AayushSabharwal:
All that's left to test now is serialization of multi-agent models. 325 tests racked up already just for JLD2 (was ~3k at one point, while I was testing `GridSpace.s` by having a `@test` macro in a loop over `eachindex`)

AayushSabharwal:
I moved daisyworld's scheduler outside of the initialization function, so it's accessible while deserializing during tests. Is this fine, or should it remain local?

Libbum:
Sure, that's not a requirement to be internal as far as I remember.

AayushSabharwal:
I'm not fully happy with how `dump_to_jld2`'s docstring is phrased. I'll work on improving it. Should it also specify what fields are not saved (`space.s`), or how they're saved (`model.agents`)?

Datseris:
I wonder whether we should use names more along the lines of "checkpoint". Isn't this how it's called in MASON?

AayushSabharwal:
I don't know about MASON, but sure. I chose the current names primarily because of how CSV integration was named. The names you suggested are more concise, and the user doesn't really need to be reminded of the save format every time they use the API.

AayushSabharwal:
Is there anything else missing apart from adding to changelog and bumping the minor version? Should there be an example for this? Maybe modify one of the existing ones?

Libbum:
I'll take a look later in the day, but it may be good to add a section to Schelling - it's our go-to in-depth example.

Datseris:
A full review, as well as an example that takes advantage of such functionality, is missing. I was thinking of an example. Can we come up with something really simple, but intuitive? Maybe we make a Hospital where doctors go and treat patients? Doesn't have to be realistic, just something that we could imagine checkpoints being important (e.g. every night a checkpoint is made, the model runs in the morning with a couple of different parameter configurations. Next night, all are compared, the most fitting to the reality is kept, and is saved as a new checkpoint).

Datseris:
+1 for also adding a section in Schelling

AayushSabharwal:
I'll work on adding a section to Schelling, and meanwhile think of another example scenario where this could be used

examples/schelling.jl
```diff
@@ -327,3 +327,88 @@ adf
...
+
+# Most of the agents have settled happily. Now, let's save the model.
+
+AgentsIO.save_checkpoint("schelling.jld2", model)
+
+# Note that we can now leave the REPL, and come back later to run the model,
+# right from where we left off. Let's clear the model variable and then
+# load it back in from the file
+
+model = nothing
```
Datseris, line 357: what does this line achieve?

AayushSabharwal: It was meant to be akin to restarting the REPL, but it isn't strictly necessary. Even I was debating whether it was worth keeping or just confusing

Datseris: i think it's just confusing

AayushSabharwal:
I'm trying to figure out why docs failed to build properly. It appears that JLD2 couldn't find SchellingAgent to reconstruct when deserializing, which is absurd since it's in the same module.

AayushSabharwal:
Interestingly, I can't reproduce this locally outside of building docs. It seems to be some sort of interaction with how Documenter handles running code blocks

AayushSabharwal:
After trying various ways of trying to make it see `SchellingAgent` and building docs more times than I can count, I'm unable to figure this out right now. I'll try again later. Should I make an issue in JLD2?

Libbum:
I bet you it's in ../path 

I'll take a look in the active branch.

Libbum:
Yep!

https://github.com/JuliaDynamics/Agents.jl/blob/gh-pages/previews/PR505/examples/schelling.jld2

It's something to do with how Documenter.jl runs things. Had the same issue with all of the video plots, worked on it for ages and finally solved it by ignoring the problem...

src/submodules/io/AgentsIO.jl
```diff
@@ -6,10 +6,12 @@ Submodule containing functionality for serialization and deserialization of mode
 to and from files.
 
 You can save and load agent data to and from CSV files using [`AgentsIO.dump_to_csv`](@ref)
```
Libbum, line 8: I think this stuff can go into `api.md` directly. It seems odd that we have sections of the API that describe what's happening all the way, then `AgentsIO` is just at the bottom. As a user: what is that? What does it do?

It'd be better if we made a section called "Checkpointing" or "(De)Serialization" or something like that.

src/submodules/io/jld2_integration.jl
```diff
@@ -0,0 +1,199 @@
...
+        t.admissibility,
+        t.walkable,
+        t.cost_metric,
+    )
+
+"""
+    AgentsIO.save_checkpoint(filename, model::ABM)
+
+Write the entire `model` to the JLD2 file specified by `filename`. Agent data, including
+multi-agent models, is also saved. Serialization capability depends on JLD2.jl.
```
Libbum, line 168: > including multi-agent models

Drop this aspect of the sentence. Sure, we need to account for that, but a user should not need to care if they have one or more agent types.

Libbum: > Serialization capability depends on JLD2.jl

This is also implied by the sentence above and considering we include JLD2 as a dependency inside Agents, this is only implicitly true, not explicitly (you don't have to `using Agents, JLD2` to get this to work`.

src/submodules/io/jld2_integration.jl
```diff
@@ -0,0 +1,199 @@
...
+Currently, serialization is also not supported for models using OpenStreetMapSpace.
```
Libbum, line 169: The following section might be better in bullet points, with a note that users must be aware of these caveats before using this method blindly.

src/submodules/io/jld2_integration.jl
```diff
@@ -0,0 +1,199 @@
+using Random
+using JLD2
+
+"""
+    AgentsIO.to_serializable(t)
+
+Return the serializable form of the passed value. This defaults to the value itself,
+unless a more specific method is defined. Define a method for this function and for
```
Libbum, line 8: This section doesn't really tell me _when_ I need to care about it. What is a property that is not handled by the Agents parser? Will there be an error if my model is not serializable?

src/submodules/io/jld2_integration.jl
```diff
@@ -0,0 +1,199 @@
...
+        t.admissibility,
+        t.walkable,
+        t.cost_metric,
+    )
+
+function from_serializable(t::SerializableABM{S,A}; kwargs...) where {S,A}
+    abm = ABM(
+        A,
+        from_serializable(t.space; kwargs...);
+        scheduler = get(kwargs, :scheduler, Schedulers.fastest),
```
Libbum, line 113: In terms of speed, `get` is surprisingly slow. We use [haskey](https://github.com/JuliaDynamics/Agents.jl/blob/cc623023755e77ca49e72e5a97156ea342b48be5/src/spaces/grid.jl#L168-L172) throughout the space generation routines because of this.

Unsure if this is truly a problem in these routines, but it may be worth checking.

src/submodules/io/jld2_integration.jl
```diff
@@ -0,0 +1,199 @@
...
+
+Load the model saved to the file specified by `filename`.
+
+The keyword argument `scheduler = Schedulers.fastest` specifies what scheduler should
+be used for the model.
+
+The keyword argument `update_vel!` specifies a function that should be used to
+update each agent's velocity before it is moved. Refer to [`ContinuousSpace`](@ref) for
+details.
+
```
Libbum, line 192: Think we should add `warn` to the handled `kwargs`. It's best to keep `warn = true` as default to make sure the loading passes checks, but if I know I have a multi-agent model, I'd prefer to suppress the warning message on-load.

examples/schelling.jl
```diff
@@ -327,3 +327,86 @@ adf
 
 # We nicely see that the larger `:min_to_be_happy` is, the slower the convergence to
 # "total happiness".
+
+# ## Saving the model state
```
Libbum, line 331: Can move this above distributed computing and ensembles, it's a simpler thing to learn and will probably be used by more people.

examples/schelling.jl
```diff
@@ -327,3 +327,86 @@ adf
...
+figure
+
+# Most of the agents have settled happily. Now, let's save the model.
+
+AgentsIO.save_checkpoint("schelling.jld2", model)
+
+# Note that we can now leave the REPL, and come back later to run the model,
+# right from where we left off.
+
+model = AgentsIO.load_checkpoint("schelling.jld2"; scheduler = Schedulers.randomly)
```
Libbum, line 356: The way I'd fix this documenter failed load for the moment would be like this:

```julia
# Note that we can now leave the REPL, and come back later to run the model,
# right from where we left off.
#
# ```julia
# model = AgentsIO.load_checkpoint("schelling.jld2"; scheduler = Schedulers.randomly)
# ```
model = AgentsIO.load_checkpoint("../schelling.jld2"; scheduler = Schedulers.randomly) # hide
```

Libbum (commented):
No real issues with implementation at all&mdash;looking good!!

There are a few suggestions to improve documentation and user interactions, but most are minor. We can perhaps work through some of the doc fixes tomorrow in the meeting.

src/submodules/io/jld2_integration.jl
AayushSabharwal: [`kwargs` isn't a `NamedTuple`](https://stackoverflow.com/questions/59296403/how-do-i-retrieve-keyword-arguments-out-of-a-field-of-splatted-kwargs), so (as far as I'm aware) accessing it's fields requires converting it to `NamedTuple`/`Dict`/similar or using `get`. I'll compare speeds of the different methods to see which one works out faster, and if there's much of a difference at all.

Libbum: Ah, sure. Yeah, I'm talking about `get` from `Dict`, so my point is moot.

AayushSabharwal: In my tests, both `get` and the entire `haskey` structure took the same time

AayushSabharwal:
It seems ../path wasn't the issue. Now it can't see the file.

Libbum:
OK, so that's a bit more complicated. Is this a problem only in CI, only when building documentation, or is it the same locally?

AayushSabharwal:
It's a problem when building docs, both in CI and locally

AayushSabharwal:
I should add, I tried creating a file with a submodule, defining an agent and model, and save/loading it within the module. Didn't create any issues. For some reason it's only when building docs

Libbum:
Alright. That'll take some investigation. I'll be out for the afternoon so don't know when I'll be able to focus on it.

docs/src/api.md
```diff
@@ -285,9 +285,12 @@ Pathfinding.HeightMap
...
-## AgentsIO
+## (De)Serialization
 ```@docs
 AgentsIO
 AgentsIO.populate_from_csv!
-AgentsIO.dump_to_csv
+AgentsIO.save_checkpoint
+AgentsIO.load_checkpoint
+AgentsIO.to_serializable
+AgentsIO.from_serializable
```
Datseris, line 295: I don't see why these two functions should be exported and discussed in the user API. Perhaps we should discuss them in the developer's docs. But the user shouldn't be using them.

docs/src/api.md
```diff
@@ -285,9 +285,12 @@ Pathfinding.HeightMap
 Building a custom metric is straightforward, if the provided ones do not suit your purpose.
 See the [Developer Docs](@ref) for details.
 
-## AgentsIO
+## (De)Serialization
```
Datseris, line 288: I think this should be named `Save/Load` instead. E.g. the CSV functionality isn't really serialization. Also, I don't even think that the average Agents.jl user would know what the world serialization means.

Libbum:
Following our weekly meeting, the decision has been made to drop the additions to the Schelling Example from this PR and merge the rest once the documentation is up to scratch. A new PR will be opened with the Schelling Example additions as-is, and we'll investigate the Documenter/JLD2 issue as time permits.

We've also decided to support OSM space by saving agent lat/long coordinates and expecting the user to know the map path to deserialize their model: we will rebuild the mapspace on-load.

Libbum:
These doc changes look good. So the only outstanding thing is OSMSpace from my perspective.

AayushSabharwal:
I'm removing `hoods` serialization, but apart from that OSMSpace is the only thing left. If I remember correctly, we planned to handle that in a separate PR

Datseris:
yes go ahead and merge when tests pass _but do not tag a new release_. New releases should be tagged only when documentation exists. Feel free to open a tracking issue with the docs code pasted there.

AayushSabharwal:
Ah. I just missed that comment.
I'll revert this commit then. The only docs missing is the example right?

AayushSabharwal:
I think this branch can be kept for now? Since OSMSpace is coming up, along with the fix to schelling

PR #506, opened 2021-06-14
Simpler ABM print

kavir1698:
Here is a small suggestion to avoid long and unhelpful ABM printing.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["Tim DuBois", "George Datseris", "Ali Vahdati"]
-version = "4.3.1"
+version = "4.3.2"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -351,7 +351,11 @@ function Base.show(io::IO, abm::ABM{S,A}) where {S,A}
     s *= "\n scheduler: $(schedulername(abm.scheduler))"
     print(io, s)
     if abm.properties ≠ nothing
-        print(io, "\n properties: ", abm.properties)
+        if typeof(abm.properties) <: Dict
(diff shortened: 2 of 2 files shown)


Datseris:
Is it correct to say that if properties is e.g.:
```julia
Base.@kwdef struct P
a = 1
b = 2
c = 3
end
```
then in this PR the values of the fields of `P` will **not** be shown when printing the model?

Can you provide an example where this is PR results in more helpful behavior?

kavir1698:
> then in this PR the values of the fields of `P` will **not** be shown when printing the model?
>
> Can you provide an example where this is PR results in more helpful behavior?

Yes, the values will not be shown. Unless `properties` has short scaler values, the print become bloated. I personally find that behavior annoying.

Libbum:
Yeah, I agree with your sentiment @kavir1698. Printing properties is only somewhat helpful if you have a small model, anything else and its a mess.

I'd even go so far as to remove the property print all together since users can always do `model.properties` directly if they need a list.

Thoughts on just dropping this all together? Or maybe just having `property count: $(length(model.properties))`?

Datseris:
I prefer the current approach where at least the property names are given.

kavir1698:
> Thoughts on just dropping this all together? Or maybe just having `property count: $(length(model.properties))`?

I have never had too many property names. I also feel printing their names doesn't hurt.

Datseris (approved):

Datseris:
Actually, I'd suggest to print the properties each in an individual line. What do you think about that? (current printing is also like this, as it prints like a dictionary)

kavir1698:
> Actually, I'd suggest to print the properties each in an individual line. What do you think about that? (current printing is also like this, as it prints like a dictionary)

Currently, a list of descriptions for `space` prints in a single line and it is clean. The names of `properties` in a single line sounds good to me.

Libbum:
Per line I don't like, it's going to be the same issue that we had in the first place: too much noise. I like Ali's recent change, it works well.

Datseris:
Okay from me.

Libbum:
We'll release this patch when the new JLD2 implementation is completely ready.

PR #511, opened 2021-06-16
OSMSpace Serialization

### LINKED ISSUE #496: Model (De)Serialization

### LINKED ISSUE #513: OSM.random_road_position and OSM.random_route! are not reproducible
**Describe the bug**
Both these functions are not reproducible since they use `generate_point_in_bounds` from OSMX, which doesn't use the model RNG.

**Minimal Working Example**
Zombies example. Two runs lead to different positions for some agents, which are ones that had a new path generated

**Agents.jl version**
4.3

AayushSabharwal:
Closes #496, Closes #513

The agent position and destination lat/lon are stored in `SerializableOSMSpace`. During deserialization, these values are used to update the corresponding agent fields.


AayushSabharwal:
Currently agent routes are rendered meaningless when it loads back in. From what I understand, there isn't really a way to recalculate them since there isn't a way to know what keywords were passed to `plan_route` to calculate it. The only other solution I can think of is to save all the intersections by converting them to lat/lon.

It might also be a good idea to deepcopy each agent and clear the routes vector before saving, to save on space

AayushSabharwal:
`OSMSpace` serialization works now. There's some overhead with having to convert between indexes and lat/lon during serialization and deserialization, but that seems unavoidable. Currently there's unnecessary space saved for agent routes in the model, when they're unnecessary and recalculated during deserialization anyway. I think that should be avoidable with a shallow copy during serialization, and clearing the `agent.route`

src/submodules/io/jld2_integration.jl
```diff
@@ -75,6 +75,7 @@ struct OSMAgentPositionData
     id::Int
     pos::NTuple{2,Float64}
     dest::NTuple{2,Float64}
+    route::Vector{NTuple{2,Float64}}
```
Libbum, line 78: Is this necessary? Why can we not just re-recreate it on-load?

AayushSabharwal: Can it be determined whether the route was :shortest or :fastest? Or if it was a return route? If so, it's not necessary.

AayushSabharwal: Also, for `:fastest`, the speeds would need to be provided

AayushSabharwal:
There's a flaw in the implementation, which I'm working on fixing

AayushSabharwal:
Is OSMX deterministic with indexes of intersections? In other words, if I load the same map twice is intersection 565 the same in both? It appears to be so in a quick test, but that invalidates the need for any special serialization code

src/submodules/io/jld2_integration.jl
Libbum: Ah, yes - I see. That needs to be added as a property somewhere probably.

Worry that I have is that the indexes in the route are not transferable due to the OSMX structure, so even adding lat/lon here it means that we'd need to make an additional bunch of methods to re-embed some synthetic route. 

Without the route type property stored that'll be a problem in general, so may make sense to consider that extension somehow. 

Bah, but this could be a per-agent property.... A bit to think about here. 

Feel free to go ahead in the direction you're taking here and we'll see if you hit a wall or not, then we can figure out if there's a simpler design.

Libbum:
> Is OSMX deterministic with indexes of intersections? In other words, if I load the same map twice is intersection 565 the same in both? It appears to be so in a quick test, but that invalidates the need for any special serialization code

No. That's the issue.

It's the same on your machine with your julia version. Change to 1.5 and it wont work, send the file to me and I can't load it.

AayushSabharwal:
Ah. I get it. This complicates things a little, but I have an idea

AayushSabharwal:
This seems to work. As of now:
- Agent positions are saved as a 3-tuple, where the first two values are the lat/lon of the corresponding intersection and the third field is identical to `OSMAgent.pos[3]`. Similarly for destination
- Routes are saved as a vector of lat/lon of the corresponding intersections

Libbum:
Alright. 

To actually review this, can you make an example using the `OSM.TEST_MAP` with a few agents on paths that you generate, save the file and email it to me. I'll load it up, plot the paths and you tell me if your model is the same.

AayushSabharwal:
There doesn't seem to be a way to easily shallow copy agents without the user having defined `copy` for that type. `deepcopy` is still an option, but it could potentially be a lot slower

Libbum:
> There doesn't seem to be a way to easily shallow copy agents without the user having defined copy for that type. deepcopy is still an option, but it could potentially be a lot slower

Indeed. There isn't one. For collection, [we tell them to be mindful](https://github.com/JuliaDynamics/Agents.jl/blob/29fb63fd05a7c14a9d17c6aade6a520368c6bc56/src/simulations/collect.jl#L97) about that. Do you think there needs to be documentation here as well?

AayushSabharwal:
I'm only looking into it to avoid storing `agent.route` since it's redundant data. It works perfectly without it, albeit with a larger file size. I have an idea with some metaprogramming magic which I'll try out

AayushSabharwal:
This reduces file size, and tests seem to pass locally

AayushSabharwal:
I expected it to merge as a single commit, but I guess this works too. Squash and merge should compress it all into one in the end.

docs/src/index.md
```diff
@@ -10,6 +10,7 @@ To get started, please read the [Tutorial](@ref) page.
 
 !!! tip "Latest news: Agents.jl v4.3"
       * New functions `dump_to_csv` and `populate_from_csv!` to save and load agent data to and from CSV files
+      * New functions `save_checkpoint` and `load_checkpoint` to save and load entire models to and from files
```
Libbum, line 13: Since this is new functionality outside of what we've already implemented, we will have to release 4.4. So this section needs to be updated accordingly.

src/submodules/io/jld2_integration.jl
```diff
@@ -80,42 +91,63 @@ struct SerializableAStar{D,P,M}
...
 end
 
-to_serializable(t::ABM) = SerializableABM(
-    collect(values(t.agents)),
-    to_serializable(t.space),
-    to_serializable(t.properties),
-    t.rng,
-    t.maxid.x,
-)
+function to_serializable(t::ABM{S}) where {S}
```
Libbum, line 94: A lot of this section should use helper functions from [here](https://github.com/JuliaDynamics/Agents.jl/blob/22076028a00cc24ac619567627967bcd76c97268/src/core/model.jl#L218-L230).

`nagents(model)` vs `length(sabm.agents)`
`allagents(model)` vs `values(t.agents)`

etc.

Libbum (approved):
Once those comments are resolved, bump the changelog and it's good for a merge/release.

Datseris (approved):
Agree with Tim!

AayushSabharwal:
I'll merge once CI finishes

Datseris:
We will need a new "update message", which will need to be done via Scratch.jl before we release, c.f. #493  (this needs to be done in a separate PR)

AayushSabharwal:
Sure. I'll spend some time going through Scratch.jl and open a PR with the changes

AayushSabharwal:
50% 🥳

PR #514, opened 2021-06-22
Use Scratch.jl for update messages

### LINKED ISSUE #493: Use Scratch.jl for version updates

AayushSabharwal:
Closes #493

### DIFF
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -63,6 +63,7 @@ function __init__()
                 \nUpdate message: Agents v$(version_number)
                 Please see the changelog online. Some key features:
     
+                * Agent data can be loaded from and saved to CSV files using `populate_from_csv!` and `dump_to_csv`
                 * Support for saving and loading entire models using `save_checkpoint` and `load_checkpoint`
                 """;
                 color = :light_magenta,
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -48,14 +48,14 @@ function __init__()
     @require Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4" begin
         include("visualization/plot-recipes.jl")
     end
-    # Get scratch space for this package
-    versions_dir = @get_scratch!("versions")
     # Update message:
     display_update = true
     version_number = "4.4"
     update_name = "update_v$(version_number)"
     
     if display_update
+        # Get scratch space for this package
+        versions_dir = @get_scratch!("versions")
         if !isfile(joinpath(versions_dir, update_name))
             printstyled(
                 stdout,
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
Since 4.3 didn't have a message, should the CSV functions also be mentioned in the message for 4.4?

Datseris:
Yes!

src/Agents.jl
```diff
@@ -47,6 +48,28 @@ function __init__()
     @require Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4" begin
         include("visualization/plot-recipes.jl")
     end
```
Datseris, line 50: Wait, what's this doing here? Guess we can remove it, right @Libbum ?

Libbum: No, that needs to stay or the docs will die. It's fine as-is.

Libbum (approved):
Yep, all looks good. Add the CSV string and release!

AayushSabharwal:
I added the message for CSV integration in the last commit. Is there anything else I'm missing?

Libbum:
Sorry, yes&mdash;was looking at the old diff.

In principle, this should be fine. Although I don't think any of us have used scratch before. Have you pulled this branch into a new environment to make sure it works? I might do that now...

AayushSabharwal:
It worked for me. `using` it the first time displayed the message, after that it didn't. I went to the scratchspaces directory and deleted the file, and it displayed the message again

src/Agents.jl
```diff
@@ -47,6 +48,29 @@ function __init__()
...
+    # Get scratch space for this package
+    versions_dir = @get_scratch!("versions")
```
Datseris, line 52: I would move this inside the `if display_update` clause

Datseris (approved):

Libbum:
Can confirm. All good after George's suggestion.

AayushSabharwal:
There's no point waiting for docs, so I'll merge once tests pass. After this is the release process the same as before?

Libbum:
Yep, exactly the same.

Issue #510, opened 2021-06-16
Fix up documenter drift


Libbum:
There are a few docstrings that are failing on CI at present (HK for example) that are due to version updates and other drift that we should contend with.

PR #518, opened 2021-06-27
Fix up doc drift

Libbum:
Should sort out #510.


Libbum:
Current failures should be fixed when InteractiveDynamics 0.15.2 is released

Libbum:
0.15.3 should fix up the top and right bounding issues.

PR #519, opened 2021-06-28
Antworld example

astenuz:
Hi! I would like to add this example, is an implementation of the original ants netlogo model (https://ccl.northwestern.edu/netlogo/models/Ants), with some adaptations to work with a grid similar to the mesa model in (https://github.com/mgoadric/ants-mesa)

Describes a colony of ants that wander off from the colony and find food to bring back.
The ants move around randomly at first, and when encountering food, return to their colony leaving a pheromone trail
Other ants when moving randomly could also follow the pheromone trail in the strongest direction.

Please let me know how this can be improved, thanks!

### DIFF
--- a/examples/ants.jl
+++ b/examples/ants.jl
@@ -1,0 +1,457 @@
+# # Ants 
+
+# This is a implementation of the original ants netlogo model (https://ccl.northwestern.edu/netlogo/models/Ants),
+# with some adaptations to work with a grid similar to the mesa model in (https://github.com/mgoadric/ants-mesa)
+
+# This model describes a colony of ants that wander off from the colony and find food to bring back.
+# The ants move around randomly at first, and when encountering food, return to their colony leaving a pheromone trail
+# Other ants when moving randomly could also follow the pheromone trail in the strongest direction
+
+using Agents: isempty
+using Base: String, Float32, Float64
+using Agents, AgentsPlots
+using Gadfly: plot as ggplot, Geom, inch, SVG
+using GLMakie
+using InteractiveDynamics
+using Match
+using ImageFiltering
+using DataFrames: rename!, stack
+
+
+"""
+Agent that represents the colony
+"""
+mutable struct Colony <: AbstractAgent
+    "The identifier number of the agent"
+    id::Int
+    "The x, y location of the agent on a 2D grid"
+    pos::Tuple{Int,Int}
+    "amount of food brought back by the ants"
+    amount::Int
+end
+
+
+"""
+Agent that represents food
+"""
+mutable struct Food <: AbstractAgent
+    "The identifier number of the agent"
+    id::Int
+    "The x, y location of the agent on a 2D grid"
+    pos::Tuple{Int,Int}
+    "food cache that the agent belongs to"
+    cache_id::Int
+    "amount of food that this agent holds"
+    amount::Int
+end
+
+
+"""
+Ants that wander around looking for food to bring back to the colony
+"""
+mutable struct Ant <: AbstractAgent
+    "the identifier number of the agent"
+    id::Int
+    "The x, y location of the agent on a 2D grid"
+    pos::Tuple{Int,Int}
+    "one of the allowed_angles"
+    angle::Float64
+    "state of the agent `:foraging` or `:going_home`"
+    state::Symbol
+    "amount of food that the agent is carrying"
+    amount::Int
+    "pheromone amount that the agent is dropping"
+    drop::Float32
+end
+
+# Utils for direction calculations
+
+"""
+Given an input angle in radians returns the closest angle 
+"""
+function closest_turn_angle(raw_angle)
+    diffs = abs.(raw_angle .- turn_angles)
+    return turn_angles[argmin(diffs)]
+end
+
+"""
+Given an angle return the direction of the neighbor that the angle points to.
+"""
+function grid_direction(angle)
+    return cmp.((cospi(angle/pi), sinpi(angle/pi)), 0)
+end
+
+"""
+Random turn angle considering a maximum absolute value `max_turn_angle`
+"""
+function random_angle(max_turn_angle)
+    allowed_angles = turn_angles[abs.(turn_angles) .<= max_turn_angle]
+    return rand(allowed_angles)
+end
+
+function random_direction(max_turn_angle)
+    return grid_direction(random_angle(max_turn_angle))
+end
+
+"""
+Rotates the `angle` by `turn_angle`. 
+
+The angle and rotations must be one of the turn angles,
+with this, we keep the rotated angle in the same 'ring' of angles
+"""
+function rotate_angle(angle, turn_angle)
+    angle_pos = findfirst(isequal(angle), turn_angles)
+    turns = round(Int, turn_angle/(pi/4))
+    new_angle_pos = mod1(angle_pos + turns, length(turn_angles))
+
+    return turn_angles[new_angle_pos]
+end
+
+"""
+Given two positions in the grid, return the closest `turn_angle` between them
+"""
+function angle_between(orig_pos, dest_pos)
+    direction = dest_pos .- orig_pos
+    
+    #adjust_angle
+    raw_angle = atan(direction[2], direction[1])
+    raw_angle = raw_angle > pi ? pi - raw_angle : raw_angle
+
+    return closest_turn_angle(raw_angle)
+end
+
+# setup
+
+"""possible angles that the ant can rotate, 8 of them"""
+turn_angles = Vector(-3*pi/4:pi/4:pi);  # multiples of pi/4
+
+"""cache positions, fractions relative to the world dims"""
+cache_positions = [
+    (3/5, 1/2), (1/5, 2/5), (1/5, 4/5)]  # fractions relative to the world dims
+
+"""directions in tuple format (x, y) ej. (1,-1)"""
+turn_directions = grid_direction.(turn_angles)
+
+"""
+Initializes ABM
+
+# Arguments
+- `num_ants::Integer`: total of ants agents.
+- `griddims::Tuple[Int]`: dimensions of the ant world.
+- `max_turn_angle::Int`: maximum angle that the ant is allowed to turn, 
+the direction is not specified so the ant will turn between ``\[-max_turn_angle, max_turn_angle\]``.
+- `evaporation_r::Float64`: evaporation rate of the pheromones.
+- `diffusion_r::Float64`: diffusion rate of the pheromones, 
+which is the portion of the tile that is spread across neighbors.
+- `init_drop::Integer`: Initial pheromone drop once the ant starts going back to the colony.
+- `drop_rate::Float64`: Decaying rate of the pheromone drop each step.
+- `random_move_prob::Float64`: probability of ignoring pheromones when foraging.
+"""
+function initialize(; 
+    num_ants = 200, 
+    griddims = (71, 71), 
+    max_turn_angle=2*pi/4, 
+    evaporation_r=0.2, diffusion_r=0.2, 
+    init_drop = 60, drop_rate=0.99, 
+    random_move_prob = 0.1)
+    
+    space = GridSpace(griddims, periodic=false, metric=:chebyshev)
+    properties = (
+        max_turn_angle = max_turn_angle,
+        evaporation_r = evaporation_r, 
+        diffusion_r=diffusion_r, 
+        init_drop = init_drop,
+        drop_rate=drop_rate,
+        random_move_prob = random_move_prob,
+        ant_pheromone = zeros(Float16, griddims),
+        colony_distance = zeros(Float16, griddims)
+    )
+    
+    model = ABM(Union{Ant, Colony, Food}, space;
+                properties = properties, scheduler = random_activation)
+
+    # add colony
+    center_pos = div.(griddims, 2)
+    setup_colony!(0, center_pos, model)
+
+    # initialize the ants
+    for _ in 1:num_ants
+        agent = Ant(nextid(model), center_pos, random_angle(pi), :foraging, 0, 0)
+        add_agent_pos!(agent, model)
+    end
+
+    for (cid, center_ratio) in enumerate(cache_positions)
+        center_pos = floor.(Int, center_ratio .* griddims)
+        setup_cache!(cid, center_pos, 5, model)
+    end
+    return model
+end
+
+"""Places the colony agent"""
+function setup_colony!(id, pos, model)
+    agent = Colony(id, pos, 0)
+    add_agent_pos!(agent, model)
+
+    @inbounds for p in positions(model) # we don't have to enable bound checking
+        model.colony_distance[p...] = sqrt(sum((pos .- p).^2))
+    end
+end
+
+"""Places the cache agents"""
+function setup_cache!(cid, center, radius, model)
+    agent = Food(nextid(model), center, cid, rand([1,2]))
+    add_agent_pos!(agent, model)
+
+    for pos in nearby_positions(center, model, radius)
+        agent = Food(nextid(model), pos, cid, rand([1,2]))
+        add_agent_pos!(agent, model)
+    end
+end
+
+# Movement actions
+
+"""
+Moves the agent 1 step to the grid position in the direction of the `new_angle`
+"""
+function walk_to_angle!(agent, new_angle, model)
+    agent.angle = new_angle
+    new_direction = grid_direction(new_angle)
+    walk!(agent, new_direction, model)
+end
+
+"""
+Randomly moves the agent to one of the neighboring positions
+"""
+function random_move!(agent, model)
+    new_angle = rotate_angle(agent.angle, random_angle(model.max_turn_angle))
+    walk_to_angle!(agent, new_angle, model)
+end
+
+"""
+Moves the agent 1 cell closer towards the colony
+"""
+function homing_move!(agent, model)
+    neighbors = collect(nearby_positions(agent, model))
+    distances = [model.colony_distance[p...] for p in neighbors]
+
+    home_angle = angle_between(agent.pos, neighbors[argmin(distances)])
+    walk_to_angle!(agent, home_angle, model)
+end
+
+"""
+Agent performs a foraging move.
+- if it detects pheromones on the current position and its not oversaturated,
+the ant follows the pheromones with probability (`1-model.random_move_prob`)
+- the ant will only try to follow the trail on its front (think of a 45 cone to both sides, so the 3 tiles in front)
+- if there are no pheromones on the front, the ant also moves randomly 
+"""
+function foraging_move!(agent, model)
+    current_pheromones = model.ant_pheromone[agent.pos...]
+
+    if (current_pheromones > 0.05) & (current_pheromones <2) & (rand() > model.random_move_prob)
+        # scans pheromones in front
+        (pheromones, _, nearby_angles) = sense_pheromones(agent, model)
+        
+        if !isempty(pheromones)
+            # follows the stronger pheromones in front
+            best_angle = nearby_angles[argmax(pheromones)]
+            walk_to_angle!(agent, best_angle, model)
+        else
+            # no detected pheromones in front
+            random_move!(agent, model)
+        end
+    else
+        random_move!(agent, model)
+    end
+end
+
+"""
+senses pheromones in a cone directly in fron of the ant, 
+this helps avoid turning back every time the ant detects pheromones
+"""
+function sense_pheromones(agent, model)
+    nearby = collect(nearby_positions(agent, model))
+    nearby_angles = angle_between.([agent.pos], nearby)
+    angle_differences = abs.(agent.angle .- nearby_angles)
+
+    positions_in_front = nearby[angle_differences .<= pi/4]  # 45 degree cone
+    angles_in_front = nearby_angles[angle_differences .<= pi/4]  # 45 degree cone
+    
+    pheromones = [model.ant_pheromone[p...] for p in positions_in_front]
+    return (pheromones, positions_in_front, angles_in_front)
+end
+
+# agent actions
+"""
+agent step for the ant, depenfing on the state it eithers forages or returns home
+"""
+function agent_step!(agent::Ant, model)
+    if agent.state == :foraging
+        agent_forage!(agent, model)
+    else
+        agent_go_home!(agent, model)
+    end
+end
+
+"""
+Foraging action, if food is detected on the same position, 
+the ant eats it and starts to go back to the colony. 
+If no food is present the ant does a foraging move, trying to sense pheromones
+"""
+function agent_forage!(agent::Ant, model)
+    agents = collect(agents_in_position(agent.pos, model))
+    dinner_array = filter!(x -> isa(x, Food), agents)
+
+    if !isempty(dinner_array)
+        food = dinner_array[0]
+        eat!(agent, food, model)
+        agent.drop = model.init_drop
+        agent.state = :going_home
+    else
+        foraging_move!(agent, model)
+    end
+end
+
+"""
+Homing action, the ant will move one step closer to the colony. 
+It also drops pheromones along the way
+
+If the ant is on the colony, it drops the carry and goes back to foraging
+"""
+function agent_go_home!(agent::Ant, model)
+    agents = collect(agents_in_position(agent.pos, model))
+    colony_array = filter!(x -> isa(x, Colony), agents)
+
+    if isempty(colony_array)
+        drop_pheromone!(agent, model)
+        homing_move!(agent, model)
+    else
+        colony = colony_array[1]
+        colony.amount += agent.amount
+
+        agent.amount = 0
+        agent.angle = rotate_angle(agent.angle, pi) # turn back
+        agent.state = :foraging
+    end
+end
+
+"""
+The ant drops a pheromone on the current position, and reduces the amount of pheromone it will drop next
+"""
+function drop_pheromone!(agent, model)
+    model.ant_pheromone[agent.pos...] += agent.drop
+    agent.drop *= model.drop_rate
+end
+
+"""
+the ant takes an amount of 1 of the food source, if then the food is depleted, the food agent is deleted
+"""
+function eat!(agent::Ant, food, model)    
+    agent.amount += 1
+    food.amount -= 1
+    
+    if food.amount<=0
+        kill_agent!(food, model)
+    end
+end
+
+"""
+Agent step for the other agent types, blank because they do no action
+"""
+function agent_step!(_, _) 
+end
+
+# model actions
+"""
+Step for the model, diffuses the pheromones
+"""
+function model_step!(model)
+    diffuse!(model)
+end
+
+"""
+Diffuses pheromones giving 1/8 of the `diffusion_r*ant_pheromone` in the tile to each neighbor
+
+Also evaporates the pheromones with rate `1 - evaporation_r`
+"""
+function diffuse!(model)
+    diffusion_kernel = (1/8) * model.diffusion_r .* ones(3,3)
+    diffusion_kernel[2,2] = 1 - model.diffusion_r
+
+    diffused_grid = imfilter(model.ant_pheromone, diffusion_kernel) # diffuse with kernel
+    diffused_grid = (1-model.evaporation_r) .* diffused_grid  # evaporate
+    diffused_grid[diffused_grid.<= 0.05] .= 0  # if its too low make it zero
+    diffused_grid[model[0].pos...] = 0  # no pheromones inside the colony
+    
+    @inbounds for p in positions(model) # we don't have to enable bound checking
+        model.ant_pheromone[p...] = diffused_grid[p...]
+    end
+end
+
+# Plotting
+
+groupcolor(_::Ant) = :black;
+groupcolor(_::Colony) = RGBA(24/255, 48/255, 60/255, 0.6);
+groupcolor(a::Food) = @match a.amount begin 
+        0 => :black
+        _ => :green
+    end;
+
+groupmarker(a) = @match a begin
+        a::Ant => '⚫'
+        c::Colony => '■'
+        f::Food => '■'
+    end;
+
+grasscolor(model) = model.ant_pheromone;
+heatkwargs = (colormap = [:white, :blue], colorrange = (0, 10));
+
+plotkwargs = (
+    ac = groupcolor, am = groupmarker, as = 6, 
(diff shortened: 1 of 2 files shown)


Datseris:
Hi there,

you will need to use `GLMakie` for the final plot, not `ggplot`. We can't afford to install a second plotting package. Can you please paste here the animations this produces?

astenuz:
Hi! This is the video output:

https://user-images.githubusercontent.com/4689127/123644604-256b2f80-d7eb-11eb-9402-840e9530dfee.mp4

astenuz:
Also, dependencies for the example should be the same as in Agents.jl right? Is there a particular reason to match the libraries? What if I add a disclaimer to the start of the example mentioning the libraries requiered, would that work?

Datseris:
The video looks good!

In general you should try to avoid as many non-Agents.jl as possible. Why do you need ImageFiltering? Why do you need to plot something with Gadfly instead of Makie?

examples/ants.jl
```diff
@@ -0,0 +1,457 @@
...
+groupcolor(a::Food) = @match a.amount begin 
+        0 => :black
+        _ => :green
+    end;
+
+groupmarker(a) = @match a begin
+        a::Ant => '⚫'
+        c::Colony => '■'
+        f::Food => '■'
+    end;
```
Datseris, line 404: Please avoid this. Don't use `Match` and instead write standard Julia code that does the same thing. Not only it is more readable, it even takes less lines of code.
```julia
groupmarker(a::Ant) = '⚫'
groupmarker(c::Colony) = '■'
groupmarker(f::Food) = '■'
```

Datseris (changes requested):

astenuz:
> The video looks good!
> 
> In general you should try to avoid as many non-Agents.jl as possible. Why do you need ImageFiltering? Why do you need to plot something with Gadfly instead of Makie?

Thanks! I used gadfly since im more familiar with something that follows the grammar of graphics. I can see how to change it, It only makes a line per food cache with a different color each. The image filtering is there because it has a method that does same-padding convolution, that is used to implement the difussion of pheromones.

image

examples/ants.jl
astenuz: oh ok, ill change that part

Datseris:
Okay then ImageFiltering cannot be avoided I guess. To use the makie just do
```julia
using CairoMakie
fig = Figure()
plot(line1)
plot(line2)
fig
```

Datseris:
May I suggest you to have a look at the predator-prey model? https://juliadynamics.github.io/Agents.jl/stable/examples/predator_prey_fast/ 

It seems to me that you will gain massive performance gains by treating the _ant trails_ as a standard Matrix which is model property, as we do with food in the predator pray. This will also be immediatelly passable to ImageFiltering.

AayushSabharwal:
Coincidentally, I have this exact model in the works and it's great to see it added here too! The convolutional filter is simply replacing each cell with the mean of it's surroundings (and a decay), right? If so, it can be implemented manually, although it requires allocating another array. I think it might be worth a look, since otherwise someone who hasn't seen `ImageFiltering` might not be able to understand what's going on.

Datseris:
We do this in Daisyworld without allocating an extra array (which comes at the cost of accuracy). Okay, so in principle we can remove ImageFiltering from the example, which will reduce build time for the documentation.

astenuz:
Hi again! I simplified dependencies removing match and gadfly. 

About the difussion, using the imgfilter was perhaps for me the cleanest way since its clear that youre running a convolution(cross-correlation?) across each cell, and I also tried to avoid reimplementing it. However, if you could point me towards a reimplementation of diffusion i could also replace that last part.  Maybe event it can be added to the core methods similar to how netlogo does it (http://ccl.northwestern.edu/netlogo/docs/dict/diffuse.html)

examples/ants.jl
astenuz: changed

Libbum:
In addition to simplification of the dependencies, I think there's some work that we could do to make the example a little more Agents.jl-like. Movement for example: we can rely on [walk!](https://juliadynamics.github.io/Agents.jl/stable/api/#Agents.walk!) heavily.

Perhaps we can add a few helper functions in the standard API to assist with turtle-like movement (which Netlogo relies on exclusively)... Or, if we're in no hurry to get this merged I can take a look at the movement needs of the agents and re-design some aspects of that.

Unsure if `Colony` and `Food` need to be Agents at all, perhaps they're better off as properties of the model?

Will think about the design a bit more in the coming days and come back to this.

astenuz:
> In addition to simplification of the dependencies, I think there's some work that we could do to make the example a little more Agents.jl-like. Movement for example: we can rely on [walk!](https://juliadynamics.github.io/Agents.jl/stable/api/#Agents.walk!) heavily.
> 
> Perhaps we can add a few helper functions in the standard API to assist with turtle-like movement (which Netlogo relies on exclusively)... Or, if we're in no hurry to get this merged I can take a look at the movement needs of the agents and re-design some aspects of that.
> 
> Unsure if `Colony` and `Food` need to be Agents at all, perhaps they're better off as properties of the model?
> 
> Will think about the design a bit more in the coming days and come back to this.

Ah that'd be nice, all the stuff with the angles is to replicate in some sense the notion of 'angle direction' and turning 'relative to the angle' so that the ant can move limited to some particular directions. It actually uses walk! but since it uses directions theres a bit of boilerplate to  transform from current angle + rotation angle to a particular direction within the neighbors.

examples/ants.jl
```diff
@@ -447,11 +441,15 @@ data, _ = run!(
...
+    c3 = lines!(ax, data.step, data.amount_cache_3, color = :green)
+    figure[1, 2] = Legend(figure, [c1, c2, c3], ["Cache 1", "Cache 2", "Cache 3"])
+    return figure
+end
 
-p = ggplot(data_melt, x=:step, y=:amount, color=:cache_number, Geom.line)
-p|> SVG("cache_amounts.svg")
\ No newline at end of file
+fig = plot_food_timeseries(data);
+save("cache_amounts.png", fig);
```
Datseris, line 455: There is no reason to save the figure, having the figure as final argument already shows it at the built docs.

astenuz: hmmm, what about the video? should it be saved?

astenuz: ah wait thats saved by default nvm

Datseris:
> Ah that'd be nice, all the stuff with the angles is to replicate in some sense the notion of 'angle direction' and turning 'relative to the angle' so that the ant can move limited to some particular directions. It actually uses walk! but since it uses directions theres a bit of boilerplate to transform from current angle + rotation angle to a particular direction within the neighbors.

That's also similar with the flocking example, so I guess perhaps there could be some helper functions that simplify both.

examples/ants.jl
astenuz: I removed the saving of that plot

astenuz:
> > Ah that'd be nice, all the stuff with the angles is to replicate in some sense the notion of 'angle direction' and turning 'relative to the angle' so that the ant can move limited to some particular directions. It actually uses walk! but since it uses directions theres a bit of boilerplate to transform from current angle + rotation angle to a particular direction within the neighbors.
> 
> That's also similar with the flocking example, so I guess perhaps there could be some helper functions that simplify both.

That'd go in the main API, right? I was thinking something like `TurtleAgent <: AbstractAgent` could perhaps be used to define things like `rotate!` and `move!` or maybe just a `walk!(turtle::TurtleAgent, rotation, distance)`.

Libbum:
> That'd go in the main API, right?

Yes. We would not want to have a specific `TurtleAgent` type like netlogo, but instead would provide helper functions for *any* agent to walk in the way a turtle agent would.

Heading is something we'd need to pay attention to in that case, since that's essentially an agent property. 

One way we can handle that without forcing a new agent type would be to make functions like `rotate!(agent, heading, angle)`, where a user could create a wrapper function `rotate_ant!(agent, angle) = rotate!(agent, agent.heading, angle)`. 

There will be a few design iterations on this I suspect. 

Essentially though, we'd like to provide an interface for Netlogo users to transition to Agents.jl and be comfortable with doing so, but not just replicating Netlogo functionality. We have a somewhat different way of thinking about ABM problems and how to attack them.

astenuz:
> > That'd go in the main API, right?
> 
> Yes. We would not want to have a specific `TurtleAgent` type like netlogo, but instead would provide helper functions for _any_ agent to walk in the way a turtle agent would.
> 
> Heading is something we'd need to pay attention to in that case, since that's essentially an agent property.
> 
> One way we can handle that without forcing a new agent type would be to make functions like `rotate!(agent, heading, angle)`, where a user could create a wrapper function `rotate_ant!(agent, angle) = rotate!(agent, agent.heading, angle)`. **Edit:** _actually this may not be sufficient..._
> 
> There will be a few design iterations on this I suspect.
> 
> Essentially though, we'd like to provide an interface for Netlogo users to transition to Agents.jl and be comfortable with doing so, but not just replicating Netlogo functionality. We have a somewhat different way of thinking about ABM problems and how to attack them.

I'm guessing you mentioned that the general api might not be sufficient since the new agents require a notion of current direction right? Perhaps that's why I was thinking of something like `TurtleAgent` or `DirectionalAgent` which composes `AbstractAgent` and has additional fields which could be direction and speed. Then we can build specific functions like `rotate(a::DirectionalAgent)` for each model space type. I'm wondering if there's a way to generate the same notion without giving the agent a current direction/angle/velocity state. 

Perhaps an apparent issue with this is that the current functions such as walk should have new methods for the directional agents that in addition to the abstract_agent's walk, update the current direction. That's a bit more tricky.

Datseris:
From what I read, this seems like a situation that `ContinuousSpace` would handle much, much better. Not only the ContinuousAgent type has already the concept of direction and speed, in the field `vel`, but there "direction" actually much more sense than the grid...

A `update_vel!` function would read the underlying grid of pheremones and update agent velocity accordingly.

Datseris:
(BTW, we could offer a trivial `rotation2D` function, which is `rotate2D(p, φ) = (cosφ*p[1] - sinφ*p[2], sinφ*p[1] + cosφ*p[2])` if this helps, I don't know)

astenuz:
> From what I read, this seems like a situation that `ContinuousSpace` would handle much, much better. Not only the ContinuousAgent type has already the concept of direction and speed, in the field `vel`, but there "direction" actually makes more sense than the grid...
> 
> A `update_vel!` function would read the underlying grid of pheremones and update agent velocity accordingly.
> 
> EDIT: But then again, the goal of this example is more like "transition from NetLogo", so perhaps we should stick to this grid approach.

I think it makes sense, to me it looks like in Netlogo the agents move as in continuous space but the stuff (food, pheromones) are specified by cell grid (the patches). Or at least it seems like that from the flexibility of real angle turnings and movements. Is there a way to do the same with Agents? For me it makes a bit more sense since the ants will move more like ants.

Datseris:
> Is there a way to do the same with Agents? 

Yes, in `ContinuousSpace`. Look at the flocking example: https://juliadynamics.github.io/Agents.jl/dev/examples/flock/

Datseris:
Specifying a cell grid on top of the continuous space is quite easy btw. A continuous space already has an underlying grid that is used for finding nearest neighbors, so you can re-use the same grid.

Datseris:
(I guess perhaps we can improve the documentation on this. Please first try to have a read at Flocking, and attempt to recreate the model in ContinuousSpace, and if you think the docs are not good, we see what we can improve)

astenuz:
Oh ok, I'll try to make some time for it :D

A first question comes to mind. What is the visual distance and how can you make sure it covers approximately one tile at most?

Another that just came to mind, how do you think I make sure that I select only the neighbors within a certain visual cone, I guess defined by (`r=visual distance, theta=[-max angle, +max_angle]`)

Datseris:
A visual cone would have to be an additional function, which I guess makes sense to implement in the "Continuous space exclusives" part of the documentation. The way I imagine it working is to add a further filter on the return value of `nearby_ids`. The function filters agents whose position angle w.r.t. the given agent's position is outside the defined cone. This is easy to do, you cancluate the angle simple as `atan((a.pos - b.pos)...)` with `a` the origin agent and `b` the agents we iterate over, found via `nearby_ids`.

Datseris:
What I'm thinking now is that the most efficient grid for finding nearest neighbors in continuous space would be different than the grid necessary for the pheromones... :( Because the neighbor optimal grid is typically about half the search radius, but I imagine your search radious would be much larger than the pheromone discretization.

astenuz:
Hi, I added the `ants_continuous.jl` example. In LOCs is almost the same since I added some stuff to rotate the vel vector and always have it unitary as direction.

For some reason I cannot get the video to work. It'd be nice If you could guide me since the error is quite arcane to me. The initial state seems to be ok. when I plot it, and the final after 1000 is actually a bit weird. seems like the pheromones are not working properly, but I cannot diagnose it well if I cannot see what's going on.

Also, I was wondering how to plot the heatmap of pheromones. I had to remove it (commenting the part in the plotkwargs) since it caused another issue with the bounds but I'm also confused as to whats the reason. In the meantime I just want to see the ants haha.

astenuz:
This is the error I get if one uncomments the `make_video(model)` line

```
ERROR: LoadError: Metadata array needs to have same length as data.
                    Found 1288 data items, and 1280 metadata items
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] #meta#116
    @ ~/.julia/packages/GeometryBasics/qasyR/src/metadata.jl:125 [inlined]
  [3] pointmeta(mesh::GeometryBasics.Mesh{2, Float32, GeometryBasics.Ngon{2, Float32, 3, Point{2, Float32}}, GeometryBasics.SimpleFaceView{2, Float32, 3, GeometryBasics.OffsetInteger{-1, UInt32}, Point{2, Float32}, GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}}}; meta_data::Base.Iterators.Pairs{Symbol, Vector{RGBA{Float32}}, Tuple{Symbol}, NamedTuple{(:color,), Tuple{Vector{RGBA{Float32}}}}})
    @ GeometryBasics ~/.julia/packages/GeometryBasics/qasyR/src/meshes.jl:251
  [4] (::Makie.var"#564#566"{GeometryBasics.Mesh{2, Float32, GeometryBasics.Ngon{2, Float32, 3, Point{2, Float32}}, GeometryBasics.SimpleFaceView{2, Float32, 3, GeometryBasics.OffsetInteger{-1, UInt32}, Point{2, Float32}, GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}}}})(meshes::Vector{GeometryBasics.Mesh{2, Float32, GeometryBasics.Ngon{2, Float32, 3, Point{2, Float32}}, GeometryBasics.SimpleFaceView{2, Float32, 3, GeometryBasics.OffsetInteger{-1, UInt32}, Point{2, Float32}, GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}}}}, colors::Vector{RGBA{Float32}}, cmap::Nothing, crange::Nothing)
    @ Makie ~/.julia/packages/Makie/c5WJV/src/basic_recipes/poly.jl:165
  [5] (::Observables.MapUpdater{Makie.var"#564#566"{GeometryBasics.Mesh{2, Float32, GeometryBasics.Ngon{2, Float32, 3, Point{2, Float32}}, GeometryBasics.SimpleFaceView{2, Float32, 3, GeometryBasics.OffsetInteger{-1, UInt32}, Point{2, Float32}, GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}}}}, GeometryBasics.Mesh{2, Float32, GeometryBasics.TriangleP{2, Float32, GeometryBasics.PointMeta{2, Float32, Point{2, Float32}, (:color,), Tuple{RGBA{Float32}}}}, GeometryBasics.FaceView{GeometryBasics.TriangleP{2, Float32, GeometryBasics.PointMeta{2, Float32, Point{2, Float32}, (:color,), Tuple{RGBA{Float32}}}}, GeometryBasics.PointMeta{2, Float32, Point{2, Float32}, (:color,), Tuple{RGBA{Float32}}}, GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}, StructArrays.StructVector{GeometryBasics.PointMeta{2, Float32, Point{2, Float32}, (:color,), Tuple{RGBA{Float32}}}, NamedTuple{(:position, :color), Tuple{Vector{Point{2, Float32}}, Vector{RGBA{Float32}}}}, Int64}, Vector{GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}}}}})(::Vector{GeometryBasics.Mesh{2, Float32, GeometryBasics.Ngon{2, Float32, 3, Point{2, Float32}}, GeometryBasics.SimpleFaceView{2, Float32, 3, GeometryBasics.OffsetInteger{-1, UInt32}, Point{2, Float32}, GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}}}}, ::Vararg{Any, N} where N)
    @ Observables ~/.julia/packages/Observables/OFj0u/src/Observables.jl:372
  [6] (::Observables.OnUpdate{Observables.MapUpdater{Makie.var"#564#566"{GeometryBasics.Mesh{2, Float32, GeometryBasics.Ngon{2, Float32, 3, Point{2, Float32}}, GeometryBasics.SimpleFaceView{2, Float32, 3, GeometryBasics.OffsetInteger{-1, UInt32}, Point{2, Float32}, GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}}}}, GeometryBasics.Mesh{2, Float32, GeometryBasics.TriangleP{2, Float32, GeometryBasics.PointMeta{2, Float32, Point{2, Float32}, (:color,), Tuple{RGBA{Float32}}}}, GeometryBasics.FaceView{GeometryBasics.TriangleP{2, Float32, GeometryBasics.PointMeta{2, Float32, Point{2, Float32}, (:color,), Tuple{RGBA{Float32}}}}, GeometryBasics.PointMeta{2, Float32, Point{2, Float32}, (:color,), Tuple{RGBA{Float32}}}, GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}, StructArrays.StructVector{GeometryBasics.PointMeta{2, Float32, Point{2, Float32}, (:color,), Tuple{RGBA{Float32}}}, NamedTuple{(:position, :color), Tuple{Vector{Point{2, Float32}}, Vector{RGBA{Float32}}}}, Int64}, Vector{GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}}}}}, Tuple{Observable{Vector{GeometryBasics.Mesh{2, Float32, GeometryBasics.Ngon{2, Float32, 3, Point{2, Float32}}, GeometryBasics.SimpleFaceView{2, Float32, 3, GeometryBasics.OffsetInteger{-1, UInt32}, Point{2, Float32}, GeometryBasics.NgonFace{3, GeometryBasics.OffsetInteger{-1, UInt32}}}}}}, Observable{Any}, Observable{Any}, Observable{Any}}})(#unused#::Vector{RGBA{Float32}})
    @ Observables ~/.julia/packages/Observables/OFj0u/src/Observables.jl:334
  [7] #invokelatest#2
    @ ./essentials.jl:708 [inlined]
  [8] invokelatest
    @ ./essentials.jl:706 [inlined]
  [9] notify
    @ ~/.julia/packages/Observables/OFj0u/src/Observables.jl:88 [inlined]
 [10] setindex!
    @ ~/.julia/packages/Observables/OFj0u/src/Observables.jl:248 [inlined]
 [11] (::Observables.var"#1#2"{Observable{Any}})(value::Vector{RGBA{Float32}})
    @ Observables ~/.julia/packages/Observables/OFj0u/src/Observables.jl:44
 [12] #invokelatest#2
    @ ./essentials.jl:708 [inlined]
 [13] invokelatest
    @ ./essentials.jl:706 [inlined]
 [14] notify
    @ ~/.julia/packages/Observables/OFj0u/src/Observables.jl:88 [inlined]
 [15] setindex!
    @ ~/.julia/packages/Observables/OFj0u/src/Observables.jl:248 [inlined]
 [16] (::Observables.var"#1#2"{Observable{Any}})(value::Vector{RGBA{Float32}})
    @ Observables ~/.julia/packages/Observables/OFj0u/src/Observables.jl:44
 [17] #invokelatest#2
    @ ./essentials.jl:708 [inlined]
 [18] invokelatest
    @ ./essentials.jl:706 [inlined]
 [19] notify
    @ ~/.julia/packages/Observables/OFj0u/src/Observables.jl:88 [inlined]
 [20] setindex!
    @ ~/.julia/packages/Observables/OFj0u/src/Observables.jl:248 [inlined]
 [21] (::Observables.var"#1#2"{Observable{Any}})(value::Vector{RGBA{Float32}})
    @ Observables ~/.julia/packages/Observables/OFj0u/src/Observables.jl:44
 [22] #invokelatest#2
    @ ./essentials.jl:708 [inlined]
 [23] invokelatest
    @ ./essentials.jl:706 [inlined]
 [24] notify
    @ ~/.julia/packages/Observables/OFj0u/src/Observables.jl:88 [inlined]
 [25] setindex!
    @ ~/.julia/packages/Observables/OFj0u/src/Observables.jl:248 [inlined]
 [26] (::Observables.var"#3#4"{Any, Observable{Any}})(value::Vector{RGBA{Float32}})
    @ Observables ~/.julia/packages/Observables/OFj0u/src/Observables.jl:58
 [27] #invokelatest#2
    @ ./essentials.jl:708 [inlined]
 [28] invokelatest
    @ ./essentials.jl:706 [inlined]
 [29] notify
    @ ~/.julia/packages/Observables/OFj0u/src/Observables.jl:88 [inlined]
 [30] setindex!(observable::Observable{Vector{RGBA{Float32}}}, val::Vector{RGBA{Float32}})
    @ Observables ~/.julia/packages/Observables/OFj0u/src/Observables.jl:248
 [31] step!(abmstepper::InteractiveDynamics.ABMStepper, model::AgentBasedModel{ContinuousSpace{2, true, Float64, typeof(Agents.defvel)}, Union{Ant, Colony, Food}, typeof(random_activation), NamedTuple{(:visual_distance, :max_turn_angle, :evaporation_r, :diffusion_r, :init_drop, :drop_rate, :random_move_prob, :ant_pheromone, :colony_distance), Tuple{Int64, Float64, Float64, Float64, Int64, Float64, Float64, Matrix{Float16}, Matrix{Float16}}}, Random.MersenneTwister}, agent_step!::Function, model_step!::Function, n::Int64)
    @ InteractiveDynamics ~/.julia/packages/InteractiveDynamics/NkWor/src/agents/stepping.jl:152
 [32] (::InteractiveDynamics.var"#64#67"{Int64, Int64, AgentBasedModel{ContinuousSpace{2, true, Float64, typeof(Agents.defvel)}, Union{Ant, Colony, Food}, typeof(random_activation), NamedTuple{(:visual_distance, :max_turn_angle, :evaporation_r, :diffusion_r, :init_drop, :drop_rate, :random_move_prob, :ant_pheromone, :colony_distance), Tuple{Int64, Float64, Float64, Float64, Int64, Float64, Float64, Matrix{Float16}, Matrix{Float16}}}, Random.MersenneTwister}, typeof(agent_step!), typeof(model_step!), InteractiveDynamics.ABMStepper, Observable{Int64}})(io::VideoStream)
    @ InteractiveDynamics ~/.julia/packages/InteractiveDynamics/NkWor/src/agents/plots_videos.jl:200
 [33] #Record#920
    @ ~/.julia/packages/Makie/c5WJV/src/display.jl:582 [inlined]
 [34] #record#919
    @ ~/.julia/packages/Makie/c5WJV/src/display.jl:576 [inlined]
 [35] abm_video(file::String, model::AgentBasedModel{ContinuousSpace{2, true, Float64, typeof(Agents.defvel)}, Union{Ant, Colony, Food}, typeof(random_activation), NamedTuple{(:visual_distance, :max_turn_angle, :evaporation_r, :diffusion_r, :init_drop, :drop_rate, :random_move_prob, :ant_pheromone, :colony_distance), Tuple{Int64, Float64, Float64, Float64, Int64, Float64, Float64, Matrix{Float16}, Matrix{Float16}}}, Random.MersenneTwister}, agent_step!::typeof(agent_step!), model_step!::typeof(model_step!); spf::Int64, framerate::Int64, frames::Int64, resolution::Tuple{Int64, Int64}, title::String, showstep::Bool, axiskwargs::NamedTuple{(), Tuple{}}, kwargs::Base.Iterators.Pairs{Symbol, Any, NTuple{4, Symbol}, NamedTuple{(:scheduler, :ac, :am, :as), Tuple{typeof(Agents.Schedulers.randomly), typeof(groupcolor), typeof(groupmarker), Int64}}})
    @ InteractiveDynamics ~/.julia/packages/InteractiveDynamics/NkWor/src/agents/plots_videos.jl:197
 [36] make_video(model::AgentBasedModel{ContinuousSpace{2, true, Float64, typeof(Agents.defvel)}, Union{Ant, Colony, Food}, typeof(random_activation), NamedTuple{(:visual_distance, :max_turn_angle, :evaporation_r, :diffusion_r, :init_drop, :drop_rate, :random_move_prob, :ant_pheromone, :colony_distance), Tuple{Int64, Float64, Float64, Float64, Int64, Float64, Float64, Matrix{Float16}, Matrix{Float16}}}, Random.MersenneTwister})
    @ Main ~/DS/introduction-to-complexity/ants/ants_continuous.jl:389
 [37] top-level scope
    @ ~/DS/introduction-to-complexity/ants/ants_continuous.jl:398
in expression starting at /Users/dagamba/DS/introduction-to-complexity/ants/ants_continuous.jl:398
```

astenuz:
Here are images of the initial state and the state after 1000 steps. For some reason the closest food cache is still there. Maybe the movement speed is too small.

Initial
image

After 1000 steps
image

astenuz:
There are also some things that perhaps could greatly simplify this code if they were on the API:
1. Functions to rotate the current vel
2. Functions o figure out the angle between the agent and nearby_positions/agents
3. a current_position or agents_in_position, I used a nearby agents with `r=model.space.spaciing/2`but I'm not sure this is the best option
4. A way to map grid positions to positions in the extent

there are also some things that made this example a bit confusing when coding
3. Its weird that when you do `nearby_positions` you  get grid positions but when you do nearby agents and you check their `.pos`you get position within the extent
4. following point 4. of the nice things. What should be the position in reference to the extent of certain grid position? should it always mark the center of the patch? 
5. I'm no sure how to for example get only the adjacent neighbors of an agent. what should `r` be?

That's it for now. It's been nice to learn how to handle better the code in Agents. Thanks for your support

examples/ants_continuous.jl
```diff
@@ -0,0 +1,434 @@
...
+
+# This is a implementation of the original ants netlogo model (https://ccl.northwestern.edu/netlogo/models/Ants),
+# with some adaptations to work with a grid similar to the mesa model in (https://github.com/mgoadric/ants-mesa)
+
+# This model describes a colony of ants that wander off from the colony and find food to bring back.
+# The ants move around randomly at first, and when encountering food, return to their colony leaving a pheromone trail
+# Other ants when moving randomly could also follow the pheromone trail in the strongest direction
+
+using Base: String, Float64, Tuple
+using Agents, AgentsPlots
```
Datseris, line 11: AgentsPlots **is deprecated since many months ago**

Datseris:
The video problem is probably a bug on Agents.jl side. We'll have a look as soon as possible.

Libbum:
All excellent questions and suggestions @astenuz. Thanks for taking the time to work on this and changing what you can. As you can see in our examples, a lot of work has gone into `GridSpace` systems but not as much in `ContinuousSpace` - but this is now one of our areas of focus.

You may have identified some issues here, or we just need some better documentation. I won't be able to look in-depth into these questions for another week or so due to other commitments, but I think we can use this example as a decent springboard to fix up a lot of these edge cases.

examples/ants_continuous.jl
```diff
@@ -0,0 +1,434 @@
...
+using ImageFiltering: imfilter as matrix_correlation
+using DataFrames: rename!, stack
+using LinearAlgebra: normalize
+using Distributions: Uniform
+
+
+"""
+Agent that represents the colony
+"""
+mutable struct Colony <: AbstractAgent
```
Libbum, line 23: I'd say that some of your issues are coming about from mixing agents that sit on `GridSpace` and `ContinuousSpace` in the same model. 

Pretty sure we should actually disallow that in general.

I think the better way to implement this example would be to have Colony and Food as model properties actually. They don't need to be agents. Similar to how Grass is made in [Predator - Prey](https://juliadynamics.github.io/Agents.jl/stable/examples/predator_prey_fast/).

reyman: Ahah interesting @Libbum , i'm working on the same example for an open source project dedicated to distributed computing (openMole) !!
You will find my work (in progress) here :  https://github.com/reyman/julia-antworld-model

reyman: I push a much much better version tonight :)
Not perfect, but that's also a start using both env for continuous and grid ... 

https://user-images.githubusercontent.com/466329/124834777-95627e00-df80-11eb-8b69-af9cb23716de.mp4

reyman: I push an ultimate version, things that miss : 
- [ ] get the front patch (like the patch-right-and-ahead 1, patch-left-and-ahead 1 from netlogo)
- [ ] better random path (linked to 1, i need to select the best pheromone in the front ant patch only)

Datseris:
With PR #526 merged, this can take advantage of better functionality based on `get_spatial_property` to have a natural antworld on continuous space.

astenuz:
> With PR #526 merged, this can take advantage of better functionality based on `get_spatial_property` to have a natural antworld on continuous space.

Sounds good, where can I check the docs on how to use it, and what functions you think will make the example simpler?

Datseris:
> Sounds good, where can I check the docs on how to use it, and what functions you think will make the example simpler? 

Unfortunately this functionality only exists on `master` branch, and not a stable release. But since this PR builds on master, that's not a problem. You need to use this function: https://juliadynamics.github.io/Agents.jl/dev/api/#Agents.get_spatial_property

but also the function `get_spatial_index`, which I unfortunately forgot to cross - reference on that list... Let me know how it goes if you have any trouble, as I wrote this functions without an application so far.

Datseris:
@astenuz thank you! In Agents.jl 5.0 all examples besides the 5 central ones to the library will be in a new repository called "Agents.jl Example Zoo". Can you please make this PR there instead? https://github.com/JuliaDynamics/Agents.jl_Example_Zoo

Datseris:

PR #520, opened 2021-06-28
Pathfinding rework

### LINKED ISSUE #515: Pathfinding extensions
I had an idea about a possible implementation for this. It's a breaking change, however. I came across @Datseris' comment in `src/spaces/grid.jl:66`:
```julia
# TODO: This is bad design. `AStar` should not be mentioned here,
# nor any `Pathfinding` business. This file should be "pure".
astar = pathfinder === nothing ? nothing : Pathfinding.AStar(d, periodic, pathfinder)
```

The only reason this is here is because `AStar` takes its dimensions and periodicity from the space. Even for `ContinuousSpace` pathfinding, none of the functionality needs any relation to the space. What if this was moved to the model? This also eliminates having to modify `ContinuousSpace` in any way. The user passes in `Pathfinder` structs, which the model turns into `AStar` structs.

This also ties into the part of my proposal for allowing better control over pathfinding, and specifically having multiple "profiles" and allowing a choice of which to use. Let the existing `Pathfinder` struct be renamed to `GridPathfinder`. `ABM` then has a field of type `Pathfinder`, which is a new struct:

```julia
struct Pathfinder{P}
    profile_map::Dict{Int,Int}  # maps agent ID to profile
    profiles::P # `AStar` or `Vector{AStar}` if there are multiple profiles
end
```

The constructor takes a single/list of `GridPathfinder` along with dimensions and periodicity and initializes itself accordingly.

There's probably a way to eliminate `profile_map` in case there is only one profile.

- `set_target!`, `set_best_target!`, `heightmap`, `walkmap` take an additional parameter `profile::Int = 1` which only needs to be specified if there are multiple profiles
- `move_along_route!`, `is_stationary`, `kill_agent!` use the appropriate profile from `profile_map` for the specified agent

To add support for `ContinuousSpace`, all that's needed is a `ContinousPathfinder` struct that the user can pass to `ABM`, which in turn passes it to the `Pathfinder` constructor. Then an additional method for `move_along_route!`, `set_target!`, since coordinates need to be converted to- and from- grid indices, and some edge cases in the path calculation.

It's a little complicated, and certainly needs both experimentation and profiling. Hopefully I've explained everything clearly

AayushSabharwal:
Closes #515 
Pathfinders will now be separate from core data structures. They've been removed from `GridSpace`, and API functions reworked/removed. Tests need to be fixed and currently don't pass. Docs aren't accurate. ~Serialization is also broken.~ Serialization works now.

Some points to consider:
- `walkmap(model)` was removed since now it can just be accessed by `pathfinder.walkable`. Should it be renamed to `pathfinder.walkmap` for consistency?
- `heightmap(model)` is now `heightmap(pathfinder)`, since it's still more convenient than `pathfinder.cost_metric.hmap`
- `kill_agent!` was removed, since we don't have access to any pathfinder object directly. Should a method `kill_agent!(agent, model, pathfinder)` be added?
- Since pathfinding for continuous space will be added too, the filename `grid_pathfinder.jl` is not appropriate. I think `grid_pathfinder.jl` should be renamed to `astar.jl` and only contain the core AStar functionality. The API functions should be in a separate file(s), such as `grid_pathfinding.jl` and `continuous_pathfinding.jl`.
- Since pathfinding will now be in multiple spaces, and considering there may be more implementations later, should a general set of functions be established as the minimum API any implementation should support? Something like `src/core/space_interaction_api.jl` for pathfinders

Todo:

- [x] Fix serialization
- [x] Fix tests
- [ ] Update docs
- [ ] Work on new example for multi-profile setup
- [x] Update existing maze and runners examples
- [x] Make `ContinuousSpace` PR

Benchmark:
GridSpace
```
100x100, 0 Admissibility
  10.913 s (51446799 allocations: 10.66 GiB)
100x100, 0.3 Admissibility
  2.033 s (10551657 allocations: 2.45 GiB)
```

ContinuousSpace
```
100x100, 0 Admissibility
  18.884 s (87439621 allocations: 17.39 GiB)
100x100, 0.3 Admissibility
  15.266 s (69817111 allocations: 13.68 GiB)
```


Datseris:
> Should a method kill_agent!(agent, model, pathfinder) be added? 

Yes

Datseris:
> Something like src/core/space_interaction_api.jl for pathfinders

Yes, and we already have most of it, e.g. set route, etc.

AayushSabharwal:
> Yes, and we already have most of it, e.g. set route, etc.

It looks like there really only are 3 necessary functions: `set_target!`, `move_along_route!` and `is_stationary`. The latter two are in `space_interaction_api.jl`, but adding `set_target!` there necessitates exposing a pathfinding-only function to the entire module.

AayushSabharwal:
This should cover the API rework. I'll work on an example for a multi-pathfinder scenario.

Should this PR also add `ContinuousSpace` pathfinding?

Libbum:
These changes are looking really good. Decoupling pathfinder is making everything simpler and adding functionality at the same time, so great to see. 

My only suggested change right now would be not to name `astar_grid_api.jl` that way. We have `space_interaction_API.jl` solely because its a developer template (which we also happen to extend ourselves). The astar grid api file is not really doing that, its just implementing the user facing functions for gridspace. Something simpler here (just `astar_grid` perhaps?) would be better.

Datseris:
I want to do a proper review on this and its impact on our two existing examples (maze / mountain runners)

Libbum:
In theory now there's no reason why `pathfinder` needs to be in `model.properties` - it could be separate entirely. Perhaps one of our two current examples can explore that.

AayushSabharwal:
Is there a specific reason/scenario where it would not be in model properties? While it's certainly possible to not include it there, I'm not sure why. Is there something I'm missing?

Libbum:
Users wanting to do other things than what we assume they will do. 
At the very least, in one of them we should not call the value `model.pathfinder` so it's clear we have not bound that name to the struct.

AayushSabharwal:
I've been thinking about the example to add for multiple profiles. Currently, this is the idea I've come up with:

- 3 types of agents: Rabbits, Foxes and Hawks/Eagles
- The world is on hilly terrain, with some deep valleys (unwalkable except for hawks) and mountains (peaks are unwalkable)
- The behaviour of agents is largely identical to Predator-Prey, except all movement is governed by pathfinders and hawks use different profiles. Rabbits also move toward regions with higher grass, and away from predators.

Datseris:
That sounds really cool!

Libbum:
If we made this a true 3D ContinuousSpace model and put some effort into extending InteractiveDynamics to support nice representations of this, I think it'd be a real showstopper for the conclusion of your GSoC term (and Agents as a whole).

AayushSabharwal:
That... would require some thought. I'll write something and see how it goes. For starters, I'm not sure how to enforce that rabbits/foxes stay on the land.

AayushSabharwal:
There one thing that I noticed in the predator-prey model that seems odd:
Why does `Δenergy` need to be a field of the struct if it's value is dependent only on the type of the agent? Couldn't there just be two model properties: `Δenergy_sheep` and `Δenergy_wolf`? Similarly for `reproduction_prob`?

Libbum:
> That... would require some thought

Certainly not a requirement, but maybe a stretch goal :P

> Why does `Δenergy` ...

I don't think there's any requirement, it's probably an afterthought&mdash;we have iterated on that example extensively. However, I have a slight recollection that leaving them there gave us a slight improvement in benchmarking. This one is our key comparison example.

AayushSabharwal:
I think it might be a good idea to add a `nearby_positions(pos, model, pathfinder, r = 1) = filter(x -> pathfinder.walkable[x], nearby_positions(pos, model, r))` function to the AStar API.

Datseris:
Oh, so in this example there isn't a "planning" phase for the pathfinding right? Because if you planned it, why would you need this function.

AayushSabharwal:
Exactly. The pathfinding is dynamic. It wasn't that way in the previous two examples, so I think that's a good demonstration here too.

Datseris:
Can you put the example in the actual examples list, so that it compiles and I can see it online here https://juliadynamics.github.io/Agents.jl/previews/PR520/ ?

examples/rabbit_fox_hawk.jl
```diff
@@ -0,0 +1,189 @@
...
+        rabbit_step!(animal, model)
+    elseif animal.type == :fox
+        fox_step!(animal, model)
+    else
+        hawk_step!(animal, model)
+    end
+end
+
+function rabbit_step!(rabbit, model)
+    if model.grass[rabbit.pos...]
```
Datseris, line 116: I don't like using numbers as booleans. I think it leads to confusing code. Use ` == 1` instead.

AayushSabharwal: Sure. I tend to only do this with `BitArrays`, since each bit has only two states. I'll change all these occurences.

AayushSabharwal:
This is the heightmap I'm using to test. I didn't add it here because I think it goes in JuliaDynamics/JuliaDynamics?
25x25heightmap

Datseris:
yes

AayushSabharwal:
heightmap2doc2
Using this heightmap with the (latest) default settings gives a dynamic equilibrium, albeit with foxes dying out (video below). I also found some issues with random pathing, initialization of hawks, and rabbits' avoidance of predators that I improved to make this better.

https://user-images.githubusercontent.com/23384717/124476424-69000380-ddc0-11eb-8082-ce2828e93915.mp4

I'll try and fiddle with the parameters to not get foxes extinct, and then work on making this 3D

Libbum:
Quite interesting! It would be nice to see the hawks covering a bit more of the map perhaps?

> issues with random pathing

Can you elaborate on that one?

AayushSabharwal:
> Can you elaborate on that one?

Previously, random movement was to any neighboring walkable square. This was problematic because the random movement often averaged out and agents stayed within a 2-3 tile radius, not really walking anywhere. Now they set a target to a random walkable tile they can see and go there, so they can actually move around and look for new food

Libbum:
Ah, yeah that's true. Because we're sampling from a uniform distribution. To make sense, random walks should really be a [Wiener Process](https://en.wikipedia.org/wiki/Wiener_process) or even [Ornstein–Uhlenbeck](https://en.wikipedia.org/wiki/Ornstein%E2%80%93Uhlenbeck_process).

I'm wondering if, rather than making random choices like that, this should be an example of how we integrate with [Distributions.jl](https://github.com/JuliaStats/Distributions.jl)?

AayushSabharwal:
I'll try and understand Weiner processes and see how to integrate it with the current setup. Meanwhile, I'll make the model 3D. `InteractiveDynamics` would then also need a revamp

AayushSabharwal:
From some experimentation, it seems that supporting 3D plotting is a matter of using `Axis3` instead of `Axis`, `Point3f0` instead of `Point2f0`, and `meshscatter!` instead of `scatter!`. This of course doesn't take into account heatmaps, `poly!`, etc.

AayushSabharwal:
In 3D, is it an issue if the vision of the agents doesn't respect line-of-sight? So they can see anything in the radius, irrespective of whether it is behind some terrain or not

AayushSabharwal:
I've made a plan for transitioning this to 3D.

- The heightmap should be scaled from 0-40, to produce smooth terrain (below)

image

- The land walkmap will be generated by setting `(x, y, heightmap(x, y))` to true across all `(x, y)`
- The air walkmap will be generated similarly, except all indexes above `heightmap(x, y)` will also be true
- The land pathfinder won't use a `HeightMap` metric, instead using `DirectDistance`, since the cost of going up and down terrain is covered by that metric. Additionally, as @Datseris mentioned, `HeightMap` should be renamed to `DensityMap`/`PenaltyMap` and is inapplicable here since there is no 3D density affecting movement
- Rabbit predator detection won't involve considering `z` coordinates in choosing the direction to move.
- Everything else should remain largely the same. `model.grass` will still be a matrix, since it doesn't depend on height

Libbum:
> In 3D, is it an issue if the vision of the agents doesn't respect line-of-sight? So they can see anything in the radius, irrespective of whether it is behind some terrain or not

For now that should not be an issue, but I think in the general case we need to consider if we provide a method to blind the neighbourhood based on such obstructions. Or if this is simple enough to do for a user, document that in an example.

> The heightmap should be scaled from 0-40, to produce smooth terrain (below)

Do you mean in general, or for the current example only?

> Everything else should remain largely the same. `model.grass` will still be a matrix, since it doesn't depend on height

If you want to be pedantic, there are snow covered mountains in your example :stuck_out_tongue:

AayushSabharwal:
> Or if this is simple enough to do for a user, document that in an example

For a heightmap, this should just be a matter of walking along the line connecting the two agents and checking if any part is below the heightmap

> Do you mean in general, or for the current example only?

Just for the current example, and how the 0-1 grayscale values scale here

> If you want to be pedantic, there are snow covered mountains in your example 😛 

They won't be affected, since the grassmap is only applicable to walkable regions

AayushSabharwal:
There seems to be a weird issue with Makie where using `zlims!` on an `Axis3` sets the ylims instead

AayushSabharwal:
image

How does this look?

Libbum:
Line of sight: not exactly, but yeah. Probably not something for us to worry about right now. There's observer height and elevation to consider on top of this....

occlusion culling: I *think* that was a relatively recent addition actually..

hawks rarely ever have a reason to leave the ground: Might be a good idea to have them feed only when 'hungry', otherwise circle around. When some timer runs to zero they look for food, dive down and take it, then fly back up to 'safety'.

AayushSabharwal:
I think I'll modify the heightmap to flatten out the shoreline, it looks weird right now. Also, I think one little outcrop on the side has separated the map into two regions

Datseris:
haha that is some pretty cool stuff! Sorry I'm away from this, had to finish some crucial stuff. going for vacation for a week, but I'd still recommend to not merge this until both tim and me give a proper review. You can open multiple PRs etc. in the mean time so don't let this hold you back!

AayushSabharwal:
I tried rendering a video of this using `abm_video`, but for some reason `GLMakie` just outputs noise and `CairoMakie` takes ~17s per frame. I'll try and see if there's anything I can do about this.

Datseris:
Be sure to report this at the Makie slack channel, you never know when you have encountered a bug. Makie plotting is ultra fast, so I'd be skeptical of slow plotting speeds.

AayushSabharwal:
https://discourse.julialang.org/t/makie-long-plotting-times-for-3d-plots-using-cairomakie/64255

It seems that CairoMakie is not designed nor optimized for 3D plotting, which means the long plot time will stay.

AayushSabharwal:
I've tried implementing  `abm_video` myself, but so far I'm unable to replicate the issue with `InteractiveDynamics.abm_video` outputting static with GLMakie

Libbum:
> How can we work with this in doc builds?

We had this set up with GLMakie and CairoMakie previously. It was clunky but it did work. We may have to go down that path, but not right now. We used to have 45 minute documentation builds, 17 minutes is fine!
The other alternative for huge 3D runs is that we could host the output on JuliaDynamics (like Zombies is). This is last resort territory. 

> The zlims! issue is resolved in master
We can build off master if you need it?

Starting to look pretty nice! I think for 3D plots in `abm_video` we could provide a motion tween for the camera, so throughout the video we get a pan around the action.

OK, so there may be some issue in `abm_stepper` then, we'll need to take a deeper look.

AayushSabharwal:
> We can build off master if you need it?

Right now, this isn't necessary since it requires `InteractiveDynamics` to also be updated with 3D plotting.

Libbum:
Let's start there at least. If it's a OS problem that generally points to it being outside of our control (and in the hands of the Makie devs), but at least we can try and widdle down the issue to something they can test.

AayushSabharwal:
Are our doc builds done on a linux machine? Because even 2D models fail to render videos for me, using the latest InteractiveDynamics version. It's highly possible I have some configuration issue on my machine, since OpenGL tends to give me issues

Libbum:
Oh, OK. Yes, we run linux on our builds. I'm on linux locally here too. I can verify later today (have not actually checked myself). So indeed - it may just be a local issue for yourself. Let's hold off on the rest of the plan until I can take a look.

AayushSabharwal:
I locally have a version of InteractiveDynamics that can work with 2D and 3D, without any changes to the existing API.

AayushSabharwal:
A thought that crossed my mind:
A* is a pathfinding algorithm applicable for any weighted graph. Can/should we make this applicable for `GraphSpace` too? LightGraphs already has an implementation for `a_star`, we would simply need to provide a wrapper. We could even use our existing implementation, a quick look suggests this should be possible with a few new cost metrics and some reworking of how we handle neighborhoods.

Libbum:
Cool Aayush, could you either PR that to InteractiveDynamics or show me how you're plotting your example at present? I'm not able to load your latest hightmap from the URL either, I'm getting a color conversion error. 

For graphs: certainly if it's straightforward. If A* is already there, then I would say the question of whether we use our implementation or theirs is whether or not we can provide more coverage. By that I mean: is there an example that we can showcase above and beyond what the LightGraphs version can do? I'm not sold on the need for a heightmap on a weighted graph network for example.

---

Also, I'm interested on your take of [D*](https://en.wikipedia.org/wiki/D*) and [LPA*](https://en.wikipedia.org/wiki/Lifelong_Planning_A*). Is there a need for such methods in Agents? I suspect so, in the case of finding better targets, updating paths on the fly...

AayushSabharwal:
> Cool Aayush, could you either PR that to InteractiveDynamics or show me how you're plotting your example at present?

I've made a PR: https://github.com/JuliaDynamics/InteractiveDynamics.jl/pull/60. I can't tag it or ask for reviewers since I'm not a collaborator on that repository

AayushSabharwal:
The example definitely needs a better name. I'll try and come up with something.

AayushSabharwal:
I've renamed `HeightMap` metric to `PenaltyMap`, as @Datseris suggested. Should the underlying `pmap` be accessed through `get_spatial_property` for generalizing it to `ContinuousSpace`?

src/submodules/pathfinding/astar.jl
```diff
@@ -246,18 +246,18 @@ Agents.is_stationary(
 ) where {A<:AbstractAgent} = isempty(agent.id, pathfinder)
 
 """
-    Pathfinding.heightmap(pathfinder)
-Return the heightmap of a [`Pathfinding.AStar`](@ref) if the
-[`Pathfinding.HeightMap`](@ref) metric is in use, `nothing` otherwise.
+    Pathfinding.penaltymap(pathfinder)
+Return the penaltymap of a [`Pathfinding.AStar`](@ref) if the
```
Datseris, line 250: penaltymap should be penalty map (two words) when not used in quotes `.

AayushSabharwal:
Additionally, I think this PR can also be an opportunity to reorganize things so that adding new pathfinding algorithms is easier. For example, `delta_cost` currently takes an `AStar`. I propose creating a new `abstract type AbstractPathfinder{D,Periodic,Moore} end`, which all pathfinders are subtypes of. This allows separating `delta_cost`, `position_delta`, etc. into a separate `pathfinding_utils.jl` file, and any `AbstractPathfinder` to use them.

Datseris:
I agree, but it should be `AbstractPathfinde` without type parameterization.

1. You gain literally nothing by adding the type parameterization at the abstract level.
2. If someone wanted to create pathfinding on `GraphSpace`, which I guess makes some sense, then all three type parameters are invalid.

AayushSabharwal:
Currently `delta_cost` dispatches based on whether `Moore` is true or not, and `position_delta` on `Periodic`. If the abstract type is not parametric, this won't be possible. Perhaps we introduce another `abstract type GridPathfinder{D,P,M} <: AbstractPathfinder`? The current metrics are only valid for grid spaces anyway, and a `GraphSpace` pathfinder would need its own metrics

AayushSabharwal:
Additionally, can we add GLMakie to docs builds, since 3D plotting effectively requires it? Currently the rabbit-fox-hawk is our only 3D one, but there could very well be more in the future and having a pre-rendered video for all of them could get impractical to maintain very quickly.

AayushSabharwal:
I moved the functions around, and added `GridPathfinder`. Although this abstract type hierarchy is yet to be finalized, this is a groundwork of sorts for wherever we choose to go with this.

Libbum:
> This would allow pmap to also be a function.

Definitely preferable.

> can we add GLMakie to docs builds,

Yeah, I think it's going to come down to that. I don't see a way around it. 
Will need to do some investigating, since the new Ubutnu LTS version is running on the Github CI now, and previously GLMakie was refusing to run on it. Let me handle that change though. I'll do it in a separate PR that we can merge into master and you can rebase here.

AayushSabharwal:
It doesn't look like using `get_spatial_property` with `pmap` will be possible, since `find_path` doesn't know what `model` is. Although we won't be able to use functions for `pmap`, I don't think this should be much of an issue since the pathfinding data is discretised anyway.

Libbum:
GLMakie is now on master, so 3D plots are possible. Use `GLMakie.activate!()` in the example.

Datseris:
can we merge master here please?

AayushSabharwal:
I'll make a PR to `JuliaDynamics/JuliaDynamics` with the heightmap for this example

AayushSabharwal:
Looks like `random_walkable` is still broken. I'll try and do something about it.

AayushSabharwal:
It appears that the test failures were due to floating-point errors: Sometimes the returned point is `2.0 + eps(2.0)` distance from the specified center, so the test fails. I added a tolerance to combat this.

Libbum:
Great. Always add tolerances for sure.

AayushSabharwal:
It looks like adding `GLMakie` brought back long build times.

Libbum:
There will be caching to be done, so that may fix some things after some initial settling. 

If it's less than 45 minutes then I think we'll be fine. All the more reason to start getting this Zoo up and running, then culling some of our examples.

AayushSabharwal:
Since this has already been running for nearly an hour, I think I'll hold off a bit on my commit just to check the time it takes. My commit will almost certainly add to build time, and this way we might be able to measure the effect.

AayushSabharwal:
Results of benchmarking script https://gist.github.com/AayushSabharwal/447dccac64023f863d7f42bf49772162
```
100x100, 0 Admissibility
  11.448 s (51446801 allocations: 10.66 GiB)
100x100, 0.3 Admissibility
  2.068 s (10551657 allocations: 2.45 GiB)
```

AayushSabharwal:
The most expensive operation seems to be the `collect` and `sort!` in `delta_cost` for `DirectDistance`

AayushSabharwal:
It looks like adding `GridPathfinder` as an abstract type turned `delta_cost` into a dynamic dispatch.

AayushSabharwal:
I've updated the benchmark results, and it seems as if the biggest hit to performance is that admissibility doesn't make all that much of a difference

Datseris:
Did you compare the discrete space benchmarks with the original implementation?

AayushSabharwal:
I reran the grid benchmarks with the new implementation, and it's just under half a second faster on the 0 admissibility case

Datseris:
Doc preview : https://juliadynamics.github.io/Agents.jl/previews/PR520/

src/submodules/pathfinding/all_pathfinders.jl
```diff
@@ -5,14 +5,14 @@ export Pathfinding
...
-[`Pathfinding.Pathfinder`](@ref) struct to the `pathfinder`
-parameter of the [`GridSpace`](@ref) constructor. During the simulation, call
-[`Pathfinding.set_target!`](@ref) to set the target destination for an agent.
-This triggers the algorithm to calculate a path from the agent's current position to the one
-specified. You can alternatively use [`Pathfinding.set_best_target!`](@ref) to choose the
-best target from a list. Once a target has been set, you can move an agent one step along
-its precalculated path using the [`move_along_route!`](@ref) function.
+You can enable path-finding and set it's options by creating an instance of a
+[`Pathfinding.AStar`](@ref) struct. This must be passed to the relevant pathfinding functions
+During the simulation, call [`Pathfinding.set_target!`](@ref) to set the target
```
Datseris, line 10: During does not have to be capitalized.

src/submodules/pathfinding/all_pathfinders.jl
```diff
@@ -5,14 +5,14 @@ export Pathfinding
 Submodule containing functionality for path-finding based on the A* algorithm.
 Currently available only for [`GridSpace`](@ref).
```
Datseris, line 6: Also available for `ContinuousSpace` now.

src/submodules/pathfinding/all_pathfinders.jl
```diff
@@ -5,14 +5,14 @@ export Pathfinding
...
 
-You can enable path-finding and set it's options by passing an instance of a
-[`Pathfinding.Pathfinder`](@ref) struct to the `pathfinder`
-parameter of the [`GridSpace`](@ref) constructor. During the simulation, call
-[`Pathfinding.set_target!`](@ref) to set the target destination for an agent.
-This triggers the algorithm to calculate a path from the agent's current position to the one
-specified. You can alternatively use [`Pathfinding.set_best_target!`](@ref) to choose the
-best target from a list. Once a target has been set, you can move an agent one step along
-its precalculated path using the [`move_along_route!`](@ref) function.
+You can enable path-finding and set it's options by creating an instance of a
```
Datseris, line 8: it's -> its.

src/submodules/pathfinding/all_pathfinders.jl
```diff
@@ -5,14 +5,14 @@ export Pathfinding
...
+[`Pathfinding.AStar`](@ref) struct. This must be passed to the relevant pathfinding functions
+During the simulation, call [`Pathfinding.set_target!`](@ref) to set the target
+destination for an agent. This triggers the algorithm to calculate a path from the agent's
+current position to the one specified. You can alternatively use
+[`Pathfinding.set_best_target!`](@ref) to choose the best target from a list. Once a target
+has been set, you can move an agent one step along its precalculated path using the
+[`move_along_route!`](@ref) function.
 
 Refer to the [Maze Solver](@ref) and [Mountain Runners](@ref) examples using path-finding
 and see the available functions below as well.
```
Datseris, line 18: Should also mention the 3D case.

src/submodules/pathfinding/astar_grid.jl
```diff
@@ -0,0 +1,87 @@
...
+    isnothing(path) && return
+    pathfinder.agent_paths[agent.id] = path
+end
+
+"""
+    Pathfinding.set_best_target!(agent, targets::Vector{NTuple{D,Int}}, pathfinder::AStar{D})
+Calculate and store the best path to move the agent from its current position to
+a chosen target position taken from `targets` using `pathfinder`.
+
+The `condition = :shortest` keyword retuns the shortest path which is shortest out of the
```
Datseris, line 23: The docstring here does not have `; kwargs...)` at the end.

src/submodules/pathfinding/astar_grid.jl
```diff
@@ -0,0 +1,87 @@
...
+    pathfinder::AStar{D},
+) where {D,A<:AbstractAgent}
+    path = find_path(pathfinder, agent.pos, target)
+    isnothing(path) && return
+    pathfinder.agent_paths[agent.id] = path
+end
+
+"""
+    Pathfinding.set_best_target!(agent, targets::Vector{NTuple{D,Int}}, pathfinder::AStar{D})
+Calculate and store the best path to move the agent from its current position to
```
Datseris, line 20: say: "Calculate, store, and return,"

src/submodules/pathfinding/astar_grid.jl
```diff
@@ -0,0 +1,87 @@
...
+a chosen target position taken from `targets` using `pathfinder`.
+
+The `condition = :shortest` keyword retuns the shortest path which is shortest out of the
+possible target positions. Alternatively, the `:longest` path may also be requested.
+
+Returns the position of the chosen target, or `nothing` if none of the supplied targets are
+reachable.
```
Datseris, line 27: Return `nothing` if none...

src/submodules/pathfinding/metrics.jl
```diff
@@ -33,23 +33,23 @@ struct MaxDistance{D} <: CostMetric{D} end
...
 
 """
-    Pathfinding.HeightMap(hmap::Array{Int,D} [, base_metric::CostMetric]) <: CostMetric{D}
+    Pathfinding.PenaltyMap(pmap::Array{Int,D} [, base_metric::CostMetric]) <: CostMetric{D}
 Distance between two positions is the sum of the shortest distance between them and the
-absolute difference in height. A heightmap of the same size as the corresponding
+absolute difference in penalty. A penalty of the same size as the corresponding
 [`GridSpace{D}`](@ref) is required. Distance is calculated using [`Pathfinding.DirectDistance`](@ref)
 by default, and can be changed by specifying `base_metric`. An example usage can be found in
 [Mountain Runners](@ref).
```
Datseris, line 47: Docstring of penalty map still seems limited to Discrete Space.

Datseris (commented):
I think in the 4.5 update message in `Agents.jl` you should write about the breaking changes and how to transition to the new interface.

---

In the docstring of the module `Pathfinder` add something about continuous space (i.e. that discretization is takern care of internally)

---

Question: why is the 3D case in Discrete space? The way I see it now, we have 3 examples for discrete space and 0 for continuous space. One must be made continuous.

---

I see that some functions could be unified. Even if not necessary, if we request `Pathfinding.set_target!(agent, target, pathfinder, model)` for both spaces, then there is only one function signature we need to display at the docs. SImilar for `set_best_target!`, `random_walkable`. Not sure if this is the best approach, but we should at least discuss it.

**Actually no, wait** Why is the continuous space pathfinding requiring the `model` as a fourth argument in all functions? When `Astar` is initialized, it takes as input a space, no? Can't you just store the space extent if the input space is continuous?

---

In the API docs, change the `Metrics` subheading to `Pathfinding Metrics`.

Datseris:
Did we loose some commits? I see that granularity is still a required argument:
```
AStar(
    space::ContinuousSpace{D,periodic},
    granularity::Dims{D};
    admissibility::Float64 = 0.0,
    walkable::BitArray{D} = trues(granularity),
    cost_metric::CostMetric{D} = DirectDistance{D}(),
) where {D,periodic} =
    AStar{Float64}(granularity; periodic, diagonal_movement = true, admissibility, walkable, cost_metric)

```

Anyways, it is not clear to me that there is no reason for a 4-argument method of e.g. `set_target!`. When `AStar` is initialized, it already has the continuous space extent available and can store it as its own field. Also, there is no reason for the `granularity` to even exist as an argument, it is deduced from the input walkable or penalty map.

Datseris:
@AayushSabharwal If possible for you, I'd really like to have this merged before Friday which I leave for a month, as this is the pretty much the PR that completes the submitted GSOC project, I'd prefer it to be merged while I'm present. If possible, please do go through my suggestions as soon as possible and I'll give another review. Sorry for putting pressure!!!

examples/rabbit_fox_hawk.jl
```diff
@@ -0,0 +1,410 @@
...
+    land_walkmap = BitArray(falses(dims...))
+    air_walkmap = BitArray(falses(dims...))
+    for i in 1:dims[1], j in 1:dims[2]
+        ## land animals can only walk on top of the terrain between water_level and grass_level
+        water_level < heightmap[i, j] < grass_level &&
+            (land_walkmap[i, j, heightmap[i, j]+1] = true)
+        ## air animals can fly at any height upto mountain_level
+        heightmap[i, j] < mountain_level &&
+            (air_walkmap[i, j, (heightmap[i, j]+1):mountain_level] .= true)
+    end
```
Datseris, line 89: This is more readable with standard `if` clauses. My suggestion is that if you need more than 1 lines, use a proper codeblock instead of going through the one-liner syntaxes like `&&` or `? :`

examples/rabbit_fox_hawk.jl
```diff
@@ -0,0 +1,410 @@
...
+    ## The spread of grass across the terrain is specified using a BitArray. A value of
+    ## 1 at a location indicates the presence of grass there.
+    ## The probability of grass growing is proportional to how close it is to the water. Grass
+    ## at or below water_level is ignored since it is not accessible through land_walkmap.
```
Datseris, line 93: This is very interesting. I think you should write this in the beginng of the example as normal text. Comments in the code should be reserved more for explaining code. Explaining the dynamics is better via text.

AayushSabharwal:
> Did we loose some commits? I see that granularity is still a required argument:

No, I simply haven't gotten around to addressing this yet. I was planning to, but the issues with `random_walkable` got in the way and this must have slipped my mind.

> When `AStar` is initialized, it already has the continuous space extent available and can store it as its own field

Yes! That should simplify a lot of things.

> If possible for you, I'd really like to have this merged before Friday which I leave for a month, as this is the pretty much the PR that completes the submitted GSOC project, I'd prefer it to be merged while I'm present.

Sure. I'll work on some of them tonight and the rest tomorrow. Barring extraordinary circumstances, I should be able to resolve all the comments by this time tomorrow, and quite possibly earlier.

AayushSabharwal:
> why is the 3D case in Discrete space? The way I see it now, we have 3 examples for discrete space and 0 for continuous space. One must be made continuous.

The example was made before the continuous space PR, and I myself have been considering whether to propose changing it to continuous space. I'll change it, and it should be a great example of the capabilities of this pathfinding system.

Datseris:
For the continuous example: one of the concerns was taking advantage of a `speed`. Why not make the mountain climbers continuous, and attribute different speeds to different climbers?

AayushSabharwal:
I feel like speed in mountain runners is a bit pointless, whereas having different speeds for different animals here adds value to the model.

Datseris:
Okay, I think it's easy to plot the terrain for continuous. Adjust the static_preplot to plot the surface providing rescaled ranges `surface(x, y, terrain)` with `x, y` having same total size as continuous space.

Libbum:
I'd say that in the long run mountain runners can go to the zoo. It's only added value is showcasing the heightmap.

examples/rabbit_fox_hawk.jl
```diff
@@ -0,0 +1,417 @@
...
+    surface!(ax, model.heightmap; colormap = :terrain)
+end
+
+## The sample heightmap used for this model
+heightmap_url =
+    "https://raw.githubusercontent.com/JuliaDynamics/" *
+    "JuliaDynamics/master/videos/agents/rabbit_fox_hawk_heightmap.png"
+model = initialize_model(heightmap_url)
+
+abm_video(
```
Libbum, line 400: Think limits will still need to be set for the main axis here too. The height seems to be off in the visualization.

Libbum: Also I'd perhaps drop the framerate a little (depending on how the limit fix goes). It's hard to see what's going on at present.

AayushSabharwal:
Now that `AStar` has a type parameter for the type of its path (T), shouldn't dispatch be done on that instead of whether the positions passed in are `Tuple{D,Int64}` or `Tuple{D,Float64}`? Since `ContinuousSpace` could use any `AbstractFloat`, it would also be better to support that as well.

Additionally, should the type of the `targets` parameter of `set_best_targets!` be restricted to a `Vector`? I don't think it needs to be typed at all, since dispatch can happen using `AStar`

examples/rabbit_fox_hawk.jl
AayushSabharwal: The visualization intentionally has a lot of empty space at the top, because that's actually how the terrain should look. Having a non-uniform scale of `(205, 205, 40)` was skewing how everything looked, since the axes were all the same length on the screen but of different scales.

src/spaces/utilities.jl
```diff
@@ -55,6 +55,29 @@ function edistance(p1::ValidPos, p2::ValidPos, model::ABM{<:GridSpace{D,true}})
     sqrt(total)
 end
 
+"""
+    get_direction(from, to, model::ABM)
+Return the direction vector from `from` to `to` taking into account periodicity of the space
```
Libbum, line 60: Return the direction vector from the position `from` to position `to`

since we don't explicitly say what from and to are.

OR you could add an agent wrapper here as well such that `get_direction(from.pos, to.pos, model.space)` is satisfied.

src/submodules/io/jld2_integration.jl
```diff
@@ -54,10 +54,9 @@ struct SerializableABM{S,A<:AbstractAgent,P,R<:AbstractRNG}
     maxid::Int64
 end
 
-struct SerializableGridSpace{D,P,W}
+struct SerializableGridSpace{D,P}
```
Libbum, line 57: Can we do another 'pass the jld2 model around via email' test to make sure this is working fine?

src/submodules/pathfinding/astar_continuous.jl
```diff
@@ -0,0 +1,199 @@
...
+function find_continuous_path(
+    pathfinder::AStar{D},
+    from::NTuple{D,Float64},
+    to::NTuple{D,Float64},
+    model::ABM{<:ContinuousSpace{D}}
+) where {D}
+    # used to offset positions, so edge cases get handled properly (i.e. (0., 0.) maps to grid
+    # cell (1, 1))
+    half_cell_size = model.space.extent ./ pathfinder.grid_dims ./ 2.
+    discrete_from = Tuple(Agents.get_spatial_index(from, pathfinder.walkable, model))
```
Libbum, line 16: Do we need to explicitly state this is an `Agents` call here? I think that's implied

src/submodules/pathfinding/astar_continuous.jl
```diff
@@ -0,0 +1,199 @@
...
+    discrete_to = Tuple(Agents.get_spatial_index(to, pathfinder.walkable, model))
+    discrete_path = find_path(pathfinder, discrete_from, discrete_to)
+    isnothing(discrete_path) && return
+    isempty(discrete_path) && return Path{D,Float64}()
+    cts_path = Path{D,Float64}()
```
Libbum, line 21: It'd be good to have a few comments in the code here to explain this logic. I know its all edge cases, but this kind of stuff is difficult to reason with six months from now when we forget about it.

examples/rabbit_fox_hawk.jl
Libbum: Ah. OK. That makes sense. Maybe we can let the eagles fly a little higher. Not ridiculously high, but over the mountain tops a bit to fill some of the dead space.

src/submodules/pathfinding/astar_continuous.jl
```diff
@@ -0,0 +1,199 @@
...
+    speed::Float64,
+    dt::Real = 1.0,
+) where {D,A<:AbstractAgent}
+    isempty(agent.id, pathfinder) && return
+    from = agent.pos
+    next_pos = agent.pos
+    while true
+        next_waypoint = first(pathfinder.agent_paths[agent.id])
+        dir = get_direction(from, next_waypoint, model)
+        dist_to_target = √sum(dir .^ 2)
```
Libbum, line 117: We have LinearAlgebra [imported](https://github.com/JuliaDynamics/Agents.jl/blob/b8c014db23551eb3a3519ffc7f6d03f1911988cc/src/Agents.jl#L12), there's little reason not to use it herein.

src/submodules/pathfinding/astar_continuous.jl
```diff
@@ -0,0 +1,199 @@
...
+        dir = dir ./ dist_to_target
+        next_pos = from .+ dir .* speed .* dt
+
+        # overshooting means we reached the waypoint
+        dist_to_next = edistance(from, next_pos, model)
```
Libbum, line 122: Not high on the priority list, but it'd be nice to see this in action with a single agent. Show the expected path, then the agent following the path with and without the correction.

AayushSabharwal: I'll try and make some sort of visualisation for this. The effect should mainly be visible when `speed*dt` is significantly larger than the size of the grid cells, or when the grid cells are fairly large

Libbum:
In the documentation half of the functions are explicitly labelled with `Pathfinder.` and the rest are not.
image

Since Documenter makes the module specific at the top of the section, I'd suggest dropping `Pathfinder.` in the docstrings to keep them short.

AayushSabharwal:
I have changes locally that I'll push once you've finished reviewing, so that the code you're viewing doesn't suddenly become out of date

Libbum (commented):
A few small bits and pieces, but this is close!

Make sure we add a CHANGELOG. We may as well bump the version number, add a new scratch and latest news message in this PR as well.

src/submodules/pathfinding/astar_continuous.jl
Libbum: Not sure if it needs to be anywhere in the documentation. I think it'll just be a good sanity check for us that the behavior is as fluid as we'd expect.

Libbum:
(My review is finished for now).

I may want to look at some performance stuff but will wait for things to settle down and we get some more recent numbers.

AayushSabharwal:
> Now that `AStar` has a type parameter for the type of its path (T), shouldn't dispatch be done on that instead of whether the positions passed in are `Tuple{D,Int64}` or `Tuple{D,Float64}`? Since `ContinuousSpace` could use any `AbstractFloat`, it would also be better to support that as well.
> 
> Additionally, should the type of the `targets` parameter of `set_best_targets!` be restricted to a `Vector`? I don't think it needs to be typed at all, since dispatch can happen using `AStar`

Can I lift the type restriction on `from`, `to`, etc. everywhere as long as its `NTuple{D}`? Or should it be kept?

AayushSabharwal:
I reran the benchmarks and put the numbers in the first comment of this thread, so they're easy to find. I'll keep rerunning when it is necessary

examples/rabbit_fox_hawk.jl
AayushSabharwal: In that case, I'll make it so that eagles can fly over any terrain, instead of being restricted by `mountain_level`

Datseris:
> Since Documenter makes the module specific at the top of the section, I'd suggest dropping Pathfinder. in the docstrings to keep them short.

Please do not do this. A function call signature should reflect exactly how the function is called. All of these functions are called with the preface `Pathfinder.whatever`. As a user I never _ever_ check the automated documenter header, because I never ever care from which module a function is coming from. And here I would also not care where the function is coming from, I only care about how to call it.

I'm willing to bet big money that if you drop `Pathfinder` from a preface, we will be getting so many issues/questions saying "ooooh I get error `set_target!` is undefined".

AayushSabharwal:
I've noticed there's some repetition in the code base, specifically around utility functions. For example, `position_delta`, `sqr_distance` (in `astar_continuous.jl`), `edistance` and `get_direction` all are similar in their functioning, just take different parameters necessitating the use of different functions. Similarly `get_spatial_index` and `to_discrete_position` (in `astar_continuous.jl`) Is there anything we can/should do to address this?

AayushSabharwal:
Additionally, is there anything we can do to make `get_spatial_property` work in the rabbit-fox-hawk example? `grass` is distributed in 2 dimensions, whereas the space is 3 dimensional. This prevents usage of `get_spatial_property`.

AayushSabharwal:
I've come across something in the example I can't (easily) replicate in continuous space: Rabbits avoid predators by finding the optimal direction away from all predators, then finding the closest walkable position to it. Since I can't iterate over nearby positions, there really isn't a way to do this. Should I add a utility function that returns the closest walkable location to a given location. It wouldn't be particularly efficient, since I'd have to iterate over the entire grid.

Libbum:
> Please do not do this.  [...]

Fair enough point. Aayush, make sure the opposite of what I said is sorted out then :stuck_out_tongue: 

I'll have to dig a bit into the other questions Aayush, which I don't have time for right away, will come back to you.

For the last one, I'm not sure this needs to be a big deal - you're the one making the rules on the example so they can be changed. I'd also say that rabbits are not continuously looking at the sky, so optimally furthest away from a predator is not representative of the real world regardless.

AayushSabharwal:
I see your point. I _think_ I have a way to make this work for the current situation, since I can use the fact that it's a heightmap to my advantage. I'll have to tinker around with things and see how it plays out.

AayushSabharwal:
I have a working version of the example in continuous space, but it's extremely slow to run. Time per step varies from 0.5s to 3.5s

AayushSabharwal:
The model is runnable now. The range of time per step is approximately the same, but skewed toward the lower end now

Datseris:
> I've noticed there's some repetition in the code base, specifically around utility functions

Avoiding code repetition is an absolute top priority for Agents.jl!

> Additionally, is there anything we can do to make get_spatial_property work in the rabbit-fox-hawk example? grass is distributed in 2 dimensions, whereas the space is 3 dimensional. This prevents usage of get_spatial_property.

Oh nooooo that's my fault. I already thought of how to design the function so this is not possible, but I didn't do it!!! I'll do it when I'm back home tonight in about 8 hours...

AayushSabharwal:
The model should run fairly well now. `set_target!` is definitely slow, and in continuous space since each timestep effectively progresses the simulation by a smaller amount, the calls need to be throttled as well. There is still some temporary code there, and the comments need reworking at places, which I'll address in some time.

AayushSabharwal:
This example is increasingly becoming a case in point of when not to use A*, especially in continuous space,

examples/rabbit_fox_hawk.jl
AayushSabharwal: This should look a bit better now

AayushSabharwal:
I think the example is in a good state now. I'll write up the update messages, both in the docs and the scratchspaces one.

CHANGELOG.md
```diff
@@ -1,4 +1,17 @@
-# v4.4 (UNRELEASED)
+# v4.5 (UNRELEASED)
+## New features and fixes
+- Add `get_spatial_property` and `get_spatial_index` for easier usage of spatially distributed properties in `ContinuousSpace`.
+- Rework the pathfinding system to be more streamlined and offer greater control over the its details. `AStar` structs are now created, instead of the old `Pathfinder` and are stored by the
```
Libbum, line 4: Just the first sentence is fine here, drop the other two.

examples/rabbit_fox_hawk.jl
Libbum: Tough this one. Slowing it down it's now easier to see that nothing is really happening. I see a lot of rabbits dying but not really why. Will play around a bit with it.

Datseris:
I'll give a final review in a couple of hours

examples/rabbit_fox_hawk.jl
```diff
@@ -0,0 +1,428 @@
...
+# walk on suitable portions of the map, hawks are capable of flight and can fly over a much
+# larger region of the map.
+#
+# Similar to the [Predator-Prey](@ref) example, agent types are distinguished using a `type`
+# field. Agents also have an additional `energy` field, which is consumed to move around and
+# reproduce. Eating food (grass or rabbits) replenishes `energy` by a fixed amount.
+using Agents, Agents.Pathfinding
+using Random
+using FileIO
+using ImageMagick # hide
```
Libbum, line 22: Not needed.

examples/rabbit_fox_hawk.jl
```diff
@@ -0,0 +1,428 @@
...
+    regrowth_chance = 0.03,  ## probability that a patch of grass regrows at any step
+    dt = 0.1,   ## discrete timestep each iteration of the model
+    seed = 42,  ## seed for random number generator
+)
+
+    ## Download and load the heightmap. The grayscale value is converted to `Float64` and
+    ## scaled from 1 to 40
+    heightmap = floor.(Int, convert.(Float64, load(download(heightmap_url))) * 39) .+ 1
+    ## The x and y dimensions of the pathfinder are that of the heightmap
+    dims = (size(heightmap)..., 100)
```
Libbum, line 83: If you change the last value here to 50 and the z dimension in space to 50.0, then the plot looks more reasonable.

CHANGELOG.md
```diff
@@ -1,4 +1,16 @@
...
+## New features and fixes
+- Add `get_spatial_property` and `get_spatial_index` for easier usage of spatially distributed properties in `ContinuousSpace`.
+- Rework the pathfinding system to be more streamlined and offer greater control over the its details.
+- Add support for pathfinding in `ContinuousSpace`.
+- New utility functions `nearby_walkable` and `random_walkable` for use in models with pathfinding.
+- Fixed bug where there was no differentiation between empty paths and paths to unreachable nodes.
+
+## Deprecated
+- The old pathfinding system is now deprecated. Pathfinding structs are not saved as part of the
+  space, and instead are stored by the user.
```
Datseris, line 11: This seems incorrect. It is not deprecated. It will error. We need "BREAKING" here.

CHANGELOG.md
```diff
@@ -1,4 +1,16 @@
-# v4.4 (UNRELEASED)
+# v4.5 (UNRELEASED)
+## New features and fixes
+- Add `get_spatial_property` and `get_spatial_index` for easier usage of spatially distributed properties in `ContinuousSpace`.
+- Rework the pathfinding system to be more streamlined and offer greater control over the its details.
+- Add support for pathfinding in `ContinuousSpace`.
+- New utility functions `nearby_walkable` and `random_walkable` for use in models with pathfinding.
+- Fixed bug where there was no differentiation between empty paths and paths to unreachable nodes.
```
Datseris, line 7: Add a smaller version of this in the start of `index.md` in the dosc, where we say latest news.

docs/src/index.md
```diff
@@ -8,8 +8,11 @@ To get started, please read the [Tutorial](@ref) page.
     If you have found this package useful, please consider starring it on [GitHub](https://github.com/JuliaDynamics/Agents.jl).
     This gives us an accurate lower bound of the (satisfied) user count.
 
-!!! tip "Latest news: Agents.jl v4.4"
-      * New functions `save_checkpoint` and `load_checkpoint` to save and load entire models to and from files
+!!! tip "Latest news: Agents.jl v4.5"
+      * New functions [`get_spatial_property`](@ref) and [`get_spatial_index`](@ref) for easier usage of spatially distributed properties in [`ContinuousSpace`](@ref).
+      * Rework of the pathfinding API to offer greater control over pathfinding, including the possiblity of multiple pathfinding profiles in the same model.
```
Datseris, line 13: Say "Breaking rework"

examples/rabbit_fox_hawk.jl
```diff
@@ -0,0 +1,428 @@
+# # 3D Predator-prey
+# ```@raw html
+# <video width="auto" controls autoplay loop>
+# <source src="../rabbit_fox_hawk.mp4" type="video/mp4">
+# </video>
+# ```
+# This model is a variation on the [Predator-Prey](@ref) example. It uses a 3-dimensional
+# [`GridSpace`](@ref), a realistic terrain for the agents, and pathfinding (with multiple
```
Datseris, line 8: ContinuousSpace now! (Also check that you don't say GridSpace in the rest of the example)

examples/rabbit_fox_hawk.jl
```diff
@@ -0,0 +1,428 @@
...
+        )
+    end
+
+    return model
+end
+
+# We create a utility function that takes the 3D position of an agent, and returns the
+# corresponding 2D position on the heightmap
+to_heightmap_index(pos, model) =
+    floor.(Int, pos[1:2] ./ size(model.space)[1:2] .* size(model.heightmap)) .+ 1
```
Datseris, line 177: I just pushed to `master` branch an update to `get_spatial_index` that will make this not necessary.

src/submodules/pathfinding/astar.jl
```diff
@@ -0,0 +1,239 @@
...
+For [`ContinuousSpace`](@ref), a walkmap or instance of [`PenaltyMap`](@ref) must be provided
+to specify the level of discretisation of the space.
+
+## Keywords
+- `diagonal_movement = true` specifies if movement can be to diagonal neighbors of a
+  tile, or only orthogonal neighbors. Only available for [`GridSpace`](@ref)
+- `admissibility = 0.0` allows the algorithm to aprroximate paths to speed up pathfinding.
+  A value of `admissibility` allows paths with at most `(1+admissibility)` times the optimal
+  length.
+- `walkable = trues(size(space))` specifies the (un)walkable positions of the
```
Datseris, line 57: What's going on here? This is so confusing, why are the call signatures so different between the two spaces??? In one it is called `walkmap` in the other it is called `walkable`. One needs them as a positional argument, the other as a keyword. Why? Make everything keyword, the same keyword, and just add a test inside that only one of the two is given. Set default `walkable` to `nothing` and the default cost_metirci as is (DirectDistance). Then check that either walkable is not nothing, or that the cost metric is PenaltyMap and say "one of these two, and onlky one, must be true". 

Also, the way the docstring is written now, it is confusing whether continuous and discrete spaces are truly the same.

Datseris (commented):
We should open a TODO issue to check performance of pathfinding. This amount of allocations seems pretty high!

src/submodules/pathfinding/astar.jl
AayushSabharwal: I'll work on changing the constructor. There is some naming conflict with `walkable`/`walkmap`, specifically because the struct field is called `walkable` and the function that exposed it in the original API was called `walkmap`. To this end, I think it's better if we keep the name as `walkmap` everywhere

AayushSabharwal: Also, wouldn't this remove the ability to have a penaltymap as well as a walkmap? For example, this would make a 2D version of rabbit-fox-hawk impossible. We could just check that their dimensions agree.

Datseris: Okay sorry I had to say "at least one of these two must be true".

AayushSabharwal:
I just realised tests won't pass, I'll update them with the `walkmap` nomenclature

examples/rabbit_fox_hawk.jl
AayushSabharwal: Yeah, I've been trying to find a workable combination of parameters too

AayushSabharwal: I think they're dying because they were affected the worst by the transition to continuous space. They only roam around randomly, so their chances of stumbling upon food is lower, and their avoidance of predators is also fairly random. It looks like they tend to die a lot after getting corralled into corners and eaten.

AayushSabharwal:
> We should open a TODO issue to check performance of pathfinding. This amount of allocations seems pretty high!

I've been trying to figure this out too. I have a few ideas on how to combat this, but it all comes down to trying it out first.

Datseris (approved):
For me this is good to go.

AayushSabharwal:
I'll create and mail another JLD2 file so we can do the test that @Libbum suggested for verifying that serialization still works across systems

AayushSabharwal:
I missed a spot

Libbum:
I'll look at that now

Libbum:
Checkpoint file worked perfectly!

---

I'd say that our example has some quirks but does the job for the moment. We can edit that as we like over the coming weeks. It's not a blocker.

---

> I've noticed there's some repetition in the code base, specifically around utility functions. For example, position_delta, sqr_distance (in astar_continuous.jl), edistance and get_direction all are similar in their functioning, just take different parameters necessitating the use of different functions. Similarly get_spatial_index and to_discrete_position (in astar_continuous.jl) Is there anything we can/should do to address this?

This is also not a blocker for merging this branch, but we may as well clean this up before releasing 4.5.

I think we can merge once the docs building now have been looked over and anything else can be done in separate PRs. 

 @Datseris, sound like a plan?

src/submodules/io/jld2_integration.jl
Libbum: All good.

examples/rabbit_fox_hawk.jl
Libbum: Let's iterate on the example outside of this PR.

src/submodules/pathfinding/astar_continuous.jl
Libbum: Outside of this PR as well I think.

Datseris:
I vote to merge **and** release 4.5. All outstanding issues are non blockers for minor release.

Libbum (approved):
Go for merge!

@AayushSabharwal, I have some local fixes that I need to push after this, so once it's in I'll get you to review them before releasing.

AayushSabharwal:
Should we wait for docs?

Libbum:
*sigh*, yeah I guess so.

Might be useful to turn off GLMakie on maze and runners if it's going to blow out like this.

AayushSabharwal:
Yes. Considering it's taking so long, I guess it's best to use it only where necessary.

Once docs build, should I switch them to Cairo before merging?

Datseris:
If GLMakie.jl indeed takes so long, we could consider saving the video output to the JuliaDynamics web and never actually using GLMakie...? Deffinitely not optimal, but if it's a last resolrt it isn't that bad.

AayushSabharwal:
Right now we only have one 3D example, so this is still an option I guess. Why does GLMakie take so long anyway? It's pretty quick on my machine, after the first plot.

AayushSabharwal:
The link to rabbit-fox-hawk is broken in index.md

AayushSabharwal:
Considering rabbit, fox, hawk adds around 15 minutes to docs build time, I'd expect that reverting maze and runners would cut 25-30 minutes off the build time.

AayushSabharwal:
Everything works now. Builds are 15 minutes faster.

Libbum:
The reason why GLMakie takes so much time is that it's emulating your GPU work on the CPU. If we want to use it that's the price we'll pay.

I'd say let's leave it for now since we're considering a reshuffle of examples in the near future anyway. We'll see how much it annoys us in the meantime.

Looks good to merge.

AayushSabharwal:
Should I tag a release?

Libbum:
Not yet!

PR #526, opened 2021-07-11
Spatial properties in continuous space

### LINKED ISSUE #524: Utilities for spatially distributed properties
Having a property spatially distributed over `GridSpace` is trivial, but not so much with `ContinuousSpace` (such as in #519). This is especially because it requires converting between coordinates repeatedly.

This would also come into picture for implementing `ContinuousSpace` pathfinding, since `walkable` and `heightmap` would need to be discretised. The path would also be in grid coordinates, and then would need to be converted back.

**Describe the solution you'd like**
It would be useful to have some features that enable such spatially distributed properties easily. I have two ideas for this:
- Functions such as `to_grid_coordinates(continuous_pos, continuous_space_size, grid_size)` (and the inverse?)
- A wrapper around `AbstractArray` that also stores the `ContinuousSpace` dimensions and can be indexed with continuous coordinates, converting them to the appropriate grid cell coordinates

I prefer the second approach, since it's much cleaner to interact with and can just be dropped into `model.properties`. It also allows the relevant `AStar` properties to be `<:AbstractArray` and agnostic of whatever is dropped in, which could even be `AbstractArray` wrappers from other packages.

Datseris:
Please provide suggestions on what examples this can be incorporated into; I'd like to crosslink one in the `get_spatial_field` docstring.

Closes #524


src/deprecations.jl
```diff
@@ -0,0 +1,27 @@
...
+    "deprecated, provide it as a keyword instead."
+    return ContinuousSpace(extend; spacing, kwargs...)
+end
+
+
+function __init__()
+    # Plot recipes
+    @require Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80" begin
+        include("visualization/plot-recipes.jl")
+    end
```
Libbum, line 16: Missing the Documenter require block

src/spaces/continuous.jl
```diff
@@ -11,32 +11,39 @@ Base.eltype(s::ContinuousSpace{D,P,T,F}) where {D,P,T,F} = T
...
-during the agent interaction, the `update_vel!` functionality targets spatial force
-fields acting on the agents individually (e.g. some magnetic field).
-By default no update is done this way.
-If you use `update_vel!`, the agent type must have a field `vel::NTuple{D, <:Real}`.
-
-There is no "best" choice for the value of `spacing`. If you need optimal performance it's
-advised to set up a benchmark over a range of choices. The value matters most when searching
-for neighbors. In [`Models.flocking`](@ref) for example, an optimal value for `spacing` is
-66% of the search distance.
+while it is strongly recommend that agents also have a field `vel::NTuple{D, <:Real}` to use
```
Libbum, line 17: while -> and

src/spaces/continuous.jl
```diff
@@ -11,32 +11,39 @@ Base.eltype(s::ContinuousSpace{D,P,T,F}) where {D,P,T,F} = T
...
+in conjunction with [`move_agent!`](@ref). Use [`ContinuousAgent`](@ref) for convenience.
+
+`ContinuousSpace` is a _true_ representation of agent dynamics on a continuous medium.
+There are no "patches". Agent position, orientation, and speed, are true floats.
```
Libbum, line 21: > There are no "patches".

I'd probably drop this from our docstring. I plan to write some documentation around 'Moving from framework X to Agents.jl' at some stage, this kind of thing should be there. Patches have no context outside of Netlogo regardless.

src/spaces/continuous.jl
```diff
@@ -11,32 +11,39 @@ Base.eltype(s::ContinuousSpace{D,P,T,F}) where {D,P,T,F} = T
...
+In addition, strong support is provided for representing spatial properties in a model
+that contains a `ContinuousSpace`. Spatial properties (which typically are contained in 
+the model properties) can either be functions of the position vector, `f(pos) = value`,
+or `AbstractArrays`, representing discretizations of 
+spatial data that may not be available as analytic forms. In the latter case,
```
Libbum, line 26: as analytic forms -> in analytic form

src/spaces/continuous.jl
```diff
@@ -189,7 +196,7 @@ function Base.show(io::IO, space::ContinuousSpace{D,P}) where {D,P}
 end
 
 #######################################################################################
-# Continuous space exclusive
+# Continuous space exclusive: agent interctions
```
Libbum, line 199: interactions?

src/spaces/continuous.jl
```diff
@@ -410,3 +417,34 @@ function Base.iterate(iter::PairIterator, i = 1)
...
+end
+
+function _convert_pos_to_cell(pos, spacesize, propertysize)
+    εs = spacesize ./ propertysize
+    idxs = floor.(Int, pos ./ εs) .+ 1
+    return CartesianIndex(idxs)
+end
+
+"""
+    get_spatial_property(pos::NTuple{D, Float64}, property::Any, model::ABM)
```
Libbum, line 447: Should `property` really be `Any` here, or `Function`?

src/deprecations.jl
Datseris: I don't understrand what you mean.

src/spaces/continuous.jl
Datseris: `Any`. Function-like objects are a huge staple of Julia, and it is not always the case that the creator would have subtyped them as `Function`. And in some cases this isn't even possible.

src/deprecations.jl
Libbum: Added in 1a8444f

Libbum (approved):
For an example, I wonder if it'd be worthwhile to reconsider daisyworld. The temperature data lies in the model and it has a complex relationship with the agents.

Datseris:
but daisyworld is discrete model.

Datseris:
I'm merging this, but we should think of some way to incorporate it to an example. I vote to modify the existing antworld candidates to use this effectively.

Datseris:
Actually I'll quickly push the change that the function form call is `f(pos, model)`, so that it can re-use existing model parameters.

PR #531, opened 2021-07-15
Continuous Space Pathfinding

AayushSabharwal:
None of the new methods are tested, and the docstrings are just copied over from `astar_grid.jl`. ~I'll work on getting `move_along_route!` working for periodic continuous spaces before addressing that~ (done). Some points to address:
- `set_target!` and `set_best_target!` for continuous space require the model to be passed in, for continuous space dimensions
- I chose to keep `find_path` as it is and add a new `find_continuous_path` method. The alternative was having `find_path` return `parent` and each `set_target!` method extract the path from that, possibly through another method such as `unravel_path`.

TODO:
- [x] Revert `ci.yml`
- [x] Tests
- [x] Docs
- [x] Example

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ## New features and fixes
 - Provide a generator function to collect `mdata` in `run!` and `ensemblerun!`.
 - Save/load entire models using `save_checkpoint` and `load_checkpoint`
+- New functions `get_spatial_property` and `get_spatial_index` that allows better handling of spatial fields present in `ContinuousSpace` that are represented via the forms of discretization over the space.
 
 # v4.3
 ## New features and fixes
--- a/docs/src/comparison.md
+++ b/docs/src/comparison.md
@@ -14,7 +14,7 @@ The results are characterised in two ways: how long it took each model to perfor
 
 Time taken is presented in normalised units, measured against the runtime of Agents.jl. In other words: the results do not depend on any computers specific hardware. If one wishes to repeat the results personally by using the scripts in the [ABM_Framework_Comparisons](https://github.com/JuliaDynamics/ABM_Framework_Comparisons) repository: they will compute the same results. For details on the parameters used for each comparison, see the `benchmark.jl` file in that repository.
 
-For LOC, we use the following convention: code is formatted using standard practices & linting for the associated language. Documentation strings and in-line comments (residing on lines of their own) are discarded, as well as any benchmark infrastructure. NetLogo is assigned two values since its files have a code base section and an encoding of the GUI. Since many parameters live in the GUI, we must take this into account. Thus `375 (785)` in a NetLogo count means 375 lines in the code section, 785 lines total in the file.
+For LOC, we use the following convention: code is formatted using standard practices & linting for the associated language. Documentation strings and in-line comments (residing on lines of their own) are discarded, as well as any benchmark infrastructure. NetLogo is assigned two values since its files have a code base section and an encoding of the GUI. Since many parameters live in the GUI, we must take this into account. Thus `375 (785)` in a NetLogo count means 375 lines in the code section, 785 lines total in the file. An additional complication to this value in NetLogo is that it stores plotting information (colours, shapes, sizes) as agent properties, and as such the number outside of the bracket may be slightly inflated.
 
 | Model/Framework | Agents 4.2 | Mesa 0.8| Netlogo 6.2 | MASON 20.0 |
 |---|---|---|---|---|
@@ -25,7 +25,7 @@ For LOC, we use the following convention: code is formatted using standard pract
 |Forest Fire|1|125.6x|53.0x|NA|
 |(LOC)|23|35|43 (545)|.|
 |Schelling|1|29.4x|8.0x|14.3x|
-|(LOC)|31|56|68 (732)|248|
+|(LOC)|31|56|60 (743)|248|
 
 ᕯ Netlogo has a different implementation to the other three frameworks here. It cheats a little by only choosing one nearest neighbor in some cases rather than considering all neighbors within vision. So a true comparison would ultimately see a slower result.
 
@@ -33,7 +33,7 @@ The results clearly speak for themselves. Across all four models, Agents.jl's pe
 
 ## Table-based comparison
 
-In an our [paper discussing Agents.jl](https://arxiv.org/abs/2101.10072), we compiled a comparison over a large list of features and metrics from the four frameworks discussed above.
+In our [paper discussing Agents.jl](https://arxiv.org/abs/2101.10072), we compiled a comparison over a large list of features and metrics from the four frameworks discussed above.
 They are shown below in a table-based format:
 
 ![Table 1](assets/table1.png)
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -1,7 +1,6 @@
 module Agents
 
 using Requires
-using Scratch
 using Distributed
 using DataStructures
 using LightGraphs
@@ -37,48 +36,35 @@ include("simulations/ensemblerun.jl")
 include("submodules/pathfinding/all_pathfinders.jl")
 include("submodules/schedulers.jl")
 include("submodules/io/AgentsIO.jl")
-
-function __init__()
-    # Plot recipes
-    @require Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80" begin
-        include("visualization/plot-recipes.jl")
-    end
-    # Workaround for Documenter.jl, so we don't need to include
-    # heavy dependencies to build documentation
-    @require Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4" begin
-        include("visualization/plot-recipes.jl")
-    end
-    # Update message:
-    display_update = true
-    version_number = "4.4"
-    update_name = "update_v$(version_number)"
-    
-    if display_update
-        # Get scratch space for this package
-        versions_dir = @get_scratch!("versions")
-        if !isfile(joinpath(versions_dir, update_name))
-            printstyled(
-                stdout,
-                """
-                \nUpdate message: Agents v$(version_number)
-                Please see the changelog online. Some key features:
-    
-                * Agent data can be loaded from and saved to CSV files using `populate_from_csv!` and `dump_to_csv`
-                * Support for saving and loading entire models using `save_checkpoint` and `load_checkpoint`
-                """;
-                color = :light_magenta,
-            )
-            touch(joinpath(versions_dir, update_name))
-        end
-    end
-end
-
-# Deprecations, that will be removed in future versions
-include("deprecated.jl")
+include("deprecations.jl")
 
 # Predefined models
 include("models/Models.jl")
 export Models
 
+# Update messages:
+using Scratch
+display_update = true
+version_number = "4.4"
+update_name = "update_v$(version_number)"
+
+if display_update
+    # Get scratch space for this package
+    versions_dir = @get_scratch!("versions")
+    if !isfile(joinpath(versions_dir, update_name))
+        printstyled(
+            stdout,
+            """
+            \nUpdate message: Agents v$(version_number)
+            Please see the changelog online. Some key features:
+
+            * Agent data can be loaded from and saved to CSV files using `populate_from_csv!` and `dump_to_csv`
+            * Support for saving and loading entire models using `save_checkpoint` and `load_checkpoint`
+            """;
+            color = :light_magenta,
+        )
+        touch(joinpath(versions_dir, update_name))
+    end
+end
 
 end # module
--- a/src/spaces/utilities.jl
+++ b/src/spaces/utilities.jl
@@ -57,7 +57,7 @@ end
 
 """
     get_direction(from, to, model::ABM)
-Returns the direction vector from `from` to `to` taking into account periodicity of the space
+Return the direction vector from `from` to `to` taking into account periodicity of the space
 """
 get_direction(from, to, model::ABM) = get_direction(from, to, model.space)
 
--- a/src/spaces/utilities.jl
+++ b/src/spaces/utilities.jl
@@ -1,4 +1,4 @@
-export edistance, walk!
+export edistance, get_direction, walk!
 
 #######################################################################################
 # %% (Mostly) space agnostic helper functions
@@ -53,6 +53,26 @@ function edistance(p1::ValidPos, p2::ValidPos, model::ABM{<:GridSpace{D,true}}) 
         total += delta^2
     end
     sqrt(total)
+end
+
+"""
+    get_direction(from, to, model::ABM)
+Returns the direction vector from `from` to `to` taking into account periodicity of the space
+"""
+get_direction(from, to, model::ABM) = get_direction(from, to, model.space)
+
+function get_direction(from::NTuple{D,Float64}, to::NTuple{D,Float64}, space::ContinuousSpace{D,true}) where {D}
+    all_dirs = [to .+ space.extent .* (i, j) .- from for i in -1:1, j in -1:1]
+    return all_dirs[argmin(map(x -> sum(x .^ 2), all_dirs))]
+end
+
+function get_direction(from::NTuple{D,Int64}, to::NTuple{D,Int64}, space::GridSpace{D,true}) where {D}
+    all_dirs = [to .+ size(space.s) .* (i, j) .- from for i in -1:1, j in -1:1]
+    return all_dirs[argmin(map(x -> sum(x .^ 2), all_dirs))]
+end
+
+function get_direction(from, to, ::Union{GridSpace,ContinuousSpace})
+    return to .- from
 end
 
 """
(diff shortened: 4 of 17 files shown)


AayushSabharwal:
Does CI not run if the PR isn't to master?

Libbum:
[Looks like it](https://github.com/JuliaDynamics/Agents.jl/blob/4362c80b35c4628ba6be7ee5da239e29c3a4c16e/.github/workflows/ci.yml#L5)...

Libbum:
Feel free to add `pathfinding_rework` into that file for this branch. We'll delete that change before we merge.

AayushSabharwal:
Tests still won't pass right now. It looks like continuous space needs a better metric of length of a path over just counting the number of nodes

Datseris:
> There was an issue with converting to discrete coordinates 

perhaps you can take advantage of the new `get_spatial_index` function I wrote?

AayushSabharwal:
> perhaps you can take advantage of the new `get_spatial_index` function I wrote?

Ah yes, I didn't notice that. I ended up doing the same thing on my own. It's probably a good idea to use `get_spatial_index` and cast the result to `Tuple`, for consistency. I'll merge master into this branch and use it.

AayushSabharwal:
Currently the pathfinding methods all use `::NTuple{D,Float64}` as the type for accepting positions. I noticed that `get_spatial_property` doesn't have such a restriction, and neither do the API methods for `ContinuousSpace`. Should this type restriction be removed from the `astar_continuous.jl` methods as well?

AayushSabharwal:
How should the functions in `astar_continuous.jl` be documented? `move_along_route!` needs its own docstring, but the other two differ only in function signature. Repeating the same or similar docstring seems redundant. So far the options I've come up with are:

- `set_target!` and `set_best_target!` in `astar_continuous.jl` get a one-liner docstring simply saying that this is the function signature for pathfinding in continuous space. `move_along_route!` has its own docstring
- Add a new file `astar_space_api.jl`, which contains definitions such as `set_target!() = nothing` and only serves the purpose of having a docstring. The docstring contains both function signatures, and in the case of `move_along_route!` an additional section for describing the new parameters.

AayushSabharwal:
I also ended up creating a `get_direction` vector for obtaining the vector from one point to another in `ContinuousSpace`, taking into account periodicity. This could potentially be useful outside of pathfinding too. Should it be added as a continuous space exclusive, or some other relevant part of the public API?

Libbum:
> Currently the pathfinding methods all use ::NTuple{D,Float64} as the type for accepting positions. I noticed that get_spatial_property doesn't have such a restriction, and neither do the API methods for ContinuousSpace. Should this type restriction be removed from the astar_continuous.jl methods as well?

What you're using there is fine that's what a position should be in continuous space, so you accept it. There's a fine line between restricting too much, not restricting enough, and being pedantic about all the little helpers. Generally speaking you should provide general methods when you can, only restrict when you know something like `::NTuple{D,Int}` wouldn't work, or it helps dispatch.

I wouldn't worry about doing anything here at the moment.

> How should the functions in astar_continuous.jl be documented?

Just add docstrings where you think they need to be. If the method is exported, Documenter will sort it out. [Take a look at add_agent!](https://juliadynamics.github.io/Agents.jl/stable/api/#Agents.add_agent!) for example: `add_agent!` from two different places are added together.

If you think it's better to have something like [`kill_agent!`](https://juliadynamics.github.io/Agents.jl/stable/api/#Agents.kill_agent!), then in principle you could use your `astar_space_api.jl` trick, but if this is just for docs I don't think it's a good idea. Better to go with your first choice and just make a simple docstring.

> get_direction ...  Should it be added as a continuous space exclusive?

Yes, that'd be a good place.

src/submodules/pathfinding/astar_continuous.jl
```diff
@@ -0,0 +1,154 @@
...
+end
+
+"""
+    get_direction(from, to, space)
+Returns the direction vector from `from` to `to` taking into account periodicity of the space
+(for continuous space)
+"""
+# TODO: Dispatch on AStar's periodicity, since it's _technically_ possible for it to be different
+# from the space
+function get_direction(from::NTuple{D,Float64}, to::NTuple{D,Float64}, space::ContinuousSpace{D,true}) where {D}
```
Libbum, line 97: If this is to be a ContinuousSpace exclusive, may want to add a `get_direction(from, to, model) = get_direction(from, to, model.space)` wrapper.

AayushSabharwal:
Considering `edistance` exists for `GridSpace` too, I added methods for `get_direction` in `GridSpace` as well. I'm not sure where to put it in docs, though

src/spaces/utilities.jl
```diff
@@ -55,6 +55,26 @@ function edistance(p1::ValidPos, p2::ValidPos, model::ABM{<:GridSpace{D,true}})
     sqrt(total)
 end
 
+"""
+    get_direction(from, to, model::ABM)
+Returns the direction vector from `from` to `to` taking into account periodicity of the space
```
Datseris, line 60: Returns -> Return, use imperative.

src/spaces/utilities.jl
```diff
@@ -55,6 +55,26 @@ function edistance(p1::ValidPos, p2::ValidPos, model::ABM{<:GridSpace{D,true}})
...
+end
+
+function get_direction(from::NTuple{D,Int64}, to::NTuple{D,Int64}, space::GridSpace{D,true}) where {D}
+    all_dirs = [to .+ size(space.s) .* (i, j) .- from for i in -1:1, j in -1:1]
+    return all_dirs[argmin(map(x -> sum(x .^ 2), all_dirs))]
+end
+
+function get_direction(from, to, ::Union{GridSpace,ContinuousSpace})
+    return to .- from
+end
```
Datseris, line 76: This seems like a rather inefficient implementation. Why do you have to make so many vectors? You make a vector with `all_dirs`, you make another vector with `map`. Two alternatives:

1. Loop over possibilities and simply keep track of minimum
2. Just use the `round` function to get the best/proper direction instead of `argmin` of all directions. This doesn't really make sense.

Also, the way this is written, it explicitly assumes 2D-space, yet any dimensional space can be given as an input.

Libbum:
> I added methods for get_direction in GridSpace as well. I'm not sure where to put it in docs, though

Maybe moving agents, together with walk!?

src/spaces/utilities.jl
AayushSabharwal: Thanks for pointing this out! It looks like I had 2D space in mind when making this and didn't generalize it. I'll rework this function

AayushSabharwal:
Is there any reason `size(space)` isn't defined for `ContinuousSpace`? It would enable avoiding repetition of functions, such as with `get_direction` where nearly the same function is repeated just because the size of the space is obtained differently

Datseris:
just go ahead and define it

AayushSabharwal:
I think docs look good, maybe a little repetitive. There's a couple things I want to test, and then I'll mark this as ready for review

AayushSabharwal:
`nearby_positions` can be used for a random walk in `GridSpace`, but there is currently no equivalent in `ContinuousSpace` if a pathfinder is used. What if `nearby_positions` returned a random position in a random walkable cell of the pathfinder?

Datseris:
>  Is there any reason for it not to be another method of nearby_positions?

That sounds fine to me.

> nearby_positions can be used for a random walk in GridSpace, but there is currently no equivalent in ContinuousSpace 

That is a bit tricky, because the number of nearby positions in `ContunuousSpace` is infinite, no matter how small the radious.

AayushSabharwal:
> That is a bit tricky, because the number of nearby positions in `ContunuousSpace` is infinite, no matter how small the radious.

True. However, computationally it's finite, since we can only store so many significant digits. Similar to how `rand` can only output finitely many values, despite being theoretically infinite. Is there a reason we can't do the same here?

Datseris:
Yes. `rand` outputs a single number, while `nearby_positions` is _an iterator over all positions_. You don't use `rand` to _iterate over all numbers between 0 and 1_. That's why we have the `random_position` function, which is defined for `ContinuousSpace`.

AayushSabharwal:
Ah yes, I got the two things confused. How should random walks be handled in continuous space pathfinding, then? Should I add methods for `walk!`, which call `set_target!` on a random position within a given radius?

Datseris:
I don't see why `walk!` is now interlaced with `set_target!`. The original definition of `walk!` is 
```
walk!(agent, direction::NTuple, model; ifempty = false)
walk!(agent, rand, model)
```
Neither is related to pathfinding. I think I've lost the thread of discussion, can you provide a summary of why `walk!` and pathfinding are related? 

For ContinuousSpace, `walk!` could have a method with a radius:
```
walk!(agent, radius::Real, model{COntinuousSpace})
```
which propagates the agent in a random direction by an amount given by the `radius`?

Datseris:
Actually Im confused by this whole business:

> What if nearby_walkable returned a random position in a random walkable cell of the pathfinder?

Isn't the entire goal of pathfinding to set proper routes between agents and targets? Why do we need this `nearby_walkable` function...?

Libbum:
In addition to George's latest question, I'd like to point out that

`walk!(agent, rand, model)`

is not really a 'random walk' in the way that term is usually understood: it's a movement in a random direction that ultimately pulls from a uniform distribution. 

Sorry, I've not had the time to look at the code yet, so I think I'm missing the thread here too.

AayushSabharwal:
`nearby_walkable` is a utility function. While making the rabbit/fox/hawk example, I ended up doing the same thing repeatedly and made it a function. It seemed like a useful function to have for grid space pathfinding.

This week, @Libbum and I discussed that it might be useful for rabbit-fox-hawk to also have a version in continuous space (similar to how there are multiple versions of predator-prey and daisyworld), as an example of why one might need it (having the concept of speed). I was going through the code for that example and checking if there's some functionality related to pathfinding in grid space that didn't have a direct alternative in continuous space, and this jumped out. Currently the method used for random walks in the example isn't a true random walk either. For that, I pick a random position in the vision radius and move there. Since the radius is large enough I think it works as a decent approximation. Replicating it for continuous space isn't easily possible with the current API

Libbum:
OK, thanks for clearing that up. I'll come back to you tomorrow morning on this - will do a proper review then.

AayushSabharwal:
I've compiled a list of all functions I think can be slightly modified to work with pathfinders:
- [ ] `random_position(model, pathfinder)` - Considers walkability of the returned position ⭐ 
- [ ] `walk!(agent, rand, model, pathfinder)` - Considers walkability of the random position
- [ ] `positions(model, pathfinder [, by])` - For discrete space, only returns walkable positions
- [ ] `fill_space!([A, ] model, args...; kwargs...)` and `fill_space!([A ,] model, f; kwargs...)` - Only fills walkable positions
- [ ] `nearby_positions(positions, model, r = 1)` - Returns only walkable positions ⭐ 
- [ ] `add_agent!([A, ] model, pathfinder, args...; kwargs...)` - The agent is added to a random walkable position ⭐ 
- [ ] The empty positions related methods - Consider walkability of the returned positions

I've marked with ⭐ the ones I think are more important

Datseris:
I vote that fundamentally central functions like `add_agent!` and `random_position` should **not** have overloads w.r.t. pathfinder, which I still consider a minor functionality. Sure, it took huge effort, and many lines of code to write, but if one consider all possible ABMs, pathfinding is minor while `add_agent!` is used in every single one.

Just notice: multiple dispatch is awesome, but everything should be taken **in moderation**. The more methods you add to `add_agents!`, the harder it becomes to learn. So you need to weight the generality of the new added method, versus the increase it gives in the methods of `add_agent!`.

Why not have `random_walkable` and `nearby_walkable` as different names instead? The function `add_agent!` can then just use the result of `random_walkable`.

Datseris:
Also, I'm completely against having "walk! with pathfinder" and I'm still unconvinced of its use. The whole purpose is setting routes via the standard approach. Why is walk here I still don't see. Just pick a random `random_walkable` and set a route to that. Perhaps you can add some kind of limiter to `random_walkable` to not give a grid cell too far away or something like that.

AayushSabharwal:
I see your point now. If all we're doing is respecting walkability, we can provide those two functions and leave it to the user to use them. The other alternative functions I proposed would boil down to simple wrappers over `random_walkable`/`nearby_walkable`, which isn't worth the added complexity, both in terms of the number of methods scattered around and documentation. I'll add these two new functions.

I think both of them can be extended to continuous space fairly easily. All it would require is generating a random grid cell in the walkable matrix and taking a random poisition inside that cell. `nearby_walkable` would additionally do a length-check on the vector, since grid cells at the edge would also have points outside the radius.

AayushSabharwal:
Although, `nearby_walkable` in continuous space wouldn't benefit from grid space's nearest neighbor optimisations, specifically the caching of hoods.

AayushSabharwal:
I think this closes the list of features that continuous pathfinding needed. I'd say it's ready for review now.

Datseris:
@AayushSabharwal this probably comes as a late notice... but can you prepare a small high level summary for the GSOC meeting tomorrow (I'll also join)?

src/submodules/pathfinding/astar_continuous.jl
```diff
@@ -0,0 +1,181 @@
+"""
+    find_continuous_path(pathfinder, from, to, model)
+Functions like find_path, but uses the output of find_path and converts it to the coordinate
```
Datseris, line 3: instead of find_path, type `find_path` (i.e. anything that is _code_ must be in backticks)

AayushSabharwal:
I just noticed a bug: In `set_best_target!` for a grid space, there is no differentiation between the current position of the agent and unreachable positions. i.e. if I pass `[agent.pos, some_unreachable_position]` to it, it will return `some_unreachable_position` since the length of the returned path to both is 0
This doesn't happen with continuous space, since `find_continuous_path` returns `nothing` if no path exists. I'll modify `find_path` to do this as well, and `set_best_target!` such that it can handle such a scenario

AayushSabharwal:
The second implementation of `random_walkable` for continuous space ended up having a lot of bugs and nuances associated with it. They should be fixed now

AayushSabharwal:
I'm not fully sure how to write tests for an RNG based function, since their output is inherently unpredictable. For now, I've taken 10 samples for each test. However, this still likely won't catch every scenario. I didn't seed the model RNG since it's possible some other seed could cause an issue that the one I use won't.

Libbum:
Yeah, they are fundamentally a problem. Just do the best you can. A sense check on a seed is fine, make sure that some property that comes out (even just the size of a vector for example) is the same.

To be completely happy with most of Agents.jl, we'd need to do some fuzz testing, so don't fret too much on having poor unit tests in this area.

Datseris:
should we merge this into the pathfinding branch and give a final review?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## New features and fixes
 - Provide a generator function to collect `mdata` in `run!` and `ensemblerun!`.
 - Save/load entire models using `save_checkpoint` and `load_checkpoint`
+- New functions `get_spatial_property` and `get_spatial_index` that allows better handling of spatial fields present in `ContinuousSpace` that are represented via the forms of discretization over the space.
```
Libbum, line 5: New functions `get_spatial_property` and `get_spatial_index`, which provide better handling of discrete representations of spatial fields present in `ContinuousSpace`.

docs/src/comparison.md
```diff
@@ -14,7 +14,7 @@ The results are characterised in two ways: how long it took each model to perfor
 
 Time taken is presented in normalised units, measured against the runtime of Agents.jl. In other words: the results do not depend on any computers specific hardware. If one wishes to repeat the results personally by using the scripts in the [ABM_Framework_Comparisons](https://github.com/JuliaDynamics/ABM_Framework_Comparisons) repository: they will compute the same results. For details on the parameters used for each comparison, see the `benchmark.jl` file in that repository.
```
Libbum, line 15: Why is this file here? This should already be in master?

src/Agents.jl
```diff
@@ -1,7 +1,6 @@
 module Agents
 
 using Requires
```
Libbum, line 3: Same deal here. Looks like you're out of sync a little.

CHANGELOG.md
Datseris: that's already from my PR

Libbum (approved):
Fine to merge into the rework branch here, there are some sync issues that are probably and artifact of branching from different locations, so that should be fine.

Don't forget to remove the `- pathfinding_rework` addition to `.github/workflows/ci.yaml`

AayushSabharwal:
Is there any reason to squash here, since we're not merging to master and maintaining history could be useful?

Datseris:
I have revisited this request on merging commits and did some research on good practices on collaborative code projects. The conclusion is: **always squash**. @AayushSabharwal I'm not sure where you heard that "you don't take credit" if PRs are squashed instead of creating a merge commit, but it is simply not true. Furthermore I believe that acknowledging contributions should be done in a dedicated place in the repo/docs, and not reliant on GitHub's statistics. I'd recommend also watching Bogumil's JuliaCon talk https://pretalx.com/juliacon2021/talk/RJE93F/

PR #532, opened 2021-07-18
Elementary cellular automata example

astenuz:
Hi! I wanted to add this simple example to the library. Its the elementary cellular automata described here https://mathworld.wolfram.com/ElementaryCellularAutomaton.html

Also, maybe to improve it, how would you approach making a video of the process. It would be nice to have a video of the model adding row by row until it forms the final picture.

This is an example with rule 30
image

### DIFF
--- a/examples/elementary_ca.jl
+++ b/examples/elementary_ca.jl
@@ -1,0 +1,111 @@
+# Implementation of an elementary cellular automata
+# according to https://mathworld.wolfram.com/ElementaryCellularAutomaton.html
+# uses random initialization and uses wolfram codes to specify the rule
+# synchronous update of the 1D lattice
+
+using Agents, Random
+using CairoMakie
+using InteractiveDynamics
+using CSV
+
+"""
+The automaton living in a 1D space
+"""
+mutable struct Cell <: AbstractAgent
+    id::Int
+    pos::Dims{1}
+    status::Int # either 0 or 1, where 1 is 'alive'
+end
+
+"""
+Returns an array with the rule for the next status configurations, 1-indexed.
+
+Thus, the first position corresponds to the rule for the cell neighborhood status (0,0,0) 
+and the last to (1,1,1)
+"""
+function rule_from_code(wolfram_code)
+    return digits(wolfram_code, base=2, pad=8)
+end
+
+"""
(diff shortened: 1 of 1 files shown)


Datseris:
Hi there, thanks for the PR! In the example library we already have the game of life example. Does this one bring any new useful information to the list of examples? What I'm thinking is that the examples are becoming a lot, but not all may be useful. So as an alternative I'm thinking of opening a new repo, AgentsExamplesZoo.jl, which will host _any_ example anyone would like to contribute, and would have no real limitation on the amount of examples.

(The only reason this is a problem for this repo is because it increases the amount of time of the CI)

@Libbum

astenuz:
That sounds good, It could also allow to have each example on its own folder/workspace in case it requires different libraries or something like that.

Libbum:
Yeah. So I've got a few research level pieces of work that I'd like to include in the Agents.jl documentation in a more curated fashion, but it's clear that our example section is growing beyond being helpful and moreso becoming an issue to manage.

I like the idea of an `Agents.jl Example Zoo`, which we can accept user submissions, maybe even build only specific doc pages on merge. This can integrate with the current documentation, but certainly be more free form and less curated by the developers.

Libbum:
Let's get moving on this repo first:
- AgentsExamplesZoo.jl
- Agents.jl_Example_Zoo
Any other naming suggestions for it?

Datseris:
AgentsExamplesZoo.jl is the way for me!

Libbum:
Why do we need the .jl for this though? Is this not a documentation repository - a compendium of models, but not a library / package within itself.

Datseris:
My thinking is that it is run via Documenter and produces its own doc pages that shows all examples being run.

Libbum:
I think running on documenter is for sure the way to go. Don't think it needs .JL to work. Guess we can play around with the repo name before publishing regardless

Datseris:
@astenuz thank you! In Agents.jl 5.0 all examples besides the 5 central ones to the library will be in a new repository called "Agents.jl Example Zoo". Can you please make this PR there instead? https://github.com/JuliaDynamics/Agents.jl_Example_Zoo

Datseris:

PR #540, opened 2021-08-06
Pathfinding performance improvements 

AayushSabharwal:
Fixed dynamic dispatch on `delta_cost` call (`@code_typed` output indicates that `pathfinder.cost_metric` is concretely typed)

Working on reducing allocations in `find_path`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["Tim DuBois", "George Datseris", "Aayush Sabharwal", "Ali Vahdati"]
-version = "4.5.1"
+version = "4.5.2"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
--- a/src/submodules/io/jld2_integration.jl
+++ b/src/submodules/io/jld2_integration.jl
@@ -81,16 +81,16 @@ struct SerializableOSMSpace
     agents::Vector{OSMAgentPositionData}
 end
 
-struct SerializableAStar{D,P,M,T}
+struct SerializableAStar{D,P,M,T,C}
     agent_paths::Vector{Tuple{Int,Vector{NTuple{D,T}}}}
     dims::NTuple{D,T}
     neighborhood::Vector{Dims{D}}
     admissibility::Float64
     walkmap::BitArray{D}
-    cost_metric::Pathfinding.CostMetric{D}
+    cost_metric::C
 end
 
-JLD2.writeas(::Type{Pathfinding.AStar{D,P,M,T}}) where {D,P,M,T} = SerializableAStar{D,P,M,T}
+JLD2.writeas(::Type{Pathfinding.AStar{D,P,M,T,C}}) where {D,P,M,T,C} = SerializableAStar{D,P,M,T,C}
 
 function to_serializable(t::ABM{S}) where {S}
     sabm = SerializableABM(
@@ -139,8 +139,8 @@ to_serializable(t::GraphSpace{G}) where {G} = SerializableGraphSpace{G}(t.graph)
 
 to_serializable(t::OSM.OpenStreetMapSpace) = SerializableOSMSpace([])
 
-JLD2.wconvert(::Type{SerializableAStar{D,P,M,T}}, t::Pathfinding.AStar{D,P,M,T}) where {D,P,M,T} =
-    SerializableAStar{D,P,M,T}(
+JLD2.wconvert(::Type{SerializableAStar{D,P,M,T,C}}, t::Pathfinding.AStar{D,P,M,T,C}) where {D,P,M,T,C} =
+    SerializableAStar{D,P,M,T,C}(
         [(k, collect(v)) for (k, v) in t.agent_paths],
         t.dims,
         map(Tuple, t.neighborhood),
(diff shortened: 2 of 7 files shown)


AayushSabharwal:
Improving performance is a more difficult task than I anticipated. I have managed to modify the algorithm so it allocates less, and it performs better than the current implementation on a simple `find_path` benchmark, but is also considerably worse when running a model with it.

Is it feasible to allocate an array `Int32` the same dimensions as the grid as part of the `AStar` struct? This be used to keep track of different connected components in the grid. This improves the worst case when no path can be found between two nodes because they are in different components and the entirety of one component ends up being searched.

Libbum:
We haven't had too many people complaining about memory issues, so maybe that's OK. But it we're trading a few miliseconds for O(n^d) memory requirements then its probably not worth it. We'd want something a bit more significant.

AayushSabharwal:
I think this is the best possible without preallocating.
Results on the `find_path` benchmark from the notebook:
Before:
```
 Range (min … max):  354.666 μs …   7.235 ms  ┊ GC (min … max): 0.00% … 89.73%
 Time  (median):     371.428 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   410.658 μs ± 318.680 μs  ┊ GC (mean ± σ):  5.20% ±  6.31%

  ▆█▇▇▆▅▄▄▄▄▄▃▃▃▂▂▁▁▁▁ ▁                                        ▂
  ████████████████████████▇█▇▇██▇█▇█▇▆▆▆▇▄▆▆▆▅▅▆▆▅▆▆▅▅▄▆▅▄▆▅▅▄▅ █
  355 μs        Histogram: log(frequency) by time        656 μs <

 Memory estimate: 350.66 KiB, allocs estimate: 1920.
```
After:
```
 Range (min … max):  351.363 μs …   5.017 ms  ┊ GC (min … max): 0.00% … 91.20%
 Time  (median):     358.365 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   384.443 μs ± 290.308 μs  ┊ GC (mean ± σ):  5.44% ±  6.59%

  ▄██▇▆▅▄▃▂▁▁▁    ▁▁▁▁▁                                         ▂
  ████████████▇▇█████████▆▃▆▄▆▆▆▅▅▅▅▄▄▅▄▄▄▅▆▅▆▁▅▁▃▃▅▁▄▃▄▃▃▄▄▃▃▄ █
  351 μs        Histogram: log(frequency) by time        509 μs <

 Memory estimate: 350.66 KiB, allocs estimate: 1920.
```
It's minor, but noticeable. It also improved runtime on a model by 0.2s over 10k steps.

AayushSabharwal:
The new ordering for the `PriorityQueue` uses the h-cost as a tiebreaker when f-cost is identical, thus searching nodes closer to the target first in case of a tie. This leads to significant improvements when there aren't many obstacles in the way.

Libbum:
> This leads to significant improvements when there aren't many obstacles in the way.

Can you show an example of this? The benchmarks you show above don't show too much difference - but maybe they're not looking at this case?

In principle the code for this PR looks good though.

AayushSabharwal:
Sure, I'll write a benchmark for this

AayushSabharwal:
I benchmarked `find_path` from `(1, 1)` to `(1000, 1000)` on a pathfinder with `trues(1000, 1000)` as the `walkmap`.

Interestingly, it seems like this reduces the nodes explored (judging by the lower memory estimate) but takes more time on average. I'm not sure why this is the case.
Before:
```
BenchmarkTools.Trial: 2344 samples with 1 evaluation.
 Range (min … max):  1.949 ms …   5.269 ms  ┊ GC (min … max): 0.00% … 57.50%
 Time  (median):     1.985 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.128 ms ± 548.100 μs  ┊ GC (mean ± σ):  5.49% ± 11.51%

  █▆▅▃                                                     ▁   
  █████▇▆▇▇▇▅▅▄▅▁▁▃▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁██▇ █
  1.95 ms      Histogram: log(frequency) by time      4.59 ms <

 Memory estimate: 3.68 MiB, allocs estimate: 13100.
```
After:
```
BenchmarkTools.Trial: 1950 samples with 1 evaluation.
 Range (min … max):  2.420 ms …   5.131 ms  ┊ GC (min … max): 0.00% … 43.91%
 Time  (median):     2.452 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.559 ms ± 437.571 μs  ┊ GC (mean ± σ):  3.24% ±  8.86%

  █▆▄▁▁                                                     ▁  
  █████▇▅▇▇▆▃▆▅▅▅▄▃▁▃▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄██ █
  2.42 ms      Histogram: log(frequency) by time      4.75 ms <

 Memory estimate: 2.76 MiB, allocs estimate: 13061.
```

AayushSabharwal:
I have been using [this article](https://theory.stanford.edu/~amitp/GameProgramming/Heuristics.html) as a reference, specifically the section on breaking ties.

AayushSabharwal:
This handles a simple edge case when `from` or `to` are out of bounds/unwalkable and instantly exits instead of exploring the entire map.

AayushSabharwal:
One single test fails occasionally when changes are made to the algorithm, since it takes the other one of two equivalent paths. I modified the check so either path passes.

Libbum:
So the modifications here are getting a little confusing. Since you swapped this from a draft I guess I'm now questioning if that implies what's here at the moment is close to merge ready. But longer times on average seems like a path that we don't want to go down?

AayushSabharwal:
I should clarify:
I removed the tiebreaker using the h-cost, so the "Before" benchmark is our current status. I also handled an edge case, where if the starting or ending positions are invalid (out-of-bounds or unwalkable) it exists straight away, instead of going through the entire grid as it did before.

Libbum:
OK, no worries. So does that mean you're happy for us to take a closer look at the code at this point and things are merge ready in your view?

AayushSabharwal:
Yes!

Libbum (approved):
Finally had a chance to look over the edge case stuff - all looks good.

AayushSabharwal:
Can I merge this, then?

AayushSabharwal:
Bumped patch version before I forget to. Is this good to go now?

Datseris:
yes, go ahead, and also tag a realese!

PR #541, opened 2021-08-09
Empty positions performance boost

Libbum:
`random_empty` and kin require a full filter of the board each time they're called regardless of how sparse the agents are populating it. This PR provides us quite a noticeable performance enhancement using guesses, that we stop using once we start failing too much.

The core concept is a loop that gives us a `random_position` and returns it, otherwise tries again.
```julia
while true
    pos = random_position(model)
    isempty(pos, model) && return pos
end
```

Benchmarking on the maximum time of our current and this proposed solution, we can find the place where we guarantee chance will win out to filtering/collecting.

```julia
function rand_mt(model)
    has_empty_positions(model) || return nothing
    while true
        pos = random_position(model)
        if isempty(pos, model)
            return pos
        end
    end
end

julia> nagents(model)
249999

julia> prod(size(model.space))
250000

function run_test(model)
    while true
        println(nagents(model))
        curr = @benchmark random_empty($model)
        prop = @benchmark rand_mt($model)
        max_curr = maximum(curr.times)*1e-6 # ms
        max_prop = maximum(prop.times)*1e-6 # ms
        println("Current: $max_curr ms, Proposed: $max_prop ms")
        if max_prop < max_curr
            return nagents(model)/prod(size(model.space))
        end
        kill_agent!(random_agent(model), model)
    end
end
```

The changeover point is close to 0.999 (!) on this model, and as we go lower, as performance matters less, 0.998.. So we set our cutoff there and we can compare our standard Schelling benchmark (which needs random_empty):

Current (my machine): 13.783 ms
This PR (my machine): 2.771 ms

Speedup: 4.98x

Comparisons against other frameworks for Schelling (est, need to bump some changes here):
Mesa: 31.5x -> 156.9x
Netlogo: 8x -> 39.8x
MASON: 14.3x -> 71.1x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
-# v4.5 (UNRELEASED)
+# v4.5
 ## New features and fixes
 - Add `get_spatial_property` and `get_spatial_index` for easier usage of spatially distributed properties in `ContinuousSpace`.
 - Rework the pathfinding system to be more streamlined and offer greater control over the its details.
 - Add support for pathfinding in `ContinuousSpace`.
 - New utility functions `nearby_walkable` and `random_walkable` for use in models with pathfinding.
 - Fixed bug where there was no differentiation between empty paths and paths to unreachable nodes.
+- Performance enhancements for `random_empty`.
 
 ## BREAKING
 - The old pathfinding system is now deprecated. Pathfinding structs are not saved as part of the
--- a/src/spaces/discrete.jl
+++ b/src/spaces/discrete.jl
@@ -86,7 +86,7 @@ function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
     # This switch assumes the worst case (for this algorithm) of one
     # agent per position, which is not true in general but is appropriate
     # here.
-    if nagents(model) / prod(size(model.space)) < cutoff
+    if nagents(model) / prod(size(model.space.s)) < cutoff
         # 0.998 has been benchmarked as a performant branching point
         # It sits close to where the maximum return time is better
         # than the code in the else loop runs. So we guarantee
--- a/src/spaces/discrete.jl
+++ b/src/spaces/discrete.jl
@@ -86,7 +86,7 @@ function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
     # This switch assumes the worst case (for this algorithm) of one
     # agent per position, which is not true in general but is appropriate
     # here.
-    if nagents(model) / prod(size(model.space.s)) < cutoff
+    if clamp(nagents(model) / prod(size(model.space.s)), 0.0, 1.0) < cutoff
         # 0.998 has been benchmarked as a performant branching point
         # It sits close to where the maximum return time is better
         # than the code in the else loop runs. So we guarantee
--- a/src/spaces/discrete.jl
+++ b/src/spaces/discrete.jl
@@ -78,14 +78,15 @@ function has_empty_positions(model::ABM{<:DiscreteSpace})
 end
 
 """
-    random_empty(model::ABM{<:DiscreteSpace})
+    random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
 Return a random position without any agents, or `nothing` if no such positions exist.
+`cutoff` switches the search algorithm from probabalistic to a filter.
 """
-function random_empty(model::ABM{<:DiscreteSpace})
+function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
     # This switch assumes the worst case (for this algorithm) of one
     # agent per position, which is not true in general but is appropriate
     # here.
-    if nagents(model) / prod(size(model.space)) < 0.998
+    if nagents(model) / prod(size(model.space)) < cutoff
         # 0.998 has been benchmarked as a performant branching point
         # It sits close to where the maximum return time is better
         # than the code in the else loop runs. So we guarantee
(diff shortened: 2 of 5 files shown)


src/spaces/discrete.jl
```diff
@@ -82,9 +82,24 @@ end
 Return a random position without any agents, or `nothing` if no such positions exist.
 """
 function random_empty(model::ABM{<:DiscreteSpace})
-    empty = empty_positions(model)
-    isempty(empty) && return nothing
-    rand(model.rng, collect(empty))
+    # This switch assumes the worst case (for this algorithm) of one
+    # agent per position, which is not true in general but is appropriate
+    # here.
+    if nagents(model) / prod(size(model.space)) < 0.998
```
Datseris, line 88: Perhaps we should put `0.998` as a second position argument? `random_empty(model, cutoff = 0.998)`? You never know, maybe in some weird other case, or higher dimensionalities, this needs to change.

Datseris:
hell yeah baby!!!!

Datseris:
Sorry, test fail because of me. I've tried to improve the printing of the model for no space but failed to adjust the tests. Can you do it quickly (I don't have Agents deved atm)

Libbum:
Yes, and yes - i'll fix both in a bit.

Datseris:
Perhaps we should also update the comparison table in this PR?

Libbum:
Yep. I need to update ABM_framework_comparison et al as well. About to go to get the second vaccine, so will get back to this later.

(Wanted to get this out before resubmit)

src/spaces/discrete.jl
```diff
@@ -78,13 +78,29 @@ function has_empty_positions(model::ABM{<:DiscreteSpace})
...
-    isempty(empty) && return nothing
-    rand(model.rng, collect(empty))
+`cutoff` switches the search algorithm from probabalistic to a filter.
+"""
+function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
+    # This switch assumes the worst case (for this algorithm) of one
+    # agent per position, which is not true in general but is appropriate
+    # here.
+    if nagents(model) / prod(size(model.space.s)) < cutoff
+        # 0.998 has been benchmarked as a performant branching point
```
rht, line 90: Is the 0.998 hardware dependent maybe?

Libbum: I don't think it'll be hardware dependent, but could be space type / dimension dependent perhaps. The 0.998 seemed to scale pretty well for 2D grid space at least, so it'll be good to take as a default.

Before doing any merging I'll be sense checking graphs and higher dimensional grid spaces.

Datseris:
just to clarify: I shouldn't merge this, right?

Libbum:
Correct. It's not ready yet. Will let you know.

Libbum:
The cutoff looks like it scales pretty well for everything:

GraphSpace
```julia
g = complete_digraph(500)
model = ABM(Agent5, GraphSpace(g))
for _ in 1:500
    add_agent_single!(model, rand(model.rng))
end
kill_agent!(random_agent(model), model)

run_test(model)
499
Current: 0.023975974358974358 ms, Proposed: 0.010073555555555554 ms
0.998
```

For 5000
```julia
julia> run_test(model)
4999
Current: 0.046383 ms, Proposed: 0.40435699999999997 ms
4998
Current: 0.052060999999999996 ms, Proposed: 0.143565 ms
4997
Current: 0.048937999999999995 ms, Proposed: 0.15190599999999999 ms
4996
Current: 0.07264799999999999 ms, Proposed: 0.052907249999999996 ms
0.9992
```
For 50
```julia
julia> run_test(model)
49
Current: 0.037711253583241455 ms, Proposed: 0.0005155497326203208 ms
0.98
```
3D GridSpace
```julia
@agent Agent3D GridAgent{3} begin end
model = ABM(Agent3D, GridSpace((10,10,10)))
for _ in 1:prod(size(model.space.s))
    add_agent_single!(model)
end
kill_agent!(random_agent(model), model)

julia> run_test(model)
999
Current: 0.007721111111111111 ms, Proposed: 0.05982014285714285 ms
998
Current: 0.01596222222222222 ms, Proposed: 0.034263166666666664 ms
997
Current: 0.007904777777777776 ms, Proposed: 0.017391666666666666 ms
996
Current: 0.007268999999999999 ms, Proposed: 0.010618924242424242 ms
995
Current: 0.015423999999999998 ms, Proposed: 0.011567 ms
0.995
```

For 50^3
```julia
julia> run_test(model)
124999
Current: 1.212627 ms, Proposed: 35.20457 ms
124998
Current: 1.047918 ms, Proposed: 21.559383 ms
124997
Current: 1.0224309999999999 ms, Proposed: 12.160535 ms
...
124954
Current: 1.023584 ms, Proposed: 1.061899 ms
124953
Current: 1.080368 ms, Proposed: 1.014422 ms
0.999624
```

Of course you can micromanage this value but we have a good default at `0.998`.

I'll run the benchmarks and update the docs soon.

AayushSabharwal:
While `cutoff` is less than 1, the `while true` loop is guaranteed to terminate (since there are less agents than the positions they can fill)
However, this isn't guaranteed if `cutoff >= 1`. Should there be some sort of guard against this? Should values larger than 1 even be allowed?

Libbum:
Good point. These tests assume that the number of agents are always less than the available spaces in the grid. The test should be clamped to 1 (values above 1 should not be allowed). I'll fix that.

CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
-# v4.5 (UNRELEASED)
+# v4.5
 ## New features and fixes
 - Add `get_spatial_property` and `get_spatial_index` for easier usage of spatially distributed properties in `ContinuousSpace`.
 - Rework the pathfinding system to be more streamlined and offer greater control over the its details.
 - Add support for pathfinding in `ContinuousSpace`.
 - New utility functions `nearby_walkable` and `random_walkable` for use in models with pathfinding.
 - Fixed bug where there was no differentiation between empty paths and paths to unreachable nodes.
+- Performance enhancements for `random_empty`.
```
Datseris, line 8: kinda missleading to put this here instead of some new header 4.5.3

Libbum: Fixed in 0e153bf1 for subsequent versions/releases.

PR #545, opened 2021-08-21
Small improvments of nearby_ids for graph space

Datseris:
TODO: Actually run a benchmark to confirm that my improvement is true.


src/spaces/graph.jl
```diff
@@ -68,21 +68,29 @@ function add_agent_to_space!(
...
 # NOTICE: The return type of `ids_in_position` must support `length` and `isempty`!
 
 #######################################################################################
 # Neighbors
 #######################################################################################
-function nearby_ids(pos::Int, model::ABM{<:GraphSpace}, args...; kwargs...)
-    np = nearby_positions(pos, model, args...; kwargs...)
-    # This call is faster than reduce(vcat, ..), or Iterators.flatten
-    vcat(model.space.s[pos], model.space.s[np]...)
+function nearby_ids(pos::Int, model::ABM{<:GraphSpace}, r; kwargs...)
```
Libbum, line 78: All other changes in the PR look good, a lot of nice simplifications.

I remember checking this `vcat` method quite extensively and it was faster than all the other methods I tried by a good margin. Will be great to see some benchmarks on this new method of yours to see how it compares!

Datseris:
Test script:
```julia
N = 10

g = complete_digraph(N)
abm = ABM(Agent5, GraphSpace(g))
for n in 1:N
    for _ in 1:N
        add_agent!(n, abm, rand(abm.rng))
    end
end

@btime nearby_ids(2, $abm, 1);
```

timings, N = 10
```
  260.116 ns (3 allocations: 1.11 KiB)

        
  305.578 ns (4 allocations: 1.20 KiB)
```

N = 100
```
  6.000 μs (3 allocations: 79.94 KiB)

  6.040 μs (3 allocations: 79.94 KiB)
```
Seems like for large numbers there isn't any noticeable performance benefit. As such, I'll revert to previous code because it is more concise.

PR #546, opened 2021-08-23
Point vortex model example

nickkeepfer:
Nobody asked for it. But this is a fairly basic and well known (at least in my field) model for point vortices which can be solved rather well with an ABM, despite actually being a system of coupled ODE's. ([wiki](https://en.wikipedia.org/wiki/Two-dimensional_point_vortex_gas))

In terms of what separates it from the other examples, it provides a simple example of how to extend the Euler propagation of `move_agent!()`, to a higher order integration method (Runge-Kutta 4th order) using a simple modification. In addition, using the Query package, it shows a way to extract and plot data from the dataframe which is returned by `Agents.run!()` based on agent field values.

The plots are rather nice and the dynamics are lovely to watch!
pvm


Libbum:
Looks really interesting Nick! There's a little refactoring we can do to simplify some bits and pieces, but the bones are certainly there. 

I haven't used Query before so I'll play around a little with that&mdash;your filters look fairly simple so I wonder if you can't do this with DataFrames directly; but that's probably besides the point of what you were going with here.

We're in the process of streamlining our examples into guided tutorials, and submissions like this will most likely end up being moved to a new set of documentation over at [Agents.jl_Example_Zoo](https://github.com/JuliaDynamics/Agents.jl_Example_Zoo). So give us a little time to work these things out. Thanks for working on this!

nickkeepfer:
Awesome, glad you like it! Of course, it definitely has room to be refactored. Actually I used Query as I wasn't familiar with DataFrames and it was the first result I found for the easy extraction of data from them. I'm sure you could grab the data without it if necessary (would help keep package imports down). 

Cool to know there's an example zoo coming! I think this example could certainly benefit from a dynamics video in the same style of the plot I added, but couldn't find a convenient way to "hold" plots using the otherwise very nice `abm_video()` function.

Libbum:
> I think this example could certainly benefit from a dynamics video in the same style of the plot I added, but couldn't find a convenient way to "hold" plots using the otherwise very nice abm_video() function.

Leave that with us! If there's something that needs to change in abm_video to get this to work, then we'll make it happen.

Datseris:
@Libbum can you explain what needs to be done?

Datseris:
@nickkeepfer thank you! In Agents.jl 5.0 all examples besides the 5 central ones to the library will be in a new repository called "Agents.jl Example Zoo". Can you please make this PR there instead? https://github.com/JuliaDynamics/Agents.jl_Example_Zoo

Datseris:

PR #552, opened 2021-09-14
Updating agent_step function

Alim-faraji:
In the model with space the code passes `agent_step!` instead of `agent_step_2d!`.


Libbum:
Great catch!

PR #553, opened 2021-09-18
Improve DaisyWorld

Datseris:
@Libbum here I've tried to make the daisy propagation part of the agent stepping function, I thought it made a bit more sense. However, the model doesn't work anymore now because no agent propagates. I wonder, did I miss something...?

### DIFF
--- a/examples/daisyworld.jl
+++ b/examples/daisyworld.jl
@@ -153,7 +153,7 @@ end
 # And if the daisies cross an age threshold, they die out.
 # Death is directly in the `agent_step!` function
 function agent_step!(agent::Daisy, model::DaisyWorld)
-    propagate!(daisy, model)
+    propagate!(agent, model)
     agent.age += 1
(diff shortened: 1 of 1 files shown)


examples/daisyworld.jl
```diff
@@ -96,75 +91,73 @@ function diffuse_temperature!(pos, model::DaisyWorld)
...
         end
-    elseif model.scenario == :change
-        model.solar_luminosity += model.solar_change
     end
 end
 
+# And if the daisies cross an age threshold, they die out.
+# Death is directly in the `agent_step!` function
+function agent_step!(agent::Daisy, model::DaisyWorld)
+    propagate!(daisy, model)
```
fbanning, line 156: Shouldn't this be `propagate!(agent, model)` instead?

Datseris, line 156:
```suggestion
    propagate!(agent, model)
```

PR #568, opened 2021-12-04
Update sir.jl

vnikoofard:
If social distancing get practiced the number of infected is reduced so the transmission probability must decrease. No?!


Datseris:
of course, thanks!

Issue #533, opened 2021-07-26
OSM alternative: LightOSM.jl


Datseris:
We wanted a new OSM pipeline, perhaps we should consider this one: https://deloittedigitalapac.github.io/LightOSM.jl/docs/ 

Apparently it has already written `nearby_positions` for us!

But they don't display any performance improvements w.r.t. OSMX.jl. I'll ask for some.

Libbum:
This looks really good.

- KDtree for neighbors
- Routing and other functions are there already
- Represented on a graph, but still contains all the underlying features which we can then incorporate.
- Nice to see my home city as an example :)

In general I think this is well worth considering.

fbanning:
Very promising, thanks for bringing this up.

> But they don't display any performance improvements w.r.t. OSMX.jl. I'll ask for some.

These shortest path benchmarks seem pretty good to me: https://deloittedigitalapac.github.io/LightOSM.jl/notebooks/benchmarks/  
Or were you thinking about other performance measurements=

Datseris:
I was thinking about loading and creating the graph times, but these are also important!

Also, cc @captchanjack

Datseris:
@fbanning or @Libbum the creator has a poster in today's juliacon poster session, might be worth dropping by and saying hello.

Libbum:
Awesome. Will try and get there.

AayushSabharwal:
This looks really cool. I'll take a look at their A* implementation later and see if we can use it in pathfinding

captchanjack:
Hey all!

Thanks for showing interest in LightOSM! Unfortunately I won't be able to host the poster session as I've been quite busy with work, nevertheless happy to answer any of your questions here.

Regarding the benchmarks, I haven't been as active but I'll try to find some time to measure the graph creation time + nearest node calcs! Also any suggestions for improvements are more than welcome :)!

jgamper:
Hey all! It seems that LightOSM pr transitioning from LightGraphs to Graphs https://github.com/DeloitteDigitalAPAC/LightOSM.jl/pull/53 has been merged recently (relating to a discussion here https://github.com/JuliaDynamics/Agents.jl/pull/564) - what needs to be done to transition to LightOSM from OpenStreetMapX? 

I suppose a complete replacement for https://github.com/JuliaDynamics/Agents.jl/blob/master/src/spaces/openstreetmap.jl needs to be written?

Datseris:
[...]

PR #569, opened 2021-12-04
[WIP] LightOSM transition

AayushSabharwal:
Still very much a work in progress. It's possible there will be some deprecations to the API, but I'll try to keep it as consistent as possible. 

Points to consider:
- ~Why does the agent store its route? If its used in a way similar to `AStar`, wouldn't it be better off in the space?~ moved to the space
- I'm still figuring out what the current implementation of `road` does. Is it faster than searching through all roads and picking the closest?
- Can we use `set_target!` (from `Pathfinding`) instead of `plan_route!`, promoting it to a function in the `Agents` module. This would, however, be breaking for `Pathfinding`, since anyone using `Pathfinding.set_target!` would have their code break.

Extra TODO tasks:
- [x] Add compat entry
- [ ] Change `plan_route!` to `set_target!` and use `const set_target! = Agents.set_target!` in `Pathfinding`


Datseris:
thanks so much for starting to work on this. tag me once you need help/review

Datseris:
> Why does the agent store its route? If its used in a way similar to AStar, wouldn't it be better off in the space?

yes, let;s do it how you do it with pathfinding so that the routes are stored in space. this is cleaner this way.

Datseris:
can't help with `road` only @Libbum knows this. I'd say for now just go with the simpler version of just searching through all roads that connect to the given intersection, it is probably okay. We can do optimizations later, stability of API is more of a priority now. Optimizations aren't breaking either.

AayushSabharwal:
Currently, `plan_route` allows choosing between `:shortest` and `:fastest` routes. We do this by taking road speeds as a keyword. However, [`LightOSM.OSMGraph`](https://deloittedigitalapac.github.io/LightOSM.jl/docs/types/#LightOSM.OSMGraph) allows road weights to be specified as `:distance` or `:time`, which is used when finding the [shortest path](https://deloittedigitalapac.github.io/LightOSM.jl/docs/shortest_path/). Doesn't this effectively work as a global setting, removing the need to differentiate between `:shortest` and `:fastest`?

AayushSabharwal:
Now that the paths will be stored in the space, I think it makes sense for `plan_route` to be `plan_route!` and not return the path (instead putting it in the space)

AayushSabharwal:
I should mention, since this is a breaking change, I thought it best to write every decision/change made so we have a record of what was done and why.

Datseris:
yes that is a good idea. You should open a PR that you introduce a _new file_ instead of making changes on the existing one.

AayushSabharwal:
I'm sorry, I'm not sure I understand what you mean by new file.

AayushSabharwal:
Does the agent need to have a `destination` field? It seems like something only used for that "last mile" where the agent needs to go to some point in the middle of a road. Creating an `OpenStreetMapPath` struct that contains this kind of metadata seems to make more sense. It can contain the path to destination, return path if required, starting and ending locations (for "last mile" traversal)

AayushSabharwal:
`road` is proving to be a bigger challenge than I thought. I'm not familiar with spherical geometry, so simple things like finding the projection of a point onto a line are not so simple any more. I think I'm on the right track, though.

Datseris:
@AayushSabharwal I'd suggest to make the new osm space in a new file, so we can compare function definitions between old and current version, but also see which API part is changing more easily. With the GitHub changes now it is a bit tricker top compare versions. I'll be checking out this locally anyway.

Datseris:
I agree with `plan_route!` changes. When you say "global setting" you mean a property of the space, rather of the route planning, right? I guess this would also close PR#429 ?

Datseris:
I agree that the agent doesn't need a destination field. If the path/routing is stored in the space, it makes no sense that the destination is stored in the agent instead.

Datseris:
What makes `road` difficult? Why isn't a simple projection good enough? Earth's spherical geometry really doesn't come into play much in a city block length scale.

AayushSabharwal:
> I'd suggest to make the new osm space in a new file, so we can compare function definitions between old and current version, but also see which API part is changing more easily. With the GitHub changes now it is a bit tricker top compare versions. I'll be checking out this locally anyway.

Ah, I understand. I'll revert my latest commit and create another file before I work on this further. The API hasn't changed too significantly, just enough to support the restructuring of where routes are stored so far. Just a thought: could we use docstrings to keep track of API changes? They haven't been touched so far.

> When you say "global setting" you mean a property of the space, rather of the route planning, right?

Yes.

> I guess this would also close PR #429 ?

I haven't looked through that PR yet, I'll do so and see if we can close it too.

> What makes `road` difficult? Why isn't a simple projection good enough? Earth's spherical geometry really doesn't come into play much in a city block length scale.

Yeah, that makes sense. I was looking into spherical because positions are all in latitude and longitude. Converting them to points on a sphere is easy enough, and I guess a cartesian approximation of distance should work.

Datseris:
> Just a thought: could we use docstrings to keep track of API changes? They haven't been touched so far.

Oh I've missed that. Please, always, always change documentation strings if you do a change. In fact, my advice for good API design is to _first write the documentation string that you want_ and then write the code that makes it possible. So, yeah, docstrings must absolutely reflect the changes as early as possible.

AayushSabharwal:
Is there a specific reason why the current `move_agent!` implementation does this:
```julia
    Agents.remove_agent_from_space!(agent, model)
    agent.pos = pos
    Agents.add_agent_to_space!(agent, model)
```
Instead of first checking if `pos[1]` has changed? If an agent moves slowly along the same road, wouldn't this result in a lot of `deleteat!` and then `push!` into the same `Vector`?

The other spaces do this as well, but it seems unlikely that `move_agent!` in `GridSpace` would be to the same location, and I'm not sure of the performance implications for `ContinuousSpace` since choosing the size of the underlying grid would have a lot of effect on this.

I also noticed `GraphSpace` does this in `move_agent!`:
```julia
    oldpos = agent.pos
    ids = ids_in_position(oldpos, model)
    splice!(ids, findfirst(a -> a == agent.id, ids))
    agent.pos = pos
    push!(ids_in_position(agent.pos, model), agent.id)
```

Instead of just calling `remove_agent_from_space!` and `add_agent_to_space!`, which is weird since the code is basically lifted verbatim from those two methods.

Datseris:
Very good catch for optimizing `move_agent` in OSM! Go ahead and change it to first check if `pos[1]` changes!

Datseris:
For `graphspace`, please simply delete the `move_agent!` method. It shouldn't exist... :D

src/spaces/openstreetmap_new.jl
```diff
@@ -0,0 +1,675 @@
...
+    neighborfn(model.space.m.g, position)
+end
+
+end # module OSM
+
+"""
+    OpenStreetMapSpace(path::AbstractString; kwargs...)
+Create a space residing on the Open Street Map (OSM) file provided via `path`.
+The functionality related to Open Street Map spaces is in the submodule `OSM`.
+
```
Datseris, line 623: Let's please move this to the top of the file! Makes the file much easier to read if you first get the summary of the API and then see the source code!

Datseris:
@AayushSabharwal I have some time on my hands, let me know if you want me to review anything specific.

Datseris:
For your comment:

> Can we use set_target! (from Pathfinding) instead of plan_route!, promoting it to a function in the Agents module. This would, however, be breaking for Pathfinding, since anyone using Pathfinding.set_target! would have their code break.

You can simply define `const set_target! = Agents.set_target!` inside the pathfinding module so it doens't have to be breaking. Generally speaking however I do agree that I dislike the missmatch of names. OSM and Pathfinding have functions that do the same thing "set a route" yet have different names. I don't see a reason for that and I'm up for having the same name.

AayushSabharwal:
> I have some time on my hands, let me know if you want me to review anything specific.

A relatively minor thing. Method definitions all use `ABM{<:OpenStreetMapSpace}`. However, `OpenStreetMapSpace` is a concrete type, so I don't see why the `<:` is necessary. Is there something I missed?

src/spaces/openstreetmap_new.jl
AayushSabharwal: Sure. I didn't notice it at the bottom of the file, which I guess makes a perfect case for why it shouldn't be there

AayushSabharwal:
It seems I interpreted the `pos` field wrong: `pos[3]` is not a fraction in `[0,1]` denoting where it is between `pos[1]` and `pos[2]`, but rather the actual distance between the position in the road and `pos[1]`

Datseris:
Nope, `<:` is not needed if it is a concrete type. It isn't incorrect either though, `Float64 <: Float64` is true. I would actually keep the `<:` in the docstrings, for the simple reason that all other space-dispatching docstrings use `<:` (I think).

AayushSabharwal:
> Nope, `<:` is not needed if it is a concrete type. It isn't incorrect either though, `Float64 <: Float64` is true. I would actually keep the `<:` in the docstrings, for the simple reason that all other space-dispatching docstrings use `<:` (I think).

Sure. I'll update the docstrings accordingly. I wonder if it makes a difference for dispatch, although I don't see why it would.

Datseris:
It doesn't.

AayushSabharwal:
Why is `OSM` its own module?

Datseris:
What do you mean? Why is pathfinding its own module?

Datseris:
Let's think about this for a second: what is more intuitive on the eyes: set_target, or plan_route? I think plan_route is better. In fact, now that I think about it, set_target isn't even a correct representation. The user _provides_ the target to the function, it's not the function that sets it. What the function actual sets is the route. So perhals it is better to use the naming system of OSM?

AayushSabharwal:
> What do you mean? Why is pathfinding its own module?

I should have phrased it better. None of the other spaces are contained inside their own modules, and `OSMSpace` isn't an "optional" submodule, so I was confused why it needed to be inside `module OSM....end`

AayushSabharwal:
> Let's think about this for a second: what is more intuitive on the eyes: set_target, or plan_route? I think plan_route is better. In fact, now that I think about it, set_target isn't even a correct representation. The user _provides_ the target to the function, it's not the function that sets it. What the function actual sets is the route. So perhals it is better to use the naming system of OSM?

Interesting. I thought of it as all agents move to their targets, so the user is telling us (`Agents`) the target, and we handle moving there.

Datseris:
The reason it was made a module initially is because it has so unique names. But if we go with the route of common name for "route planning" then its okay if it is not I guess.

> so the user is telling us (Agents) the target, and we handle moving there

Yes but `set_target!` doesn't do any moving though. I'm not sure which name is the best, but it isn't trivial either. However, given that the move function is `move_along_route!`, then it makes more sense to "plan the route" than to "set the target". It's okay if `plan_route` does nothing.

AayushSabharwal:
Just to update: I'm working on `plan_route!` and `move_along_path!`, and there are a bunch of edge cases to handle so taking a little while to ensure I get everything covered. I'll push my changes once I'm fairly certain it works.

AayushSabharwal:
This _should_ work, but will need tests to confirm. I tried to be very liberal with comments since it's a long winded function

AayushSabharwal:
I don't fully understand `nearby_ids`, but it (and the utility functions it uses) don't directly interact with the map, so nothing should be broken there. `nearby_positions` needed a minor fix, and now both should work.

AayushSabharwal:
I tried `include`ing `openstreetmap_new.jl`, and it compiles! However, there's a warning:
```
WARNING: Method definition __init__() in module Agents at /home/aayush/d/Julia/Agents.jl/src/deprecations.jl:12 overwritten at /home/aayush/d/Julia/Agents.jl/src/Agents.jl:51.
  ** incremental compilation may be fatally broken for this module **
```
Can we merge the two `__init__` functions?

AayushSabharwal:
Tests expect there to be a `move_agent!` for `GraphSpace`. That function was removed. What should be done about this?

Datseris:
> Tests expect there to be a move_agent! for GraphSpace. That function was removed. What should be done about this?

LOL hahahaha yeah remember this function I told you to just delete? Well, turns out we simply had to replace its source to call the other two functions.

Datseris:
> Can we merge the two __init__ functions?

??? Why we have two `__init__` functions? This sounds wrong.

fbanning:
This is the WIP for #533 , right? Do you have a list of things that still need to be looked at?

Datseris:
@AayushSabharwal I now realize the problem regarding move agent missing. It is here: https://github.com/JuliaDynamics/Agents.jl/blob/master/src/core/space_interaction_API.jl#L48 

The `move_agent!` function should just be:
```
function move_agent(agent, pos, model)
remove_agent_from_space!(agent, model)
agent.pos = pos
end
```

This is the generic implementation that should work for any space. Of course, most spaces want to write their own `move_agent!` for performance reasons, but in GrapHSpace that is obvviously not necessary.

Datseris:
@fbanning one thing, that is not crucial for the 5.0 release, but that is certainly beneficial in the long run is to port open street map space plotting to Makie.jl. So far we have been using https://github.com/pszufe/OpenStreetMapXPlot.jl  but it is really slow and of course now it won't work as we don't use this OpenStreetMapX.jl

fbanning:
> port open street map space plotting to Makie.jl

I'll have a look if I understand what needs to be done for this to happen.

AayushSabharwal:
> LOL hahahaha yeah remember this function I told you to just delete? Well, turns out we simply had to replace its source to call the other two functions.

haha sure

> ??? Why we have two __init__ functions? This sounds wrong.

I'm not sure, first time I've actually taken a look at `deprecations.jl`

> one thing, that is not crucial for the 5.0 release

Are we releasing 5.0 after this PR? I don't think we have any other breaking changes to make, but might be a good idea to make sure.

AayushSabharwal:
> This is the WIP for #533 , right? Do you have a list of things that still need to be looked at?

I'm pretty sure all of this works, but tests and `zombies.jl` need to be fixed. LightOSM doesn't ship with a default map like OSMX did, so I guess we need to download one during tests. Is there a way to get the same one OSMX had, somehow?

I also upgraded to 1.7, and the new RNG seems to be failing other tests that passed previously

Datseris:
> I'm not sure, first time I've actually taken a look at deprecations.jl

deprecations.jl goes away, unless it is replaced by new ones.

> Are we releasing 5.0 after this PR? I don't think we have any other breaking changes to make, but might be a good idea to make sure.

Well not _immediatelly_ after, but as soon as possible. Yes, of course we should check.

Datseris:
> LightOSM doesn't ship with a default map like OSMX did, so I guess we need to download one during tests. Is there a way to get the same one OSMX had, somehow?

We don't have to use the same!

> I also upgraded Julia to 1.7, and the new RNG seems to be failing other tests that passed previously

Yeah but that needs to be done in a different PR. Let's try to keep PRs as clean as possible.

fbanning:
> > port open street map space plotting to Makie.jl
> 
> I'll have a look if I understand what needs to be done for this to happen.

This looks doable. I will invest some time into this and see how far I can take it, possibly culminating in a new `osmplot` recipe for us to use.

AayushSabharwal:
Should we include a default map with Agents.jl? Perhaps as an artifact only downloaded when `OSM` module is loaded? It allows us to maintain the `OSM.TEST_MAP` feature, speeds up testing, and prevents our tests from relying on something out of our control

Datseris:
I really don't follow you. Your original suggestion was that OSM should no longer be an idependent module due to the merging of planning functions to the central API. So how do you intended to do your suggestion _practically_, like code-wise? 

I admit the suggestion to not have the tests rely on external means is a good one but:

1. We should try to avoid adding large files on `pkg> add Agents`
2. We really, really need to test that it is possible to download a map, load it into Julia, and make it an OSMSpace, so we might as well do it for the tests anyways. So this eliminates the need to "bundle a map" with Agents.
3. The `TEST_MAP` feature can still be achieved as a function: `OSM.test_map()` which downloads a test map and transforms it into a space.

AayushSabharwal:
> Your original suggestion was that OSM should no longer be an idependent module due to the merging of planning functions to the central API. So how do you intended to do your suggestion practically, like code-wise?

I should clarify, I just meant to support the functionality of having a default map available to users. It doesn't have to be a constant. It could be a function `OSM_test_map()`, like you suggested, which downloads if it isn't already in archives

> We should try to avoid adding large files on `pkg> add Agents`

This way, it isn't installed unless someone explicitly needs it and calls the function.

> We really, really need to test that it is possible to download a map, load it into Julia, and make it an OSMSpace

Downloading maps and loading into Julia is handled by LightOSM, and they test that it all works, so is it necessary for us to test that aspect too? We could just test turning it into OSMSpace, and all of the functionality we build around it.

Datseris:
if you provide a function `OSM_test_map()`, which I agree with, then you need to test it. Given that this function downloads an OSM map and then turns it to our format, then that's exactly what I said we should test, and that exactly what you would be testing as well if you test that `OSM_test_map()` works. So we seem to completely agree.

Don't forget that even though this is handed by LightOSM, we still need to at least document how a user will actually go from the website of openstreetmap to an instance of an Agents.jl space!

AayushSabharwal:
I get it now. Sounds like a good idea. I'll look for a map that isn't too big. London turned out to be ~80MB

AayushSabharwal:
It seems intersections are not nodes. There are multiple nodes along a single road between two intersections. This conflicts with both the API naming, and the tests, since it looks like OSMX used intersections.

AayushSabharwal:
Additionally, the edge weights in LightOSM are not necessarily distances. They can be `:distance`, `:time`, or `:lane_efficiency`. This basically invalidates the positions generated from `road` whenever the weights are not distances.

Datseris:
>  I'll look for a map that isn't too big. London turned out to be ~80MB

I don't see a reason why we would need more than 2 city blocks for testing purposes. Please don't go for a "realistic" map, rather for an efficient one. 

---

I think you are spending a lot of effort trying to make the things be the same as the previous api. You don't have to do this. This is a breaking change. Previous API was made on a suboptimal mapping version, we now use something completely different. There is no reason for us to guarantee that distances need to be kilometers. The "radious" in this new map has the same units as what the user chose when importing the map. Be it time or space. Lane efficiency is weighted space / weighted time. Personally I see no reason for _us_ to convert things to length.

`nearby_ids(pos, model, r)`

`r` is time if lane units are time. Simple and clean.

Datseris:
This means we also dont need #429 anymore

AayushSabharwal:
Our positions are of the form `(x, y, z)`, where `x` and `y` are node IDs on either side of the road and `z` is the distance from `x` traveled along this road. How would this be handled if the edge weights are not distances? Could we just interpolate the time/efficiency values? If we go down this route, `move_along_route!` takes a distance to be moved along the route. Do we rename this to something more generic, since the "distance" could be a time or efficiency value?

Datseris:
We  can make `x` be the _percentage_ of the "distance" between the nodes already travelled instead of what it is now. Then we don't have to worry what units this distance has. Is this easy to do in code?


There is no reason to change anything for `move_along_route!`. Distance is given in units of time, it is not such a wild concept, so we just go with it. Move something for distance of 2 seconds. It's totally okay. It gives so much power to the API and solves somethjing users that use OSM already want, i.e. #429

Datseris:
@AayushSabharwal can you do a video call now? It's proablyt easier to talk live

fbanning:
> Currently, `random_position` gets a random point in bounds and then finds the closest intersection to that point. Is this really necessary? Can't we just get a random intersection?

I think it would even be enough to place agents on a random node of `OSMGraph.nodes`. That's already pretty random placement.

AayushSabharwal:
Yeah, that's what I meant. I guess I should stop calling them intersections, since now I know they aren't.

fbanning:
Yep, it's good to stick with graph terminology: nodes and edges are what we deal with here. Those nodes and edges can have special properties which `OSMGraph ` saves in its `tags` dictionaries. :)

AayushSabharwal:
It seems Julia 1.7 segfaults if I try to run `Graph.neighbors` on the graph from `OSM_test_map`. I'll try and see if this happens on 1.6

AayushSabharwal:
For some reason, the latlon `(51.5328328, 9.9351811)` gives a nearby node with `OSM.intersection` but a much farther road with `OSM.road`. I'm still confused as to why this is the case, but I'll keep investigating

AayushSabharwal:
I think it might just be better to have agent positions be graph vertices over node indices. It would save several dictionary lookups, since most data we use from `OSMGraph` is indexed by graph vertex.

AayushSabharwal:
It's worth mentioning that `road` is extremely slow, to the point where I doubt it would ever be worth using in a model. I'm not sure what to do to improve it, though.

AayushSabharwal:
TODO Some issues I noticed with `plan_route!` + `move_along_route!`,
- [ ] It doesn't work when `agent.pos[1] == agent.pos[2]`
- [ ] Since the route is planned from `agent.pos[1]`, it inherently assumes that the road the agent is currently on is two-way, and the agent can turn around. It should be planned from `agent.pos[2]`, since that is the direction the agent is going in. This, however, has the disadvantage that if the agent is near the beginning of a long two-way road, a better route might be obtained by starting from `agent.pos[1]`.

Datseris:
i agree with the proposed solution.

Datseris:
What does `road` do now?

Datseris:
> I think it might just be better to have agent positions be graph vertices over node indices. It would save several dictionary lookups, since most data we use from OSMGraph is indexed by graph vertex, and space.s is as well. I'll leave it for now, and change it once everything works.

I don't understand what it means. To my knowledge "graph vertex" and "node index" are exactly the same thing. But your sentence suggests they are different things.

fbanning:
> > I think it might just be better to have agent positions be graph vertices over node indices. It would save several dictionary lookups, since most data we use from OSMGraph is indexed by graph vertex, and space.s is as well. I'll leave it for now, and change it once everything works.
> 
> I don't understand what it means. To my knowledge "graph vertex" and "node index" are exactly the same thing. But your sentence suggests they are different things.

I think I can enlighten you a bit. The use of "graph vertex" and "node index" is pretty confusing and one might think that it's the same, if you're not used to working with those two packages. `OSMGraph` fundamentally provides two ways to access the nodes, so let me stick to the terminology used in `LightOSM` and `Graphs`:

- `OSMGraph.nodes` is a `Dict{Int64, Node{Int64}}` and its keys are the node indices in the form of unique OpenStreetMap identifiers. It is a vector that might look like this `5875062472, 1550763764, 471488472, 9030155750, 4087644130, 1939558651, 817321616, 3533830193, .....`.
  
  If you access `OSMGraph.graph` you will get the specific graph constructed via the Graphs package.
- `Graphs` provides a different way of accessing those nodes through its graph structs, e.g. `{3026, 4753} directed simple Int64 graph` where the first number is nodes and the second number is edges. We can access the nodes via `vertices(graph)` and this returns a `Base.OneTo(3026)` iterator.

This means that they're fundamentally different and this is probably why @AayushSabharwal made the distinction between vertex and index. If we want to access the OSM properties of a given `graph vertex`, then we first need to find its corresponding `node index` inside `OSMGraph` via the built-in `OSMGraph.index_to_node` dictionary.

Long story short: While both point to the same underlying node (in a general sense), there's a distinction to be made because of the implementation. Performance wise it's just an O(n) (because of the provided static dictionary) but still something to keep in mind.

AayushSabharwal:
> I don't understand what it means. To my knowledge "graph vertex" and "node index" are exactly the same thing. But your sentence suggests they are different things.

Exactly what @fbanning said. "node index" refers to the OpenStreetMap ID for that node, which comes from the downloaded map data. LightOSM loads this as a graph, so each node has a corresponding graph vertex, which consequently has a different index.

I arbitrarily chose the node index to be used for agent positions, since it was hard to tell which one would be better without actually writing the code. Now that a lot of it is written, it looks like we can avoid a lot of dictionary lookups by using graph vertices instead (There are lookups in `road_length`, and even `move_agent!`). The change won't be difficult.

> What does `road` do now?

`road` takes in a `(latitude, longitude)` and returns the closest point on a road. Currently, this just iterates through all of the edges in the underlying graph, calculates the distance from the input point to the corresponding line segment, and takes the closest one. This is absurdly slow, since there are so many edges even in the small testing graph.

fbanning:
> it looks like we can avoid a lot of dictionary lookups by using graph vertices instead

Makes sense to me. I'd say go ahead.

AayushSabharwal:
Ah, that's a typo
I meant "bug fixes"
:P

Datseris:
graph vertices should be the way to go. The integers from 1 to all amount of graph vertices in the constructed `Graph`.

fbanning:
Quick heads up: I've started a new package for plotting OpenStreetMap data with Makie. It's very heavily work in progress but the basics seem to be up and running. Have a look, if you want: https://github.com/fbanning/OSMMakie.jl

AayushSabharwal:
I don't think `map_coordinates` would be required now that plotting is in another package?

Datseris:
yes, its not, but lon/lat coordinates would be.

AayushSabharwal:
This test:
```
# @test sort!(nearby_ids(model[6], model, 2.0)) == [1, 2, 3, 4, 5]
```
Doesn't pass, since it recurses infinitely, eventually leading to a stack overflow. I'll try and see if I can figure out what's happening. I'm not entirely sure why this is as complex as it is, since it should just be a simple Breadth-First Search on a graph.

AayushSabharwal:
`search_outward_ids` goes recursively and doesn't keep track of nodes it has already visited. Since individual node separations are also small, even a small value of `distance` causes it to recurse for a long time. I should be able to refactor this into one BFS, which doesn't recurse.

Project.toml
```diff
@@ -11,8 +11,9 @@ Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 GraphRecipes = "bd48cda9-67a9-57be-86fa-5b3c104eda73"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
+LightOSM = "d1922b25-af4e-4ba3-84af-fe9bea896051"
```
Datseris, line 14: Don't forget to add a version of this in the `[compat]` otherwise registrator will complain.

AayushSabharwal:
The JLD2 serialization test for OSMSpace doesn't pass because the map isn't strongly connected. Randomly generating paths doesn't work since LightOSM throws an error if a path doesn't exist. Serialization should be functional, otherwise.

AayushSabharwal:
I don't think OSMSpace needs to be a `DiscreteSpace`. It doesn't seem to support the `DiscreteSpace` exclusives anywhere.

AayushSabharwal:
I've updated the documentation. It could probably still do with some work.

Should the `OSM` module be removed in this PR and `Pathfinding` modified accordingly, or should that be separate?

AayushSabharwal:
`random_route!` is not guaranteed to be successful in finding a path. In this case, should it return `false` like it currently does, or should it keep trying until it is successful (with a cap on the number of attempts)?

Datseris:
> random_route! is not guaranteed to be successful in finding a path. 

Why?

AayushSabharwal:
The underlying graph is directed, and not guaranteed to be strongly connected. So, there isn't a guarantee any two random nodes will have a path from one to the other. Running `Graphs.strongly_connected_components` on our test map shows this. Most of the nodes are in the same SCC, but not all of them.

Datseris:
okay just put a try limit as keyword like try 10 times by default, Return `nothing` if it failed.

AayushSabharwal:
Even after the refactoring, OSMSpace exports a lot of new function names. Maybe it's still worth keeping `OSM` as its own module? We can still pull `plan_route!` out to the `Agents` module and refactor pathfinding accordingly.

AayushSabharwal:
There's an issue with downloading the map on the fly. The data was updated since I last downloaded it locally, and tests were updated accordingly. Now, since the data is changed, it downloads a slightly different map in CI, which fails.

Datseris:
Perhaps store a downloaded version in the JuliaDYnamics/JuliaDynamics repo and download it from there directly? We really, _really_ want deterministic tests that never change.

Datseris:
> OSMSpace exports a lot of new function names. Maybe it's still worth keeping OSM as its own module? We can still pull plan_route! out to the Agents module and refactor pathfinding accordingly. 

Ok.

In OSM space I am not sure what "set best target" means. To be honest, I'm not even sure what it means in pathfinding :D

AayushSabharwal:
> In OSM space I am not sure what "set best target" means. To be honest, I'm not even sure what it means in pathfinding :D

In pathfinding, it takes a vector of possible targets, and plans a route to the closest (or farthest) one. We could add a similar function in OSM Space, where the user passes in a vector of destination positions and a path to the closest one is planned.

AayushSabharwal:
Once this fix to artifact downloading is added, I think this PR should be ready for review.

Datseris:
> We could add a similar function in OSM Space, where the user passes in a vector of destination positions and a path to the closest one is planned.

Good to have, but not a priority either. Feel free to decide yourself if you want to add this.

What I'd certainly like to have for sure is a function that given two points on the space, it returns the distance between them.

Datseris:
> Once this fix to artifact downloading is added, I think this PR should be ready for review.

Tests don't pass yet

AayushSabharwal:
> What I'd certainly like to have for sure is a function that given two points on the space, it returns the distance between them.

Should the distance be in terms of the shortest path joining the two points, or the difference in latitude/longitude?

> Tests don't pass yet

Those tests were broken by Julia v1.7 due to the major RNG change, not LightOSM. Should they be fixed in this PR or another one?

Datseris:
> Should the distance be in terms of the shortest path joining the two points, or the difference in latitude/longitude?

In terms of the shortest path!

> Those tests were broken by Julia v1.7 due to the major RNG change, not LightOSM. Should they be fixed in this PR or another one?

God Fucking Damn it. So from now own we will just initialize a single global instance of a number from ReproducibleRNGs.jl and use this in every test. Damn it. But this is also a good reminder that we really need to produce better tests. Tests that rely on randomness are just not good.

Datseris:
But yes, the RNG related tests need to be fixed in a different PR. You have tested locally the OSM tests and they work? Then I can take a final review.

AayushSabharwal:
> But yes, the RNG related tests need to be fixed in a different PR. You have tested locally the OSM tests and they work? Then I can take a final review.

Yeah, they work.

> In terms of the shortest path!

Sure, I'll add it as soon as I can

> God Fucking Damn it. So from now own we will just initialize a single global instance of a number from ReproducibleRNGs.jl and use this in every test. Damn it. But this is also a good reminder that we really need to produce better tests. Tests that rely on randomness are just not good.

I think most of what broke relied on the model initializing its own RNG, which is the same type as the global RNG. Since that changed, everything broke. If we fix that, we should be good. I'll look into ReproducibleRNGs.jl too.

Datseris:
> Sure, I'll add it as soon as I can

Ok I wait for this, and in the mean time I try to look at the code as much as possible.

Datseris:
@AayushSabharwal have you updated the zombie example? Have you had a look at the plotting repository that @fbanning has created?

AayushSabharwal:
> Step 1 would be to make the Zombie example work again.

I did make some changes since I last updated it, so I'll check again. A quick peek at docs says the functions compile, so hopefully it works.

> Have you had a look at the plotting repository that `@fbanning` has created?

I saw it, but haven't actually used it. I didn't update the Zombies example to use this because I thought it would be added to InteractiveDynamics. Should it be directly plotted here, for now?

AayushSabharwal:
Additionally, how should the extra `__init__` method in `deprecations.jl` be handled? It causes the following warning on precompilation:
```
WARNING: Method definition __init__() in module Agents at /path/to/Agents.jl/src/deprecations.jl:12 overwritten at /path/to/Agents.jl/src/Agents.jl:51.
  ** incremental compilation may be fatally broken for this module **
```

src/spaces/openstreetmap_new.jl
```diff
@@ -358,6 +358,127 @@ end
 plan_route!(agent::A, dest::Int, model; kwargs...) where {A<:AbstractAgent} =
     plan_route!(agent, (dest, dest, 0.0), model; kwargs...)
 
+"""
+    function distance(pos_1, pos_2, model::ABM{<:OpenStreetMapSpace})
```
Datseris, line 362: you shouldn't include `function` in the call declaration. I think a better name is `osmdistance`. Distance is missleading as it typically means euclidean distance.

src/spaces/openstreetmap_new.jl
```diff
@@ -358,6 +358,127 @@ end
...
+
+    return dist
+end
+
+function distance(
+    pos_1::Int,
+    pos_2::Tuple{Int,Int,Float64},
+    model::ABM{<:OpenStreetMapSpace}
+)
+    distance((pos_1, pos_1, 0.), pos_2, model)
```
Datseris, line 471: Please do not write `0.` but `0.0` instead. In Julia I think it is not a good idea to end literals with `.`, because the `.` has other syntaxes as well, specifically broadcasting or field access.

src/spaces/openstreetmap_new.jl
AayushSabharwal: If we're keeping the `OSM` module, doesn't `OSM.distance` already accomplish this?

Speaking of, all the module docstrings should be `OSM.(...)` to be in line with other submodules.

Datseris: Oh yeah, then yes, if the module OSM is kept, which I agree with, all docstrings must start with `OSM.`...

AayushSabharwal:
OSMMakie isn't actually published to the registry, so using it requires the GitHub URL

AayushSabharwal:
@fbanning There seems to be some bug with `OSMMakie`:
```
Error showing value of type Makie.FigureAxisPlot:
ERROR: Buffer position does not have the same length as the other buffers.
        offset has length 38390
        position has length 20521
        scale has length 38390
```

Datseris:
> OSMMakie isn't actually published to the registry, so using it requires the GitHub URL

For now let's just try to make it work locally :D But I think plotting the space should be done in a different PR that @fbanning can participate as well, because I'm sure this will need parallel development on both packages. @fbanning do you mind moving OSMMakie into JuliaDynamics just for these initial stages of development, so we can all work on it along a PR on Agents.jl?

Datseris:
@AayushSabharwal have you done all renaming necessary etc. in pathfinding and wherever else?

docs/src/api.md
```diff
@@ -109,10 +109,10 @@ OSM.latlon
 OSM.intersection
 OSM.road
 OSM.random_road_position
-OSM.plan_route
+OSM.plan_route!
```
Datseris, line 112: I'm just a bit confused here. I thought `plan_route!` and `move_along_route!` would become top-level API functions. Why are they still in the OSM module then?

src/spaces/openstreetmap_new.jl
```diff
@@ -0,0 +1,944 @@
...
+    OpenStreetMapSpace(path::AbstractString; kwargs...)
+Create a space residing on the Open Street Map (OSM) file provided via `path`.
+The functionality related to Open Street Map spaces is in the submodule `OSM`.
+
+This space represents the underlying map as a *continuous* entity choosing accuracy over
+performance. The map is represented as a graph, consisting of nodes connected by edges. Nodes
+are not necessarily intersections, and there may be multiple nodes on a road joining two
+intersections. The length of an edge between two nodes is specified in the units of the
+map's `weight_type` as listed in the documentation for
+[`LightOSM.OSMGraph`](https://deloittedigitalapac.github.io/LightOSM.jl/docs/types/#LightOSM.OSMGraph).
```
Datseris, line 49: I think it is worth metioning here all possible types of `weight_type`, even if it is copy pasting from the OSMGraph docs.

src/spaces/openstreetmap_new.jl
```diff
@@ -0,0 +1,944 @@
...
+    return_route::Vector{Int}   # node IDs along path from `dest` to `start`
+    has_to_return::Bool
+end
+
+# NOTE: All positions are indexed by vertex number and _not_ node id
+
+"""
+    OpenStreetMapSpace(path::AbstractString; kwargs...)
+Create a space residing on the Open Street Map (OSM) file provided via `path`.
+The functionality related to Open Street Map spaces is in the submodule `OSM`.
```
Datseris, line 42: I think this docstring should say where/how you can obtain this file.

src/spaces/openstreetmap_new.jl
```diff
@@ -0,0 +1,944 @@
...
+Current `pos`ition tuple is represented as
+`(start intersection index, finish intersection index, distance travelled)`.
+The distance travelled is in the units of `weight_type`.
+
+Further details can be found in [`OSMAgent`](@ref).
+
+## Routing
+
+There are two ways to generate a route, depending on the situation.
+1. Use [`OSM.plan_route!`](@ref) to plan a route from an agent's current position to a target
```
Datseris, line 80: `plan_route!` will be a top-level API, right? So no `OSM.`.

src/spaces/openstreetmap_new.jl
```diff
@@ -0,0 +1,944 @@
...
+        io,
+        "OpenStreetMapSpace with $(length(s.map.highways)) ways " *
+        "and $(length(s.map.nodes)) nodes",
+    )
+end
+
+"""
+    OSM.test_map()
+
+Download a small test map of [`Göttingen`](https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=191361&class=boundary)
```
Datseris, line 112: Göttingen shouldn't be in code ticks `

Datseris (commented):
This is looking excellent. Some minor corrections and then we can merge. Have you tried to produce a rudimentary video of agents moving around and changing color with Makie? It is relatively simple:
```
using GLMakie
pos = Observable([Point2f0(lonlat(agent.pos)) for agent in allagents(model)])
colors = Obsrvable([agent.infected ? :green : :black for agent in allagents(model)])

fig, ax = scatter(pos; color = colors)
function update!()
    pos[] = [Point2f0(lonlat(agent.pos)) for agent in allagents(model)]
    colors[] = [agent.infected ? :green : :black for agent in allagents(model)]
end

record(...) do i
step!(model, ...)
update!()
end
```

---

In the docstring of `nearby_ids` you need to change the phrase that says that in OSM space the distance is in meters to be in `weight_type`.

---

After we see the video that "prooves" that OSMSpace works, this is ready to merge in my eyes.

AayushSabharwal:
> After we see the video that "prooves" that OSMSpace works, this is ready to merge in my eyes.

It works!

https://user-images.githubusercontent.com/23384717/147637609-b7bd62c8-138c-41d0-8199-09ca0da32756.mp4

How should the zombies example be handled, considering OSMMakie.jl isn't actually released?

Datseris:
> How should the zombies example be handled, considering OSMMakie.jl isn't actually released?

For now, can you just do this simple animation as above? (but agents change color as well)

Datseris:
in the plot, what are the axis? They cannot be longitude latitude. The city center of Goettingen is not that large :D

AayushSabharwal:
It's latitude-longitude, but I guess the search API returned more than just the city center. [This](https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=191361&class=boundary) is the map used.

Datseris:
Oh that's humongous... I can't make sense if the agent paths are correct (i.e., they are actually on a road) or not! Shouldn't we have something really small for the default map, like e.g., this: https://www.openstreetmap.org/export#map=16/51.5333/9.9363 ?

AayushSabharwal:
https://user-images.githubusercontent.com/23384717/147732677-fe8204a2-6ab4-49ed-92ce-9d2cf4dcd841.mp4

This is the video on the new map. I used red for humans instead of black just for testing, since it is easier to see.

Datseris:
looks great to me. Shall we merge and then finalize #574 ? In the same pr we can make sure all tests pass correctly etc.

fbanning:
> @fbanning There seems to be some bug with `OSMMakie`:
> 
> ```
> Error showing value of type Makie.FigureAxisPlot:
> ERROR: Buffer position does not have the same length as the other buffers.
>         offset has length 38390
>         position has length 20521
>         scale has length 38390
> ```

Did you use the [development branch](https://github.com/fbanning/OSMMakie.jl/tree/inspection)? The master branch currently has this problem but I think I've fixed it on the other branch. Not totally sure though because it's still so heavily work in progress.

fbanning:
> > OSMMakie isn't actually published to the registry, so using it requires the GitHub URL
> 
> For now let's just try to make it work locally in a single script :D But I think plotting the space should be done in a different PR that @fbanning can participate as well, because I'm sure this will need parallel development on both packages. @fbanning do you mind moving OSMMakie into JuliaDynamics just for these initial stages of development, so we can all work on it along a PR on Agents.jl?

Haven't really thought about this thoroughly, so maybe I'll change my mind in the future, but I think I'd like to finish the recipe in its basic form and for that time leave it where it currently is. After its somewhat functional (i.e. a 0.1.0 release), it would make most sense to me if I would ask for it to be added to [JuliaPlots](https://github.com/JuliaPlots). Don't know if they have any special requirements. But that's a problem for future me.

Regarding interaction with Agents, I'd propose the following: Once the basics of OSMMakie are finished, I can publish it on the registry. Then we can extend the `ABMPlot` recipe in InteractiveDynamics to work with OSMSpace. It can then make proper use of OSMMakie (i.e. wrap it in `ABMPlot`) and we can conveniently plot our agents on the same canvas. This way we will have only one very convenient way for users to plot their ABMs, no matter the underlying space, namely through the `abmplot` function provided by InteractiveDynamics.

So from my perspective, this is the tl;dr:

1. OSMMakie should remain separate and ultimately find its home in JuliaPlots (or similar).
2. Once this PR here is merged, we have our ABM framework for OSMSpace up and running. When it comes to plotting, we don't need any  changes in Agents but instead should work on a PR in InteractiveDynamics.

Feel free to (dis)agree. Oh, and a happy new year to all of you! :)

Datseris:
i guess we have to try out this other branch and see whether it works! Haven't seen it so far. Ultimately, I agree the only choice it makes sense is to put it to JuliaPlots. I've never used recipes before, but as long as the package can plot the map, that's all we care about, right? For InteractiveDynamics we just need an extra `if` clause that plots some background stuff, similarly with how `static_preplot` works. There shouldn't be any more interaction between the two packages.

Oh, actually, scratch that... There is some crucial interaction: that the "units" and "coordinates" used for plotting the street and agents is the same. But I guess this shouldn't be hard to coordinate? We'll have to check! I'm working on #574 at the moment, so it will be a while before I can check!

Enjoy the coming year! Let's hope it is much better than the current one!

AayushSabharwal:
> But I guess this shouldn't be hard to coordinate?

It's not difficult, `OSM.latlon(agent.pos, model)` does the trick.

AayushSabharwal:
With the above PR in JuliaDynamics/JuliaDynamics, `OSM.test_map` should download only the city-center map.

fbanning:
> There is some crucial interaction: that the "units" and "coordinates" used for plotting the street and agents is the same. But I guess this shouldn't be hard to coordinate?

It should actually be pretty simple if I'm not mistaken. The `abmplot` recipe for an OSMSpace model might look something like this:
- We first use the `osmplot` recipe provided by OSMMakie to plot the background map. I will made sure that the package provides some sensible defaults for how the map will be plotted.
- Then we use `scatter` to plot the agents onto the same canvas according to their current positions (properly adjusted via `OSM.latlon`. There should be no conflicts between the positions of the nodes of the map and the positions of the agents.

In my head this is actually going to be as straightforward as it gets. But now that I've said it out loud, I've probably jinxed it. :D

PR #570, opened 2021-12-11
Update schelling.jl

MaxKerney:
Currently the example only mentions using the CairoMakie backend which, if people are following along with the code, leads to the interactive application not working. So I've added that GLMakie has to be used for that section.


fbanning (approved):

fbanning:
Ah, good catch. Yes, that's necessary for the `abm_data_exploration` function to work.

PR #572, opened 2021-12-16
Alternative version of Opinion spread example

MaxKerney:
If the examples are just to showcase package functionality, and the details of the models don't really matter, then feel free to reject this PR. It's just I found the behaviour of the opinion spread example a bit odd, so made some changes to it.

Firstly, because `rand()` uses replacement, using it to generate agents' opinions means that they can end up with multiples of the same opinion, which doesn't really make sense. So I changed this to use `StatsBase.sample(replace = false)` instead.

Secondly, `model[neighbor].opinion .== agent.opinion` is sensitive to the order of opinions. So even if the neighbour and agent have exactly the same opinions, this function could say that they don't if their opinion vectors happen to be ordered differently. This leads to a lot of the model run time being spent reshuffling the orders of the opinion vectors, when this isn't really relevant to the spread of opinions. I've changed the `adopt!` function to be agnostic to the ordering of the opinion vectors.

Finally, I changed the animation, because the specification of RGB values also relies on the ordering of the opinion vectors.

Again, though, if these details don't really matter for the purposes of the example, feel free to reject this. I just thought I'd offer an alternative in case it was helpful.


AayushSabharwal:
I think this looks really cool. Let's see how it turns out in the doc build.

MaxKerney:
Hmm, any idea why it's failing?

AayushSabharwal:
Tests fail partly because we're still migrating from OSMX.jl to LightOSM.jl (see #569) and because Julia v1.7 had major RNG changes which breaks some tests that depend on it. Documentation still builds, but I'm not sure why it isn't deploying so we can see a preview.

Datseris:
@MaxKerney thank you! In Agents.jl 5.0 all examples besides the 5 central ones to the library will be in a new repository called "Agents.jl Example Zoo". Can you please make this PR there instead? https://github.com/JuliaDynamics/Agents.jl_Example_Zoo 

PR #574, opened 2021-12-28
[WIP] start fixing tests & eliminating random dependence

Datseris:


Datseris:
@AayushSabharwal whatever deprecations need to be put we need to now put them into the `deprecations.jl` file.

Datseris:
wtffffff coverage dropped by 15%?????

AayushSabharwal:
#569 duplicated `openstreetmap.jl` into `openstreetmap_new.jl`, and needed to be removed before merging. I guess that's a major reason why. Also, shouldn't it have been a squash and merge?

Datseris:
can't you just fix it in this PR? It is a PR that fixes tests after all. 

I wanted to squash and merge but you complained in the past that this "doesn't give you credit" or something like that...

Datseris:
I've already merged your fix in. But I fear we have forgotten to write some deprecations...?

AayushSabharwal:
> I wanted to squash and merge but you complained in the past that this "doesn't give you credit" or something like that...

Oh, apologies for that. I was mistaken then, and must have not mentioned that I don't mind squash and merge at all especially after you explained why it is the preferred option. Either way, I wouldn't want to mess with the existing convention for merging PRs.

AayushSabharwal:
> But I fear we have forgotten to write some deprecations...?

I'll take a look soon. In #569 there was an issue with precompilation because of the duplicated `__init__` functions in `deprecations.jl` and `Agents.jl`. That would need fixing too

Datseris:
YEah but I removed the second `init` function because it was in `deprecations.jl` anyways

Datseris:
Okay, this PR fixes the tests so I merge it. We can do deprecations in a different PR.

PR #578, opened 2022-01-02
Link to examples zoo

### LINKED ISSUE #577: Agents Zoo Example docs
We now have a hosted documentation dedicated to Agents.jl examples: https://juliadynamics.github.io/AgentsExampleZoo.jl/dev/ 

we need to:

- [ ] Remove all duplicate examples from main repo.
- [ ] Link these new docs in the main docs, preferably in several places.
- [ ] Some documentation strings link to some of the examples that will be removed. We need to fix those links by adding explicit links to the outside docs.
- [ ] Decide what to do with the `Models` sublibrary. I guess we should remove all models that are not explicitly part of the main docs.

Datseris:
Closes #577

Needs a thorough review before merging.

- [x] Remove all duplicate examples from main repo.
- [x] Link these new docs in the main docs, preferably in several places.
- [x] Some documentation strings link to some of the examples that will be removed. We need to fix those links by adding explicit links to the outside docs.
- [x] Decide what to do with the `Models` sublibrary. I guess we should remove all models that are not explicitly part of the main docs.
- [x] Test suite should not use Models if possible.


AayushSabharwal:
> Decide what to do with the Models sublibrary. I guess we should remove all models that are not explicitly part of the main docs.

- Some tests rely on the Models
- Could we turn the zoo into a package, so anyone who wants the extra models can get it?

Datseris:
> Could we turn the zoo into a package, so anyone who wants the extra models can get it?

We would have to be a bit smarter about the code though... So far each model as a Literate.jl version and then a copy pasted version in the source that exports everything. Can we modify the Literate.jl version so that this is possible without having duplicate files...?

AayushSabharwal:
It's an interesting idea. Worth checking out, since it makes maintaining easier. How should the models used in tests be handled?

Datseris:
> How should the models used in tests be handled?

Oh we simply have to remove the unecessary model usage in tests, or define it explicitly in the test suite when necessary. I just haven't gotten myself around to it yet...

Datseris:
@AayushSabharwal feel free to push commits here if you want, I don't have an exclusive over this PR :P I'll be working on it a bit this Sunday for sure though!

AayushSabharwal:
I just chucked `forest_fire` in the test because it's so small. Can't do the same with `daisyworld`.

Datseris:
Why is Daisyworld used in the test suite though? The test suite should test the API. I'm pretty sure we can test the API Daisyworld tests without using such a complex model.

AayushSabharwal:
Yeah. It was unnecessary, I removed it in favour of a fake model that accomplishes a similar goal with regard to tests.

AayushSabharwal:
It doesn't really make a difference here, but in the flocking model why are all the `*_factor` and `visual_distance` part of the agent? Wouldn't these be better off as model properties?

Datseris:
because you may want some randomization/diversity for each bird, but that doesn't matter in the test suite for sure.

Datseris:
I'm listing here some things that don't build in the docs so I can have a look later:

- no videos in Social networks with Graphs.jl
- no videos in Flocking
- @AayushSabharwal in the documentation of OSM download functions from LightOSM.jl I see no `weight_type` argument. Where did you see this from?
- I think we should re-export the download function from LightOSM.jl, and also list it in the docs under the OSM submodule
- @AayushSabharwal the link to Goettingen in the docstring of `test_map` I think no longer correctly shows the map we use, right?

I'd like to devote sunday to get OSM plotting integrated in InteractiveDynamics, so I already finished this one. I'll merge this in, and we should have deprecations and/or update messages in a different PR.

AayushSabharwal:
> in the documentation of OSM download functions from LightOSM.jl I see no weight_type argument. Where did you see this from?

It's [here](https://deloittedigitalapac.github.io/LightOSM.jl/docs/types/#LightOSM.OSMGraph) in the description of the `weights` field of `OSMGraph`.

> I think we should re-export the download function from LightOSM.jl, and also list it in the docs under the OSM submodule

Sure

> the link to Goettingen in the docstring of test_map I think no longer correctly shows the map we use, right?

Yeah, that needs a fix. I'll open a PR for all of these.

Issue #566, opened 2021-11-15
nearby_ids returns the same ID multiple times when r is bigger than the space size


jarbus:
**Describe the bug**
`nearby_ids` returns the same ID multiple times

**Minimal Working Example**
Modified from flock.jl
```julia
using Agents
mutable struct Bird <: AbstractAgent
    id::Int
    pos::NTuple{2,Float64}
end

space2d = ContinuousSpace((100,100),5.0)
model = ABM(Bird, space2d)
for _ in 1:3
    add_agent!( model)
end

println(collect(nearby_ids(model.agents[2], model, 100; exact=false)))
```

```
julia> include("flock2.jl")
[1, 3, 3, 1, 3]
AgentBasedModel with 3 agents of type Bird
 space: periodic continuous space with 30×30 divisions
 scheduler: randomly
```
**Agents.jl version**
4.5.6

jarbus:
This seems to be oriented around `model.space.grid.s` having lots of duplicate ids in a single vector.

AayushSabharwal:
This is interesting. Your interpretation of `model.space.grid.s` is correct. I'm not entirely sure why this happens, but it's definitely something to look into. I'll try running this on my machine in a while and see what happens.

AayushSabharwal:
![image](https://user-images.githubusercontent.com/23384717/142810502-bb365d44-2baf-4691-9d4e-a717041ff034.png)
I get a similarly incorrect output. `model.space.grid.s` as shown here seems to be fine, so it's likely an issue with `nearby_ids`

Datseris:
This is a problem with `periodic = true`. If I run exactly the same thing with `periodic = false` I get the correct answer. I think we need an extra clause to not count periodically found neighbors if they are already found in a non-periodic manner. 

To be fair, I'm not sure whether we should do this... Adding an extra `if` clause increases compute time. But, how often do you really need `nearby_agents` to search with a radious so large that it covers all of space? Because they only scenario where a neighbor is both periodically nearby, but also non periodically, is when the search radius is literally all of space...

Datseris:
Yeah, this is non-fixavle:
```
function _grid_space_neighborhood(
    α::CartesianIndex,
    space::GridSpace{D,true},
    hood,
) where {D}
    return ((mod1.(Tuple(α + β), hood.whole.maxi)) for β in hood.βs)
end
```
we will lose some crazy amount of optimization if we try and fix this issue, and in my eyes it is just not worth it.

[...]

PR #579, opened 2022-01-06
Documentation Update

jarbus:
Explains unintuitive output for `nearby_ids`. See #566 for explanation if needed


Datseris:
Sorry, but the way you say this in the documentation is missleading. This "issue" happens on the extreme case of using a radius so large that it will include a cell both periodically but also non-periodically. It doesn't just occur if you use a periodic space. If you think this is a general phenomenon of periodic space, then please provide a different MWE the one in #566 and I'd be happy to look at it.

PR #580, opened 2022-01-06
Documentation update

jarbus:
Created a new PR to keep the revised documentation in a single commit @Datseris.

### DIFF
--- a/src/core/space_interaction_API.jl
+++ b/src/core/space_interaction_API.jl
@@ -62,7 +62,9 @@ remove_agent_from_space!(agent, model) = notimplemented(model)
     nearby_ids(position, model::ABM, r; kwargs...) → ids
 
 Return an iterable of the ids of the agents within "radius" `r` of the given `position`
-(which must match type with the spatial structure of the `model`).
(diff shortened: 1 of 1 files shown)


jarbus:
Made another change with a force-push to clarify that this case is very rare, still uses one commit.

Datseris:
There is no reason to be concerned about the commit numbers xD I will anyways squash and merge, so the actual PR will be pushed as 1 commit to master no matter how many commits it will have xD

jarbus:
Thanks lol, I'm still pretty new to PRs and thought something like that should be possible but didn't want to assume

Datseris (changes requested):
I'd like two changes:

1) This message should be at the very end of the documentation string. It is certainly not important to be at the very first sentences, which serve to only give a concise description of the function. You can put it after the text regarding the `exact` keyword.

2) We can make the message a bit more precise:
"""
In periodic discrete or continuous spaces, when used with a radius larger than half of the entire space, this function may find the same agent(s) more than once.
See Issue #566 online for more information.
"""

PR #582, opened 2022-01-08
Minor OSM fixes

AayushSabharwal:
Waiting for docs to build to see how this looks


AayushSabharwal:
Hmm I wonder if we can re-export the docstring somehow

Datseris:
? Just put `OSM.download_osm_network` in the `@docs` block?

AayushSabharwal:
> ? Just put OSM.download_osm_network in the @docs block?

I did. It reports the docstring as missing.

Datseris:
Ah sorry sorry, in the `makedocs` function in make.jl you need to add the LightOSM module in the `modules` argument.

Datseris:
@AayushSabharwal sorry for yet again annoying you with the Goettingen map, but can we please get the pedestrian version? :) :) The version we have now seems like the car-oriented version, and I really hate cars :) 

image


(there are no "roads" in the city center)

AayushSabharwal:
image

This looks better.

AayushSabharwal:
There's a minor annoyance with this: it requires downloading with `network_type = :none`. As a result, using this map requires `network_type = :none` to be passed into the `OSMSpace` constructor (for `graph_from_file`) otherwise the extra edges are simply filtered out. I'll add this to the docstring for `OSM.test_map`.

Datseris:
Do you plot this without errors...?

AayushSabharwal:
Nope, same error. It still plots, though.

AayushSabharwal:
It's nice that doc builds take minutes now instead of an hour

AayushSabharwal:
That's just...absurd. Why is windows downloading the old map?

Datseris:
yeah for me it also downloaded the wrong map :P

AayushSabharwal:
Could you please try deleting the contents of `Artifacts.toml` and downloading?

AayushSabharwal:
All tests pass for me on windows.

AayushSabharwal:
Is there some sort of caching going on behind the scenes for windows? I'm at a bit of a loss, because I can't replicate this locally.

Issue #581, opened 2022-01-07
Allocations during stepping


Datseris:
### Discussed in https://github.com/JuliaDynamics/Agents.jl/discussions/559

<div type='discussions-op-text'>

<sup>Originally posted by **fbanning** September 28, 2021</sup>
Here's a broader question that's been boggling my mind for a while now:  
Is there any way for us to get rid of allocations during stepping an ABM?

Say we have the following:

```julia
using Agents

model, agent_step!, model_step! = Models.schelling()
```

If we then track allocations during stepping, we can see that there are new allocations every time:

```julia
julia> @allocated step!(model, agent_step!, model_step!, 1)
115760

julia> @allocated step!(model, agent_step!, model_step!, 1)
40160

julia> @allocated step!(model, agent_step!, model_step!, 1)
15776

julia> @allocated step!(model, agent_step!, model_step!, 1)
7328

julia> @allocated step!(model, agent_step!, model_step!, 1)
4784
```

(Of course the number of allocations decreases here because there's less and less movement happening during the simulation.)

I normally would expect that we have done all the necessary pre-allocations during creation of *this* model because we only have `id::Int`, `pos::Dims{D}`, `mood::Bool` and `group::Int` as agent variables. All of those fields have a fixed length in memory unlike for example a mutable vector that might grow over the course of the simulation.

Regular stepping in this case (i.e. no removing/adding of agents) should therefore not result in new allocations, or am I hugely mistaken here? Why do we still observe allocations?</div>

Datseris:
@fbanning raises a good point here. I think `nearby_agents` is allocating. And this shouldn't be, because we tried hard to design this function with iterators instead of collecting data. So, a proper profiling of the `step!` function needs to be done, so that we identify which function allocates.

fbanning:
> I think `nearby_agents` is allocating.

Could be the case but I think the issue lies deeper within the structure of our agents container.
Trying to replicate it in a MWE could look sth like this:

```julia
julia> mutable struct MyType
           id::Int
       end

julia> d = Dict(1 => MyType(1), 2 => MyType(1))
Dict{Int64, MyType} with 2 entries:
  2 => MyType(1)
  1 => MyType(1)

julia> @allocated d[2].id = 2
875338
```

As you can see, this simple value change allocates.
This doesn't happen when we change values of a simpler dictionary:

```julia
julia> d = Dict(1 => 1, 2 => 1)
Dict{Int64, Int64} with 2 entries:
[...]

PR #584, opened 2022-01-09
Scheduler Rework

### LINKED ISSUE #576: Turn schedulers into function-like objects
Scheduler's are currently normal functions. Turning them into function-like objects has the main benefit of allowing them to be serialized with the model using the `AgentsIO` submodule. It would end up being something like this:

```julia
abstract type AbstractScheduler end

struct Fastest <: AbstractScheduler end

(::Fastest)(model::ABM) = keys(model.agents)
```

and so on. It also makes schedulers that store data a bit easier, so `partially` can be

```julia
struct Partially <: AbstractScheduler
    p::Float64
end

(sched::Partially)(model::ABM) = randomsubseq(model.rng, collect(keys(model.agents)), sched.p)
```
It's a breaking change, so should be good to get in before v5

AayushSabharwal:
Closes #576

Changing `ByType` to returning an iterator instead of `vcat(sets...)` should help with allocations (#581)

TODO:
- [ ] `UserScheduler`
- [ ] Global search and replace for names of schedulers
- [ ] Fix JLD2
  - [ ] Warnings for `UserScheduler`
- [ ] Docs
- [ ] Tests
- [ ] Examples

### DIFF
--- a/src/submodules/schedulers.jl
+++ b/src/submodules/schedulers.jl
@@ -28,117 +28,108 @@ module Schedulers
 using Agents
 using Random: shuffle!, randsubseq
 
-export randomly, by_id, fastest, partially, by_property, by_type
+export AbstractScheduler,
+    Fastest,
+    ByID,
+    Randomly,
+    Partially,
+    ByProperty,
+    ByType
+
+abstract type AbstractScheduler end
 
 ####################################
 # Schedulers
 ####################################
 """
-    Schedulers.fastest
+    Schedulers.Fastest
 A scheduler that activates all agents once per step in the order dictated by the
 agent's container, which is arbitrary (the keys sequence of a dictionary).
 This is the fastest way to activate all agents once per step.
 """
-fastest(model::ABM) = keys(model.agents)
+struct Fastest <: AbstractScheduler end
+(::Fastest)(model::ABM) = keys(model.agents)
 
 """
-    Schedulers.by_id
+    Schedulers.ByID
 A scheduler that activates all agents agents at each step according to their id.
 """
-function by_id(model::ABM)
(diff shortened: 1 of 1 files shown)


AayushSabharwal:
I'm not entirely sure if `ByType` will serialize properly, but I think it should.

src/submodules/schedulers.jl
```diff
@@ -28,117 +28,108 @@ module Schedulers
...
 using Random: shuffle!, randsubseq
 
-export randomly, by_id, fastest, partially, by_property, by_type
+export AbstractScheduler,
+    Fastest,
+    ByID,
+    Randomly,
+    Partially,
+    ByProperty,
+    ByType
```
Datseris, line 37: All the old names need to be put in the deprecations! Also warnings need to be thrown in the ByType and ByProperty when used with a function!

Datseris:
@AayushSabharwal I'm really thinking whether such a large change is worth it... User will always have to provide some functions after load, and will again be the same for any advanced scheduler such as user defined. The load function anyways has a default scheduler value that is the same as the default scheduler of `ABM`. So, maybe we are changing too much for too little gain.

AayushSabharwal:
After beginning to write this, I think I see what you mean now. It ends up with too many edge cases. `ByProperty` might end up with a function, which means that needs to be handled separately. Unsure if `ByType` will work. If the user has a custom scheduler that uses a function, they'd need to implement `from_serializable`. It's leading to more code instead of less.

Datseris:
Yeah, let's close this. It was a valid attempt but unless JLD2 gets to know how to save functions, we're not really in a good situation. BSON can save functions but I guess it has many other drawbacks versus JLD2.

PR #585, opened 2022-01-12
Add lonlat function

fbanning:
Added a `lonlat` function which basically reverses the output of `latlon`. The function is also exported.

This is useful for plotting agents on north-oriented maps (see https://github.com/fbanning/OSMMakie.jl/issues/3#issuecomment-1011005059).


fbanning:
@AayushSabharwal Did you manage to fix the windows test problem? Seems to happen here again.

AayushSabharwal:
Not yet, but I have a feeling #586 should do it.

Datseris:
@fbanning can you please just remove `latlon` entirely in this PR, and only leave `lonlat`? I never noticed that the latitude was first but this doesn't make sense, `lon` is the horizontal coordinate and hence the first to be returned.

fbanning:
I would advise to use `lonlat` as well because it's just more common. Still, `latlon` isn't totally unusual and often used in communication. Having both options might be useful for some users, so why not leave it in there? Doesn't really cost us anything.

Datseris:
`latlon(args...) = reverse(lonlat(args...))` so no for me it is not worth it to export two names that are so simply connected. Every name exported by the API needs to be weighted in gold is my philosophy!

fbanning:
Sure, I get your point. Although by that measure we could also just leave in `latlon` instead of `lonlat` because reversing is so simple. I don't have a hard opinion on this topic but only see that it's not something that's up to us to decide.

If @AayushSabharwal doesn't object, I'm going to rewrite the PR and delete `latlon` entirely. Will this break tests and such?

Datseris:
> Although by that measure we could also just leave in latlon instead of lonlat because reversing is so simple.

No, I disagree, there are fundamental reasons to have `lonlat`. In science in general we are used to the horizontal dimension being the first dimension. x precedes y, and whatever is plotted on the horizontal axis in a plot precedes what is plotted on the vertical axis. That is just the universal, field-agnostic scientific convention. Longitude precedes latitude in the current convention of earth sciences in general. I would be extremely surprised to learn that in e.g., urban transport scientific research latitude precedes longitude.

That is the reason to have `lonlat`. The reason to not have `latlon` as well is because it is simply not worth exporting an extra name.

fbanning:
I've just remembered that `LightOSM` also uses `latlon` format in their `OSMGraph.node_coordinates` field. This is probably why @AayushSabharwal chose it in the first place. How do we want to proceed here? Ask `LightOSM` to also change to `lonlat`? Do we want to fully commit to `lonlat` and solely use it in our functions like `OSM.road` or `OSM.intersection`? That would require us to carefully go through the OSM module and look where we use any coordinates.

Datseris:
Wait, why does LIghtOSM matter suddenly? I thought that `lonlat` (previously `latlon`) was just a convenience function that we exported for users. I didn't realize that we use it in the source code. In the source code we can do whatever, use latlon if this is necessary to play well with LightOSM.

AayushSabharwal:
Yeah, LightOSM uses `latlon` internally, so I just stuck with that. The simplest way to remove `latlon` would just be not exporting it and instead exporting `lonlat`. Then optionally `latlon(...) = reverse(lonlat(...))`. I'm not sure of the performance implications of this.

Datseris:
Yeap, just don't export latlon. 

>  I'm not sure of the performance implications of this.

None. Julia is a compiled language.

fbanning:
> Yeah, LightOSM uses `latlon` internally, so I just stuck with that. The simplest way to remove `latlon` would just be not exporting it and instead exporting `lonlat`. Then optionally `latlon(...) = reverse(lonlat(...))`. I'm not sure of the performance implications of this.

Good idea, that's the way to go. Will adjust the PR accordingly.

fbanning:
I've added back the unexported function `latlon`.

Furthermore I think that the exported functions `road` and `intersection` should take args from the exported interface `lonlat`. Therefore I've modified their docstrings to reflect this and also reversed the lonlat input in place to be able to use it in the functions themselves.

Datseris (approved):
LGTM, thank you :)

PR #586, opened 2022-01-12
Fix Artifacts not working

AayushSabharwal:
It seems this is the correct way to use artifacts, except it expects a gzipped file. I'm trying to figure out if there's a way to tell Pkg that the file isn't gzipped.


fbanning:
Have you had a look at https://github.com/simeonschaub/ArtifactUtils.jl ? Maybe it's helpful for us as well.

AayushSabharwal:
I did _not_ know that existed. Definitely worth a shot.

Datseris:
If we can avoid the extra dependency to the extra package, then we should.

AayushSabharwal:
I guess we should also update the Manifest.toml to the new format, so it stops throwing a warning?

Datseris:
> I guess we should also update the Manifest.toml to the new format, so it stops throwing a warning?

? There shouldn't be a Manifest.toml file in the source code. If it is please delete it :)

AayushSabharwal:
> There shouldn't be a Manifest.toml file in the source code. If it is please delete it :)

Ah yes, my bad. The file isn't committed, I forgot that it's generated and has stuck around since I first cloned this repository.

AayushSabharwal:
https://docs.julialang.org/en/v1/stdlib/LazyArtifacts/

This recommends adding `LazyArtifacts` as a dependency, similarly to `Pkg`. How would I find the UUID for this?  `LazyArtifacts.jl` is a different package on the registry. (https://github.com/JuliaPackaging/LazyArtifacts.jl)

AayushSabharwal:
That's weird. `using LazyArtifacts` works without any dependencies for me.

AayushSabharwal:
Turns out that _is_ the package, and we need a dependency to support Julia < 1.6

Datseris:
yeah ok then. You have already done that?

AayushSabharwal:
> yeah ok then. You have already done that?

Yep, this fixes everything. Tests seem to pass on windows too!

PR #588, opened 2022-01-14
Update `test_map` doc, Fix zombies example

### LINKED ISSUE #587: Better documentation of `test_map()`
At the moment the ddocstring says:
"  Using this map requires network_type = :none to be passed as a keyword to OSMSpace."

But in the zombies example this is *not* done, and yet the example works. So is the statement true?

Also, can you please add what is the distance unit for the default map?

AayushSabharwal:
Closes #587 

The example also didn't run before because `using Random` wasn't there.

### DIFF
--- a/examples/zombies.jl
+++ b/examples/zombies.jl
@@ -9,6 +9,7 @@
 #
 # ## Constructing the end of days
 using Agents
+using Random
 
 # We'll simulate a zombie outbreak in a city. To do so, we start with an agent which
 # satisfies the OSMSpace conditions of having a `pos`ition of type
@@ -37,7 +38,7 @@ end
 
 function initialise(; seed = 1234)
     map_path = OSM.test_map()
-    model = ABM(Zombie, OpenStreetMapSpace(map_path); rng = Random.MersenneTwister(seed))
+    model = ABM(Zombie, OpenStreetMapSpace(map_path; network_type = :none); rng = MersenneTwister(seed))
 
     for id in 1:100
         start = random_position(model) # At an intersection
--- a/examples/zombies.jl
+++ b/examples/zombies.jl
@@ -18,6 +18,7 @@ using Random
 
 @agent Zombie OSMAgent begin
     infected::Bool
+    speed::Float64
 end
 
 # To be explicit, this macro builds the following type:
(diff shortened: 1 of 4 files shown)


Datseris:
What does units of "time" mean? Like, there is a physical space things exist in. To translate physical space to temporal distance you need some velocity. What is this velocity? Docs says that the units are "hours".

Also, is it _our choice_ what distance type we use? Can we please choose units of :distance instead, and also please document what are the actual physical units of this distance, such as meters or kilometers?

Also, why do we need network type :none instead of :walk ?

Datseris:
Also, wouldn't it be simpler if we had the 3-lines-of-code snippet that actually downloads a map after 
https://github.com/JuliaDynamics/Agents.jl/blob/master/src/spaces/openstreetmap.jl#L42-L45 ?

It would make life easier and make it explicit what we should do. LightOSM.jl has no such thing as an overview or a tutorial or some kind of guidance how to use it. You need to connect yourself the dots. We have already done it, so let's just put the information there.

Datseris:
@AayushSabharwal can you also make the zombies move slower in this PR? Would also be nice to highlight how easy it is to make some agents move slower than others (simplu multiply the distance to travel by a new field `agent.speed`).

Sorry I'm adding more things to do in this PR that had a simpler purpose, but they fit here so better we do them here anyways!

AayushSabharwal:
> Also, is it our choice what distance type we use?

Yeah, it looks like the default measure is `:time`, and changing it to `:distance` updates the weights to reflect that. It's the user's choice which unit to use.

> What does units of "time" mean? Like, there is a physical space things exist in. To translate physical space to temporal distance you need some velocity. What is this velocity? Docs says that the units are "hours".

It seems that a speed of 30km/h is used to calculate the times.

> Can we please choose units of :distance instead

Sure, that does make more sense.

> also please document what are the actual physical units of this distance, such as meters or kilometers?

This is done in the docstring for `OSMSpace`

> Also, why do we need network type :none instead of :walk ?

The network_type is simply a filter on the nodes/edges. The filter can be applied while downloading, and it can also be applied when loading the network from file. I just happened to try `:none` to download the map since we wanted more edges, so that's what I stuck with.

> Also, wouldn't it be simpler if we had the 3-lines-of-code snippet that actually downloads a map after
https://github.com/JuliaDynamics/Agents.jl/blob/master/src/spaces/openstreetmap.jl#L42-L45 ?

We already re-export `download_osm_network`, so simply adding a 3-line example for that should suffice, right?

Datseris:
> It seems that a speed of 30km/h is used to calculate the times.

Is that map-agnostic? For any map I download that would be the same number? And is this road agnostic? Is the same speed used for each road? If so, the `:time` unit is **absolutely useless**. It just multiplies the `:distance` with a number. I thought that `:time` would be the time you need by going at the maximum allowed speed of the road. Maybe we should open an issue at LightOSM.jl to support this instead. And to actually document how the make the `:time` version, because they don't.

> Instead, what if we add a dt to model.properties and move each agent by agent.speed * model.dt?

Yes, okay, this makes more sense.

AayushSabharwal:
> Is that map-agnostic? For any map I download that would be the same number? And is this road agnostic? Is the same speed used for each road? If so, the :time unit is absolutely useless. It just multiplies the :distance with a number. I thought that :time would be the time you need by going at the maximum allowed speed of the road. Maybe we should open an issue at LightOSM.jl to support this instead. And to actually document how the make the :time version, because they don't.

It seems I was mistaken. I had picked a few random edges, and coincidentally they all used 30kmph. However, this is not constant across roads. I'm not entirely sure how the speeds are chosen.

AayushSabharwal:
Is there anything else required in this PR?

Datseris (approved):
Sorry @AayushSabharwal for forgetting this, this is LGTM

PR #591, opened 2022-01-16
move to v5

Datseris:
Add changelogs and other entries in docs.


AayushSabharwal:
Is there anything else required here? What else needs to be done before we can release v5?

Datseris:
If you don't mind, I'd like to have native osmplotting released. We are currently working on that with Fredrerik and I believe it will be done very very soon. THen, the OSM example will build its own plot. For this PR in particular I don't think there is anything more. Did you check if I forget any breakign change?

Datseris:
The deprecations file is also empty, we can put there the names that were changed from the Pathfinding

AayushSabharwal:
> If you don't mind, I'd like to have native osmplotting released. We are currently working on that with Fredrerik and I believe it will be done very very soon. THen, the OSM example will build its own plot.

Okay, that is for the best.

Datseris:
Actually @AayushSabharwal I'm thinking of releasing Agents.jl v5 without the plotting for open street map. The OSM plotting may take too long and is unecessarily holding back a very important release that goes into Graphs.jl. What do you think?

AayushSabharwal:
InteractiveDynamics should still work with the current setup, bar OSMSpace right? If OSM plotting will take that long, I don't see the issue in releasing this now. Either way, that will be a new version of InteractiveDynamics and not Agents.jl, if I'm not mistaken?

Datseris:
Yes you are correct. My rationale so far is that it would be good to state in the release message that "we also have native OSM plotting now". Let me see how this weekend goes and then I'll act accordingly.

AayushSabharwal:
I have a couple hours now with not much to do. I've still not been able to figure out `@deprecate`. Is there anything else that needs cleaning up here?

Datseris:
> . I've still not been able to figure out `@deprecate`

Can you please remind me what this issue is?

AayushSabharwal:
> Can you please remind me what this issue is?

I _think_ I understand what I was doing wrong now. It seems the old method used in `@deprecate` needs to be defined as well. I'll add dummy methods to the `OSM` submodule and then try add deprecations.

Datseris:
@AayushSabharwal please help! In the zombies.jl example, I do 
```
step!(model, agent_step!, 1)
```
and I get 
```
julia> step!(model, agent_step!, 1)
ERROR: Not implemented for space type OpenStreetMapSpace
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:33
 [2] notimplemented(model::AgentBasedModel{Agents.OSM.OpenStreetMapSpace, Zombie, typeof(Agents.Schedulers.fastest), Dict{Symbol, Float64}, MersenneTwister})
   @ Agents C:\Users\datse\.julia\dev\Agents\src\core\space_interaction_API.jl:30
 [3] move_along_route!(agent::Zombie, model::AgentBasedModel{Agents.OSM.OpenStreetMapSpace, Zombie, typeof(Agents.Schedulers.fastest), Dict{Symbol, Float64}, MersenneTwister}, args::Nothing; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Agents C:\Users\datse\.julia\dev\Agents\src\core\space_interaction_API.jl:132
 [4] move_along_route!(agent::Zombie, model::AgentBasedModel{Agents.OSM.OpenStreetMapSpace, Zombie, typeof(Agents.Schedulers.fastest), Dict{Symbol, Float64}, MersenneTwister}, args::Nothing)
   @ Agents C:\Users\datse\.julia\dev\Agents\src\core\space_interaction_API.jl:132
 [5] agent_step!(agent::Zombie, model::AgentBasedModel{Agents.OSM.OpenStreetMapSpace, Zombie, typeof(Agents.Schedulers.fastest), Dict{Symbol, Float64}, MersenneTwister})
   @ Main c:\Users\datse\.julia\dev\Agents\examples\zombies.jl:82
 [6] step!(model::AgentBasedModel{Agents.OSM.OpenStreetMapSpace, Zombie, typeof(Agents.Schedulers.fastest), Dict{Symbol, Float64}, MersenneTwister}, agent_step!::typeof(agent_step!), model_step!::typeof(dummystep), n::Int64, agents_first::Bool)
   @ Agents C:\Users\datse\.julia\dev\Agents\src\simulations\step.jl:55
 [7] step!(model::AgentBasedModel{Agents.OSM.OpenStreetMapSpace, Zombie, typeof(Agents.Schedulers.fastest), Dict{Symbol, Float64}, MersenneTwister}, agent_step!::Function, n::Int64, agents_first::Bool) (repeats 2 times)
   @ Agents C:\Users\datse\.julia\dev\Agents\src\simulations\step.jl:45
 [8] top-level scope
   @ REPL[6]:1

```
!!! Why? It is because of the `distance_left = move_along_route!(agent, model, agent.speed * model.dt)`. Without the `distance_left`, i.e., using the version in `src/models/zombies.jl`, I don't get the error!

Datseris:
I think `is_stationary` is bugged and doesn't give false even if `move_along_route!` returns `nothiung`.

AayushSabharwal:
I'm afraid I was asleep when you pinged me. I'll try and take a look in some time, maybe even try and see what's causing #595. Good to see v5 is out though, it's been a long time coming!

Datseris:
Notice that I've fixed it by making the function return 0 instead of missing. This means I didn't "actually" fix the issue. There were still occassions where the agent reached the destinatiion (remianing distance 0), but it was not identified as `is_stationary`. ANd yeah, I've noticed #595 as well when building the docs!

PR #594, opened 2022-01-23
Rename `OSM.intersection` and `OSM.road`

### LINKED ISSUE #593: Rename `OSM.intersection` and `OSM.road`
I realised the names are misnomers. They no longer return intersections/roads. What about `OSM.nearest_node` and `OSM.nearest_edge`?

AayushSabharwal:
Closes #593

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -4,7 +4,6 @@ The API of Agents.jl is defined on top of the fundamental structures [`AgentBase
 In this page we list the remaining API functions, which constitute the bulk of Agents.jl functionality.
 
 ## `@agent` macro
-
 The [`@agent`](@ref) macro makes defining agent types within Agents.jl simple.
 
 ```@docs
@@ -16,7 +15,6 @@ OSMAgent
 ```
 
 ## Agent/model retrieval and access
-
 ```@docs
 getindex(::ABM, ::Integer)
 getproperty(::ABM, ::Symbol)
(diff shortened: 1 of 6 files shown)


docs/src/api.md
```diff
@@ -15,6 +16,7 @@ OSMAgent
 ```
 
 ## Agent/model retrieval and access
+
```
Datseris, line 19: Why have you added all these empty lines between headers and docs? they are not necessary, right? everything compiles without them.

Datseris (changes requested):
Please add these breaking changes in the CHANGELOG.md. Probably also in deprecations.jl.

docs/src/api.md
AayushSabharwal: I forgot about that. I installed a markdown autoformatter recently, and I forgot that it was enabled and formats documents on save. I'll remove these lines

AayushSabharwal:
I've fixed the line spacing, but am still unable to figure out `@deprecate`. I'll add to the changelog in #591 since adding it here would possibly cause a merge conflict.

AayushSabharwal:
There's something wrong with docs...

Datseris:
> There's something wrong with docs... Both here and https://github.com/JuliaDynamics/Agents.jl/pull/591

Don't worry about docs for now, I still need to merge OSM plotting in InteractiveDynamics.jl before re-doing the docs anyways. Just be sure we have the deprecations for the changed names in place and all is good!

PR #600, opened 2022-03-29
Fix OSM pathfinding

### LINKED ISSUE #595: Agents get stuck in OpenStreetMapSpace example
In our example for `OpenStreetMapSpace` I've observed the following situation after about 3000 steps:

Screenshot_20220207_172811

A big share of the agents are stuck in dead ends of our map and can't seem to find their way out.
@AayushSabharwal Could it be the case that the wayfinding algorithm is broken?

AayushSabharwal:
Closes #595

I have a plan to make the code a bit more sensible and easy to reason with, which should help with the bugs since there's currently a lot of scope to miss edge cases.

### DIFF
--- a/src/spaces/openstreetmap.jl
+++ b/src/spaces/openstreetmap.jl
@@ -23,6 +23,7 @@ export test_map,
     lonlat,
     nearest_node,
     nearest_road,
+    same_position,
     download_osm_network    # re-exported from LightOSM.jl
 
 
@@ -214,9 +215,9 @@ function Agents.plan_route!(
 ) where {A<:AbstractAgent}
     delete!(model.space.routes, agent.id)   # clear old route
 
-    identical_position(agent.pos, dest, model) && return true
+    same_position(agent.pos, dest, model) && return true
 
-    if identical_edge(agent.pos, dest)
+    if same_road(agent.pos, dest)
         if agent.pos[1] == dest[2] # opposite orientations
             dest = get_reverse_direction(dest, model)
         end
@@ -268,7 +269,7 @@ function Agents.plan_route!(
         end
         # start and end in middle of edge, but edges have common node
         start_node == agent.pos[1] &&
-            move_agent(agent, get_reverse_direction(agent.pos, model), model)
+            move_agent!(agent, get_reverse_direction(agent.pos, model), model)
         end_node == dest[2] &&
             (dest = get_reverse_direction(dest, model))
         model.space.routes[agent.id] = OpenStreetMapPath(
@@ -372,10 +373,10 @@ function distance(
     kwargs...
 )
     # positions are identical
-    identical_position(pos_1, pos_2, model) && return 0.0
+    same_position(pos_1, pos_2, model) && return 0.0
 
     # positions on same road
-    if identical_edge(pos_1, pos_2)
+    if same_road(pos_1, pos_2)
         if pos_1[1] == pos_2[1]
             return abs(pos_1[3] - pos_2[3])
         else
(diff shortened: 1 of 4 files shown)


src/spaces/openstreetmap.jl
```diff
@@ -112,7 +111,7 @@ function OpenStreetMapSpace(
     path::AbstractString;
     kwargs...
 )
-    m = graph_from_file(path; weight_type = :distance, kwargs...)
+    m = graph_from_file(path; weight_type=:distance, kwargs...)
```
Datseris, line 114: Why? No, please leave spaces between `=`.

AayushSabharwal: Thanks for pointing it out. Looks like the Julia extension is acting up, I'll just turn off the auto-formatter and do it manually.

AayushSabharwal:
I _think_ the issue should be fixed now. I did the following inside the `agent_step!` function in zombies

> old = agent.pos
    distance_left = move_along_route!(agent, model, agent.speed * model.dt)
    npos = agent.pos
    if OSM.identical_position(old, npos, model) && !is_stationary(agent, model)
        println(agent.id, old)
    end

And it doesn't print anything for a couple thousand steps

Datseris (approved):
Very nice. I like a lot the decision to split the huge function into more. Actually, I think we could do it even more, but not necessary for me to submit the approval. One thing I would like a change though is to increase the code clarity of the smaller functions as I've outlined in the comments below.

AayushSabharwal:
Sure. I'll address these in a while.

One question: Is it worth extracting `identical_position` and/or `identical_edge` to the public API? It's probably a common enough operation to check if two agents are at the same location/on the same edge. Perhaps the names can be refactored

AayushSabharwal:
I also wonder if there's a sensible way to split this into multiple files, since it's getting a bit monolithic.

Datseris:
Splitting into multiple files is not strictly necessary. You can have similarly useful outcome by putting appropriate headers like
```
#############################
# Section 1 with appropriate name
#############################
```
But splitting is also a good idea.

and yes for adding `OSM.same_position` and `OSM.same_road`

AayushSabharwal:
Would `same_road` be misleading? Since there are multiple graph edges on the same road, and the function only tests for points being on the same edge.

Datseris:
For me road == edge. For you, what is a road? I am not sure I can imagine how "multiple edges can be in the same road"...?

AayushSabharwal:
In LightOSM, what we call "roads" are `Way`s. A `Way` is composed of multiple `Node`s connected by edges in the graph. A `Node` is not an intersection of "roads", but rather a location with a latitude and longitude (possibly with some metadata). This, for example, allows curved roads to be represented by discretization into multiple segments.

Datseris:
But the function `nearest_road` returns the nearest edge, no?

Datseris:
Futhermore I find it unintuitive to call Ways "roads", because the word "road" in real life isn't used like that. If I set up a google maps journey from here to another place, I would not say that I am on one single "road". On the other hand, the word "road" is also what an edge is in code. Because even a real life straight road would be several edges simply because they are mandatorily broken by intersections.

So perhaps, we should consider dropping the word road alltogether.

AayushSabharwal:
That it does, but it's documented as returning the nearest location on a road, so I think it's okay? Since the nearest location on a road is on one of the edges that compose it.

AayushSabharwal:
> So perhaps, we should consider dropping the word road alltogether.

Hmm. What can we use instead? `road_length` becomes `edge_length`, `nearest_road` becomes `nearest_location_on_edge`? (Off the top of my head those are the only two public functions using "road")

Datseris:
For me it is okay to use the word road for edge. It makes sense. Let's just be clear in the documentation that throughout the OSM module, "road" refers to a directional edge on the graph. Even if in the real world the same "road" may be broken up by multiple intersections, in the OSM module all these would different "roads".

Where is the word `Way` used in either code or documentation? Is it actually used anywhere? I know that we have `OpenStreetMapPath`, but that is `path` not `way`.

AayushSabharwal:
We haven't used `Way` anywhere, it's what LightOSM uses: https://deloittedigitalapac.github.io/LightOSM.jl/docs/types/#LightOSM.Way

Datseris:
Go ahead and merge and tag after increasing version in Project.toml!

AayushSabharwal:
This would be 5.1.0 right? Since `same_position` and `same_road` are new? As opposed to 5.0.1

PR #602, opened 2022-04-08
in schelling example, set agent mood to false if count_neighbors_same…

jochenkrattenmacher:
…_group < minhappy

otherwise the sum of happy agents always increases monotonously when running a model, even if that's not really the case (e.g. when min_to_be_happy == 6)


Datseris (approved):
Yes, legitimate!

PR #603, opened 2022-04-08
fix latitude and longitude of first zombie in zombie example

jochenkrattenmacher:
Hello,
according to the zombie outbreak tutorial, OSM.nearest_road and OSM.nearest_node take a position in form of (latitude, longitude).

However, it seems to be the case that they actually take data in form of (longitude, latitude) (see docstrings of the functions as defined in openstreetmap.jl). Consequently, the video which is produced by the tutorial is different from what one would expect from the tutorial code. The zombie starts at the top left corner, while from (51.5328328, 9.9351811) one would expect the zombie to start on the right, if the format was (latitude, longitude).

The reason why I discovered this issue was because I wanted to let a zombie start from my house - I don't think I will be the last one to try this :D.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati"]
-version = "5.1.1"
+version = "5.1.2"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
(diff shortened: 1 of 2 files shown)


Datseris:
Thanks a lot. Can you please increment patch version number in Project.toml?

jochenkrattenmacher:
so I am new to this. What exactly should I do? Change version = "5.1.1" to version = "5.1.2"?

How can I conveniently do that (I mean create a corresponding pull request)?

Datseris:
No, you can do this in the existing pull request. Go to your fork and the branch from which you've started the pull request: https://github.com/jochenkrattenmacher/Agents.jl/tree/master  then go to the file you want to change and change the version .yes, you need t o put 5.1.2

jochenkrattenmacher:
wow, sorry for the mess, but I suppose I did it

Datseris:
why did you close the PR...?

Datseris (approved):

jochenkrattenmacher:
sorry for the further mess, that wasn't actually my intention, I thought that would just close the discussion... 🙄 

Anyway so I guess for any further pull requests I will always increment the version number (let me know if there's something else I should know in this respect) cheers

PR #606, opened 2022-04-13
simplify union_types and edit docstring

LilithHafner:
The code changes should be a no-op except for dropping support for the (now unused) two-argument version.

Docstring changes: 
1) The argument must be a type
2) The function returns a tuple, not a set
3) `Union`s don't have order. For example, 
```
julia> Agents.union_types(Union{Grass, Wolf, Sheep})
(Grass, Sheep, Wolf)
julia> Union{Grass, Wolf, Sheep} === Union{Grass, Sheep, Wolf}
true
```


Datseris:
thanks, straightforward change

PR #607, opened 2022-04-13
Allow integers to specify stop condition

LilithHafner:
This allows folks to write `step!(model, agent_step!, sqrt(1729))`

### DIFF
--- a/src/simulations/step.jl
+++ b/src/simulations/step.jl
@@ -39,10 +39,10 @@ Use instead of `agent_step!` in [`step!`](@ref) if no function is useful to be d
 """
 dummystep(agent, model) = nothing
 
-until(s, n::Int, model) = s < n
+until(s, n::Real, model) = s < n
 until(s, n, model) = !n(model, s)
 
-step!(model::ABM, agent_step!, n::Int=1, agents_first::Bool=true) = step!(model, agent_step!, dummystep, n, agents_first)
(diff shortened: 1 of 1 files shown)


Datseris:
What's the motivation for this be possible and/or allowed?

AayushSabharwal:
The comparison is with an integer anyway. This effectively implicitly takes the `floor` of the passed-in `Real`. Wouldn't it be better for the user to `floor`/`ceil` it manually and pass in an integer, so they know exactly what is happening? At the very least, this forces a comparison of a `Real` and an `Int`, which if nothing else doesn't sound like a great idea for performance.

Even worse, what if a user passes in something like `12.999`? It might be a fair assumption that some floating-point error could cause this and make it hard to debug.

LilithHafner:
> The comparison is with an integer anyway. This effectively implicitly takes the floor of the passed-in Real. Wouldn't it be better for the user to floor/ceil it manually and pass in an integer, so they know exactly what is happening?

Yes.

> At the very least, this forces a comparison of a Real and an Int, which if nothing else doesn't sound like a great idea for performance.

For Reals like Float64, this difference should cost less than a nanosecond per timestep.

> Even worse, what if a user passes in something like 12.999 when they really intended it to be 13? It might be a fair assumption that some floating-point error could cause this and make it hard to debug.

Yeah. This is why we should force the user to do the rounding. It's much easier for them to spot floating point errors due to rounding in their own code than in ours

I agree with you that reals should not be allowed.

I still think we should widen the type to Integers.

Datseris:
Switching to `Integer` is no problem :)

Issue #445, opened 2021-02-28
Check performance of multi-agent models for dynamic dispatch


Datseris:
EDIT: See https://github.com/JuliaDynamics/Agents.jl/issues/445#issuecomment-1187730099 for a potential solution!!!

---

In Julia when iterating over a non-concretely typed container (e.g. Dict{Int, Union}) the iterables are not type stable. If they are given to a function (e.g. `agent_step!`) Julia does dynamic dispatch instead of static, which is a big hit for performance.

We should see if our simple Wolf/Sheep/Grass model does dynamic or static dispatch. Set up the model in a way that no agents will be created or killed, but only moved and change values. Be sure that no structs/arrays are created during model evolution. 

Then, benchmark the following:

```julia
sheeps = collect_all_model_sheeps
wolfs = # same
grasses = #same 

function custom_step(sheeps, wolfs, grasses, model)
    for s in sheeps; agent_step!(s, model); end
    for w  in wolfs; agent_step!(w, model); end
    for g in grasses; agent_step!(g, model); end
end

```

and compare it with:
```
function our_step(model)
    for a in schedule(model)
         agent_step!(a, model) # uses Multiple Dispatch
    end
end
```
and check the allocation and time differences between the two. Do some schedulers (e.g. the scheduling by Type) improve this?

Should we re-work the internals of `step!` for multi-agent models so that it explicitly separates agents into types first? Is it worth it? Otherwise, we can simply put a note on the "Performance tips" section of the docs of users writing their own custom model stepping function that collects each agent type first and then steps each one individually.

(this Performance section is still to be written and will also have stuff like immutable types, and using a single agent type for maximal performance)

Libbum:
Sounds like a decent plan. If it works we can make a macro to do the step expansion, so worth it: probably. Depending on how much speedup we see. But it shouldn't take too much effort at least.

AayushSabharwal:
I benchmarked this, using the following setup:
https://gist.github.com/AayushSabharwal/f5870d837635a4510c41e57087b31a64

```julia
julia> include("src\\predator_prey.jl")
julia> m = initialize_model()
AgentBasedModel with 550 agents of type Union{Grass, Sheep, Wolf}
 space: GridSpace with size (20, 20), metric=chebyshev and periodic=false
 scheduler: by_union
julia> sheep = [a for a in allagents(m) if typeof(a) == Sheep];
julia> wolves = [a for a in allagents(m) if typeof(a) == Wolf];
[...]

PR #608, opened 2022-04-13
Faster scheduling

LilithHafner:
This already eliminates a lot of dynamic dispatch and helps performance substantially, but I'd like to make it a bit more cohesive and maintainable before merging.

### DIFF
--- a/src/submodules/schedulers.jl
+++ b/src/submodules/schedulers.jl
@@ -26,7 +26,7 @@ Notice that schedulers can be given directly to model creation, and thus become 
 """
 module Schedulers
 using Agents
-using Random: shuffle!, randsubseq
+using Random: shuffle!, randsubseq, randperm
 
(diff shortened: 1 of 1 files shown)


Datseris:
test failure is easy to fix, but how did you measure the performance improvement?

LilithHafner:
I followed along with the method described here: https://github.com/JuliaDynamics/Agents.jl/issues/445#issuecomment-796578504. IIRC I had to do a touch of debugging for it to work on the latest version of Agents.

Datseris:
The code here https://juliadynamics.github.io/Agents.jl/stable/performance_tips/#Avoid-Unions-of-many-different-agent-types-(temporary!)-1 is run via CI so it should work. It also produces a plot for you automatically.

LilithHafner:
The changes I'm making may end up depending on base's `sortperm` so I'm going to work on that before returning here.

PR #609, opened 2022-04-13
Optimize for heterogeneous agents

### LINKED ISSUE #445: Check performance of multi-agent models for dynamic dispatch

LilithHafner:
This already eliminates a lot of dynamic dispatch and helps performance substantially, but I'd like to do a deeper analysis and consider further improvements before merging.

Intended to fix #445

### DIFF
--- a/src/simulations/step.jl
+++ b/src/simulations/step.jl
@@ -44,18 +44,33 @@ until(s, n, model) = !n(model, s)
 
 step!(model::ABM, agent_step!, n::Int=1, agents_first::Bool=true) = step!(model, agent_step!, dummystep, n, agents_first)
 
-function step!(model::ABM, agent_step!, model_step!, n = 1, agents_first=true)
+function step!(model::ABM{S,A}, agent_step!, model_step!, n = 1, agents_first=true) where {S, A}
     s = 0
+    types = Agents.union_types(A)
     while until(s, n, model)
         !agents_first && model_step!(model)
         if agent_step! ≠ dummystep
-            activation_order = schedule(model)
-            for index in activation_order
-                haskey(model.agents, index) || continue
-                agent_step!(model.agents[index], model)
-            end
+            activation_order = schedule(model) # TODO make scheduling type stable
+            # until then, a function barrier can recover most of the performance loss
+            _inner_step!(activation_order, model, types, agent_step!)
         end
         agents_first && model_step!(model)
         s += 1
+    end
+end
+
+function _inner_step(activation_order, model, types, agent_step!)
+    for index in activation_order
+        haskey(model.agents, index) || continue
+        agent = model[index]
+        stepped = false
+        for t in types
+            if agent isa t
+                agent_step!(agent, model)
+                stepped = true
+                break
+            end
+        end
+        @assert stepped
     end
 end


src/simulations/step.jl
```diff
@@ -44,18 +44,33 @@ until(s, n, model) = !n(model, s)
...
         !agents_first && model_step!(model)
         if agent_step! ≠ dummystep
-            activation_order = schedule(model)
-            for index in activation_order
-                haskey(model.agents, index) || continue
-                agent_step!(model.agents[index], model)
-            end
+            activation_order = schedule(model) # TODO make scheduling type stable
+            # until then, a function barrier can recover most of the performance loss
+            _inner_step!(activation_order, model, types, agent_step!)
```
Datseris, line 55: I don't understand these comments. Scheduling is **always** type stable, because it returns a vector of integers. It is using these integers that make things not stable. So, how is `_inner_step` recovering most of the performance lost, given that the loss of performance comes from the command `agent = model[index]`?

src/simulations/step.jl
```diff
@@ -44,18 +44,33 @@ until(s, n, model) = !n(model, s)
...
+        haskey(model.agents, index) || continue
+        agent = model[index]
+        stepped = false
+        for t in types
+            if agent isa t
+                agent_step!(agent, model)
+                stepped = true
+                break
+            end
+        end
```
Datseris, line 73: Also not clear to me what these two loops achieve. Why are we checking if `agent isa t`?

src/simulations/step.jl
LilithHafner: > Scheduling is **always** type stable

It should be but isn't.

> it returns a vector of integers

Integers are not a concrete type. It should return a vector of ints instead.

```
julia> @code_warntype Agents.schedule(m)
MethodInstance for Agents.schedule(::AgentBasedModel{GridSpace{2, false}, Union{Grass, Sheep, Wolf}, Agents.Schedulers.var"#by_union#7"{Bool, Bool}, Nothing, TaskLocalRNG})
  from schedule(model::AgentBasedModel) in Agents at /Users/x/.julia/dev/Agents/src/submodules/schedulers.jl:7
Arguments
  #self#::Core.Const(Agents.schedule)
  model::AgentBasedModel{GridSpace{2, false}, Union{Grass, Sheep, Wolf}, Agents.Schedulers.var"#by_union#7"{Bool, Bool}, Nothing, TaskLocalRNG}
Body::Union{Vector{Any}, Vector{Integer}}
1 ─ %1 = Base.getproperty(model, :scheduler)::Agents.Schedulers.var"#by_union#7"{Bool, Bool}
│   %2 = (%1)(model)::Union{Vector{Any}, Vector{Integer}}
└──      return %2
```

I might fix all this in https://github.com/JuliaDynamics/Agents.jl/pull/608

Datseris: Okay, yes, I should have been more specific, because some schedulers indeed don't return a vector of integers. But this effect you see depends strongly on the scheduler chosen and I do not know which scheduler you have chosen above. Because e.g., `by_id` should not care about the agent types.

> Integers are not a concrete type

Not sure why you think I wouldn't already know this, but thanks for clarifying I guess. When using human words, I refer to the concepts. If I really wanted to imply that they return a vector of `Integer` instead of `Int` I would have used `Vector{Integer}` exactly. Using code with backticks makes it clear if you want to precisely refer to an object existing in code, like `Int`s or `Integer`s.

Nevertheless, schedulers don't have to return `Vector{Int}` exactly anyways, each scheduler should return an _iterator_ over `Int`, whether this is a `Vector` or an `AbstractRange` or a generated iterator. The point is, calling `Agents.schedule(m)` _should_ always return an iterator over integers in a type stable manner. But it doesn't, so it's great that you spotted that. The next question now is which schedulers have this deficit.

Datseris:
Scheduling is now type stable as far as I can tell for single-agent-models. Feel free to open a new PR if you want to work on this based on the new schedulers. The dynamic dispatch on agent types still exists for multi agent models. It still exists in this PR as well.

PR #612, opened 2022-04-15
Random nearby agents

### LINKED ISSUE #575: Random nearby agent
At the moment getting a random nearby agent is highly inefficient. The user needs to collect the output of `neaby_agents` and then pick a random one via tha `rand` function. I propose that we implement the function `random_nearby` which returns a random nearby agent as per the rules of `nearby_agents`. Crucial to make this function as performant as possible. If there are no agents nearby we return `nothing`.

AayushSabharwal:
Closes #575 

The algorithm used is from https://en.wikipedia.org/wiki/Reservoir_sampling#An_optimal_algorithm

It's a bit better, but not much:
```julia
julia> using Agents
julia> @agent Foo GridAgent{2} begin end
julia> sp = GridSpace((10,10))
julia> model = ABM(Foo, sp)
julia> for i in 1:10, j in 1:10
           add_agent!((i, j), model)
       end
julia> @benchmark rand($(model).rng, collect(nearby_agents($(model)[1], $(model), 10))) # old method
 Range (min … max):  17.719 μs …  2.865 ms  ┊ GC (min … max): 0.00% … 98.77%
 Time  (median):     18.125 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   20.492 μs ± 60.028 μs  ┊ GC (mean ± σ):  6.43% ±  2.19%

  ▇█▆▄▂▂▁          ▂▂▁▂▂▁▁▁                                   ▂
  ████████▇▇▇▇▇▆▆▅▇█████████▇▇▆▅▅▄▄▇▇▇▇▇▇▅▅▄▄▄▃▁▁▁▅▇▆▅▃▁▁▁▁▃▆ █
  17.7 μs      Histogram: log(frequency) by time      34.2 μs <

 Memory estimate: 28.25 KiB, allocs estimate: 447.

julia> @benchmark Agents.random_nearby_agent($(model)[1], $(model), 10) # new method
 Range (min … max):  12.697 μs …  2.528 ms  ┊ GC (min … max): 0.00% … 98.59%
 Time  (median):     13.082 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   15.201 μs ± 47.829 μs  ┊ GC (mean ± σ):  6.21% ±  1.97%

  ▆█▅▂             ▃▃▂▃▃▁▁                                    ▂
  █████▇▇▆▆▅▆▆▅▆▆▆▇████████▇▇▅▅▆▅▆▇█▇▇▇▆▆▅▅▄▁▃▁▁▁▃▅▆▆▄▃▄▅▁▃▄▅ █
  12.7 μs      Histogram: log(frequency) by time        29 μs <

 Memory estimate: 20.67 KiB, allocs estimate: 441.
```

My guess is the remaining allocations are the ones inside `nearby_ids`

### DIFF
--- a/src/core/space_interaction_API.jl
+++ b/src/core/space_interaction_API.jl
@@ -361,9 +361,11 @@ nearby_agents(a, model, r = 1; kwargs...) =
     (model[id] for id in nearby_ids(a, model, r; kwargs...))
 
 """
-    random_nearby_id(agent, model::ABM, r = 1; kwargs...) -> agent
+    random_nearby_id(agent, model::ABM, r = 1; kwargs...) → id
 
-Return the `id` of a random agent near the position of the given `agent`.
+Return the `id` of a random agent near the position of the given `agent` using an optimized
+algorithm from [Reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling#An_optimal_algorithm).
+Return `nothing` if no agents are nearby.
 
 The value of the argument `r` and possible keywords operate identically to [`nearby_ids`](@ref).
 """
@@ -396,11 +398,15 @@ function random_nearby_id(a, model, r = 1; kwargs...)
 end
 
 """
-    random_nearby_id(agent, model::ABM, r = 1; kwargs...) -> agent
+    random_nearby_agent(agent, model::ABM, r = 1; kwargs...) → agent
 
-Return the a random agent near the position of the given `agent`.
+Return the a random agent near the position of the given `agent`. Return `nothing` if no agent
+is nearby.
 
 The value of the argument `r` and possible keywords operate identically to [`nearby_ids`](@ref).
 """
-random_nearby_agent(a, model, r = 1; kwargs...) =
-    model[random_nearby_id(a, model, r; kwargs...)]
+function random_nearby_agent(a, model, r = 1; kwargs...)
+    id = random_nearby_id(a, model, r; kwargs...)
+    isnothing(id) && return
+    return model[id]
+end
--- a/src/core/space_interaction_API.jl
+++ b/src/core/space_interaction_API.jl
@@ -22,6 +22,8 @@ export move_agent!,
     nearby_positions,
     nearby_ids,
     nearby_agents,
+    random_nearby_id,
+    random_nearby_agent,
     plan_route!,
     plan_best_route!,
     move_along_route!,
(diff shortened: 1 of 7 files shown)


src/core/space_interaction_API.jl
```diff
@@ -357,3 +359,48 @@ The value of the argument `r` and possible keywords operate identically to [`nea
 """
 nearby_agents(a, model, r = 1; kwargs...) =
     (model[id] for id in nearby_ids(a, model, r; kwargs...))
+
+"""
+    random_nearby_id(agent, model::ABM, r = 1; kwargs...) -> agent
```
Datseris, line 364:
```suggestion
    random_nearby_id(agent, model::ABM, r = 1; kwargs...) → id
```

src/core/space_interaction_API.jl
```diff
@@ -357,3 +359,48 @@ The value of the argument `r` and possible keywords operate identically to [`nea
...
+            _, state = res
+            skip_counter -= 1
+        end
+    end
+
+    return choice
+end
+
+"""
+    random_nearby_id(agent, model::ABM, r = 1; kwargs...) -> agent
```
Datseris, line 399:
```suggestion
    random_nearby_agent(agent, model::ABM, r = 1; kwargs...) → agent
```

src/core/space_interaction_API.jl
```diff
@@ -357,3 +359,48 @@ The value of the argument `r` and possible keywords operate identically to [`nea
 """
 nearby_agents(a, model, r = 1; kwargs...) =
     (model[id] for id in nearby_ids(a, model, r; kwargs...))
+
+"""
+    random_nearby_id(agent, model::ABM, r = 1; kwargs...) -> agent
+
+Return the `id` of a random agent near the position of the given `agent`.
```
Datseris, line 366:
```suggestion
Return the `id` of a random agent near the position of the given `agent`
using an optimized algorithm from [Reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling#An_optimal_algorithm).
```

Datseris (commented):
This is huge! It definitely is faster, 25% or so. But i also adds a new API function, which is the best part. You still need to:

- Increase project toml
- add docstring in the `@docs` block in docs in local area ("local area" sounds so weird, maybe time to rename it to just "nearby agents").

AayushSabharwal:
While writing a couple simple tests for these, I noticed `space_test.jl` relies on `BadAgent` which is defined in `model_creation_tests.jl`. Shouldn't any structs that are common across files be defined in `runtests.jl`, since otherwise there's an implicit dependency between the files, and one _has_ to run before the other?

Datseris:
yes please, move the definition as with all the other `AgentX` definitions.

src/core/space_interaction_API.jl
```diff
@@ -357,3 +359,48 @@ The value of the argument `r` and possible keywords operate identically to [`nea
...
+
+The value of the argument `r` and possible keywords operate identically to [`nearby_ids`](@ref).
+"""
+function random_nearby_id(a, model, r = 1; kwargs...)
+    # Uses Reservoir sampling (https://en.wikipedia.org/wiki/Reservoir_sampling)
+    iter = nearby_ids(a, model, r; kwargs...)
+
+    res = iterate(iter)
+    isnothing(res) && return    # `iterate` returns `nothing` when it ends
```
Datseris, line 375: We should also say that the function returns `nothing` if no agents are neargby.

AayushSabharwal: Ah, that would also make `random_nearby_agent` error. I'll fix that

Datseris:
In my review forgot to say the most important part which is to add tests, but I'm sure you're on it anyways :D

AayushSabharwal:
:D Yes, they're nearly done, along with all you comments.

AayushSabharwal:
That's weird

Datseris (approved):
Please add a CHANGELOG.md entry as well, and then go ahead and merge and tag :)

PR #613, opened 2022-04-17
Refactor schedulers, reduce allocations

### LINKED ISSUE #581: Allocations during stepping

AayushSabharwal:
Closes #581 

- [x] turn schedulers into function-like objects
- [ ] update docs
- [x] update examples
- [ ] update tests
- [ ] Update models

So far I've chosen to keep the names of the schedulers the same, which isn't the best practice for code style but would require less API changes. `by_type` now also takes the type-union in the constructor.

There's some weird behaviour with `by_type` where `shuffle_types` only works if you `collect` the output of the iterator, I'm not really sure why. Adding `print` statements before and after the `shuffle!` line shows that it works, but the test doesn't pass.

### DIFF
--- a/examples/schelling.jl
+++ b/examples/schelling.jl
@@ -97,7 +97,7 @@ function initialize(; numagents = 320, griddims = (20, 20), min_to_be_happy = 3,
     rng = Random.MersenneTwister(seed)
     model = ABM(
         SchellingAgent, space;
-        properties, rng, scheduler = Schedulers.randomly
+        properties, rng, scheduler = Schedulers.randomly()
     )
 
     ## populate the model with agents, adding equal amount of the two types of agents
--- a/examples/flock.jl
+++ b/examples/flock.jl
@@ -57,7 +57,7 @@ function initialize_model(;
     spacing = visual_distance / 1.5,
 )
     space2d = ContinuousSpace(extent, spacing)
-    model = ABM(Bird, space2d, scheduler = Schedulers.randomly)
+    model = ABM(Bird, space2d, scheduler = Schedulers.randomly())
     for _ in 1:n_birds
         vel = Tuple(rand(model.rng, 2) * 2 .- 1)
         add_agent!(
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -243,7 +243,7 @@ e.g. `(agent1, agent7, agent8)`. `order` must be larger than `1` but has no uppe
 Index order is provided by the [`Schedulers.by_id`](@ref) scheduler by default, but can be altered
 with the `scheduler` keyword.
 """
-iter_agent_groups(order::Int, model::ABM; scheduler = Schedulers.by_id) =
+iter_agent_groups(order::Int, model::ABM; scheduler = Schedulers.by_id()) =
     Iterators.product((map(i -> model[i], scheduler(model)) for _ in 1:order)...)
 
 """
@@ -270,9 +270,9 @@ map_agent_groups(order::Int, f::Function, model::ABM, filter::Function; kwargs..
     index_mapped_groups(order::Int, model::ABM, filter::Function; scheduler = Schedulers.by_id)
 Return an iterable of agent ids in the model, meeting the `filter` criterea if used.
 """
-index_mapped_groups(order::Int, model::ABM; scheduler = Schedulers.by_id) =
+index_mapped_groups(order::Int, model::ABM; scheduler = Schedulers.by_id()) =
     Iterators.product((scheduler(model) for _ in 1:order)...)
-index_mapped_groups(order::Int, model::ABM, filter::Function; scheduler = Schedulers.by_id) =
+index_mapped_groups(order::Int, model::ABM, filter::Function; scheduler = Schedulers.by_id()) =
     Iterators.filter(filter, Iterators.product((scheduler(model) for _ in 1:order)...))
 
 #######################################################################################
--- a/src/models/flocking.jl
+++ b/src/models/flocking.jl
@@ -37,7 +37,7 @@ function flocking(;
     spacing = visual_distance / 1.5,
 )
     space2d = ContinuousSpace(extent, spacing)
-    model = ABM(Bird, space2d, scheduler = Schedulers.randomly)
+    model = ABM(Bird, space2d, scheduler = Schedulers.randomly())
     for _ in 1:n_birds
         vel = Tuple(rand(model.rng, 2) * 2 .- 1)
         add_agent!(
--- a/src/models/schelling.jl
+++ b/src/models/schelling.jl
@@ -17,7 +17,7 @@ function schelling(; numagents = 320, griddims = (20, 20), min_to_be_happy = 3)
     @assert numagents < prod(griddims)
     space = GridSpace(griddims, periodic = false)
     properties = Dict(:min_to_be_happy => min_to_be_happy)
-    model = ABM(SchellingAgent, space; properties, scheduler = Schedulers.randomly)
+    model = ABM(SchellingAgent, space; properties, scheduler = Schedulers.randomly())
     for n in 1:numagents
         agent = SchellingAgent(n, (1, 1), false, n < numagents / 2 ? 1 : 2)
         add_agent_single!(agent, model)
--- a/src/submodules/schedulers.jl
+++ b/src/submodules/schedulers.jl
@@ -26,13 +26,21 @@ Notice that schedulers can be given directly to model creation, and thus become 
 """
 module Schedulers
 using Agents
-using Random: shuffle!, randsubseq
+using Random: shuffle!, randsubseq!
 
 export randomly, by_id, fastest, partially, by_property, by_type
 
 ####################################
 # Schedulers
 ####################################
+
+function get_ids!(ids::Vector{Int}, model::ABM)
+    resize!(ids, nagents(model))
+    for (i, id) in enumerate(keys(model.agents))
+        ids[i] = id
+    end
+end
+
 """
     Schedulers.fastest
 A scheduler that activates all agents once per step in the order dictated by the
@@ -45,9 +53,15 @@ fastest(model::ABM) = keys(model.agents)
     Schedulers.by_id
 A scheduler that activates all agents agents at each step according to their id.
 """
-function by_id(model::ABM)
-    agent_ids = sort(collect(keys(model.agents)))
-    return agent_ids
+struct by_id
+    ids::Vector{Int}
+end
+
+by_id() = by_id(Int[])
+
+function (sched::by_id)(model::ABM)
+    get_ids!(sched.ids, model)
+    sort!(sched.ids)
 end
 
 """
@@ -55,20 +69,31 @@ end
 A scheduler that activates all agents once per step in a random order.
 Different random ordering is used at each different step.
 """
-function randomly(model::ABM)
-    order = shuffle!(model.rng, collect(keys(model.agents)))
+struct randomly
+    ids::Vector{Int}
+end
+
+randomly() = randomly(Int[])
+function (sched::randomly)(model::ABM)
+    get_ids!(sched.ids, model)
+    shuffle!(model.rng, sched.ids)
 end
 
 """
     Schedulers.partially(p)
 A scheduler that at each step activates only `p` percentage of randomly chosen agents.
 """
-function partially(p::Real)
-    function partial(model::ABM)
-        ids = collect(keys(model.agents))
-        return randsubseq(model.rng, ids, p)
-    end
-    return partial
+struct partially{R<:Real}
+    p::R
+    all_ids::Vector{Int}
+    schedule::Vector{Int}
+end
+
+partially(p::R) where {R<:Real} = partially{R}(p, Int[], Int[])
+
+function (sched::partially)(model::ABM)
+    get_ids!(sched.all_ids, model)
+    randsubseq!(model.rng, sched.schedule, sched.all_ids, sched.p)
 end
 
 
@@ -79,40 +104,58 @@ with agents with greater `property` acting first. `property` can be a `Symbol`, 
 just dictates which field of the agents to compare, or a function which inputs an agent
 and outputs a real number.
 """
-function by_property(p)
-    function property(model::ABM)
-        ids = collect(keys(model.agents))
-        properties = [Agents.get_data(model[id], p) for id in ids]
-        s = sortperm(properties)
-        return ids[s]
+struct by_property{P}
+    p::P
+    properties::Vector{Float64} # TODO: don't assume Float64
+    ids::Vector{Int}
+    perm::Vector{Int}
+end
+
+by_property(p::P) where {P} = by_property{P}(p, Float64[], Int[], Int[])
+
+function (sched::by_property)(model::ABM)
+    get_ids!(sched.ids, model)
+    resize!(sched.properties, length(sched.ids))
+
+    for (i, id) in enumerate(sched.ids)
+        sched.properties[i] = Agents.get_data(model[id], sched.p)
     end
+
+    initialized = true
+    if length(sched.perm) != length(sched.ids)
+        resize!(sched.perm, length(sched.ids))
+        initialized = false
+    end
+
+    sortperm!(sched.perm, sched.properties; initialized)
+    return Iterators.map(i -> sched.ids[i], sched.perm)
 end
 
 
+
+struct by_type
+    shuffle_types::Bool
+    shuffle_agents::Bool
+    type_inds::Dict{DataType,Int}
+    ids::Vector{Vector{Int}}
+end
+
 """
-    Schedulers.by_type(shuffle_types::Bool, shuffle_agents::Bool)
-A scheduler useful only for mixed agent models using `Union` types.
+    Schedulers.by_type(shuffle_types::Bool, shuffle_agents::Bool, agent_union)
+A scheduler useful only for mixed agent models using `Union` types (`agent_union`).
 - Setting `shuffle_types = true` groups by agent type, but randomizes the type order.
 Otherwise returns agents grouped in order of appearance in the `Union`.
 - `shuffle_agents = true` randomizes the order of agents within each group, `false` returns
 the default order of the container (equivalent to [`Schedulers.fastest`](@ref)).
 """
-function by_type(shuffle_types::Bool, shuffle_agents::Bool)
-    function by_union(model::ABM{S,A}) where {S,A}
-        types = Agents.union_types(A)
-        sets = [Integer[] for _ in types]
-        for agent in allagents(model)
-            idx = findfirst(t -> t == typeof(agent), types)
-            push!(sets[idx], agent.id)
-        end
-        shuffle_types && shuffle!(model.rng, sets)
-        if shuffle_agents
-            for set in sets
-                shuffle!(model.rng, set)
-            end
-        end
-        vcat(sets...)
-    end
+function by_type(shuffle_types::Bool, shuffle_agents::Bool, agent_union)
+    types = Agents.union_types(agent_union)
+    return by_type(
+        shuffle_types,
+        shuffle_agents,
+        Dict(t => i for (i, t) in enumerate(types)),
+        [Int[] for _ in 1:length(types)]
+    )
 end
 
 """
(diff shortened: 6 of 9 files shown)


examples/schelling.jl
```diff
@@ -97,7 +97,7 @@ function initialize(; numagents = 320, griddims = (20, 20), min_to_be_happy = 3,
     rng = Random.MersenneTwister(seed)
     model = ABM(
         SchellingAgent, space;
-        properties, rng, scheduler = Schedulers.randomly
+        properties, rng, scheduler = Schedulers.randomly()
```
Datseris, line 100: Unfortunately this is a **massively breaking change** that we simply cannot do without releasing 6.0. I'd recommend to make a new scheduler `Randomly`.

Datseris: (and similarly for all other schedulers that need to change)

Datseris (changes requested):

examples/schelling.jl
AayushSabharwal: Isn't that still breaking? Or should the old schedulers also be kept, with something like a warning to use the new ones?

Datseris: Oh yes, sorry I wasn't clear. ALl old schedulers need to be kept exactly as they are :D 

Warning would be nice, sure.

AayushSabharwal:
The only consistent way to add a warning (without printing it _every step_) would be to do it in the `ABM` constructor, which feels a bit dirty. Should I add that or skip it altogether?

Datseris:
Adding it in the ABM constructor is totally valid and a good idea. I don't see a problem with it.

Datseris:
@AayushSabharwal I've added one more todo that we need to add a benchmarkng file in the benchmarks that shows that indeed this is more performant!

AayushSabharwal:
I've never taken a look at our benchmarks before. It'll take some reading up, but it doesn't look particularly difficult.

Datseris:
I think it is useful to prioritize getting some benchmarks in before doing the rest of really large changes. I believe you that this is faster, but it is important to have the numbers as well, and to also see by how much it is faster, and whether there are cases where it isn't faster.

AayushSabharwal:
Schelling with `Schedulers.randomly`:
```julia
julia> @benchmark step!($(old), as, 1000)
BenchmarkTools.Trial: 600 samples with 1 evaluation.
 Range (min … max):  7.807 ms …  11.291 ms  ┊ GC (min … max): 0.00% … 25.70%
 Time  (median):     8.060 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.329 ms ± 611.080 μs  ┊ GC (mean ± σ):  1.43% ±  4.69%

     ▃█▃
  ▂▆▇███▅▄▃▃▂▃▃▃▃▅▄▄▄▃▃▃▂▁▁▂▂▂▂▂▂▂▂▂▂▁▁▂▁▁▂▂▁▂▂▁▁▁▂▂▂▂▂▂▂▃▂▂▂ ▃
  7.81 ms         Histogram: frequency by time        10.7 ms <

 Memory estimate: 2.56 MiB, allocs estimate: 1003.
```

Schelling with `Schedulers.Randomly()`
```julia
julia> @benchmark step!($(now), as, 1000)
BenchmarkTools.Trial: 611 samples with 1 evaluation.
 Range (min … max):  7.896 ms …  10.365 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     8.063 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.191 ms ± 315.656 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▁▂▂█▆▁
  ▅▄▆███████▃▂▄▃▃▄▃▄▃▃▃▃▃▃▃▂▂▂▃▃▂▂▂▃▃▃▂▂▃▂▂▂▃▂▂▂▃▃▂▁▂▃▁▁▂▂▂▁▂ ▃
  7.9 ms          Histogram: frequency by time        9.23 ms <

 Memory estimate: 176 bytes, allocs estimate: 3.
```

---

Schelling with `Schedulers.by_id`:
```julia
julia> @benchmark step!($(old), as, 1000)
BenchmarkTools.Trial: 659 samples with 1 evaluation.
 Range (min … max):  6.770 ms … 15.487 ms  ┊ GC (min … max): 0.00% … 26.63%
 Time  (median):     7.110 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   7.583 ms ±  1.065 ms  ┊ GC (mean ± σ):  3.15% ±  6.89%

   ▂▅█▇▁
  ▃█████▄▃▄▃▃▄▄▃▃▄▃▃▃▃▃▃▂▂▂▂▂▂▁▂▂▃▃▃▄▃▂▂▃▂▂▂▃▂▂▂▂▃▂▁▃▂▁▁▁▁▂▂ ▃
  6.77 ms        Histogram: frequency by time        11.1 ms <

 Memory estimate: 5.13 MiB, allocs estimate: 2003.
```

Schelling with `Schedulers.ByID`:
```julia
julia> @benchmark step!($(now), as, 1000)
BenchmarkTools.Trial: 779 samples with 1 evaluation.
 Range (min … max):  6.120 ms …   7.452 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     6.360 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   6.421 ms ± 203.978 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

          ▁▁▃▂▄▇█▅▃
  ▃▃▃▅▃▅▆▆█████████▆▄▃▄▅▃▄▄▄▅▃▃▃▃▄▃▃▄▃▃▄▃▃▃▃▃▃▁▃▂▃▃▂▂▃▃▁▂▂▃▂▂ ▃
  6.12 ms         Histogram: frequency by time        7.11 ms <

 Memory estimate: 176 bytes, allocs estimate: 3.
```

---

Flocking with `Schedulers.partially(0.7)`:
```julia
julia> @benchmark step!($(old), as, 1000)
BenchmarkTools.Trial: 4 samples with 1 evaluation.
 Range (min … max):  1.440 s …    1.774 s  ┊ GC (min … max): 16.51% … 25.37%
 Time  (median):     1.536 s               ┊ GC (median):    17.92%
 Time  (mean ± σ):   1.571 s ± 144.305 ms  ┊ GC (mean ± σ):  19.70% ±  4.21%

  █          █         █                                   █
  █▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  1.44 s         Histogram: frequency by time         1.77 s <

 Memory estimate: 1.71 GiB, allocs estimate: 23831258.
```

Flocking with `Schedulers.Partially(0.7)`:
```julia
julia> @benchmark step!($(now), as, 1000)
BenchmarkTools.Trial: 6 samples with 1 evaluation.
 Range (min … max):  829.740 ms …    1.542 s  ┊ GC (min … max): 14.54% … 23.66%
 Time  (median):     882.250 ms               ┊ GC (median):    17.00%
 Time  (mean ± σ):   993.958 ms ± 274.066 ms  ┊ GC (mean ± σ):  18.01% ±  5.61%

  ▁█     ▁    ▁                                               ▁
  ██▁▁▁▁▁█▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  830 ms           Histogram: frequency by time          1.54 s <

 Memory estimate: 970.84 MiB, allocs estimate: 11951443.
```

---

Schelling with `Schedulers.by_property(:group)`:
```julia
julia> @benchmark step!($(old), as, 1000)
BenchmarkTools.Trial: 288 samples with 1 evaluation.
 Range (min … max):  16.124 ms … 33.498 ms  ┊ GC (min … max): 0.00% … 46.68%
 Time  (median):     16.601 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   17.402 ms ±  2.386 ms  ┊ GC (mean ± σ):  1.81% ±  6.73%

  █▆▃▃▄▃▁   ▂▂
  ███████▆▅███▅▄▇▆▁▁▄▁▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▆ ▅
  16.1 ms      Histogram: log(frequency) by time      31.2 ms <

 Memory estimate: 10.42 MiB, allocs estimate: 6003.
```

Schelling with `Schedulers.ByProperty(:group)`:
```julia
julia> @benchmark step!($(now), as, 1000)
BenchmarkTools.Trial: 280 samples with 1 evaluation.
 Range (min … max):  17.212 ms …  25.579 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     17.517 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   17.853 ms ± 937.090 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂▁▂█
  ████▇▃▄▆▄▃▄▄▅▅▄▄▃▃▄▃▃▂▁▂▃▁▂▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▁▁▁▁▁▂▁▁▁▁▂▁▃ ▃
  17.2 ms         Histogram: frequency by time         21.6 ms <

 Memory estimate: 15.83 KiB, allocs estimate: 1003.
```

I'll work on something for `by_type`/`ByType`

AayushSabharwal:
> I am not sure what's the number of agents used in these examples

The defaults for the models used.

> compare once with a model that has very little amount of agents, and once with a model that has a huge amount. Just to be extra safe. Perhaps pick two of the schedulers and make a 10x10 schelling and then a 1000x1000 and compare.

> Also, please put the benchmark files explicitly in the benchmark folder in an appropriate subfolder.

Sure.

AayushSabharwal:
Should the benchmarks be added to the same suite or in a different file and different suite.

Datseris:
same suite, but new _group_. There is only one benchmark suite to run with PkgBenchmarks, but it can have an arbitrary amount of groups. 

(And yes, a new file that is included in the benchmark suite file)

AayushSabharwal:
https://gist.github.com/AayushSabharwal/a6e0845c8d138d610ead6534f8e7e1a8

Here's the benchmark results

Datseris:
thanks a lot! Okay, I can see an improvement, but it isn't massive, right? It seems to me like a 10% boost at best, or am I missing something?

AayushSabharwal:
I wonder why the difference in my manual benchmarks was so much larger. I definitely ran code to precompile everything.

benchmark/schedulers/schedulers.jl
```diff
@@ -0,0 +1,58 @@
...
+        (Schedulers.fastest, "fastest"),
+        (Schedulers.by_id, "by_id"),
+        (Schedulers.ByID(), "ByID"),
+        (Schedulers.randomly, "randomly"),
+        (Schedulers.Randomly(), "Randomly"),
+        (Schedulers.by_property(:group), "by_property"),
+        (Schedulers.ByProperty(:group), "ByProperty"),
+    ]
+    model, agent_step!, model_step! = schelling_with_scheduler(; scheduler)
+    SCHED_SUITE[name] = @benchmarkable step!($model, $agent_step!, $model_step!)
```
Datseris, line 17: Sorry for taking so long, I did an eye surgery. I can see barely now, I have 200% zoom on firefox. Alrighty, so, here is the thing. These benchmarks are inappropriate. They do not time the amount of time necessary to do scheduling, which is what their name suggests. They time the amount of time necessary to do schedulling and stepping. Now, if stepping takes 99.9% of time, then even if you've made schedulling 100x faster, it will barely show in the benchmarks. Furthermore, it is just in general good practice to benchmark _only_ the thing you care about. Functions such as nearby agents (which are in `step!` here) are benchmarked elsewhere.

So, can you please do the appropriate changes to the benchmarks? Given that in the end of the day all benchmark operations are `scheduler(model)`, you can reduce the source code of this file a lot as well via loop :)

AayushSabharwal: I really hope you're okay, and that you get better soon.

I haven't had the time to look at this either, since I have exams coming up shortly. I'll try to get to it in a week or so, when they're mostly over. Thanks for pointing out the issues here, I guess I have stuff to learn when it comes to writing benchmarks! :P I'll fix these as soon as I can.

Datseris: Things are okay, I had some small complications during the surgery and have an inflammation, but the doctor says nothing to worry. But then again I don't trust doctors :D but all is well :)

AayushSabharwal:
```julia
function add_agent!(A::Type{<:AbstractAgent}, model::ABM, properties...; kwargs...)
    add_agent!(random_position(model), A, model, properties...; kwargs...)
end
```
It appears this signature of `add_agent!` (L294 in `space_interaction_API.jl`) can't be used with `nothing` space since it looks for a random position

Datseris:
okay, but that seems irrelevant with this PR and also easy to fix right? We add another method
```julia
function add_agent!(A::Type{<:AbstractAgent}, model::ABM{Nothing}, properties...; kwargs...)
    add_agent!(A, model, properties...; kwargs...)
end
```

(_not_ in this PR, in a different one!)

AayushSabharwal:
> okay, but that seems irrelevant with this PR and also easy to fix right?

Yes, I just thought I'd mention it. Best open an issue

Datseris:
Seems like `Randomly` is significantly slower than `randomly` in the benchmarks

AayushSabharwal:
It seems that no matter what I do it doesn't count GC time. Allocs are measured, but somehow even with 65MB of allocations its not counting the GC time.

Datseris:
I don't understand what the problem is or where we are stuck here and what is needed to progress. Can you please explain?

AayushSabharwal:
The benchmarks currently measure the time that each scheduler function takes. However, the benefit of reducing allocations is also that it doesn't trigger the GC nearly as often. As far as I can tell, the benchmarks are not measuring the GC time.

benchmark/benchmarks.jl
```diff
@@ -85,8 +85,8 @@ SUITE["graph"]["add_union"]["agent"] =
     @benchmarkable add_agent!($graph_agent, $graph_union_model) samples = 100
 SUITE["graph"]["add_union"]["agent_pos"] =
     @benchmarkable add_agent_pos!($graph_agent, $graph_union_model) samples = 100
-SUITE["graph"]["add_union"]["agent_single"] =
-    @benchmarkable add_agent_single!($graph_agent, $graph_union_model) samples = 100
+# SUITE["graph"]["add_union"]["agent_single"] =
+#     @benchmarkable add_agent_single!($graph_agent, $graph_union_model) samples = 100
```
Datseris, line 89: Why is this commented?

Datseris:
ok, so the new design is "better" then. But by very little, right? What do we do with the randomly, which seems to be worse in the new version?

benchmark/benchmarks.jl
AayushSabharwal: It wasn't working, I haven't looked into it yet

AayushSabharwal:
> so the new design is "better" then. But by very little, right?

For some cases, it's better (`by_type`/`ByType`). In other cases, it still allocates less, which should still improve performance, although we can't really measure it yet since GC time doesn't want to show up in benchmarks. If I run the same thing manually, there's a notable difference:
```julia
julia> m = fake_model(scheduler = Schedulers.randomly)
AgentBasedModel with 500 agents of type FakeAgent
 space: nothing (no spatial structure)
 scheduler: randomly

julia> @benchmark Agents.schedule($m)
BenchmarkTools.Trial: 10000 samples with 3 evaluations.
 Range (min … max):   8.818 μs … 680.257 μs  ┊ GC (min … max): 0.00% … 95.55%
 Time  (median):      9.609 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   10.153 μs ±   9.472 μs  ┊ GC (mean ± σ):  1.27% ±  1.35%

        ▂▅▇█▆▂
  ▁▁▁▂▃▇██████▇▄▃▂▂▁▁▁▁▁▁▂▃▃▄▄▄▃▃▃▃▃▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  8.82 μs         Histogram: frequency by time         13.4 μs <

 Memory estimate: 4.06 KiB, allocs estimate: 1.

julia> m = fake_model(scheduler = Schedulers.Randomly())
AgentBasedModel with 500 agents of type FakeAgent
 space: nothing (no spatial structure)

julia> @benchmark Agents.schedule($m)
BenchmarkTools.Trial: 10000 samples with 4 evaluations.
 Range (min … max):  7.466 μs …  13.487 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     7.992 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.015 μs ± 269.025 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

               ▁▂▃▆▆▇▇██▇▅▅▄▃▁
  ▂▂▂▂▂▃▃▄▄▄▆▇█████████████████▇▆▅▅▃▃▃▃▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▄
  7.47 μs         Histogram: frequency by time        8.95 μs <

```

Datseris:
ok so even `Randomly` is vetter. Cool. So we need documentation + deprecations and then this is good to go!

docs/src/api.md
```diff
@@ -228,10 +228,15 @@ Schedulers
 Some useful schedulers are available below as part of the Agents.jl API:
 ```@docs
 Schedulers.fastest
+Schedulers.ByID
 Schedulers.by_id
```
Datseris, line 232: Notice that deprecated things should not be in the public documentation! (but should still be exported and documented)

src/Agents.jl
```diff
@@ -59,22 +59,8 @@ if display_update
...
-              open street map spaces, which is integrated in all interactive apps as well!
-            * The majority of examples of the documentation have now moved to a new repository,
-              JuliaDyamics/AgentsExampleZoo.jl. From now on, all additional examples
-              will be hosted in that repo, to keep the main documentation lean.
-            * Functions for planning and moving along routes have had their names
-              unified across Pathfinding and OpenStreetMap modules. The names now are
-              `plan_route!` and `move_along_route!` and are accessible from the top
-              level scope.
+            * Schedulers have been reworked to be more performant and allocate
+              less
```
Datseris, line 63:
```suggestion
            * Schedulers have been reworked to be more performant and allocate
              less. This means that most scheduler names have been replaced from
              functions to types, such as `by_type` -> `ByType`. See changelog for
              full list!
```

Datseris: Can you also add full list of changed names in changelog please?

Datseris (approved):
Legiiiiiiit

PR #622, opened 2022-05-09
added batch_size kwarg to ensemblerun!

mt-digital:
This change enables users to specify the number of simulations in each batch distributed to processors by the `Distributed.pmap` call in `src/simulation/ensemblerun.jl`, when `parallel = true`. Currently, `ensemblerun!` uses the default `pmap` batch size, which is 1. I explain in the documentation that this is useful when each model run finishes quickly compared to the time it takes to run all desired model runs.

I recognized this was a problem by profiling my code that uses `ensemblerun!` and found poor scaling in memory allocations as the number of trials was increased. With a more appropriate `batch_size` setting I was able to get `ensemblerun!` performance to scale linearly with the number of trials.

### DIFF
--- a/src/simulations/ensemblerun.jl
+++ b/src/simulations/ensemblerun.jl
@@ -14,7 +14,9 @@ value counting the ensemble member. The function returns `agent_df, model_df, mo
 
 The keyword `parallel = false`, when `true`, will run the simulations in parallel using
 Julia's `Distributed.pmap` (you need to have loaded `Agents` with `@everywhere`, see
-docs online).
+docs online). Use keyword `batch_size` to specify the `pmap` batch size, which can
+reduce parallelization overhead especially when each model run finishes
+quickly relative to the time it takes to run all models.
 
 All other keywords are propagated to [`run!`](@ref) as-is.
 
@@ -30,10 +32,12 @@ function ensemblerun!(
     model_step!,
     n;
     parallel = false,
+    batch_size = 1,
     kwargs...,
 )
     if parallel
-        return parallel_ensemble(models, agent_step!, model_step!, n; kwargs...)
+        return parallel_ensemble(models, agent_step!, model_step!, n, batch_size; 
+                                 kwargs...)
     else
         return series_ensemble(models, agent_step!, model_step!, n; kwargs...)
     end
@@ -74,10 +78,11 @@ function series_ensemble(models, agent_step!, model_step!, n; kwargs...)
     return df_agent, df_model, models
 end
 
-function parallel_ensemble(models, agent_step!, model_step!, n; kwargs...)
+function parallel_ensemble(models, agent_step!, model_step!, n, batch_size; kwargs...)
     all_data = pmap(
         j -> run!(models[j], agent_step!, model_step!, n; kwargs...),
-        1:length(models),
+        1:length(models);
+        batch_size = batch_size
     )
 
     df_agent = DataFrame()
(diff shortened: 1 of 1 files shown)


Datseris:
Thank you very much. I admit, I know little about parallelization, so can you please clarify: You are asupposed to **increase** `batch_size` when the model run finishes quickly, or the opposite? Would be nice to say this more clearly in the docstring by explicitly using the word "increase" or "decrease".

src/simulations/ensemblerun.jl
```diff
@@ -74,10 +78,11 @@ function series_ensemble(models, agent_step!, model_step!, n; kwargs...)
     return df_agent, df_model, models
 end
 
-function parallel_ensemble(models, agent_step!, model_step!, n; kwargs...)
+function parallel_ensemble(models, agent_step!, model_step!, n, batch_size; kwargs...)
     all_data = pmap(
         j -> run!(models[j], agent_step!, model_step!, n; kwargs...),
-        1:length(models),
+        1:length(models);
+        batch_size = batch_size
```
Datseris, line 85:
```suggestion
        batch_size
```

Datseris: (This is exactly the reason we require julia 1.5+)

mt-digital: My pleasure! Thank you for the excellent library. I am an avid Agents.jl user and evangelist and was hoping to find a way to contribute.

I agree, the explanation of `batch_size` was unclear in my message here and in the documentation. I have expanded the docstring with an example. I think it's good to include more detail since this is an important parameter that has yielded big performance gains for me. 

Basically `batch_size` should be increased when there are many more models to run than the number of processors available and each model run takes seconds or less to run. It's expensive for worker processors to communicate with the coordinating processor, so we should avoid making the processors do this excessively. Increasing batch size increases the number of models run by each worker before communicating, reducing communication overhead in the situation listed above. I know a bit, but I'm no expert on parallel computing; [this Julia parallel computing tutorial](https://techytok.com/lesson-parallel-computing) had a lot of helpful information for me (some [specifically about `pmap`](https://techytok.com/lesson-parallel-computing/#pmap), which I found via this helpful Discourse thread: https://discourse.julialang.org/t/parallelism-understanding-pmap-and-the-batch-size-parameter/15604.

I also fixed the `batch_size` kwarg redundancy you suggested with the latest push.

Datseris: If you want to contribute more, here is something that you can do with your expertise: add a new section in the Performance Tips page, that shows the impact of this parameter. Set up some real simple model where you can produce a plot of overall run time versus the `batch_size` keyword, and present the plot.

(if producing such a plot takes too much time, you can produce it and save it and then link it to the page instead)

Datseris: BTW, in the Performance Tips section you can have this more detailed description of the `batch_size` that you have here. It is just that my philosophy in documentation strings is that every sentence much be weighted in gold!

mt-digital: No problem, I had a feeling it was too much. I updated the docs as you suggested.

I'd be happy to add a new section in Performance Tips. The Schelling model comes to mind as an example model, since that example uses `ensemblerun!`. Would a demonstration about different `batch_size` choices be useful in the Schelling example, or maybe just a one-sentence note about it in Schelling, with a link to the Performance Tips page?

Datseris: One sentence about it in Schelling, and linking to Performance Tips is a better approach I feel.

mt-digital: fyi, still working on this, just doing some more tests and studying to make sure `batch_size` is the best way to accomplish this. I'm looking into approaches that automatically handle task batches sent to the workers. These seem to be greatly outperforming the current implicit default `batch_size = 1`, and could possibly save users from having to think about what batch size to set.

Datseris: thanks a lot, that is great!

mt-digital: So I tried a different approach I think will work better and opened a new PR since it's a bit different solution: #624. Hope that's ok!

Datseris:
#624 superseeded this one

PR #623, opened 2022-05-26
Add progress meter to `run!`

### LINKED ISSUE #542: Progress meter
I vaguely remember there used to be an option to print the progress of a simulation in the `run!` function. Am I correct? Any reason it has been taken out? This is an important feature for heavy simulations.

kavir1698:
Closes #542

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati"]
-version = "5.3.0"
+version = "5.3.1"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati"]
-version = "5.3.1"
+version = "5.4.0"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -113,13 +113,13 @@ function run!(
     agent_step!,
     model_step!,
     n;
-    when=true,
-    when_model=when,
-    mdata=nothing,
-    adata=nothing,
-    obtainer=identity,
-    agents_first=true,
-    showprogress=true
+    when = true,
+    when_model = when,
+    mdata = nothing,
+    adata = nothing,
+    obtainer = identity,
+    agents_first = true,
+    showprogress = true
 )
 
     df_agent = init_agent_dataframe(model, adata)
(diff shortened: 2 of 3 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati"]
-version = "5.3.0"
+version = "5.3.1"
```
Datseris, line 4: This should be 5.4.0 because it adds a new feature, it does not fix a bug.

src/simulations/collect.jl
```diff
@@ -112,12 +113,13 @@ function run!(
     agent_step!,
     model_step!,
     n;
-    when = true,
-    when_model = when,
-    mdata = nothing,
-    adata = nothing,
-    obtainer = identity,
-    agents_first = true,
+    when=true,
```
Datseris, line 116: I don't understand these changes. It changes the more readable format of `x = y` to the less readable one of `x=y`. Regardless, even though this style can be debated of which one is better, such formatting style changes shouldn't be done in a PR that adds a new feature, but in an independent one.

src/simulations/collect.jl
```diff
@@ -145,6 +148,7 @@ function run!(
         end
         step!(model, agent_step!, model_step!, 1, agents_first)
         s += 1
+        next!(p)
```
Datseris, line 151: Do `showprogress && next!(p)` here.

src/simulations/collect.jl
```diff
@@ -136,6 +138,7 @@ function run!(
     end
 
     s = 0
+    p = showprogress ? typeof(n) <: Int ? ProgressMeter.Progress(n, 1) : [0] : nothing
```
Datseris, line 141: This is not the correct way to use the `ProgressMeter` package. Instead you need to initialize a proper progress meter as indicated by the package's API. `p = ProgressMeter.Progress(n, 0, "Performing `Agents.run!`...")` and later  down the line call `next!(p)`.

Datseris: Furthermore, the ProgressMeter.jl package has a `ProgressUnknown` type that we can use when `n` is a function: https://github.com/timholy/ProgressMeter.jl#progress-meters-for-tasks-with-an-unknown-number-of-steps

src/simulations/collect.jl
```diff
@@ -510,3 +514,20 @@ collect_model_data!(df, model, properties::Nothing, step::Int = 0; kwargs...) =
...
+
+
+###################################################
+# ProgressMeter 
+###################################################
+
+"""
+Show number of passed steps when `n` in `run!` is a function.
+"""
+function next!(p::Vector{Int})
```
Datseris, line 526: This is type piracy and definitely not something we can do in a Julia package. It extends a method whose both function, and type of input arguments, we do not own. The ProgressMeter.jl package says how we should use it by initializing a `Progress`.

Datseris (changes requested):
Adding progress meter report to `run!` is a great idea, but at the moment the PR uses ProgressMeter.jl inappropriately. Once the documentation of ProgressMeter.jl has been consulted and the PR has been modified to use the expected API, I'm happy to merge!

src/simulations/collect.jl
kavir1698: This was something Julia addon on VSCode changed automatically. I reverted them.

src/simulations/collect.jl
kavir1698: Thanks for your feedback. I wasn't aware of `ProgressUnknown`.

Datseris (approved):
Awesome! I didn't know that you could set visibility from within the Progress type.

Datseris:
@kavir1698 forgot to ask: did you test this also with parallel runs and it works? And what happens when we do ensemvle runs? Does it show several progress bars in paralell?

kavir1698:
> @kavir1698 forgot to ask: did you test this also with parallel runs and it works? And what happens when we do ensemvle runs? Does it show several progress bars in paralell?

I just checked it on `paramscan`. It prints one progress bar for each simulation. The best solution is to `showprogress` to `false` by default? so that no other change to code is needed.

Datseris:
yeah, then I assume on `ensemblerun` it would do the same, one progressbar per simulation. So it is probably better to have it false by default. Then, in ensemble run, we should have a progressbar that counts the full simulations, not steps.

PR #624, opened 2022-05-27
Use DistributedArrays for faster ensemblerun! parallelism

mt-digital:
This PR evolved from #622 which was attempting to more efficiently use available processors by specifying the `batch_size` kwarg provided by `pmap` as an `ensemblerun!` kwarg. That approach would have forced Agents.jl users to figure out what `batch_size` is optimal, which is not necessarily easy. In this PR I use an alternate approach with the [`DistributedArrays`](https://juliaparallel.org/DistributedArrays.jl/stable/) package that effectively automates setting `batch_size`. In my tests (explained below) I found that the DistributedArray approach outperforms the current `pmap` implementation of `ensemblerun!`. I included both the current `pmap` implementation and my new DistributedArray implementation for now, and will replace `pmap` entirely if you indeed agree that is the correct move.

Using the branch associated with this PR, I ran [this script](https://github.com/mt-digital/JuliaJunk/blob/f28aa721f8c0a8d8785e1849d8851d653f3defff/scripts/schelling_ensemblerun_performance.jl) to test how long different test runs took with the Schelling model, using serial execution, the current parallel execution `pmap` implementation, and a new `darray` version. The Schelling model was run for 100 values of `numagents` and three different `(nreplicates,nsteps)` pairs: `(2, 100)`, `(4,100)`, and `(10,2000)`. Here is how the `models` are constructed that are passed to `ensemblerun!`; `numagents_low = 200` and `numagents_high = 300`:
```julia
basemodels = [Models.schelling(; numagents)[1] 
                          for numagents in numagents_low:numagents_high]
models = repeat(basemodels, nreplicates)
```

The script can be run on the command line, where `-p3` indicates assigning three worker cores for parallelization, like so
```sh
julia -p3 scripts/schelling_ensemblerun_performance.jl
```

Below is the result I obtained on my 8-CPU M1 running Julia 1.7.2 (I obtained similar results on a 2015 MacBook Pro with 4 intel cores on Julia 1.7.1). It shows that with the current implementation of `ensemblerun!`, serial execution outperforms parallel due to excessive data transfer overhead between workers and the coordinating core. However, the `darray` parallel implementation included in this PR is close to serial execution for small numbers of replicates. With ten replicates and 2000 time steps we finally see time savings from the `darray` parallel implementation. This suggests that `darray` is the right approach that will "just work" better at this time. I observed similar speedups in a research model I'm developing. Tests should continue, though, to optimize parallelism in Agents.jl. Happy to clarify or try to simplify anything.
```
ENSEMBLE RUN TEST, n_replicates=2, nsteps=100
serial finished in 0.582458541s
current finished in 6.815945s
darray finished in 1.012393666s
Parallel-to-serial ratio (current): 11.70202601595982
Parallel-to-serial ratio (darray version): 1.738138587961748

ENSEMBLE RUN TEST, n_replicates=4, nsteps=100
serial finished in 0.212536959s
current finished in 11.789640625s
darray finished in 0.234458416s
Parallel-to-serial ratio (current): 55.47101398491356
Parallel-to-serial ratio (darray version): 1.1031418587296151

ENSEMBLE RUN TEST, n_replicates=10, nsteps=2000
serial finished in 9.284566584s
darray finished in 4.186712875s
Parallel-to-serial ratio (darray version): 0.45093250580106986
```


Datseris:
Thank you very much @mt-digital this is a truly wonderful contribution. I read your PR in detail and I agree that we should follow therefore your approach forwards. Here are two comments:

- Please also include a run with many more models, but much less steps, like 500 replicates and 50 steps 
- For completeness, in your original post in this PR, please edit it and include the above run, but also the results of "current" pmap implementation for the 10 replicates and 2000 steps, because these numbers are not included there.
- In the source code you use the `@DArray` macro. Is it possible to change this version so that it doesn't use macros but normal function and type calls, without any loss of performance? if so, that would be preferred, because macros make source code less readable. If it is not possible to achieve same performance without macros, then it is okay to use the `@DArray` macro.

Besides this comments, I'd say go ahead and remove the old version and add necessary documentation if any, and extra tests if any necessary.

Datseris:
Ah also I guess this approach needs to be used also for `paramscan` and `ensemlerun` when parallelization is on?

mt-digital:
Thanks for the feedback. I updated the benchmarks and my initial comment as requested, with some other minor edits.

I do think it's possible to replace the `@DArray` macro; I'll try that in the next few days at most (but hopefully later today).

Regarding `paramscan`, on a first look it appears it won't be possible to keep the progressbar for parallel paramscan and use DistributedArrays. Also, I would probably look to re-use `ensemblerun!` by first initializing an array of models using the `initialize` function argument to `paramscan`, then pass the models array to `ensemblerun!`. I think this would clean up `paramscan.jl` a bit since it could replace the `run_single` helper in both the parallel and serial versions of `paramscan`. 

I think I could restore the progressbar for parallel ensembleruns in a future PR, but for now I would suggest dropping the progressbar for parallel runs. Part of the problem I see is that the [`SharedArrays` of the standard library](https://docs.julialang.org/en/v1/stdlib/SharedArrays/) only allow bits-type shared arrays, which means we can't share `DataFrame` arrays across processors. If `SharedArrays` could be of generic types then we could use a `Distributed` for-loop and track progress there. I believe a `Distributed` for-loop would achieve the same performance as using `DistributedArrays`.

mt-digital:
Regarding the progressbar for parallel runs, I would be willing to work on that for this PR, but it might take some time.

Datseris:
i like your suggestion to clean paramscan via utilizing `ensemblerun!` go ahead with that!

I don't fully understand what is the problem with progress bars of parallel runs: my suggestion would be to add a progress bar that counts the overall amount of runs that have been done, not the individual steps within one run (which is what the progress bar in `run!` does)

Datseris:
oh an also put the benchmarking script you used in a subfolder of the `benchmarks` folder please, for future reference!

mt-digital:
OK on it

Datseris:
I am not sure how long it should take, because there are a lot of stuff there... But several minutes seems reasonable. Several hours does not! :D

Project.toml
```diff
@@ -19,18 +19,18 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
...
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 Scratch = "6c6a2e73-6563-6170-7368-637461726353"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 CSV = "0.9.7, 0.10"
 DataFrames = "0.21, 0.22, 1"
 DataStructures = "0.18"
-Graphs = "1.4"
```
Datseris, line 28: Why is Graphs and ProgressMeter removed from Project...?

Project.toml
```diff
@@ -19,18 +19,18 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
Datseris, line 22: Neither revise nor StableRNGs need to be in project. If StableRNGs needs to be in Project.toml, it must be a test-only dependency.

mt-digital:
Sorry about the messy push. It was just a wip checkpoint, should have used a different branch probably. I'll clean up the dependencies you noted and other stuff.

mt-digital:
Latest push is cleaned up, with everything mostly done. I'll improve documentation and inline comments, then maybe it's close to ready? Let me know if anything is unclear.

test/collect_tests.jl
```diff
@@ -468,42 +516,78 @@ end
...
-            n = n,
+            n,
             model_step! = forest_model_step!,
             include_constants = true,
             mdata,
         )
         # 3 is the number of combinations of changing params,
         # 5 is 3+2, where 2 is the number of constant parameters
-        @test size(data) == ((n + 1) * 3, 5)
+        @test size(mdf) == ((n + 1) * 3, 5)
```
mt-digital, line 540: It seems the additional `paramscan` below is not necessary, I'll remove.

mt-digital:
I'll add to the docs and fix some minor problems by tomorrow. 

The one test on Windows is failing due to an out of memory error on a parallel run, it seems: https://github.com/JuliaDynamics/Agents.jl/runs/6751345338?check_suite_focus=true#step:7:184. Can we bump up the memory or is this fixed? We can reduce memory usage in the test if necessary.

mt-digital:
OK, I'll reduce the parameter set size and we'll see if that fixes it.

The latest push updates the docs and fixes some coding style issues.

mt-digital:
Just pushed updated `ensemblerun!` test with reduced memory usage. Hopefully the Windows test will be happy now.

Datseris:
Thanks, I'll give a final review as soon as I can.

AayushSabharwal (commented):
I'm no expert on parallelism, but this PR looks amazing. The benchmark results look great.

There doesn't seem to be any issue with the code itself, just a few minor changes about clarity and documentation.

Some parts of the code (long function calls/definitions that wrap to multiple lines) are not formatted how I would do it. However, it's still easily readable and I don't want to be picky when I'm not sure what the norm is for this library, so I haven't marked them. @Datseris do we follow any style guide (even roughly) or should code just be clear and readable?

AayushSabharwal:
Oh, and the merge conflicts also need to be looked into.

Datseris:
> @Datseris do we follow any style guide (even roughly) or should code just be clear and readable?

Not yet unfortunately. Soon though I will do a massive PR that restyles all code. For now we can use the SciML one: https://github.com/SciML/SciMLStyle . I agree with almost everything there.

Datseris:
I'm going to finish this up real quick, as I'm under time pressure to get 5.4 out due to an upcoming presentation.

Datseris (approved):
Thank you very much. This is LGTM, and amazing. Thanks!

mt-digital:
Sorry that was just a wip push to save progress, I'll clean up and let you
know when it's done, there's a lot to clean up

Datseris:
@mt-digital I am confused :D This Pull Request is already merged. What is a "wip push" here? or, where is it?

PR #625, opened 2022-05-28
Roll back version once

AayushSabharwal:
I just noticed 5.3 was never tagged, so there's no point in rolling out 5.4 now. I've merged the changelog for both, and after this PR I'll tag it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
-# v5.4
-- New keyword `showprogress` in `run!` function that displays a progress bar.
-
 # v5.3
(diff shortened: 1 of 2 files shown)


Datseris:
Agents 5.3 is released: https://juliahub.com/ui/Packages/Agents/nTsV8/5.3.0  its just that stupid TagBot is not working. It is not creating the automatic releases. I do not know what to do to fix it, noone has replied to me when I asked how to fix it...

I am not tagging 5.4 yet because I want to add the same kind of proigress bar in parameter scanning and ensemble runs

Issue #528, opened 2021-07-11
Update Agents.jl logo to new (Makie.jl based) plotting


Datseris:
Not much to say here, pretty much up-for-grabs. Old code is in `docs\logo.jl`.

PR #626, opened 2022-05-30
Update Agents.jl logo 

jgamper:
Still some work needs to be done, but would appreciate feedback if this is what you meant in #528 @Datseris 

Also, where does one get access to the font that is referred to in the old code? https://github.com/JuliaDynamics/Agents.jl/blob/dbddf76e89176159e44c65090b8a85aead371ea0/docs/logo.jl#L24


Datseris:
By the way, why can't the logo generation simply be the output of running `abmvideo`? It seems to me that it should be as simple as that, once we have defined the model and stepping functions. https://juliadynamics.github.io/Agents.jl/stable/agents_visualizations/#ABM-Videos-1

Libbum:
[Moon2.0](https://github.com/JuliaDynamics/Agents.jl/blob/dbddf76e89176159e44c65090b8a85aead371ea0/docs/logo.jl#L12)

Datseris:
also cc @AayushSabharwal

AayushSabharwal:
Just saw the email notification for your ping. The new logo looks AWESOME. The pink-green contrast is much nicer than the old red-green, and I'm always down for a dark theme.

kavir1698:
> This is also our chance to change the font if we would like. Suggestions are welcome. @Libbum , @fbanning , @kavir1698 you guys should also chime in on the black background business, so far I see 3 votes for, none against!

White text against black background increases the emphasis of the text. Therefore, it's a good choice for the logo. I would change the colors such that the text ends up in white which provides the maximum contrast. Right now it feels like Agents.jl is becoming weaker.

fbanning:
No objection regarding the new background colour. I definitely agree with @kavir1698 's remark about the weird feel that the logo "blends in" with the background over time. We could play around a bit with different constellations to bring about another dynamic. There are endless possibilities to approach this. No idea which one works best.

For example, how about making the dots of the logo have the background colour at first (so it's not visible). Then we have bouncing dots in green and purple all around it. Once the logo dots get hit, they take the colour of the moving dot (green or purple). Once a logo dot got hit twice (or at least once by each colour), it turns white. This might take a longer time to "realise" the logo in the video, so maybe the movement speed should be adjusted. 

Regarding the font, I think it's actually quite nice. However, it only allows for "personal use" in its license. Maybe open source projects are included in this as they're non-commercial but I wouldn't bet on it. Another font with broad strokes that might work well could be Squada One (https://fonts.google.com/specimen/Squada+One?category=Display&thickness=7#standard-styles). It is published under the open font license, therefore definitely no problems with using it.

Datseris:
I am wondering, could we perhaps make an "open street map" version of the logo, with roads connecting and stuff? Honestly sounds too hard, and I don't even know how. @fbanning I like a lot the font you suggested, I will go ahead and use it. I am not sure about the subsequent suggestion, because it goes too much into the billiards regime, and Agents.jl isn't particularly suited for billiard simulations. HardSphereDynamics.jl is much better. The hope was that the logo illustrated the "covid spread" animations people have seen many times. Of course, now with the inverted colors I guess the connection may not be as obvious...

Datseris:
Okay here is the updated version, where white is the recovered color. I also made it so agents that move always die when recovering, so that the figure clears up and reveals the text. I don't think I can get it to make all agents white in the end, unless I actually do it by hand.

https://user-images.githubusercontent.com/19669089/171996200-0d9217fc-de86-4ec5-b6dc-247aec6da2b0.mp4

fbanning:
> Okay here is the updated version, where white is the recovered color. I also made it so agents that move always die when recovering, so that the figure clears up and reveals the text.

This looks insanely good! Awesome work 👌

fbanning (approved):

PR #633, opened 2022-06-13
Add `:manhattan` metric and `manhattan_distance` utility function

### LINKED ISSUE #632: Manhattan distance for grid space
There are only Euclidean distance and Chebyshev distance for the grid space and I'm trying to make code with Manhattan distance. It is convenient to use when you want to include first neighbors and their neighbors in ¨von Neumann¨ configuration.

**Describe alternatives you've considered**
There isn´t a way to do what I´d like, exactly, using existing distances.

fbanning:
Closes #632 

I've added some utility functions to calculate manhattan distance between two agent positions. Should work for both periodic/non-periodic. This allows users to easily use the distance metric in their models.

I have not integrated the manhattan distance as a new `metric` for GridSpace. Would that be necessary?

*Sidenote*: We should maybe change the `periodic` keyword to `toroidal` at some point because that word is a bit more common in this context as far as I know.

## To do:

- [x] Test `manhattan_distance`
- [x] Test `:manhattan` metric 
- [ ] Add plot explaining distances (task for @Datseris)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Manifest.toml
 *.osm.cache
 src/update_*
 agents_visualizations.md
+.vscode
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # v5.4
 - New keyword `showprogress` in `run!` function that displays a progress bar.
+- New `:manhattan` metric for `GridSpace` models.
+- New `manhattan_distance` utility function.
 
 # v5.3
 - Rework schedulers to prefer returning iterators over arrays, resulting in fewer allocations and improved performance. Most scheduler names are now types instead of functions:
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -125,7 +125,8 @@ nearby_agents
 nearby_positions
 random_nearby_id
 random_nearby_agent
-edistance
+euclidean_distance
+manhattan_distance
 ```
 
 ## A note on iteration
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -34,6 +34,11 @@ GraphSpace
 GridSpace
 ```
 
+Here is a specification of how the metrics look like:
+```@example
+include("distances_example_plot.jl") # hide
+```
+
 ### Continuous spaces
 ```@docs
 ContinuousSpace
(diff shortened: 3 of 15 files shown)


gasparalfaro:
I think integrating it as a metric is necessary since I need Von Neuman neighbours and the nearby_agents function uses a metric to define the distance. Thank you!

Datseris:
@fbanning it is easy (I think), to make this a formal metric for `GridSpace`. You just add here an extra clause for it, and the rest of the source code should work out of the box.

https://github.com/JuliaDynamics/Agents.jl/blob/master/src/spaces/grid.jl#L121-L130

(Given `r`, you need to find all Cartesian Indices within distance `r` from origin).

Datseris:
Furthermore, I'd like us to rename `edistance` to `euclidean_distance`, and `mdistance` to `manhattan_distance` in this PR. The `edistance` change needs to be deprecated.

fbanning:
OK, this should work now, I think. For the calculation of the Hood in the case of `:manhattan` metric we can just reuse the existing code from `:euclidean` because it provides the same results for a GridSpace as far as I can tell.

I've deliberately left out the method for `r::NTuple{D,Real}` until somebody explicitly asks for it. It's a bit of a hassle to implement this and I would rather spend my time on other things in the meantime. ;-)

fbanning:
Ah no, this doesn't work for distances >= 3. OK, let me write up a separate filter for `:manhattan` then. :)

fbanning:
Generic functions for arbitrary dimensions is always better and generally what a package should aim for such that its functionality can be reused in as many ways as possible by its users. (Wow, that was a mouthful of a sentence. 😆 )

I suppose I should also add some tests for :`manhattan` metric in general, right?

Datseris:
yes, tests are need for the new metric. You only need to test `nearby_ids` and `nearby_positions`.

Datseris (approved):
LGTM, thanks a lot for the contribution!

Datseris:
@fbanning if you don't mind I won't merge this yet. I want to add a plot to the documentation that displays the three distances, so that people that come from this "Moore and Von Neumann" kind of background terminology can understand what is going on.

fbanning:
> @fbanning if you don't mind I won't merge this yet. I want to add a plot to the documentation that displays the three distances, so that people that come from this "Moore and Von Neumann" kind of background terminology can understand what is going on.

That's a nice idea. Do you have time for this or should I take care of it?

Datseris:
I plan to do it on the weekend where I have a lot of free time allocated for software dev!

PR #638, opened 2022-06-17
Fix `random_nearby_id`

AayushSabharwal:
It's basically just moving a couple lines down to fix the bug where the first element of the iterator never gets picked.


Datseris:
We should add a test for this as well. Make a setting where an agent has very little amount of neighbors, 2 (make a model with three agents) and loop over calling this function some times and ensure that all ids are picked.

AayushSabharwal:
Thanks for the test idea. Fixed another bug with the algorithm.

Datseris (approved):
LGTM, thanks for the fix!

Issue #634, opened 2022-06-13
Reform tests according to Good Scientific Code practices


Datseris:
I am preparing a huge workshop that I'll give next month on Good Scientific Code practices. (will upload to youtube and share here)

Unfortunately, our tests miss some good practices regarding a good unit test suite. Specifically:

- They are not "clean slate", which dictates that every test file should be runnable as-is, by itself, without depending on the global state, and without leaving something on the global state to be used later. Obviously, our test suite completely fails on that given the massive amount of definitions in `runtests.jl` and the lack of `using` statements in the individual files.
- They aren't organized into subfolders for better navigation. In fact, in general the organization of tests suffers with weird file names and content. E.g., what would you expect to be tested in the file `graph_tests.jl`? I'd expect all of `GraphSpace` features, but that's nowhere near the case. Each space should have its own test file.
- They rely way too much on random numbers, where we should be writing analytic tests instead. Deterministic, analytically resolvable tests are the most accurate ones. The expected outcome of a test comes from pen and paper instead of running code. Right now we have a large amount of tests that were generated like "seed an rng, run some model code, write down the result as a test suite.". What we should be doing instead is "devise a distribution of agents. Analytically resolve their neighbors and actions given some rules and distances. Write tests that test again this analytically resolved output." I recall we had problems many times throught the lifespan of Agents.jl with random tests being broken etc. (obviously, random stuff like the RNG in the model should also be tested, but if you can avoid randomness in the tests, you should)
- They are not **computationally minimal**, which means that  you should be doing the least amount of computer operations to confirm whether some functionality works. E.g., look at the start of `continuousSpace_tests.jl`. 8 spaces are initialized, but as far as the subsequent tests are concerned, you only need 3.
- They don't use [SafeTestsets](https://github.com/YingboMa/SafeTestsets.jl). In the `runtests.jl`, the test files should be included like 
  ```
  @safetestset "Benchmark Tests" begin include("benchmark_tests.jl") end
  ```
  and the only modules actually used in the `runtests.jl` file should be `Test, SafeTestsets`.

hbsmith:
@Datseris I'd be interested to know what was discussed at the workshop if you have more information on that

Datseris:
Here's the workshop: https://github.com/JuliaDynamics/GoodScientificCodeWorkshop

Datseris:
here's a checklist with what properties good tests should have: https://github.com/JuliaDynamics/ChaosTools.jl/issues/264

[...]

PR #643, opened 2022-06-27
Optimized grid space with only 1-agent-per-position

### LINKED ISSUE #640: Simplest Schelling implementation about 2x as fast as our example
Below I will attach two code snippets that create an ABM of the Schelling model. One is pure Julia, written in as simple possible way, and the other is in Agents.jl. As you will see, the pure/simple Julia version is about 2x fast. Before anyone gets too panicked: **it is completely expected that the pure version is faster**. Many reasons exist, the ones that come to mind immediatelly: 
1. simple version is actually a cellular automaton, so it utilizes a defacto knowledge of one agent per position (this gives many benefits such as much more efficient iteration over neighbors because of dropping nested `flatten` calls)
2. Due to the above, the simple version does not need the two extra fields of id and position in the agent struct.
3. simple version doesn't do generalized neighborhood searches, does a hardcoded Moore neighborhood.

Still, the pieces of code should be compared in detail. While it is true that we cannot expect to match the pure performance, we can ask: how close can we get? Which are the parts that actually differ more? 

(I haven't analyzed the codes in detail yet and just compared stepping. It is a todo to see in more detail which parts has the most performance difference. Is it finding neighbors? Is it moving agents?)

Here is the code:

```julia
using Random # reproducibility
# shared constants
const min_to_be_happy = 3 # how many nearby agents you need of same group
const grid_occupation = 0.8 # percentage of space occupied by agents
const grid_size = (30, 30)

######## PURE VERSION
# model is a dictionary, mapping filled positions (i, j) to agents
Random.seed!(1234)
mutable struct ExampleAgent
    group::Int
    happy::Bool
end

function initialize_pure()
    model = Dict{NTuple{2, Int}, ExampleAgent}()
    N = grid_size[1]*grid_size[2]*grid_occupation
    for n in 1:N
        group = n < N / 2 ? 1 : 2
        pos = (rand(1:grid_size[1]), rand(1:grid_size[2]))
        while haskey(model, pos)
            pos = (rand(1:grid_size[1]), rand(1:grid_size[2]))
        end
        agent = ExampleAgent(group, false)
        model[pos] = agent
    end
    return model
end

function simulation_step!(model)
    # crucial to collect keys here
    filled_positions = collect(keys(model))
    for pos in filled_positions # loop over all locations on grid
        agent = model[pos]
        same = 0
        neighboring_indices = [pos .+ n for n in neighborhood]
        for npos in neighboring_indices
            if !haskey(model, npos)
                continue
            end
            neighbor = model[npos]
            if neighbor.group == agent.group
                same += 1
            end
        end
        if same ≥ min_to_be_happy
            agent.happy = true
        else # generate new unoccupied position
            newpos = (rand(1:grid_size[1]), rand(1:grid_size[2]))
            while haskey(model, newpos) # careful here, use `newpos`, not `pos`!
                newpos = (rand(1:grid_size[1]), rand(1:grid_size[2]))
            end
            model[newpos] = agent
            delete!(model, pos)
        end
    end
end

model_pure = initialize_pure()
println("benchmarking pure step")
@btime simulation_step!($model)

######## AGENTS.JL VERSION
mutable struct SchellingAgent <: AbstractAgent
    id::Int
    pos::NTuple{2, Int} # Notice that position type depends on space-to-be-used
    group::Int
    happy::Bool
end

function initialize_agents()
    space = GridSpace(grid_size; periodic = false)
    properties = Dict(:min_to_be_happy => min_to_be_happy)
    rng = Random.Xoshiro(1234)
    model = ABM(SchellingAgent, space; properties, rng)
    N = grid_size[1]*grid_size[2]*grid_occupation
    for n in 1:N
        group = n < N / 2 ? 1 : 2
        agent = SchellingAgent(n, (1, 1), group, false)
        add_agent_single!(agent, model)
    end
    return model
end

    nearby_same = 0
    for neighbor in nearby_agents(agent, model)
        if agent.group == neighbor.group
            nearby_same += 1
        end
    end
    if nearby_same ≥ model.min_to_be_happy
        agent.happy = true
    else
        move_agent_single!(agent, model)
    end
    return
end

model_agents = initialize_agents()
println("Benchmarking Agents.jl version")
@btime step!($model_agents, agent_step!)
```

with output
```
benchmarking pure step
  140.000 μs (721 allocations: 146.38 KiB)
Benchmarking Agents.jl version
  288.500 μs (5448 allocations: 255.38 KiB)
```

Datseris:
Closes #640

This PR is very ambitious does the following:

- Completely rework internals of `GridSpace` and remove `Hood` stuff.
- Significantly increase performance of nearby searches of `GridSpace` using a dedicated iterator struct.
- Introduce a new space, `GridSpaceSingle` that enforces only 1 agent per position.
- Re-write the tests for grid spaces in accordance to #634 
- Resolve all problems with COntinuousSpace with new internals, make it rely as much as possible in public API.
- Re-write the tests for continouos space in accordance to #634
- Turns out simply filtering all distances is much faster because its type stable, that's what the exact version now does in continuous space
- See CHANGELOG changes for more

---

**This top level comment will be edited with latest information when necessary.**

---

I've pushed a version that uses `SoloGridSpace` that uses an Array{Int} to store agent IDs, using ID = 0 as absence of agents. I made it a formal part of the space API. I've added the Schelling model in the benchmarks for this version. Current benchmarks are:

```
Benchmarking GridSpace version
  202.800 μs (0 allocations: 0 bytes)
Benchmarking GridSpace version: count nearby same
  123.804 ns (0 allocations: 0 bytes)
Benchmarking GridSpaceSingle version
  76.300 μs (0 allocations: 0 bytes)
Benchmarking GridSpaceSingle version: count nearby same
  62.921 ns (0 allocations: 0 bytes)
Benchmarking abusive unkillable version
  14.900 μs (0 allocations: 0 bytes)
Benchmarking abusive unkillable version: count nearby same
  42.929 ns (0 allocations: 0 bytes)
```

---

For reference, the very first benchmarks were:
```
Benchmarking GridSpace version
  227.200 μs (5448 allocations: 255.38 KiB)
Benchmarking SoloGridSpace version
  202.300 μs (0 allocations: 0 bytes)
```


Datseris:
cc @AayushSabharwal

Datseris:
Update, I've added benchmarking of exclusively timing the time spent into counting nearby IDs. It seems most of the timing discrepency is coming from there. We have:
```
Benchmarking SoloGridSpace count nearby same
  256.338 ns (1 allocation: 80 bytes)
Benchmarking abusive unkillable version: count nearby same
  37.399 ns (1 allocation: 32 bytes)
```
it seems like our code for finding nearby agents is highly problematic...?

Datseris:
Perhaps this whole `Hood` business is too complex. Also, there is no actual reason to use `CartesianIndex`, because we exclusively use Tuples for positions. I think I'll write down some quick code that just makes a vector of Tuples as the neighborhood, just like the `moore` struct in our code, and then use the approach of `checkbounds` that @AayushSabharwal used.

Datseris:
I've coded the version of neighborhoods that just uses `Vector{NTuple}`. It is much faster. Total runtime is now multiplied with a factor of 0.75, which is much better. Now we have 143 μs for new SoloGridSpace, while the abusive version is a far fetched dream of 14.9 μs. Counting nearby same neighbors is still the reason for the performance deficit, with 163.241 ns for the SoloGrid version and 37.8 ns for the abusive version.

I do not believe we can reach the abusive version while maintaining a high quality, no-code-duplication codebase. Here are some notes:

-  I'll completely re-write the `GridSpace` searches to use the new `Vector{NTuple}` approach and `checkbounds`, since this is much faster.
- When searching for nearby ids, we have an additional filtering step that filters out the id of the agent that we started from. To get rid of this step we would have to write both a `nearby_ids(agent...)` and a `nearby_ids(pos...)` function for `SoloGridSpace`. I guess we should do that.
- We have essentially two nested calls to `Iterators.filter`: one for existing agents in position (i.e., id is not 0), and one for checking bounds for the array. In the abusive version these are hardcoded and merged into one filtering step.

---

**IMPORTANT UPDATE**

I've considered the last bullet point. I've re-written the code to do this version. This improved performance even further.

```
# before merging filtering steps:

Benchmarking SoloGridSpace version
  140.700 μs (0 allocations: 0 bytes)
Benchmarking SoloGridSpace version: count nearby same
  172.779 ns (1 allocation: 64 bytes)

# after merging filtering steps:
Benchmarking SoloGridSpace version
  107.900 μs (0 allocations: 0 bytes)
Benchmarking SoloGridSpace version: count nearby same
  144.444 ns (1 allocation: 64 bytes)
```

As always, top-level-comment has the benchmarks of the latest pushed code.

Datseris:
Seems like I succeeded in writing clauses that skip origin posiiton for agent searches. How I did that, I have no idea. I re-wrote the code, but I wrote exactly the same things. What the hell. Really, WHAT THE FUCK, I've spent 3 hours on this shit. Anyways, here are current standings, for the first time our code is faster than the dict version.

```
benchmarking dict-based step
  95.200 μs (1 allocation: 11.38 KiB)
Benchmarking GridSpace version
  231.500 μs (5510 allocations: 258.28 KiB)
Benchmarking SoloGridSpace version
  77.500 μs (0 allocations: 0 bytes)
Benchmarking SoloGridSpace version: count nearby same
  84.216 ns (0 allocations: 0 bytes)
Benchmarking abusive unkillable version
  14.700 μs (0 allocations: 0 bytes)
Benchmarking abusive unkillable version: count nearby same
  41.473 ns (0 allocations: 0 bytes)
```
We see now that even though we are only 2x slower in counting nearby neighbors, we are still 4x slower in overall model simulation versus the abusive version. So, we can at most gain a factor of 2 from optimizing neighbor searches further, the rest must rely on the model structure.

Datseris:
I've updated both spaces to use the vector of positions, and now they share code implementation. That reduced code duplication a lot, but there is still some for performance reasons. Now we have:
```
benchmarking dict-based step
  96.100 μs (1 allocation: 11.38 KiB)
Benchmarking GridSpace version
  267.100 μs (5458 allocations: 255.84 KiB)
Benchmarking GridSpace version: count nearby same
  215.294 ns (2 allocations: 96 bytes)
Benchmarking GridSpaceSingle version
  76.500 μs (0 allocations: 0 bytes)
Benchmarking GridSpaceSingle version: count nearby same
  76.907 ns (0 allocations: 0 bytes)
Benchmarking abusive unkillable version
  14.700 μs (0 allocations: 0 bytes)
Benchmarking abusive unkillable version: count nearby same
  41.818 ns (0 allocations: 0 bytes)
```

So, simply iterating over nearby agents is allocating in `GridSpace`. **THIS IS BAD**. I need to figure out why and where this is happening.

Datseris:
I have created a dedicated, and smart iterator for `GridSpace`. It **reduces allocations to zero** and speeds up more things even more. We now have:
```
Benchmarking GridSpace version
  202.800 μs (0 allocations: 0 bytes)
Benchmarking GridSpace version: count nearby same
  123.804 ns (0 allocations: 0 bytes)
Benchmarking GridSpaceSingle version
  76.300 μs (0 allocations: 0 bytes)
Benchmarking GridSpaceSingle version: count nearby same
  62.921 ns (0 allocations: 0 bytes)
Benchmarking abusive unkillable version
  14.900 μs (0 allocations: 0 bytes)
Benchmarking abusive unkillable version: count nearby same
  42.929 ns (0 allocations: 0 bytes)
```

This was a massive achievement. It took me 4 hours of very intense programming to get it done.

Datseris:
oh man current grid tests are just as bad as it can get... we test non exported stuff, we test internal implementations rather than the purpose of functions, we rely on global state on pretty muich every line of tests.... I'll try to re-write some. I'm unfortunatley also time pressured to get this merged as soon as possible.

Datseris:
@AayushSabharwal There is one test that is not passing here and is related to Astar. Maybe it relied to some deep internals of ContinuousSpace with the internals of GridSpace also assumed. (really deeply nested stuff, we should have used the API as much as possible).

The problem is, now at line 104 of the `tests/astar_tests.jl` file, where we test continuous space, the very first test fails. See the description at the line numbers. I get some really weird stuff, we try to move agent to position (-0.25, -0.25), which is of course outside the space extent.

I was not able to figure out why. But I must leave this for another time where you will be available to help, because now I am under a lot of pressure to merge this PR and move forwards with finish my presentation.

AayushSabharwal:
I'll try and get to this, and the ensemble PR you mentioned me in. Unfortunately, I'm absurdly busy right now, so I'm not sure when I'll have the time to solve all of this. Might be able to take a look tomorrow, and see if I can locate the issue.

Datseris:
> Unfortunately, I'm absurdly busy right now

That makes two of us sadly, but nevertheless I appreciate a lot whatever time you can spend!

This PR is done BTW, I'm merging it now. It is one of the most massivest changes in Agents.jl, making absolutely massive performance increases in continuous and grid spaces. E.g., for continuous:
```
# Current state
Continuous space count nearby ids, spacing=0.05, r=0.1, inexact
  839.130 ns (2 allocations: 288 bytes)
Continuous space count nearby ids, spacing=0.05, r=0.1, exact
  4.243 μs (60 allocations: 5.56 KiB)

  Continuous space count nearby ids, spacing=0.1, r=0.1, inexact
  1.220 μs (2 allocations: 288 bytes)
Continuous space count nearby ids, spacing=0.1, r=0.1, exact
  4.843 μs (58 allocations: 5.22 KiB)


# master
Continuous space count nearby ids, spacing=0.05, r=0.1, inexact
  3.612 μs (71 allocations: 4.91 KiB)
Continuous space count nearby ids, spacing=0.05, r=0.1, exact
  10.600 μs (164 allocations: 30.15 KiB)

Continuous space count nearby ids, spacing=0.1, r=0.1, inexact
  5.900 μs (118 allocations: 8.05 KiB)
Continuous space count nearby ids, spacing=0.1, r=0.1, exact
  10.300 μs (218 allocations: 32.88 KiB)

```

AayushSabharwal:
That is... incredible

AayushSabharwal:
I can't find the test that fails. Did you fix it?

Datseris:
I've commented it out, it is here: https://github.com/JuliaDynamics/Agents.jl/blob/main/test/astar_tests.jl#L104

Issue #654, opened 2022-07-16
`walk!` can be made generic and space agnostic with a simple normalizing function


Datseris:
If we define a simple function `normalize_position(pos, space)` that takes in a position, and normalizes it so that it is appropriate for the space's extents, then the walk function becomes the trivial:
```julia
function walk!(agent, direction, model)
    newpos = agent.pos .+ direction
    newpos = normalize_position(newpos, model.space)
    move_agent!(agent, newpos, model)
end
```

and works for any space that defines `normalize_position`. Furthermore, it appears to me that we can reduce source code duplications by defining this `normalize_position` function. We should also make it part of the public api.

Datseris:
(and yes, we would also need to have a check for `ifempty` by just using `isempty(newpos, model)` as a clause)

AayushSabharwal:
I'm working on this in `fix_api`, possibly along with a couple other issues with unnecessary functions and API. On a related node, should `spacesize` also be made part of the public API? I think it would be useful to be able to get the size of a space without knowing the internals

Datseris:
YES!!!

Datseris:
can't beleive i didn't do this already in the 5.4 pR!

AayushSabharwal:
Suppose `a=(1.3, 2.5)`, `b=(1.2, 2.3)`. So it turns out `clamp.(a, (zero(eltype(b)) for _ in 1:length(b)), b)` allocates because of generating the second argument (a tuple of zeros of the appropriate type and length). Is there a better way to do this?

Datseris:
yes, use `map` ! In general prefer `map` when working with tuples, unless it is straightforward to do with broadcasting `.`

AayushSabharwal:
That turns out to be slower by 3x if we cast the result to `Tuple`, but faster by 2x if we leave it as a `Vector`

Datseris:
I am confused, what was initialy a `Vector` to begin with? Nothing should be a vector, everything is a tuple. Can we please move the suggestion to the PR and on specific code lines, so I can see where things become a vector to begin with?

PR #656, opened 2022-07-19
Fix astar test

AayushSabharwal:
The issue was wrapping in periodic space. The shortest direction from `(0,0)` to `(4.75,4.75)` in a periodic space is `(-0.25,-0.25)`, which is where it tried to move to. Fixed by using the new `Agents.spacesize` function.


Datseris (approved):
This is yet another thing that could be simplified by #654 by using this normalizing function. Task for another day :)

PR #658, opened 2022-07-22
Generalize `walk!`

### LINKED ISSUE #654: `walk!` can be made generic and space agnostic with a simple normalizing function

AayushSabharwal:
Fixes #654 

- [x] Add tests
- [x] Update docs

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -79,6 +79,12 @@ genocide!
 sample!
 ```
 
+## Space utility functions
+```@docs
+normalize_position
+spacesize
+```
+
 ## Discrete space exclusives
 ```@docs
 positions
--- a/src/spaces/utilities.jl
+++ b/src/spaces/utilities.jl
@@ -232,3 +232,11 @@ walk!(agent, ::typeof(rand), model::ABM{<:AbstractGridSpace{D}}; kwargs...) wher
 
 walk!(agent, ::typeof(rand), model::ABM{<:ContinuousSpace{D}}) where {D} =
     walk!(agent, Tuple(2.0 * rand(model.rng) - 1.0 for _ in 1:D), model)
+
+"""
+    spacesize(model::ABM)
+
+Return the size of the model's space. Works for [`AbstractGridSpace`](@ref) and
+[`ContinuousSpace`](@ref).
+"""
+spacesize(model::ABM) = spacesize(model.space)
--- a/src/spaces/utilities.jl
+++ b/src/spaces/utilities.jl
@@ -166,7 +166,7 @@ end
 
 function normalize_position(pos, space::ContinuousSpace{D,false}) where {D}
     ss = spacesize(space)
-    return Tuple(clamp.(pos, zeros(D), ss .- eps.(ss)))
+    return Tuple(clamp.(pos, 0.0, prevfloat.(ss)))
 end
 
 function normalize_position(pos, space::AbstractGridSpace{D,true}) where {D}
--- a/src/spaces/utilities.jl
+++ b/src/spaces/utilities.jl
@@ -166,7 +166,7 @@ end
 
 function normalize_position(pos, space::ContinuousSpace{D,false}) where {D}
     ss = spacesize(space)
-    return Tuple(clamp.(pos, zeros(D), ss .- eps(ss)))
+    return Tuple(clamp.(pos, zeros(D), ss .- eps.(ss)))
 end
 
 function normalize_position(pos, space::AbstractGridSpace{D,true}) where {D}
@@ -174,7 +174,7 @@ function normalize_position(pos, space::AbstractGridSpace{D,true}) where {D}
 end
 
 function normalize_position(pos, space::AbstractGridSpace{D,false}) where {D}
-    return Tuple(clamp.(pos, ones(D), spacesize(space)))
+    return Tuple(clamp.(pos, ones(Int, D), spacesize(space)))
 end
 
 """
@@ -214,7 +214,7 @@ function walk!(
     direction::NTuple{D,Float64},
     model::ABM{<:ContinuousSpace}
 ) where {D}
-    target = normalize_position(agent.pos + direction, model)
+    target = normalize_position(agent.pos .+ direction, model)
     move_agent!(agent, target, model)
 end
 
--- a/src/spaces/utilities.jl
+++ b/src/spaces/utilities.jl
@@ -204,7 +204,7 @@ function walk!(
     ifempty::Bool = true
 ) where {D}
     target = normalize_position(agent.pos .+ direction, model)
-    if !ifempty || isempty(model.space.s[agent.pos...])
+    if !ifempty || isempty(model.space.stored_ids[agent.pos...])
         move_agent!(agent, target, model)
     end
 end
--- a/src/spaces/utilities.jl
+++ b/src/spaces/utilities.jl
@@ -1,4 +1,4 @@
-export euclidean_distance, manhattan_distance, get_direction, walk!
+export euclidean_distance, manhattan_distance, get_direction, normalize_position, walk!, spacesize
 
 #######################################################################################
 # %% Distances and directions in Grid/Continuous space
@@ -154,6 +154,28 @@ end
 # %% Walking
 #######################################################################################
 """
+    normalize_position(pos, space::Union{AbstractGridSpace,ContinuousSpace})
+
+Return the position `pos` normalized for the extents of the given `space`.
+"""
+function normalize_position(pos, space::ContinuousSpace{D,true}) where {D}
+    return mod.(pos, spacesize(space))
+end
+
+function normalize_position(pos, space::ContinuousSpace{D,false}) where {D}
+    ss = spacesize(space)
+    return clamp.(pos, zeros(D), ss .- eps(ss))
+end
+
+function normalize_position(pos, space::AbstractGridSpace{D,true}) where {D}
+    return mod1.(pos, spacesize(space))
+end
+
+function normalize_position(pos, space::AbstractGridSpace{D,false}) where {D}
+    return clamp.(pos, ones(D), spacesize(space))
+end
+
+"""
     walk!(agent, direction::NTuple, model; ifempty = false)
 
 Move agent in the given `direction` respecting periodic boundary conditions.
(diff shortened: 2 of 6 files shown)


src/spaces/utilities.jl
```diff
@@ -153,6 +153,28 @@ end
 #######################################################################################
 # %% Walking
 #######################################################################################
+"""
+    normalize_position(pos, space::Union{AbstractGridSpace,ContinuousSpace})
```
Datseris, line 157: Given that in general we do not let users handle and access spaces directly, I think the top level and exported version of htis should be with `ABM` instead.

AayushSabharwal:
```julia
julia> a = (1.3, 2.5)
julia> b = (1.2, 2.3)
julia> f(a, b::NTuple{N,T}) where {N,T} = map(t -> clamp(t[1], zero(T), t[2]), zip(a,b))
julia> f(a,b)
2-element Vector{Float64}:
 1.2
 2.3
```

AayushSabharwal:
Conveniently, the tests for `walk!` should just cover `normalize_position`.

AayushSabharwal:
Wait a minute. Why does `spacesize` need to exist if `Base.size` is already overloaded?

AayushSabharwal:
So there's a change in behaviour of `walk!`: Before, in a periodic continuous space of size `(10.0, 10.0)`, moving to `(5.0, 10.0)` clamped to `(5.0, prevfloat(10.0))` but now it wraps to `(5.0, 0.0)`. Personally, it seems like the latter makes more sense: the "last" value in that dimension is `prevfloat(10.0)` after which it should wrap, similar to how `GridSpace` would.

AayushSabharwal:
I ran a quick set of benchmarks:
```julia
julia> using StaticArrays
julia> x = @SVector[1.3, 2.5]
julia> y = @SVector[1.2, 2.3]
julia> a = Tuple(x)
julia> b = Tuple(y)
julia> f1(x, y) = clamp.(x, zero(y), y)
julia> f2(x,y::NTuple{N,T}) where {N,T} = Tuple(clamp.(x, (zero(T) for _ in 1:N), y))
julia> f3(x,y::NTuple{N,T}) where {N,T} = Tuple(map(t -> clamp(t[1], zero(T), t[2]), zip(x,y)))

julia> @benchmark f1($x,$y)
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  2.639 ns … 8.874 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.652 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.684 ns ± 0.202 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▆▁                                                       ▁
  ███▃▅▆▃▄▄▄▃▁▃▁▃▄▃▃▄▄▁▁▃▃▁▃▁▄▁▁▁▄▁▄▁▁▃▁▃▅▅▁▄▅▆▅▅▆▆▆▇▆▆▇▇▆▆ █
  2.64 ns     Histogram: log(frequency) by time     3.63 ns <

julia> @benchmark f2($a,$b)
BenchmarkTools.Trial: 10000 samples with 269 evaluations.
 Range (min … max):  290.405 ns …  14.432 μs  ┊ GC (min … max): 0.00% … 96.26%
 Time  (median):     294.770 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   321.498 ns ± 468.748 ns  ┊ GC (mean ± σ):  5.20% ±  3.49%

  ▇█▄▁▁▁                ▁▁▁                                     ▁
  ████████▇█▇▇▇▅▅▆▅▆▅▇▆████▇▆▆▇▆▆▃▅▄▅▄▃▃▃▄▅▂▃▃▄▃▅▃▃▄▃▄▄▅▄▄▂▂▃▃▂ █
  290 ns        Histogram: log(frequency) by time        508 ns <

 Memory estimate: 224 bytes, allocs estimate: 5.

julia> @benchmark f3($a,$b)
BenchmarkTools.Trial: 10000 samples with 383 evaluations.
 Range (min … max):  249.371 ns …  11.116 μs  ┊ GC (min … max): 0.00% … 96.79%
 Time  (median):     252.815 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   270.655 ns ± 343.287 ns  ┊ GC (mean ± σ):  4.35% ±  3.35%

  ▄██▅▃▂▁▁▁▁▁▂▁▁                ▁ ▁                             ▂
  ████████████████████▇▆▇▆▇▆▇▇▇█████▇▆█████▆▆▅▆▆▆▇▆▆▄▅▄▄▁▃▄▄▃▅▅ █
  249 ns        Histogram: log(frequency) by time        345 ns <

 Memory estimate: 144 bytes, allocs estimate: 4.
```

Is there something I'm missing with `map` that makes it return a `Vector`?

Also, I've always preferred positions to be `StaticArrays` and not `NTuple`, especially considering basically everything we do with them bar indexing arrays treats positions as mathematical vectors. So far, I hadn't come across a reason to suggest the change. As the saying goes if it isn't broke(n), don't fix it. While actually changing this isn't the goal of this PR, I thought I'd bring it up just in case we actually can't beat that performance.

Datseris:
...? 

```
julia> f(a, b::NTuple{N,T}) where {N,T} = clamp.(a, zero(T), b)

julia> f(a,b)
(1.2, 2.3)

julia> using BenchmarkTools

julia> @btime f($a, $b);
  1.500 ns (0 allocations: 0 bytes)
```

AayushSabharwal:
...
All the lengths I went to, to not think of this.
Be right back, Ima go punch a wall.

Datseris:
> Why does spacesize need to exist if Base.size is already overloaded?

I've thought about that, and concluded that `Base.size(continuous space)` doesn't make sense. `Base.size` is about array/collection sizes, while that's not what we want for continuous space. For discrete space it just so happens that "space eize" and "array size" coincide.

Datseris:
To clarify: `map` would always work, as it is _always_ more stable than broadcasting. You just need to map each element rather than zipping:
```
julia> f(a, b::NTuple{N,T}) where {N,T} = map((x, y) -> clamp(x, zero(T), y), a, b)

julia> @btime f($a, $b);
  1.500 ns (0 allocations: 0 bytes)
```
(remember multi-argument form of `map`)

AayushSabharwal:
I had absolutely no idea `map` worked that way. Learn something new every day, I guess.

Datseris:
> Before, in a periodic continuous space of size (10.0, 10.0), moving to (5.0, 10.0) clamped to (5.0, prevfloat(10.0)) but now it wraps to (5.0, 0.0). 

I agree with this change, but let me see the code as well for certainty. Will review later in more detail.

docs/src/api.md
```diff
@@ -79,6 +79,12 @@ genocide!
 sample!
 ```
 
+## Space utility functions
```
Datseris, line 82: Putting these functions here is okay with me.

src/spaces/utilities.jl
```diff
@@ -153,6 +153,30 @@ end
 #######################################################################################
 # %% Walking
 #######################################################################################
+"""
+    normalize_position(pos, model::ABM{<:Union{AbstractGridSpace,ContinuousSpace}})
+
+Return the position `pos` normalized for the extents of the space of the given `model`.
```
Datseris, line 159: I think here we should add more info, like "for periodic spaces this wraps the position along each dimension, while for non-periodic spaces this clamps the position to the space extent."

src/spaces/utilities.jl
```diff
@@ -153,6 +153,30 @@ end
...
+"""
+normalize_position(pos, model::ABM) = normalize_position(pos, model.space)
+
+function normalize_position(pos, space::ContinuousSpace{D,true}) where {D}
+    return mod.(pos, spacesize(space))
```
Datseris, line 164: I was using the `mod1` function, and here you use `mod`, what's the change for?

src/spaces/utilities.jl
```diff
@@ -176,46 +200,21 @@ Example usage in [Battle Royale](
...
-    target = min.(target, prevfloat.(spacesize(model)))
-    move_agent!(agent, target, model)
-end
 function walk!(
     agent::AbstractAgent,
     direction::NTuple{D,Float64},
-    model::ABM{<:ContinuousSpace{D,false}}
+    model::ABM{<:ContinuousSpace}
 ) where {D}
-    target = min.(max.(agent.pos .+ direction, 0.0), prevfloat.(spacesize(model)))
```
Datseris, line 218: Okay, so instead of the `min, max` approach, you are using `clamp`. Have you tested which version is faster?

Datseris (commented):
Just for clarity, can you please summarize which versions you swapped with which? It appears at the moment that all previous versions have been completely replaced (no min max, no mod1, but clamp and mod)

src/spaces/utilities.jl
AayushSabharwal: `mod1(0.0, 2.0) == 2.0`, but continuous space doesn't include it's upper bound, and does include 0

src/spaces/utilities.jl
AayushSabharwal: Clamp, by almost 2x

AayushSabharwal:
> Just for clarity, can you please summarize which versions you swapped with which? It appears at the moment that all previous versions have been completely replaced (no min max, no mod1, but clamp and mod)

min max is slower, so I used clamp. mod1 is for Grid Space, since it is basically like mod for 1-indexed arrays. mod is for continuous space, since it wraps around like you'd expect. I just realised there's no need for `prevfloat`, I'll fix that.

src/spaces/utilities.jl
AayushSabharwal:
```julia
julia> a = (1.3, 2.5)
julia> b = (1.2, 2.3)
julia> f1(a,b) = clamp.(a, 0., b)
julia> f2(a,b) = min.(max.(a, 0.), b)

julia> @benchmark f1($a, $b)
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  1.557 ns … 6.759 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.566 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.591 ns ± 0.136 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▇▃  ▁                                                    ▂
  ███▅▃█▇▃▆▄▄▁▃▃▃▃▁▁▁▁▁▁▃▃▁▁▁▁▁▁▁▁▁▄▅▄▅▇▆▇▇▇▇▆▇▇▇▇▇▇▇▆▇██▇▇ █
  1.56 ns     Histogram: log(frequency) by time     2.07 ns <

julia> @benchmark f2($a, $b)
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  2.279 ns … 42.178 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.292 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.454 ns ±  1.387 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▅▄                                                        ▁
  ████▆▄▅▄▃▁▃▃▃▃▃▁▄▄▃▃▃▁▁▄▅▃▄▅▄▅▅▅▆▆▅▆▅▅▆▆▆▆▆▆▅▆▆▆▆▆▆▆▆▆▅▆▆▆ █
  2.28 ns      Histogram: log(frequency) by time     4.39 ns <

```

Datseris (approved):
Does `spacesize` have a documentation string...?

AayushSabharwal:
I moved `spacesize(model::ABM)` to `utilities.jl` since it didn't seem to make sense having it in `continuous.jl`, and added the docstring.

src/spaces/utilities.jl
```diff
@@ -232,3 +232,11 @@ walk!(agent, ::typeof(rand), model::ABM{<:AbstractGridSpace{D}}; kwargs...) wher
 
 walk!(agent, ::typeof(rand), model::ABM{<:ContinuousSpace{D}}) where {D} =
     walk!(agent, Tuple(2.0 * rand(model.rng) - 1.0 for _ in 1:D), model)
+
+"""
+    spacesize(model::ABM)
+
+Return the size of the model's space. Works for [`AbstractGridSpace`](@ref) and
+[`ContinuousSpace`](@ref).
```
Datseris, line 240: Is there an easy way to make this also work for OSM space, which would return an ecnlosing rectangle with min max lon and min max lat?

Datseris:
We can possibly extent `spacesize` to e.g., `OSMSpace` in a different PR.

Datseris:
@AayushSabharwal as this `spacesize` function is now part of the public API, it also needs to be explicitly tested in the test suite. So please add a couple of simple tests and then merge :)

AayushSabharwal:
> We can possibly extent `spacesize` to e.g., `OSMSpace` in a different PR.

The simplest method I can think of is looping over all nodes and finding the min/max lat/lon, and caching the result in the struct.

> @AayushSabharwal as this `spacesize` function is now part of the public API, it also needs to be explicitly tested in the test suite. So please add a couple of simple tests and then merge :)

Sure

AayushSabharwal:
Will merge once tests pass. They work locally, and should pass here too but I'd rather make sure.

src/spaces/utilities.jl
```diff
@@ -176,46 +202,21 @@ Example usage in [Battle Royale](
 function walk!(
     agent::AbstractAgent,
     direction::NTuple{D,Int},
-    model::ABM{<:AbstractGridSpace{D,true}};
-    kwargs...,
+    model::ABM{<:AbstractGridSpace};
+    ifempty::Bool = true
```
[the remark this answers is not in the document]
Datseris: It's unfortunate that we didn't catch this. But please feel free to make a PR that fixes the Battle Royale example. Although I'm not sure what you mean by "it breaks the example in the docs", as the example runs fine here: https://juliadynamics.github.io/AgentsExampleZoo.jl/dev/examples/battle/

PR #661, opened 2022-07-28
Fix `GridSpaceSingle` deserialization, add tests

### LINKED ISSUE #660: Schelling.jl example returns an error on `AgentsIO.load`
**Describe the bug**
Julia produces the following error after running the example/schelling.jl
```
ERROR: LoadError: MethodError: no method matching GridSpaceSingle{2, false}(::Matrix{Int64}, ::Symbol, ::Dict{Any, Any})
Closest candidates are:
  GridSpaceSingle{D, P}(::Any, ::Any, ::Any, ::Any) where {D, P} at ~/XXX/Agents.jl/src/spaces/grid_single.jl:12
Stacktrace:
 [1] from_serializable(t::Agents.AgentsIO.SerializableGridSpaceSingle{2, false}; kwargs::Base.Pairs{Symbol, Agents.Schedulers.Randomly, Tuple{Symbol}, NamedTuple{(:scheduler,), Tuple{Agents.Schedulers.Randomly}}})
   @ Agents.AgentsIO ~/XXX/Agents.jl/src/submodules/io/jld2_integration.jl:146
 [2] from_serializable(t::Agents.AgentsIO.SerializableABM{Agents.AgentsIO.SerializableGridSpaceSingle{2, false}, SchellingAgent, Dict{Symbol, Int64}, MersenneTwister}; kwargs::Base.Pairs{Symbol, Agents.Schedulers.Randomly, Tuple{Symbol}, NamedTuple{(:scheduler,), Tuple{Agents.Schedulers.Randomly}}})
   @ Agents.AgentsIO ~/XXX/Agents.jl/src/submodules/io/jld2_integration.jl:121
 [3] load_checkpoint(filename::String; kwargs::Base.Pairs{Symbol, Agents.Schedulers.Randomly, Tuple{Symbol}, NamedTuple{(:scheduler,), Tuple{Agents.Schedulers.Randomly}}})
   @ Agents.AgentsIO ~/XXX/Agents.jl/src/submodules/io/jld2_integration.jl:224
 [4] top-level scope
   @ ~/XXX/Agents.jl/examples/schelling.jl:298
in expression starting at ~/XXX/Agents.jl/examples/schelling.jl:298
```

**Minimal Working Example**
just run `julia examples/schelling.jl` in the repo package

**Agents.jl version**
[46ada45e] Agents v5.4.2

AayushSabharwal:
Fixes #660

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati"]
-version = "5.4.2"
+version = "5.4.3"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
--- a/src/submodules/io/jld2_integration.jl
+++ b/src/submodules/io/jld2_integration.jl
@@ -143,7 +143,7 @@ function from_serializable(t::SerializableGridSpace{D,P}; kwargs...) where {D,P}
 end
 function from_serializable(t::SerializableGridSpaceSingle{D,P}; kwargs...) where {D,P}
     s = zeros(Int, t.dims)
-    return GridSpaceSingle{D,P}(s, t.metric, Dict())
+    return GridSpaceSingle{D,P}(s, t.metric, Dict(), Dict())
 end
 
 function from_serializable(t::SerializableContinuousSpace{D,P,T}; kwargs...) where {D,P,T}
--- a/test/jld2_tests.jl
+++ b/test/jld2_tests.jl
@@ -16,6 +16,12 @@
         @test space.metric == other.metric
     end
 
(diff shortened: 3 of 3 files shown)


Datseris:
Maybe we can have foresight here: is there some space that is not tested for IO? Could we write a simple loop that loops over the space types and their agent types and does a trivial IO?

(You can use `GraphAgent` as an actual agent `ABM(GraphAgent, GraphSpace(...))` so we could loop over a zip of space instances and default agent tyupes)

AayushSabharwal:
When a new space is added, at the very least it would require a `test_space` function to be implemented, so it's not like we can just add the `(Space, AgentType)` tuple to the list and have it work. OSMSpace needs to be initialized differently, with some paths defined for the agents so we can ensure they are also saved properly. In effect, we'd need to loop over a list of function calls that create the model, which is basically the same as what we have currently.

Datseris:
Fair point. Can you please increment patch version in Project.toml?

PR #664, opened 2022-07-29
New `@agent` macro

### LINKED ISSUE #651: Extend `@agent` macro to allow inheritance
It would be great to extend the `@agent` macro to allow for field inheritance.

```julia
@agent Animal{I} GridAgent{2} begin
    age::I
end

@agent Bird{F} Animal{I} begin
    wingspread::F
end
```

The current implementation errors out when trying to define the agent subtype:

```julia
julia> @agent Bird{F} Animal{I} begin
           wingspread::F
       end
ERROR: LoadError: UndefVarError: Animal not defined
Stacktrace:
 [1] top-level scope
   @ none:1
 [2] eval(m::Module, e::Any)
   @ Core ./boot.jl:373
 [3] var"@agent"(__source__::LineNumberNode, __module__::Module, name::Any, base::Any, fields::Any)
   @ Agents ~/.julia/packages/Agents/sza9W/src/core/agents.jl:62
in expression starting at REPL[8]:1
```

Datseris:
Closes #651

This PR aims to do:
- [x] Completely rework `@agent` to do what #651 wants
- [x] Allow optional supertype hierarchy that is not directly from `AbstractAgent`.
- [x] Re-write the docs of `@agent` and in general `AbstractAgent`.
- [x] use `@agent` to create the default agent types
- [x] Make it clear that **the way** to make agent types is with `@agent`. 
- [x] Fully remove from the docs making agents with the `mutable struct` way. We need a full library search for this.
- [x] Add tests
- [x] Add a declaration in the docs that in the future internal fields may be attributed to agents are not part of public API.
- [x] Add changelog entry and an update message for 5.3
- [x] Test that this works with Distributed computing.


Datseris:
cc @fbanning

AayushSabharwal:
I love this. I'll look over the macro code later, although I've quite literally never written macros before. Enforcing that `id` be `const` in >=1.8 is nice. I just have one question: Once 1.8 is out, I assume we'll update the `NoSpaceAgent` to have that `const` (otherwise our own structs violate the rules). However, this breaks support for anything below 1.7, which might pose an issue for users. Is there a way to change the code at compile time based on the Julia version? Is this something we want to look into?

Also, isn't this breaking, since we've dropped the manual struct method of creating agents? It would still _work_, it's just that we've dropped it from our API. And once 1.8 rolls out it would be breaking either way.

AayushSabharwal:
I also think we can define all our `GridAgent`, `ContinuousAgent`... through the `@agent` macro, so they "inherit" the `id` field from `NoSpaceAgent`?

Datseris:
Notice that this doesn't actually work. Putting it in a module does not define the types outside the module. https://discourse.julialang.org/t/metaprogramming-obtain-actual-type-from-symbol-for-field-inheritance/84912/15 .

This:
```julia
@agent Person{T} GridAgent{2} begin
    age::Int
    moneyz::T
end
```
DOesn't actually create a `Person` in `Main` module.

Datseris:
> However, this breaks support for anything below 1.7, which might pose an issue for users

Yes there is a way to define different code for different Julia versions.

Datseris:
Fixed it. The macro now works exactly as described in its documentation string. I am super proud that I got this right. It was a really big fight with the metaprogramming system.

Datseris:
This PR is now done. I'd really appreciate a formal review @fbanning  or @AayushSabharwal given the extent of this change.

CHANGELOG.md
```diff
@@ -1,4 +1,16 @@
...
+- The `@agent` macro is now THE way to create agent types for Agents.jl simulations.
+  Directly creating structs by hand is no longer mention in the documentation at all.
+  This will allow us in the future to utilize additional fields that the user does not
+  have to know about, which may bring new features or performance gains by being
+  part of the agent structures.
+- The minimal agent types like `GraphAgent` can be used normally as standard agent
+  types that only have the mandatory fields. This is now clear in the docs.
+  (this was possible also before v5.4, just not clear)
+- In the future, making agent types by hand may be completely dissalowed, resulting
+  in error. Therefore, making agent types manually is considered deprecated.
```
AayushSabharwal, line 13: Out of curiosity, is there any way we can actually enforce this?

docs/src/tutorial.md
```diff
@@ -4,49 +4,45 @@
...
 
 
 ## [1. The space](@id Space)
 Agents.jl offers several possibilities for the space the agents live in.
-In addition, it is straightforward to implement a fundamentally new type of space, see [Developer Docs](@ref).
+In addition, it is straightforward to implement a fundamentally new type of space, see [Creating a new space type](@ref).
 
-The available spaces are:
+The available spaces are listed in the [Available spaces](@ref) part of the API.
+An example of a space is [`OpenStreetMapSpace`](@ref). It is based on Open Street Map, where agents are confined to move along streets of the map, using real-world meter values for the length of each street.
```
AayushSabharwal, line 30: > using real-world meter values for the length of each street

This is not true. The unit of length in OSMSpace is arbitrary, and can be length, time to cross, or another quantity altogether.

docs/src/tutorial.md
```diff
@@ -4,49 +4,45 @@
...
 
-One simply initializes an instance of a space, e.g. with `grid = GridSpace((10, 10))` and passes that into [`AgentBasedModel`](@ref). See each individual space for all its possible arguments.
-
-
-## 2. The agent type
 
+## 2. The agent type(s)
 ```@docs
+@agent
 AbstractAgent
```
AayushSabharwal, line 38: Should `AbstractAgent` still be in the docs if the user never directly interacts with it?

docs/src/tutorial.md
```diff
@@ -4,49 +4,45 @@
...
 ```
 
-Once an agent is created it can be added to a model using e.g. [`add_agent!`](@ref).
+## 3. The model
+Once an agent is created with [`@agent`](@ref), it can be added to a model using e.g. [`add_agent!`](@ref).
```
AayushSabharwal, line 42: I think this is a little confusing. Agent _types_ are created with `@agent`, agents themselves are created just like instantiating another struct, and these instances are added to the model with `add_agent!`. This line gives the impression that I can do something like this:
```julia
@agent Foo GridAgent{2} begin
    bar::int
end
add_agent!(Foo, model)
```

docs/src/tutorial.md
```diff
@@ -55,16 +51,18 @@ AgentBasedModel
 
 ## 4. Evolving the model
 
-Any ABM model should have at least one and at most two step functions.
+Any ABM model should have at least one and at most two stepping functions.
```
AayushSabharwal, line 54: This is really a minor gripe, but "ABM model" is a bit repetitive, is it not? It's like saying OSM map.

examples/rabbit_fox_hawk.jl
```diff
@@ -24,18 +24,19 @@ using Random
 import ImageMagick
 using FileIO: load
 
-mutable struct Animal <: AbstractAgent
+@agent Animal ContinuousAgent{3} begin
     id::Int
     pos::NTuple{3,Float64}
```
AayushSabharwal, line 29: `id` and `pos` don't need to be here.

examples/schoolyard.jl
```diff
@@ -26,11 +26,9 @@ using SimpleWeightedGraphs: SimpleWeightedDiGraph # will make social network
...
 using Random: MersenneTwister                     # reproducibility
 
-# And create a very simple agent without any extra properties
-mutable struct Student <: AbstractAgent
-    id::Int
-    pos::Tuple{Float64,Float64}
-end
+# And create an alias to `ContinuousAgent{2}`,
+# as our agents don't need additional properties.
+Student = ContinuousAgent{2}
```
AayushSabharwal, line 31: This should be `const`, since it shouldn't be reassigned, and non-const globals are typically slower.

src/core/agents.jl
```diff
@@ -1,54 +1,68 @@
...
-```julia
-mutable struct ExampleAgent <: AbstractAgent
-    id::Int
-    pos::Int
-    weight::Float64
-    happy::Bool
-end
-```
+The [`@agent`](@ref) macro ensures that all of these constrains are in place
+and hence it is the expected way to generate new agent types.
```
AayushSabharwal, line 15: We should mention here that it is the only supported way of generating new types, not just expected.

src/core/agents.jl
```diff
@@ -57,60 +71,168 @@ mutable struct Person{T} <: AbstractAgent
...
-to the start of the agent type.
-"""
-mutable struct ContinuousAgent{D} <: AbstractAgent
-    id::Int
-    pos::NTuple{D,Float64}
-    vel::NTuple{D,Float64}
+macro agent(new_name, base_type, extra_fields)
+    # This macro was generated with the guidance of @rdeits on Discourse:
+    # https://discourse.julialang.org/t/
+    # metaprogramming-obtain-actual-type-from-symbol-for-field-inheritance/84912
```
AayushSabharwal, line 162:
```suggestion
    # https://discourse.julialang.org/t/metaprogramming-obtain-actual-type-from-symbol-for-field-inheritance/84912
```

So it's easier to copy-paste the link

src/core/agents.jl
```diff
@@ -1,54 +1,68 @@
...
 
-Create a struct for your agents which includes the mandatory fields required to operate
-in a particular space. Depending on the space of your model, the `AgentSupertype` is
-chosen appropriately from [`GraphAgent`](@ref), [`GridAgent`](@ref),
-[`ContinuousAgent`](@ref), [`OSMAgent`](@ref).
+Define an agent struct which includes all fields that `AnotherAgentType` has,
+as well as any additional ones the user may provide via the `begin` block.
+See below for examples.
+
+Using `@agent` is **the recommended way to create agent types** for using in Agents.jl.
```
AayushSabharwal, line 30: Similar to above, I feel like mentioning that it is the recommended way leaves the question that there are other ways that we support, which isn't the case.

src/core/agents.jl
```diff
@@ -57,60 +71,168 @@ mutable struct Person{T} <: AbstractAgent
...
+        end
+    end
 end
+# TODO: I do not know how to merge these two macros to remove code duplication.
+# There should be away that only the 4-argument version is used
+# and the 3-argument version just passes `AbstractAgent` to the 4-argument.
+macro agent(new_name, base_type, super_type, extra_fields)
+    # This macro was generated with the guidance of @rdeits on Discourse:
+    # https://discourse.julialang.org/t/
+    # metaprogramming-obtain-actual-type-from-symbol-for-field-inheritance/84912
```
AayushSabharwal, line 197:
```suggestion
    # https://discourse.julialang.org/t/metaprogramming-obtain-actual-type-from-symbol-for-field-inheritance/84912
```

Same as above, to make this copy-pasteable

docs/src/tutorial.md
AayushSabharwal: Never mind, I forgot we actually _can_ do that if we provide the value of `bar`: `add_agent!(Foo, model, 123)` 😅 . However, I still think it's misleading.

AayushSabharwal:
In `Creating a new space type` section in the docs, we should add that a new agent "template" (what are we calling these exactly?) should also be created.

CHANGELOG.md
Datseris: Yeap.

examples/schoolyard.jl
Datseris: Yes, but this is not true for type aliasing.

docs/src/tutorial.md
Datseris: Yes, it is an exported name.

Datseris:
@AayushSabharwal I believe I've address all of your review comments. Would you please review again, and if no comments are necessary, post an approving review.

AayushSabharwal (approved):

AayushSabharwal:
Why does the log test fail though? The strings look the same

src/core/agents.jl
```diff
@@ -57,60 +73,168 @@ mutable struct Person{T} <: AbstractAgent
...
+mutable struct Baker{T} <: AbstractAgent
     id::Int
-    pos::Int
+    pos::NTuple{2, Int}
+    age::Int
+    moneyz::T
+    breadz_per_day::T
 end
+```
+### Exaple with optional hierachy
```
fbanning, line 92:
```suggestion
### Example with optional hierarchy
```

src/core/agents.jl
```diff
@@ -57,60 +73,168 @@ mutable struct Person{T} <: AbstractAgent
...
+An alternative way to make the above structs, that also establishes
+a user-specific subtyping hierachy would be to do:
```
fbanning, line 94:
```suggestion
a user-specific subtyping hierarchy would be to do:
```

src/core/agents.jl
```diff
@@ -57,60 +73,168 @@ mutable struct Person{T} <: AbstractAgent
...
+The minimal agent struct for usage with `nothing` as space (i.e., no space).
+It has the field `id::Int`, and potentially other internal fields that
+are not documentated as part of the public API. See also [`@agent`](@ref).
 """
-mutable struct OSMAgent <: AbstractAgent
+mutable struct NoSpaceAgent <: AbstractAgent
     id::Int
-    pos::Tuple{Int,Int,Float64}
-end
+end
```
fbanning, line 240: I've tried to create a new agent via

```julia
@agent MyAgent NoSpaceAgent begin end
```

but it seems the exporting of `NoSpaceAgent` doesn't properly work and

```julia
ERROR: UndefVarError: NoSpaceAgent not defined
```

This is a bit of a weird error because this kind of subtyping works just fine internally, e.g. for `ContinuousAgent`.

src/core/agents.jl
```diff
@@ -57,60 +73,168 @@ mutable struct Person{T} <: AbstractAgent
...
+            end
+            # @show expr # uncomment this to see that the final expression looks as desired
+            # It is important to evaluate the macro in the module that it was called at
+            Core.eval($(__module__), expr)
+        end
+    end
 end
+# TODO: I do not know how to merge these two macros to remove code duplication.
+# There should be away that only the 4-argument version is used
+# and the 3-argument version just passes `AbstractAgent` to the 4-argument.
```
fbanning, line 195: This would indeed be very nice. I'll try and play around with the macro a bit and see if I can make it work.

src/core/agents.jl
```diff
@@ -1,54 +1,70 @@
...
-To make an agent with two additional properties, `weight, happy`, we'd write
-```julia
-mutable struct ExampleAgent <: AbstractAgent
-    id::Int
-    pos::Int
-    weight::Float64
-    happy::Bool
-end
-```
+The [`@agent`](@ref) macro ensures that all of these constrains are in place
```
fbanning, line 15:
```suggestion
The [`@agent`](@ref) macro ensures that all of these constraints are in place
```

src/core/agents.jl
```diff
@@ -57,60 +73,168 @@ mutable struct Person{T} <: AbstractAgent
...
-[`GridSpace`](@ref). It attributes the fields `id::Int, pos::NTuple{D,Int}`
-to the start of the agent type.
-"""
-mutable struct GridAgent{D} <: AbstractAgent
-    id::Int
-    pos::Dims{D}
+@agent Worker GridAgent{2} AbstractHuman begin
+    age::Int
+    moneyz::Float64
 end
```
fbanning, line 101: Running this example code throws an error for me:

```julia
julia> abstract type AbstractHuman <: AbstractAgent end

julia> @agent Worker GridAgent{2} AbstractHuman begin
              age::Int
              moneyz::Float64
          end
ERROR: LoadError: MethodError: no method matching var"@agent"(::LineNumberNode, ::Module, ::Symbol, ::Expr, ::Symbol, ::Expr)
Closest candidates are:
  var"@agent"(::LineNumberNode, ::Module, ::Any, ::Any, ::Any) at ~/.julia/packages/Agents/9Rd7T/src/core/agents.jl:61
in expression starting at REPL[3]:1
```

fbanning:
Github is painfully slow today. I will continue to skim through the changes made in this PR on another day.

Issue #659, opened 2022-07-24
Integration with CellListMap.jl


Datseris:
I just became aware of this package: 

https://github.com/m3g/CellListMap.jl

Its goal is to accelerate computations of forces, or any kind of interactions, that occur between pairs of "agents" in 2D or 3D continuous space. Simply put, it does both what `interacting_pairs` currently does, **and** also computes the forces, which a user would have to do by hand after getting the interacting pairs. Performance benchmarks also look like its better than competing alternatives (common story of Julia packages it would seem).

I think the best place to start integrating this package with Agents.jl is to write an **integration example** that showcases using it to accelerate computations occurring in an ABM. From there we can see what can be made even more general and integrated more formally into the source of Agents.jl.

ehsanirani:
Hi. I have a simple code to mimic run & tumble dynamics of bacteria in porous media (implemented for an unpublished researh project). CellListMap is used to find collisions between agents in the ABM model and an array of random fixed obstacles. Bacteria are modeled as sphero-cylinders, that adds a bit complexity to the code to find real distances and calculate torques. Would it be a good integration example (checkout the video)? 

I can also simplify the code, to use disks instead of cylinders. Then it would be much simpler, similar to the SIR example with addition of random obstacles (another research project of mine).

Maybe @lmiq has a better suggestion for the integration example.

https://user-images.githubusercontent.com/26183665/180777800-2e1452a1-7240-4542-8917-1b760ab6c859.mp4

lmiq:
I was looking at the `Agents` examples, and found that probably the `Flocking` example of birds is the closer to the typical application of CellListMap. 

Yet, as I understand, the function `agend_step` applies to **one bird**. To effectively use a neighbor list (or function to compute the interactions directly), the way I'm used to see simulations is to have a "double"-loop over the elements (the birds), to compute the interactions between neighboring birds. That double loop can be replaced by a single call to the `map_pairwise!` function from CellListMap. 

That loop, is however, sort of implicit in the way `Agents` is implemented, as the first loop over agents is hidden (at least in that interface), and the function that is called for each agents computes the neighbors. 

Thus, I think that I need to interact to something more internal, or access an API which is not apparent at first sight. Any help is appreciated.

lmiq:
A typical simulation written on top of CellListMap is here, if anyone wants to take a look: https://github.com/m3g/2021_FortranCon/blob/main/celllistmap_vs_namd/simulate.jl

[...]

PR #669, opened 2022-08-02
included celllismap example

lmiq:
Followup of: https://github.com/JuliaDynamics/Agents.jl/issues/659

I included here the files for the `CellListMap.jl` integration example. I tried to follow the style of the other examples (but included a SVG figure in the examples dir, not sure if that is what you would like - I linked the final raw image in the markdown). 

The video that should be produced is similar to this one: https://github.com/lmiq/agents_with_celllistmap/blob/main/celllistmap.mp4


Datseris:
@lmiq thanks, I am working on this now. Question. Why are `forces` in the `Properties` struct, and positions in the `CellListMapData` struct? What decides which goes where?

Datseris:
I think there is no reason for `Properties` to exist. `forces, cutoff` are also "required data" for CellListMap.jl so it doesn't make sense to separate them.

Datseris:
@lmiq I removed the .svg image. We try to not include images or any large files in the Agents.jl repo as they accumulate over time. Especially in this case I do not think this image added much useful information to the integration example. It's a plot of a function someone can quite easily create themselves if need be.

Datseris:
@lmiq from a design point of view, why would the user have to care about updating these threaded data structures...?
Can't you create a central structure that would contain things like 
```julia
mutable struct CellListMapData{B,C,A,O}
    positions::Vector{SVector{2,Float64}}
    forces::Vector{SVector{2,Float64}}
    cutoff::Float64
    box::B
    cell_list::C
    aux::A
    output_threaded::O
end
```
and provide functions like `reset_forces!(datastructure)` that does the zero part automatically, and in general handles things automatically? 

Perhaps it is too much glue code for the user to have to carry around things like the `Box`. The user never actually cares to use the box. Just the positions and forces. Wouldn't it make sense to wrap all these stuff into a struct and provide functions that operate on this struct, and take in positions and forces vectors and update them...?

Datseris:
(also notice that I think `fill!(model.clmap.forces, SVector(0.0, 0.0))` is both more efficient and more intuitive way to reset the forces to 0)

lmiq:
Yes, in this example I can do that (in general I'm not sure, there are some customized problems that may require the current interface). 

I will do that here and maybe add that as a possibility in the package.

lmiq:
> Why are `forces` in the `Properties` struct, and positions in the `CellListMapData` struct?

What I tried is to separate what would be necessary for the use of CellListMap from what is required in an implementation that would not use it. As far as I understand, even without CellListMap, it would be interesting to have a `forces` vector to contain the forces acting on each particle, to be updated at each iteration. The `positions` vector, on the other side, is redundant, because the same information is contained in `agent.pos` fields, but CellListMap needs that additional array in order to not allocate a new instance of it at every iteration.

> and provide functions like `reset_forces!(forces, datastructure)`

I agree with this one in this example, particularly from the point of view of the above. With that the CellListMap "part" would be more self-contained still. I can do that. 

> I think there is no reason for `Properties` to exist. `forces, cutoff` are also "required data" for CellListMap.jl

This is more "controversial". I think that what is needed for CellListMap should be separated from what is needed to run the simulation in general. The `cutoff` has one additional detail: It could be used as an input parameter in the definition of the space, to set the the finer subspace possible to tune the performance of the `interacting_pairs` you have implemented. We are not using it now, but it could be used for that (although I'm not sure how much that can help).

lmiq:
> Wouldn't it make sense to wrap all these stuff into a struct and provide functions that operate on this struct, and take in positions and forces vectors and update them...?

I started to do that... but here what we actually get is that `model_step!` is identical to what an `update_forces!` function would look like. We need, inside the function, the `model` data structure, because the parameters of the particles (radii and force constants) are properties that are carried on the agents. 

Thus, although the code looks somewhat more organized, there is no real gain in conciseness, unless I put some of that into a new interface of CellListMap (which is not out of question and won't be breaking, so I can certainly see if I arrive to something simpler in these lines).

I will push the changes, but I'm not sure if they are in the right direction.

lmiq:
@Datseris:

I have significantly changed the example, after creating a new interface for the use of CellListMap which will simplify its, particularly in these examples. 

The branch containing the new interface is: https://github.com/m3g/CellListMap.jl/tree/PeriodicSystem

I have still to add some functionality and test the new interface, but in general that is what I arrived at.

I would like to know if you have other suggestions for this interface (concerning specifically how the example looks like now here), before I actually merge all that into CellListMap.

For the moment, I could do that without introducing breaking changes, so that's fine.

examples/celllistmap.jl
```diff
@@ -0,0 +1,249 @@
...
+#    the same type as the positions. These forces will be updated by the `CellListMap.map_pairwise!`
+#    function.
+#
+# Additionally, the computation with `CellListMap.jl` requires the definition of a `cutoff`,
+# which will be twice the maximum interacting radii of the particles, and the geometry of the
+# the system, given by the `sides` of the periodic box. 
+#
+# Just for clarity, we define an alias for the function that fetches the forces from 
+# the `CellListMap.PeriodicSystem` structure:
+const get_forces = CellListMap.get_output
```
Datseris, line 87: I'd advise to not do this and use the full call name instead.

examples/celllistmap.jl
```diff
@@ -0,0 +1,249 @@
...
+# The keywords here control number of particles and sizes.
+function initialize_model(;
+    number_of_particles=10_000,
+    sides=SVector(500.0, 500.0),
+    dt=0.001,
+    max_radius=10.0,
+    parallel=true
+)
+    ## initial random coordinates
+    coordinates = [sides .* rand(SVector{2,Float64}) for _ in 1:number_of_particles]
```
Datseris, line 101: Hm, I find the name `coordinates` confusing, I think `positions` is much better. The latter more clearly reflects that this is the positions of all particles. Coordinates, even if plural, may mean the coordinate system. At first I though that this variable was about the internal cell size of the cell list.

examples/celllistmap.jl
```diff
@@ -0,0 +1,249 @@
...
+
+    ## We will use CellListMap to compute forces, with similar structure as the coordinates
+    forces = similar(coordinates)
+
+    ## Space and agents
+    space2d = ContinuousSpace(Tuple(sides); periodic=true)
+
+    ## default maximum radius is 10.0 thus cutoff is 20.0
```
Datseris, line 109:
```suggestion
```
I don't think this comment has reason to exist.

examples/celllistmap.jl
```diff
@@ -0,0 +1,249 @@
...
+        output=forces,
+        parallel=parallel,
+    )
+
+    ## define the model properties
+    properties = Properties(
+        dt=dt,
+        number_of_particles=number_of_particles,
+        clmap_system=clmap_system,
+    )
```
Datseris, line 126: There is no reason to make the additional struct `Properties` here. Instead, you should use a standard named tuple. These three "properties" are constants, they do not change during the simulation, hence a standard named tuple is enough.

examples/celllistmap.jl
```diff
@@ -0,0 +1,249 @@
...
+# it exports some functions with conflicting names with `Agents`.
+
+# 
+# ## Model properties.
+# 
+# The `clmap_system` field will store the data required for `CellListMap`, and again we 
+# use a parametric type to guarantee the concrectness of the types, aoviding type
+# instabilities.  
+#
+Base.@kwdef struct Properties{CL<:CellListMap.PeriodicSystem}
```
Datseris, line 64: This struct can be removed, see comment below.

examples/celllistmap.jl
Datseris: (Named Tuples are also strictly typed)

Datseris (approved):
This looks fantastic, and is exactly what I was suggesting! I think this has simplified the interface to the bare minimum necessary for integration. Great.

Notice that you have committed an .mp4 file in the git history. Please remove this, and then add .mp4 into the .gitignore for safety.

examples/celllistmap.jl
Datseris: Actually, forget what I said. You are right, this makes for a more intuitive name call.

Project.toml
```diff
@@ -5,11 +5,14 @@ version = "5.5.0"
...
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
+CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
+CellListMap = "69e1c6dd-3888-40e6-b3c8-31ac5f578864"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
+InteractiveDynamics = "ec714cd0-5f51-11eb-0b6e-452e7367ff84"
```
Datseris, line 15: WOOPS, I just saw this. All of these MUST be removed. You want to add CellListMap in the `docs/Project.toml`. NOT the top most level project toml, these are the formal package dependencies!!!!

lmiq: Yes, I know, it was not the idea to merge this now. I have yet to finish and document the new interface and release the CellListMap side. 

(The problem here is that dev'ing two packages at once is cumbersome without adding the dependency one).

Datseris:
@lmiq please ping me when you think I should give another review and merge

lmiq:
> @lmiq please ping me when you think I should give another review and merge

Sure, I'll tell you when I release the corresponding version of CellListMap.

lmiq:
@Datseris I think its done now.

examples/celllistmap.jl
```diff
@@ -43,7 +43,7 @@ using Agents
 end
 Particle(; id, pos, vel, r, k, mass) = Particle(id, pos, vel, r, k, mass)
 
-# ## Required and data structures for CellListMap.jl
+# ## Required and data structures for CellListMap.jl (requires version ≥7.22)
```
Datseris, line 46:
```suggestion
# ## Required and data structures for CellListMap.jl (requires version ≥0.7.22) 
```

lmiq:
As a side note, I've seen now that you have a "elastic collisions" function implemented. It would be nice to have an example of that as the function to mapped among pairs. I may try to add that example as an integration example on my side, than we can link the examples.

Datseris:
Notice that the docs are erroring: https://github.com/JuliaDynamics/Agents.jl/runs/7838758532?check_suite_focus=true#step:8:160 

don't forget to put the package into the `docs/Project.toml` file!

lmiq:
> Notice that the docs are erroring: https://github.com/JuliaDynamics/Agents.jl/runs/7838758532?check_suite_focus=true#step:8:160
> 
> don't forget to put the package into the `docs/Project.toml` file!

Added the new doc dependencies. I'm not sure if the error in the docs is related to this pull request, it seemed to be a documenter key access problem.

examples/celllistmap.jl
```diff
@@ -0,0 +1,233 @@
...
+# to tuples to conform the `Agents` API for the positions and velocities
+# of the agents. Finally, the positions within the `CellListMap.PeriodicSystem`
+# structure are updated.
+function agent_step!(agent, model::ABM)
+    id = agent.id
+    dt = model.properties.dt
+    ## Retrieve the forces on agent id
+    f = model.system.forces[id]
+    a = f / agent.mass
+    # Update positions and velocities
```
Datseris, line 187:
```suggestion
    ## Update positions and velocities
```

Datseris:
Thank you very much @lmiq ! I merged this in, because if we want to alter the "elastic collision stuff", it is probably better to do in a different PR anyways. Or you can do this on your end and then I can link your docpage here!

PR #673, opened 2022-08-07
Use SVector for vel and pos fields of ContinuousAgents

### LINKED ISSUE #672: Should we just go ahead and use `SVector` in position and velocity fields of ContinuousSpace...?
Alright, so, plenty of times in many use cases of `ContinuousSpace`, it becomes an obvious pain to use `NTuple` instead of `SVector`. So, should we just go ahead and change the types to static vectors? Problem is, I am not sure how breaking this is. Unless a user for whatever reason annotatted the position type, everything the user did before will work now as well. This includes changing velocity via a tuple, `agent.vel = (1, 2)` would still work even if `vel isa SVector`. 

For example, this problem came when making the Flocking example, in the discussion in #671 , in the integration example with CellListMap.jl, the bacteria example, I mean, practically everywhere with `ContinuousSpace`.

mastrof:
Closes #672

### DIFF
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -329,12 +329,12 @@ function elastic_collision!(a, b, f = nothing)
     # mass weights
     m1 == m2 == Inf && return false
     if m1 == Inf
-        @assert v1 == (0, 0) "An agent with ∞ mass cannot have nonzero velocity"
+        @assert v1 == SVector(0, 0) "An agent with ∞ mass cannot have nonzero velocity"
         dot(r1, v2) ≤ 0 && return false
         v1 = ntuple(x -> zero(eltype(v1)), length(v1))
         f1, f2 = 0.0, 2.0
     elseif m2 == Inf
-        @assert v2 == (0, 0) "An agent with ∞ mass cannot have nonzero velocity"
+        @assert v2 == SVector(0, 0) "An agent with ∞ mass cannot have nonzero velocity"
         dot(r2, v1) ≤ 0 && return false
         v2 = ntuple(x -> zero(eltype(v1)), length(v1))
         f1, f2 = 2.0, 0.0
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -537,5 +537,5 @@ function get_spatial_index(pos, property::AbstractArray{T,D}, model::ABM) where 
     usize = ssize[1:D]
     εs = usize ./ propertysize
     idxs = floor.(Int, upos ./ εs) .+ 1
-    return CartesianIndex(idxs)
+    return CartesianIndex(Tuple(idxs))
 end
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -95,7 +95,7 @@ function ContinuousSpace(
 end
 
 function random_position(model::ABM{<:ContinuousSpace})
-    map(dim -> rand(model.rng) * dim, spacesize(model))
+    map(dim -> rand(model.rng) * dim, SVector(spacesize(model)))
 end
 
 "given position in continuous space, return cell coordinates in grid space."
@@ -110,7 +110,7 @@ end
 
 distance_from_cell_center(pos, model::ABM) =
     distance_from_cell_center(pos, cell_center(pos, model))
-function distance_from_cell_center(pos::SVector, center::SVector)
+function distance_from_cell_center(pos::SVector, center::Union{SVector,Tuple})
     sqrt(sum(abs2.(pos .- center)))
 end
 
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -20,22 +20,22 @@ function Base.show(io::IO, space::ContinuousSpace{D,P}) where {D,P}
 end
 
 @agent ContinuousAgent{D} NoSpaceAgent begin
-    pos::NTuple{D,Float64}
-    vel::NTuple{D,Float64}
+    pos::SVector{D,Float64}
+    vel::SVector{D,Float64}
 end
 
 @doc """
     ContinuousAgent{D} <: AbstractAgent
 The minimal agent struct for usage with `D`-dimensional [`ContinuousSpace`](@ref).
-It has the additoinal fields `pos::NTuple{D,Float64}, vel::NTuple{D,Float64}`.
+It has the additional fields `pos::SVector{D,Float64}, vel::SVector{D,Float64}`.
 See also [`@agent`](@ref).
 """ ContinuousAgent
 
 """
     ContinuousSpace(extent::NTuple{D, <:Real}; kwargs...)
 Create a `D`-dimensional `ContinuousSpace` in range 0 to (but not including) `extent`.
-Your agent positions (field `pos`) must be of type `NTuple{D, <:Real}`,
-and it is strongly recommend that agents also have a field `vel::NTuple{D, <:Real}` to use
+Your agent positions (field `pos`) must be of type `SVector{D, <:Real}`,
+and it is strongly recommend that agents also have a field `vel::SVector{D, <:Real}` to use
 in conjunction with [`move_agent!`](@ref). Use [`ContinuousAgent`](@ref) for convenience.
 
 `ContinuousSpace` is a representation of agent dynamics on a continuous medium
@@ -78,7 +78,7 @@ If you want exact searches use the slower [`nearby_ids_exact`](@ref).
   You can of course change the agents' velocities
   during the agent interaction, the `update_vel!` functionality targets spatial force
   fields acting on the agents individually (e.g. some magnetic field).
-  If you use `update_vel!`, the agent type must have a field `vel::NTuple{D, <:Real}`.
+  If you use `update_vel!`, the agent type must have a field `vel::SVector{D, <:Real}`.
 """
 function ContinuousSpace(
     extent::NTuple{D,X};
@@ -103,7 +103,7 @@ pos2cell(pos::Tuple, model::ABM) = @. floor(Int, pos/model.space.spacing) + 1
 pos2cell(a::AbstractAgent, model::ABM) = pos2cell(a.pos, model)
 
 "given position in continuous space, return continuous space coordinates of cell center."
-function cell_center(pos::NTuple{D,<:AbstractFloat}, model) where {D}
+function cell_center(pos::SVector{D,<:AbstractFloat}, model) where {D}
     ε = model.space.spacing
     (pos2cell(pos, model) .- 1) .* ε .+ ε/2
 end
@@ -504,7 +504,7 @@ end
 #######################################################################################
 export get_spatial_property, get_spatial_index
 """
-    get_spatial_property(pos::NTuple{D, Float64}, property::AbstractArray, model::ABM)
+    get_spatial_property(pos::SVector{D, Float64}, property::AbstractArray, model::ABM)
 Convert the continuous agent position into an appropriate `index` of `property`, which
 represents some discretization of a spatial field over a [`ContinuousSpace`](@ref).
 Then, return `property[index]`. To get the `index` directly, for e.g. mutating the
@@ -516,7 +516,7 @@ function get_spatial_property(pos, property::AbstractArray, model::ABM)
 end
 
 """
-    get_spatial_property(pos::NTuple{D, Float64}, property::Function, model::ABM)
+    get_spatial_property(pos::SVector{D, Float64}, property::Function, model::ABM)
 Literally equivalent with `property(pos, model)`, provided just for syntax consistency.
 """
 get_spatial_property(pos, property, model::ABM) = property(pos, model)
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -99,7 +99,7 @@ function random_position(model::ABM{<:ContinuousSpace})
 end
 
 "given position in continuous space, return cell coordinates in grid space."
-pos2cell(pos::Tuple, model::ABM) = @. floor(Int, pos/model.space.spacing) + 1
+pos2cell(pos::SVector, model::ABM) = @. floor(Int, pos/model.space.spacing) + 1
 pos2cell(a::AbstractAgent, model::ABM) = pos2cell(a.pos, model)
 
 "given position in continuous space, return continuous space coordinates of cell center."
@@ -110,7 +110,7 @@ end
 
 distance_from_cell_center(pos, model::ABM) =
     distance_from_cell_center(pos, cell_center(pos, model))
-function distance_from_cell_center(pos::Tuple, center::Tuple)
+function distance_from_cell_center(pos::SVector, center::SVector)
     sqrt(sum(abs2.(pos .- center)))
 end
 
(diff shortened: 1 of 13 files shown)


Datseris:
@mastrof try to make the tests pass, this will show us how many things are really breaking.

mastrof:
The tests (at least most of them) fail because they pass tuples to `add_agent!`, construct agents using tuples, call other distance evaluation functions using tuples etc...
Should I just put explicit conversions `SVector(tuple)` wherever needed to allow tuples to be passed in without errors?
Some tests will still fail since they will be checking equality between `SVector` and `Tuple` e.g.
https://github.com/JuliaDynamics/Agents.jl/blob/83129042f01673f832e4a32de53b93ecd6af80ab/test/continuous_space_tests.jl#L32

Datseris:
No, you simply make `pos0` an `SVector`. Similarly, all 'passing tuples' that happens in the test suite needs to become 'passing SVectors'.

mastrof:
As a start I just focused on [continuous_space_tests.jl](https://github.com/JuliaDynamics/Agents.jl/blob/83129042f01673f832e4a32de53b93ecd6af80ab/test/continuous_space_tests.jl) Made some more updates to the functions that I originally missed and converted tuples to svectors in these tests. Some of the previous errors are gone, but everything that calls `nearby_ids` fails; the stacktraces all start the same:
```
Got exception outside of a @test
  type GridSpace has no field space
  Stacktrace:
    [1] getproperty(x::GridSpace{2, true}, f::Symbol)
      @ Base ./Base.jl:33
    [2] notimplemented(model::GridSpace{2, true})
      @ Agents ~/.julia/dev/Agents/src/core/space_interaction_API.jl:32
    [3] nearby_ids(position::SVector{2, Int64}, model::GridSpace{2, true}, r::Float64)
      @ Agents ~/.julia/dev/Agents/src/core/space_interaction_API.jl:74
    [4] nearby_ids(pos::SVector{2, Float64}, model::AgentBasedModel{ContinuousSpace{2, true, Float64, typeof(Agents.no_vel_update)}, SpeedyContinuousAgent, typeof(Agents.Schedulers.fastest), Nothing, MersenneTwister}, r::Float64; exact::Bool)
      @ Agents ~/.julia/dev/Agents/src/spaces/continuous.jl:205
    [5] nearby_ids(pos::SVector{2, Float64}, model::AgentBasedModel{ContinuousSpace{2, true, Float64, typeof(Agents.no_vel_update)}, SpeedyContinuousAgent, typeof(Agents.Schedulers.fastest), Nothing, MersenneTwister}, r::Float64)
      @ Agents ~/.julia/dev/Agents/src/spaces/continuous.jl:195
```

Datseris:
The key point is that `ContinuousSpace` internally has a `GridSpace` that compartisizes it for accelerating nearest neighbor searches. The grid space takes tuples of integers, and all its internal searching does the same. So you just need to be careful when you pass things to the `GridSpace` searching functionality. 

I'll review this PR properly soon, but having passing tests would be really nice before I review.

mastrof:
Just realised that `nearby_ids` works with the `GridSpace` that underlies the `ContinuousSpace`, so everything going through to `GridSpace` must be a tuple.
It seems like a simple tuple conversion here https://github.com/JuliaDynamics/Agents.jl/blob/83129042f01673f832e4a32de53b93ecd6af80ab/src/spaces/continuous.jl#L204-L205 
redefining `focal_cell = Tuple(pos2cell, model)` fixes most errors.

<pre><font color="#F1F1F0"><b>Test Summary:          | </b></font><font color="#5AF78E"><b>Pass  </b></font><font color="#FF5C57"><b>Error  </b></font><font color="#9AEDFE"><b>Total</b></font>
ContinuousSpace        | <font color="#5AF78E">  69  </font><font color="#FF5C57">    1  </font><font color="#9AEDFE">   70</font>
  space initialization | <font color="#5AF78E">   4  </font>       <font color="#9AEDFE">    4</font>
  add/kill/move agent  | <font color="#5AF78E">   9  </font>       <font color="#9AEDFE">    9</font>
  nearby ids           | <font color="#5AF78E">  21  </font>       <font color="#9AEDFE">   21</font>
  Interacting pairs    | <font color="#5AF78E">  21  </font>       <font color="#9AEDFE">   21</font>
  nearest neighbor     | <font color="#5AF78E">   4  </font>       <font color="#9AEDFE">    4</font>
  walk                 | <font color="#5AF78E">   9  </font>       <font color="#9AEDFE">    9</font>
  collisions           | <font color="#5AF78E">   1  </font><font color="#FF5C57">    1  </font><font color="#9AEDFE">    2</font>
<font color="#FF5C57"><b>ERROR: </b></font>LoadError: <font color="#FF5C57">Some tests did not pass: 69 passed, 0 failed, 1 errored, 0 broken.</font></pre>

Still need to check the errors in the collisions testset.

Datseris:
@mastrof test still fail fyi

mastrof:
I forgot to include a `using StaticArrays` in `runtests.jl`.
This is what I got now from runtests. There are still some errors happening with pathfinding and gridspaces but I didn't look into those
<pre><font color="#F1F1F0"><b>Test Summary:              | </b></font><font color="#5AF78E"><b>Pass  </b></font><font color="#FF5C57"><b>Error  </b></font><font color="#9AEDFE"><b>Total</b></font>
Agents.jl Tests            | <font color="#5AF78E"> 893  </font><font color="#FF5C57">    5  </font><font color="#9AEDFE">  898</font>
  @agent macro             | <font color="#5AF78E">  11  </font>       <font color="#9AEDFE">   11</font>
  Model construction       | <font color="#5AF78E">  23  </font>       <font color="#9AEDFE">   23</font>
  add_agent! (discrete)    | <font color="#5AF78E">  10  </font>       <font color="#9AEDFE">   10</font>
  move_agent!              | <font color="#5AF78E">   6  </font>       <font color="#9AEDFE">    6</font>
  kill_agent!              | <font color="#5AF78E">   8  </font>       <font color="#9AEDFE">    8</font>
  genocide!                | <font color="#5AF78E">   9  </font>       <font color="#9AEDFE">    9</font>
  fill space               | <font color="#5AF78E"> 503  </font>       <font color="#9AEDFE">  503</font>
  model step order         | <font color="#5AF78E">   2  </font>       <font color="#9AEDFE">    2</font>
  Higher order groups      | <font color="#5AF78E">  20  </font>       <font color="#9AEDFE">   20</font>
  Random Number Generation | <font color="#5AF78E">   5  </font>       <font color="#9AEDFE">    5</font>
  sample!                  | <font color="#5AF78E">  12  </font>       <font color="#9AEDFE">   12</font>
  random agent             | <font color="#5AF78E">   3  </font>       <font color="#9AEDFE">    3</font>
  Scheduler tests          | <font color="#5AF78E">  46  </font>       <font color="#9AEDFE">   46</font>
  Model Access             | <font color="#5AF78E">  33  </font>       <font color="#9AEDFE">   33</font>
  Space                    | <font color="#5AF78E">  24  </font>       <font color="#9AEDFE">   24</font>
  GridSpace                | <font color="#5AF78E">  92  </font><font color="#FF5C57">    2  </font><font color="#9AEDFE">   94</font>
    size, dim=1            | <font color="#5AF78E">   6  </font>       <font color="#9AEDFE">    6</font>
    size, dim=3            | <font color="#5AF78E">   6  </font>       <font color="#9AEDFE">    6</font>
    add/move/kill          | <font color="#5AF78E">   6  </font>       <font color="#9AEDFE">    6</font>
    positions + empty      | <font color="#5AF78E">  11  </font>       <font color="#9AEDFE">   11</font>
    Distances              |       <font color="#FF5C57">    2  </font><font color="#9AEDFE">    2</font>
      Euclidean distance   |       <font color="#FF5C57">    1  </font><font color="#9AEDFE">    1</font>
      Manhattan Distance   |       <font color="#FF5C57">    1  </font><font color="#9AEDFE">    1</font>
    Nearby pos/ids/agents  | <font color="#5AF78E">  36  </font>       <font color="#9AEDFE">   36</font>
    Random nearby          | <font color="#5AF78E">   5  </font>       <font color="#9AEDFE">    5</font>
    walk!                  | <font color="#5AF78E">  22  </font>       <font color="#9AEDFE">   22</font>
  GridSpaceSingle          | <font color="#5AF78E">  86  </font><font color="#FF5C57">    2  </font><font color="#9AEDFE">   88</font>
    size, dim=1            | <font color="#5AF78E">   6  </font>       <font color="#9AEDFE">    6</font>
    size, dim=3            | <font color="#5AF78E">   6  </font>       <font color="#9AEDFE">    6</font>
    add/move/kill          | <font color="#5AF78E">   6  </font>       <font color="#9AEDFE">    6</font>
    positions + empty      | <font color="#5AF78E">   5  </font>       <font color="#9AEDFE">    5</font>
    Distances              |       <font color="#FF5C57">    2  </font><font color="#9AEDFE">    2</font>
      Euclidean distance   |       <font color="#FF5C57">    1  </font><font color="#9AEDFE">    1</font>
      Manhattan Distance   |       <font color="#FF5C57">    1  </font><font color="#9AEDFE">    1</font>
    Nearby pos/ids/agents  | <font color="#5AF78E">  36  </font>       <font color="#9AEDFE">   36</font>
    Random nearby          | <font color="#5AF78E">   5  </font>       <font color="#9AEDFE">    5</font>
    walk!                  | <font color="#5AF78E">  22  </font>       <font color="#9AEDFE">   22</font>
</pre>

AayushSabharwal:
Both the issues seem to be that the `add_agent!` method is passing `NTuple` for velocity instead of `SVector`

lmiq:
Chimming in:

`StaticArrays` also support `FieldVector`, which is a very practical way to represent positions. As far as I can see, mostly of what is changed here are signatures, so why not define at the top:

```
const PositionVector = Union{NTuple, SVector, FieldVector}
```
such that it will be much easier to extend this functionality if needed? 

Also it seems to me that a better approach would be to make the functions return the type inferred from the input types, not with conversions on return. For instance, would be too bad to define:
```
mutable struct Agent6{Vec<:PositionVector} <: AbstractAgent
    id::Int
    pos::Vec
    vel::Vec
    weight::Float64
end
```
and then use the the propagated `Vec` to infer the return types?

src/spaces/continuous.jl
```diff
@@ -110,7 +110,7 @@ end
 
 distance_from_cell_center(pos, model::ABM) =
     distance_from_cell_center(pos, cell_center(pos, model))
-function distance_from_cell_center(pos::Tuple, center::Tuple)
+function distance_from_cell_center(pos::SVector, center::SVector)
     sqrt(sum(abs2.(pos .- center)))
```
lmiq, line 114: If only `SVectors` will be supported, this is simply `norm(pos - center)`.

src/spaces/utilities.jl
```diff
@@ -163,12 +163,12 @@ spaces this clamps the position to the space extent.
 normalize_position(pos, model::ABM) = normalize_position(pos, model.space)
 
 function normalize_position(pos, space::ContinuousSpace{D,true}) where {D}
-    return mod.(pos, spacesize(space))
+    return mod.(pos, SVector(spacesize(space)))
 end
 
 function normalize_position(pos, space::ContinuousSpace{D,false}) where {D}
```
lmiq, line 169: This functions could be written as:
```
function normalize_position(pos::T, space::ContinuousSpace{D,false}) where {T,D}
    ss = spacesize(space)
    return T(clamp.(pos, 0.0, prevfloat.(ss)))
end
```

to support Tuples, SVectors, and possibily FieldVectors as well.

src/spaces/utilities.jl
```diff
@@ -232,7 +232,7 @@ walk!(agent, ::typeof(rand), model::ABM{<:AbstractGridSpace{D}}; kwargs...) wher
     walk!(agent, Tuple(rand(model.rng, -1:1, D)), model; kwargs...)
 
 walk!(agent, ::typeof(rand), model::ABM{<:ContinuousSpace{D}}) where {D} =
-    walk!(agent, Tuple(2.0 * rand(model.rng) - 1.0 for _ in 1:D), model)
+    walk!(agent, SVector{D}(2.0 * rand(model.rng) - 1.0 for _ in 1:D), model)
```
lmiq, line 235:
```
walk!(agent, typeof(agent.pos)(2.0 * rand(model.rng) - 1.0 for _ in 1:D), model)
```

Datseris:
```
mutable struct Agent6{Vec<:PositionVector} <: AbstractAgent
```
Agents are no longer defined manually by users. Only with the `@agent` macro. With the macro one can't inherit Union types... Furthermore, remember that we are writing generic functions for _any_ agent. I don't see how we could use the `Vec` parameterization in `move_agent!` for example. Like, how would we identify where is `Vec`? All agents have `Vec` as first parameterization? If we enfoce it why not enforce the position type already?

But most importantly, I am not convinced that there is a benefit in supporting general position types. Why should a position be different than `SVector`? Arguments for are welcomed, because I see many arguments against, e.g., the kinda pretty large development effort and the larger test suite.

lmiq:
I don't have very strong feelings about that.  FieldVectors are convenient because you access each position by name (i.e. `pos.x`, `pos.y`). 

Otherwise is to allow more ecosystem integration. For example, recently Julio Hoffimann, from JuliaEarth, was implemnting a custom type of SVector for their use. 

Anyway, if the `agent.pos` field must exist, it is also possible to use `typeof` everywhere, or just add an internal field in the model that carries the type.

Just for thought.

Datseris:
i'm ok with using `typeof` wherever necessary. In many cases only simple broadcast should be used, which preservee types. Only when using `spacesize()` which must return a concrete SVector and not care about agent position type, only then an explicit usage of `typeof(agent.pos)` is necessary.

We'd also need to make one more version of `ContinuousAgent{VectorType}` that intializes a type with specific vector values. So then `ContinuousAgent{D<:Number}` becomes the SVector version.

mastrof:
Sorry I have been away for a while due to conferences and holidays.
I have some interest in moving forward with this pr. I can start by addressing the reviews by @lmiq and fixing the errors in pathfinder, but not sure how to deal with the support for more general vector types.

Datseris:
Don't consider general vector types at the moment. This is something for a different PR. The goal of this PR is simple and focused on SVector{Float64}.

lmiq:
Without complicating too much, the only thing that I would consider more in depth is to support `FieldVector`. Then those suggestions are useful, and the signature could be an alias for the Union.

mastrof:
I get 1749/1749 passed tests as of last commit.
Even if everything seems to work, pathfinder is still somewhat messy. Since the algorithms rely on
https://github.com/JuliaDynamics/Agents.jl/blob/8f199bb3733389181c91bbba3040603b5d555b34/src/submodules/pathfinding/astar.jl#L6
there are some objects being converted back and forth between SVector and Tuple.
Maybe a distinction should be made between a DiscretePath (or GridPath) and a ContinuousPath?

AayushSabharwal:
I'm busy and a little sick, but I'll take a look at this as soon as I can. As long as pathfinding works, it being messy is something I can look into later if necessary.

Datseris:
I would say to keep this open because it is defeinitely the plan to make continuous space use static vectors, the question is for everyone to find the time to implement, review, and propagate changes everywhere. So I'd recommend to leave this open @mastrof

mastrof:
Whoops I didn't plan to close, my bad.

PR #676, opened 2022-08-14
Fix elastic collision

### LINKED ISSUE #675: The social distancing model does not look like elastic collision
https://juliadynamics.github.io/AgentsExampleZoo.jl/dev/examples/social_distancing/

It seems the balls are going faster and faster.

sosiristseng:
fix #675

I also renewed the facing logic to:

```julia
!(dot(r2, v1) > 0 && dot(r1, v2) > 0) && return false
```

### DIFF
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -346,7 +346,7 @@ function elastic_collision!(a, b, f=nothing)
         f1 = (2m2 / (m1 + m2))
         f2 = (2m1 / (m1 + m2))
     end
-
+    dv = a.vel .- b.vel
     a.vel = v1 .- f1 .* (dot(dv, r1) / n) .* (r1)
     b.vel = v2 .+ f2 .* (dot(dv, r2) / n) .* (r2)
     return true
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -13,8 +13,8 @@ Base.eltype(s::ContinuousSpace{D,P,T,F}) where {D,P,T,F} = T
 no_vel_update(a, m) = nothing
 spacesize(space::ContinuousSpace) = space.extent
 function Base.show(io::IO, space::ContinuousSpace{D,P}) where {D,P}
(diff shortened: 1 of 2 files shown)


sosiristseng:
However, some disks would pass each other. Is there any tip so I can fix that?

https://user-images.githubusercontent.com/40054455/184536419-411f9e61-955f-410d-9c84-5c8e32e67caf.mp4

sosiristseng:
The current facing logic does not consider the lower one.
Elastischer_stoß2

sosiristseng:
The new logic using dv looks better but fails to pass one collision test.

https://user-images.githubusercontent.com/40054455/184538283-f857a050-07db-4954-abb0-0ea46592ef16.mp4

sosiristseng:
I'm reading these lines in tests.

https://github.com/JuliaDynamics/Agents.jl/blob/83129042f01673f832e4a32de53b93ecd6af80ab/test/continuous_space_tests.jl#L324-L330

In the tests, only 50 out of 100 agents are movable, so `x` is at most 50. However, they can collide many times in the simulation (167 times in my case). Should we change the test case `@test x ≥ model.c/2`?

Datseris:
Thanks so much, this is a great fix!

Yes please, go ahead and correct the test.

Datseris:
Ummm @sosiristseng are you perhaps using an automated linter? Can you please revert the linter induced changes? We do not use a linter yet in Agents.jl, but will in the future. Until then, it is best to not include linting changes in pull requests.

sosiristseng:
I left my formatter on. Will fix that, too.

Datseris (approved):
Thank you very much, we appreciate the fix!

sosiristseng:
Hello, since this is a correctness fix, would you deploy a patch version (like 5.4.4) containing this PR? Or would you like this PR to go into a bigger new version (like 5.5.0) with other PRs?

Datseris:
I will release 5.5 soon with this in. Can't promise when though! At the moment the `main` branch is already too far away from 5.4 to publish a patch release. The mean time, feel free to use the development version!

PR #677, opened 2022-08-16
some doc formatting updates

lmiq:
After seeing the final rendered page, I think these formatting updates will look better. I really apologize in advance because I don't know how to see the rendered page before it is published, using this system, so I'm not sure, for example, if the `\textrm` LaTeX command will render correctly.


Datseris:
oh don't worry at all. I also merged because I wanted to see the rendered page and I was too lazy to do it locally. For the future, it is super easy for you to render the docs. Just run `docs/make.jl` locally and in the `built` folder there will be the html pages.

PR #678, opened 2022-08-17
fixed subindices in docs

lmiq:
Just fixed the subindices in the formula of the potential energy function.


lmiq:
sorry, looks sort of messy, I'll try again

PR #682, opened 2022-09-14
Use `LightOSM.nearest_way` for `nearest_road`

### LINKED ISSUE #681: Improve `nearest_road` using new LightOSM function `nearest_way`
Referring to https://github.com/DeloitteDigitalAPAC/LightOSM.jl/pull/80

This issue is mostly for my own reference. The change is likely trivial.

AayushSabharwal:
Closes #681 

Should we allow the user to tune `search_radius` (third parameter to [nearest_way](https://deloittedigitalapac.github.io/LightOSM.jl/stable/nearest_way/#LightOSM.nearest_way)), potentially causing the method to return `nothing` (and document it, of course)

### DIFF
--- a/src/spaces/openstreetmap.jl
+++ b/src/spaces/openstreetmap.jl
@@ -540,7 +540,8 @@ end
 """
     OSM.nearest_road(lonlat::Tuple{Float64,Float64}, model::ABM{<:OpenStreetMapSpace})
 
-Return a location on a road nearest to **(longitude, latitude)**. 
+Return a location on a road nearest to **(longitude, latitude)**. Slower, but more
+precise than [`OSM.nearest_node`](@ref).
 """
 function nearest_road(ll::Tuple{Float64,Float64}, model::ABM{<:OpenStreetMapSpace})
     geoloc = GeoLocation(ll[2], ll[1], 0.0)
--- a/src/spaces/openstreetmap.jl
+++ b/src/spaces/openstreetmap.jl
@@ -540,47 +540,23 @@ end
 """
     OSM.nearest_road(lonlat::Tuple{Float64,Float64}, model::ABM{<:OpenStreetMapSpace})
 
-Return a location on a road nearest to **(longitude, latitude)**.
-Significantly slower, but more precise than [`OSM.nearest_node`](@ref).
+Return a location on a road nearest to **(longitude, latitude)**. 
 """
 function nearest_road(ll::Tuple{Float64,Float64}, model::ABM{<:OpenStreetMapSpace})
-    ll = reverse(ll)
-    best_sq_dist = Inf
-    best = (-1, -1, -1.0)
-    pt = LightOSM.to_cartesian(GeoLocation(ll..., 0.0))
-    for e in edges(model.space.map.graph)
-        s = LightOSM.to_cartesian(GeoLocation(
-            model.space.map.node_coordinates[src(e)]..., 0.0))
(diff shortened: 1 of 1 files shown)


Datseris:
The possibility to return `nothing` exists regardless of whether the user can tune `search_radius` so we should definitely allow it and document it. I didn't undewrstand why you removed the docstring line "Significantly slower, but more precise than [`OSM.nearest_node`](@ref)."

AayushSabharwal:
> I didn't undewrstand why you removed the docstring line

I intend to run benchmarks to verify if this is actually significantly slower now or not.

> The possibility to return nothing exists regardless of whether the user can tune search_radius so we should definitely allow it and document it.

Does it, though? The default `search_radius` is the distance to the nearest node from the given position. The comment linked in the code mentions that for now, every node would be attached to a way, so there will always be a way within the default search bounds.

Datseris:
Okay, you are right. I have misunderstood. So then I wait the benchmark numbers!

AayushSabharwal:
```julia
# Setup from osm_test.jl
julia> goettingen = OSM.test_map()
julia> space = OpenStreetMapSpace(goettingen; network_type = :none, weight_type = :time)
julia> mutable struct AgentOSM <: AbstractAgent
           id::Int
           pos::Tuple{Int,Int,Float64}
       end
julia> model = ABM(AgentOSM, space)
julia> start_lonlat = (9.9351811, 51.5328328)
julia> @benchmark OSM.nearest_road($start_lonlat, $model)
 Range (min … max):  12.412 μs …  6.999 ms  ┊ GC (min … max): 0.00% … 99.36%
 Time  (median):     13.200 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   14.989 μs ± 69.950 μs  ┊ GC (mean ± σ):  4.64% ±  0.99%

  ▅██▆▃▁  ▂▃▃▁                                                ▁
  ██████▇▇█████▇▇██▆▆▅▅▅▅▅▄▄▄▅▄▅▅▅▄▅▆▅▅▆▆▆▆▅▅▅▆▅▅█▇▆▅▆▅▄▆▄▄▄▅ █
  12.4 μs      Histogram: log(frequency) by time      33.7 μs <

 Memory estimate: 5.28 KiB, allocs estimate: 114.

julia> @benchmark OSM.nearest_node($start_lonlat, $model)
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.682 μs … 664.672 μs  ┊ GC (min … max): 0.00% … 99.11%
 Time  (median):     1.754 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.136 μs ±   6.718 μs  ┊ GC (mean ± σ):  3.08% ±  0.99%

  ██▄▂▁▅▄▂▁▃▂                             ▃▁                  ▂
  ████████████▇▇▇▅▆▆▆▆▆▆▆▆▆▆▆▇█████▇▇▆▇▇▅▆██▇▆██▇█▇█▆▇▆▆▆▅▄▄▅ █
  1.68 μs      Histogram: log(frequency) by time      4.77 μs <

 Memory estimate: 928 bytes, allocs estimate: 16.
```

Datseris:
Okay but still, a difference of 100x is significant, doens't matter how fast both methods are, so we should say it. Feel free to re-update the docstring and merge, as far as I can tell this is LGTM and tests pass.

Datseris (approved):

Issue #684, opened 2022-09-15
nearby_ids output changes with translated positions; wrong output


hbsmith:
**Describe the bug**
`nearby_ids` and `nearby_ids_exact` produce incorrect output in some situations, although I can't figure out what triggers it

**Minimal Working Example**
```julia
using Revise
using Agents, Plots, Random
using Test

mutable struct TestAgent <: AbstractAgent
    id::Int
    pos::NTuple{2,Float64}
end

distance(a1::TestAgent,a2::TestAgent) = let δ = a1.pos .- a2.pos
  hypot(δ...)
end
  
positions = [
  (122.91984594569145, 129.27537099976024),
  (119.8316511853557, 120.63347075797228),
  (120.88133497820543, 122.44001918308183),
  (118.33581146746528, 121.94704974817539),
  (120.22776179237353, 121.74193334411757)]

space = ContinuousSpace((300,300),periodic=true)
model = ABM(TestAgent, space)

for i in positions
  add_agent!(i,model)
end
```
Now looking at the calculated distances between agents compared to the what is returned by `Agents.nearby_ids` and `Agents.nearby_ids_exact`
```julia

julia> for a in 1:4
  d = distance(model[5],model[a])
  println("Distance between 5 and $a: $d")
end

Distance between 5 and 1: 8.000000000000027
Distance between 5 and 2: 1.1771121093064787
Distance between 5 and 3: 0.9562853903521156
Distance between 5 and 4: 1.9030367235379475

julia> collect(Agents.nearby_ids_exact(model[5],model,2))
1-element Vector{Int64}:
 3
## Expecting this to be [2, 3, 4]

julia> collect(Agents.nearby_ids(model[5],model,2))
2-element Vector{Int64}:
 1
 3
## Expecting this to be [2, 3, 4] (or [1, 2, 3, 4] based on the ContinuousSpace docstring)
```

Changing to `periodic=false` gives the same results.

But sometimes translating the positions gives the expected answer...
```julia
positions = [i .- 100 for i in positions]

## Rerun initialization

julia> for a in 1:4
  d = distance(model[5],model[a])
  println("Distance between 5 and $a: $d")
end

Distance between 5 and 1: 8.000000000000027
Distance between 5 and 2: 1.1771121093064787
Distance between 5 and 3: 0.9562853903521156
Distance between 5 and 4: 1.9030367235379475

julia> collect(Agents.nearby_ids_exact(model[5], model, 2))
[...]

PR #685, opened 2022-09-22
add nearest_neighbor_ties

hbsmith:
pretty simple, this just adds a function in continuous space that does the same thing as `nearest_neighbor`, except ties are allowed, and it returns a vector of agents instead of a single agent (or returns an empty vector).

tests added as well.

running a simple benchmark to one of the tests (which I stole from benchmark/benchmarks.jl) we see they hover about within 500ns of each other, and of course returning a vector means that we allocate a tiny bit of memory:
```bash
julia> @btime nearest_neighbor($a, $continuous_model, 5)
  54.917 μs (0 allocations: 0 bytes)
ContinuousAgent(138, (3.33, 8.325, 6.66), (0.8, 0.7, 1.3), 6.5, false)

julia> @btime nearest_neighbor_ties($a, $continuous_model, 5)
  55.372 μs (3 allocations: 208 bytes)
2-element Vector{ContinuousAgent}:
 ContinuousAgent(138, (3.33, 8.325, 6.66), (0.8, 0.7, 1.3), 6.5, false)
 ContinuousAgent(132, (3.33, 6.66, 8.325), (0.8, 0.7, 1.3), 6.5, false)
```

benchmark:
```julia
mutable struct AgentNNBContinuous <: AbstractAgent
    id::Int
    pos::NTuple{3,Float64}
    vel::NTuple{3,Float64}
    one::Float64
    two::Bool
end

continuous_model = ABM(AgentNNBContinuous, ContinuousSpace((10.0, 10.0, 10.0); spacing = 0.5))
continuous_agent = AgentNNBContinuous(1, (2.2, 1.9, 7.5), (0.5, 1.0, 0.01), 6.5, false)

for x in range(0, stop = 9.99, length = 7)
    for y in range(0, stop = 9.99, length = 7)
        for z in range(0, stop = 9.99, length = 7)
            add_agent!((x, y, z), continuous_model, (0.8, 0.7, 1.3), 6.5, false)
        end
    end
end

a = continuous_model[139]
pos = (7.07, 8.10, 6.58)
```

See https://github.com/JuliaDynamics/Agents.jl/issues/684 for some discussion on the issue.

### DIFF
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -306,7 +306,8 @@ function nearest_neighbor_ties(agent::A, model::ABM{<:ContinuousSpace,A}, r) whe
     for id in n
         dnew = euclidean_distance(agent.pos, model[id].pos, model)
         if dnew < d
-            d, j = dnew, Int[id]
+            empty!(j)
+            d, j = dnew, push!(j, id)
         elseif dnew == d && id ∉ j
             push!(j, id)
         end
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -1,6 +1,6 @@
 export ContinuousSpace, ContinuousAgent
 export nearby_ids_exact, nearby_agents_exact
-export nearest_neighbor, elastic_collision!, interacting_pairs
+export nearest_neighbor, nearest_neighbor_ties, elastic_collision!, interacting_pairs
 
 struct ContinuousSpace{D,P,T<:AbstractFloat,F} <: AbstractSpace
     grid::GridSpace{D,P}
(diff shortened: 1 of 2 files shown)


Datseris:
I thought you were talking about discrete space... 

How possible is it even that two neighbors would have **exactly the same** distance in **continuous space**? Isn't the probability for that as close to zero as possible? 

...Do we really need this function? For such a low low probability event...? Like how often would a user have to worry about it?

Datseris:
I'm really sorry to decline a PR, it always makes me feel bad, but sometimes it doesn't fit well or it is not super necessary :/

PR #686, opened 2022-09-27
Reverted `n` parameter in ensemblerun! to correct pre-5.4.x version

mt-digital:
I added a small bug in my PR #624 for updating `ensemblerun!` for improved parallelism and progressbars. My change (reverted here) makes dispatching work incorrectly when `models` is a vector/tuple and `n` is a stopping function. This PR reverts that change so expected behavior is recovered. Sorry for this needless hassle!

### DIFF
--- a/src/simulations/ensemblerun.jl
+++ b/src/simulations/ensemblerun.jl
@@ -31,7 +31,7 @@ function ensemblerun!(
     models::Vector_or_Tuple,
     agent_step!,
     model_step!,
-    n::Int;
+    n;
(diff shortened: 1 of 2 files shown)


Datseris:
No worries, but this means we need to add a test that catches the case that fails that you just described! Please add that extra 1-liner test? :)

mt-digital:
OK, I think that should be good. Sorry to make you ask ;)

Datseris (approved):
Lovely, thanks!

PR #690, opened 2022-10-11
Update utilities.jl

feyereisl:
The documentation of walk!() and the default value of parameter 'ifempty' are inconsistent with each other. One states 'false', the other 'true'.


Datseris:
Oh yeah thanks! cc also @mrchaos because I guess that's the reason the wolf sheep example broke.

PR #691, opened 2022-10-13
In the wolf `walk!`, parameter update

mrchaos:
In the wolf `walk!`, `ifempty=false` is required for the wolf to eat the sheep. Without `ifempty=false` the wolf would starve to death.


sooyeon0901 (commented):
I tried same thing, this is right way.

sooyeon0901 (approved):

Datseris:
Thanks both!

PR #692, opened 2022-10-24
Correction of "position_delta"

mrchaos:
In case of poistion from (2,1) to (4,4), correct delta is (2,3) 
But current `position_delta` result is (2,2)
So I have correct the `position_delta`

### DIFF
--- a/src/submodules/pathfinding/astar_continuous.jl
+++ b/src/submodules/pathfinding/astar_continuous.jl
@@ -2,9 +2,12 @@ to_discrete_position(pos, pathfinder) =
     floor.(Int, pos ./ pathfinder.dims .* size(pathfinder.walkmap)) .+ 1
 to_continuous_position(pos, pathfinder) =
     pos ./ size(pathfinder.walkmap) .* pathfinder.dims .-
-    pathfinder.dims ./ size(pathfinder.walkmap) ./ 2.
-sqr_distance(from, to, pathfinder::AStar{D,true}) where {D} =
-    sum(min.(abs.(from .- to), pathfinder.dims .- abs.(from .- to)) .^ 2)
+    pathfinder.dims ./ size(pathfinder.walkmap) ./ 2.    
+function sqr_distance(from, to, pathfinder::AStar{D,true}) where {D} 
+    Δ = abs.(to .- from)
+    Δ= any(Δ .> pathfinder.dims) ? Δ .% pathfinder.dims : Δ
+    return sum(Δ .^ 2)
+end
 sqr_distance(from, to, pathfinder::AStar{D,false}) where {D} =
     sum((from .- to) .^ 2)
 
--- a/src/submodules/pathfinding/astar_continuous.jl
+++ b/src/submodules/pathfinding/astar_continuous.jl
@@ -5,7 +5,7 @@ to_continuous_position(pos, pathfinder) =
     pathfinder.dims ./ size(pathfinder.walkmap) ./ 2.    
 function sqr_distance(from, to, pathfinder::AStar{D,true}) where {D} 
     Δ = abs.(to .- from)
-    Δ= any(Δ .> pathfinder.dims) ? Δ .% pathfinder.dims : Δ
+    Δ = any(Δ .> pathfinder.dims) ? Δ .% pathfinder.dims : Δ
     return sum(Δ .^ 2)
 end
 sqr_distance(from, to, pathfinder::AStar{D,false}) where {D} =
--- a/src/submodules/pathfinding/pathfinding_utils.jl
+++ b/src/submodules/pathfinding/pathfinding_utils.jl
@@ -5,10 +5,7 @@ periodicity of `pathfinder`.
 """
 function position_delta(pathfinder::GridPathfinder{D,true}, from::Dims{D}, to::Dims{D}) where {D} 
     Δ = abs.(to .- from)
-    walkmap_size = size(pathfinder.walkmap)
-    if any(Δ .> walkmap_size)
-        Δ = Δ .% walkmap_size
-    end
+    Δ = any(Δ .> pathfinder.dims) ? Δ .% pathfinder.dims : Δ    
     return Δ
 end
 
--- a/src/submodules/pathfinding/pathfinding_utils.jl
+++ b/src/submodules/pathfinding/pathfinding_utils.jl
@@ -3,8 +3,14 @@
 Returns the absolute difference in coordinates between `from` and `to` taking into account
 periodicity of `pathfinder`.
 """
-position_delta(pathfinder::GridPathfinder{D,true}, from::Dims{D}, to::Dims{D}) where {D} =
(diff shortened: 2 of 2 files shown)


Datseris:
All tests fail here.

AayushSabharwal:
I'll check the code in a while, but this statement is wrong:
> In case of poistion from (2,1) to (4,4), correct delta is (2,3)
> But current position_delta result is (2,2)
So I have correct the position_delta

Assuming you mean a grid of size (5,5), the delta is (2,2): move 2 down, and 2 left (wrap around the left edge)

AayushSabharwal:
So the only change is `position_delta`? It seems like you reverted your own wolf-sheep change

Datseris:
Yeah something tells me the currect code is correct and the PR code is wrong :/

mrchaos:
position_delta(pathfinder::GridPathfinder{D,true}, from::Dims{D}, to::Dims{D}) where {D} =
    min.(abs.(to .- from), size(pathfinder.walkmap) .- abs.(to .- from))
    
 D = 2
 size(pathfinder.walkmap) = (5,5)
 from = (1,4)
 to = (5,5)
 
 `min.(abs.(to .- from), size(pathfinder.walkmap) .- abs.(to .- from)) = (1,1) `
 
 Is (1,1) right ?

AayushSabharwal:
> Is (1,1) right ?

Yes, 1 up 1 right

> I think the answer is (4,1), is this wrong?

Please note that the grid is periodic (at least, you're quoting the method for a periodic grid). This means opposite edges are adjacent, so going up from (1,4) lands you in (5,4). The answer would be (4,1) if the grid were not periodic.

> If from=(1,4), to=(7,7) then current position_delta = (-1,2) , is right ?

I'm not sure what you mean. (7,7) is outside a grid of size (5,5). And `position_delta` documents that it returns the *absolute* distance, which means that the returned value cannot have a negative component.

mrchaos:
When the grid size is (5,5), when going from (1,1) to (3,4), the position_delta should return the same value whether periodic or not because both (1,1) and (3,4) are within the grid size. .
But the current position_delta is wrong because it is yielding different values.

And if the grid is periodic, if coordinates that exceed the grid size are input, clipping should be done to calculate the correct coordinate values, but now the wrong values ​​are calculated.
By default, periodic coordinates use a modulus calculation, but no such calculation is shown anywhere.

If you look at the code below, you can see that the output is different.

  from = (1,1)
  to = (3,4)
  
  pathfinder = AStar(GridSpace((5, 5)))  
  @show position_delta(pathfinder,from,to)
  ==> (2,2)
  
  pathfinder = AStar(GridSpace((5, 5);periodic=false))
  @show position_delta(pathfinder,from,to)
  ==> (2,3)

mrchaos:
If you look at the aster_test.jl file, there is the following test code.
@test a.pos == (1, 3)
@test plan_best_route!(a, [(5, 1), (1, 1), (3, 3)], model.pf) == (5, 1)
What do you think of the test code result above?
The condition parameter of plan_best_route! is :shortest by default, how can the result be (5,1)?

In this case, if delta_position is the same as the current one, the above incorrect result is produced.

mrchaos:
> > Is (1,1) right ?
> 
> Yes, 1 up 1 right
> 
> > I think the answer is (4,1), is this wrong?
> 
> Please note that the grid is periodic (at least, you're quoting the method for a periodic grid). This means opposite edges are adjacent, so going up from (1,4) lands you in (5,4). The answer would be (4,1) if the grid were not periodic.
> 
> > If from=(1,4), to=(7,7) then current position_delta = (-1,2) , is right ?
> 
> I'm not sure what you mean. (7,7) is outside a grid of size (5,5). And `position_delta` documents that it returns the _absolute_ distance, which means that the returned value cannot have a negative component.

실제 소스코드를 확인하고 실행해봐라 그러면 음수 값이 나올 것이다.
periodic인 이유는 입력값이 grid size를 넘어가도 grid size안으로 들어 올수 있게 하기 위한 것이다. 그런데  grid size에 안들어 오는 값을 산출한다는것은 틀린것 이다.

Datseris:
offtopic, but in the future @mrchaos please use backticks ` to format code, it makes your comments much easier to read.

Multiple lines use three backticks, inline uses one.

AayushSabharwal:
> When the grid size is (5,5), when going from (1,1) to (3,4), the position_delta should return the same value whether periodic or not because both (1,1) and (3,4) are within the grid size

> And if the grid is periodic, if coordinates that exceed the grid size are input, clipping should be done to calculate the correct coordinate values, but now the wrong values ​​are calculated.

`position_delta` is only designed to work with valid coordinates inside the grid. Periodicity does not care about whether input coordinates are inside the grid, and a periodic grid does not mean that it accepts coordinates outside the grid. Periodicity means that the agent is free to move toward one edge of the grid and appear on the opposite edge, which modifies how distances are calculated. (1, 1) and (1, 5) are adjacent in a periodic grid of size (5, 5), whereas normally they wouldn't be. `position_delta` is designed to calculate absolute distances between coordinates while taking this into account.

Since `position_delta` is not part of the public API, there is not a question of what it "should" do. It does what is necessary to be useful where it is used internally. It is not intended to be used directly, and is not guaranteed to work for all inputs: just the inputs it is expected to receive internally. It is documented to make internal code easier to understand. If you think the docstring is misleading as to what it does, you are more than welcome to suggest changes.

> If you look at the aster_test.jl file, there is the following test code.
> ```
> @test a.pos == (1, 3)
> @test plan_best_route!(a, [(5, 1), (1, 1), (3, 3)], model.pf) == (5, 1)
> ```
> What do you think of the test code result above?
> The condition parameter of plan_best_route! is :shortest by default, how can the result be (5,1)?

`plan_best_route!` with `:shortest` takes the path with the lowest number of edges. In other words, the path where the agent moves to an adjacent cell the fewest number of times. In a periodic grid, all the targets in the test have a path length of 2, so `(5, 1)` is picked since it is the first one in the input vector. I can see that this may not be clear from the docstring, and is certainly a welcome addition.

I hope I've addressed your concerns adequately. If you have any more questions, I'd be happy to answer them.

mrchaos:
```
@test a.pos == (1, 3)
@test plan_best_route!(a, [(4,4), (5, 1), (1, 1), (3, 3)], model.pf) == (5, 1)
```
When (4,4) is added to the test code above, `plan_best_route!` answers (4,4) as the shortest path, not (1,1), (5,1), or (3,3). This means that `plan_best_route!` is producing the wrong result.
Please review the `plan_best_route!` code in detail. 
Correctly modifying the `position_delta` code will give you the correct answer.

AayushSabharwal:
(1, 3) -> (5, 4) -> (4, 4)

(4, 4) is also two steps away, so the code is working as intended.

mrchaos:
You are right.

But I wish it was more intuitive. When I set the grid size to (10,10), I confirmed that your code works well.
It would be better if periodic and non-periodic could give the same result when the agent's coordinates are in the grid range. In the meantime, thanks for your kind reply.

PR #694, opened 2022-10-29
Fix `nearest_way` not working

AayushSabharwal:
Default `OSMGraph` constructor functions don't call `LightOSM.create_rtree!`


Datseris:
increment patch version

PR #702, opened 2022-11-15
CI: Include run using Julia nightly

EwoutH:
Add a run to the test job of the CI which uses the latest nightly build of Julia. This allows to detect issues with upcoming Julia releases in advance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,11 @@ jobs:
         os:  [ubuntu-latest, windows-latest, macos-latest]
         arch:
           - x64
+        include:
+          - version: nightly 
+            os: ubuntu-latest
+            arch: x64
+            
     steps:
       # Cancel ongoing CI test runs if pushing to branch again before the previous tests
       # have finished


Datseris:
Hi, sorry but I disagree with this. I think it's a waste of energy. 

A minor or patch increase of Julia version (which is what the nightly builds are) should not be breaking Agents.jl in any way. Sure, there could be bugs, but I think it is super wasteful for all of the 1000s packages of Julia to run their test suite one additional time to detect bugs somewhere else; it's not their role either in my humble opinion. Furthermore the majority of times the nightly builds are broken anyways for unrelated reasons, providing such a large number of false positives that they aren't even helpful in the end.

So to conclude, and especially give the status quo of climate and energy landscape, I don't think the benefits of a nightly test run justify its cost.

EwoutH:
Thanks for your quick and extensive reply. I agree that blown out CI matrices and the energy that consumes are a huge waste and a serious problem.

What we could do is only run on pushes to the main branch or once a week. That saves from running on each PR. The test job is about 6 minutes long, on a single-core cloud instance, and it's only a single job, so I think that would be proportional to the costs.

In this PR the CI found [a single case](https://github.com/JuliaDynamics/Agents.jl/actions/runs/3468749088/jobs/5794918540#step:7:205) which fails with Julia v1.10.0-DEV.2 (the current nightly), which passes on the latest stable version (1.8.2). I do think that's interesting to have detected early.

fbanning:
> But the question that remains is who will use this output? [...] the majority of the signal being noise (false positives)

I have to fully agree with @Datseris on this topic. If there's not a concrete use case for testing nightly builds, then there's also nothing to analyse for us. Generating data just for the sake of having *more* data available is nothing I would recommend doing.

> In this PR the CI found [a single case](https://github.com/JuliaDynamics/Agents.jl/actions/runs/3468749088/jobs/5794918540#step:7:205) which fails with Julia v1.10.0-DEV.2 (the current nightly), which passes on the latest stable version (1.8.2). I do think that's interesting to have detected early.

I think this is a great example of the point above. Knowing that a highly unstable development version of `Julia v1.10.0` will create some failing tests with our current test suite that's specifically tailored towards stable releases on the three major OSs doesn't really add anything of value. We could take action on that but it will be a moving target and we would have to keep track of what was changed when and it would just bind resources that could be used otherwise. I seriously don't think that testing nightly builds will result in any tangible or even beneficial outcomes. Hence, why I would also vote to reject adding it to our CI tests.

EwoutH:
Thanks for both your feedback. For the part of who's going to look at it, it doesn't have to happen in real time. CI environments and dependencies change all the time, sometimes it's great when debugging a new CI failure to be able to bisectect when it's first happened. If there are multiple weeks between CI runs a lot could have changed which could cause the failure, so that's when looking back at periodically CI runs can be usesful.

I agree that testing against the nightly development versions is not of the most used, since it's both far away in the future and can be fixed relatively soon. For that you would ideally have a run that's allowed to fail, but unfortunately GitHub is not implementing this for unknown reason (see https://github.com/actions/toolkit/issues/399).

I think it would be more useful to test against Julia alpha and beta releases, because those are both more stable and will be in production relatively soon. I openend a feature request for setup-julia yesterday (https://github.com/julia-actions/setup-julia/issues/126), and a maintainer received that positively.

The end goal of such a run is to catch errors and deprecations in new Julia versions before they get stable. I think that can add value, but agree the noise and overhead should be as low as possible and proportional to the effort.

fbanning:
So this is going to be a personal statement and everyone's free to disagree (as usual), but...

> For that you would ideally have a run that's allowed to fail, but unfortunately GitHub is not implementing this for unknown reason

Allowing tests to fail is like allowing condoms to have holes. It doesn't make any sense. Tests, just like condoms, are there to prevent unwanted things from happening. If you poke a hole in the test suite (condom) you willingly accept that it doesn't do what it's meant to do in the first place. Making failing test suites still pass CI doesn't really help to take action on failing tests because nobody will take action if there are a dozen green check marks instead of a single red cross.

> The end goal of such a run is to catch errors and deprecations in new Julia versions before they get stable. I think that can add value, but agree the noise and overhead should be as low as possible and proportional to the effort.

For that you would have to have a *really* extensive test suite that covers all the eventualities that *might* arise. I'm not totally against this if there's interest from anyone to extend our tests with more use case. Our current tests only deal with basic models to test that fundamental functionalities work correctly.

But tbh it's virtually impossible to include every conceivable combination in our tests and it may sound harsh but we indeed have the best test suite in the world: our users and developers. Agent-based models are complex by nature and only the way that users implement them and make use of our Agents.jl framework will show us the real problems that are rooted in the inner workings of our package.

If Julia's minor or patch versions (`v0.x.0` or `v0.0.x`) are incremented, nothing breaking should happen in our package because that's how [semantic versioning](https://semver.org/) is conceived. According to our `Project.toml` we currently allow every Julia version from `1.5` upwards. If there's really a problem with a new minor/patch version of Julia, then that's a breaking change and not backwards compatible and that should be labeled as such (namely a major version change).

And even if you disagree with all that I've just said, @Datseris pointed out one of the core issues with this early on:

> So to conclude, and especially give the status quo of climate and energy landscape, I don't think the benefits of a nightly test run justify its cost.

Running CI tests that aren't really helpful nor used by anyone will result in (i) more unnecessary energy use which we should always try to avoid and (ii) bound computational resources that could be used for more productive uses.

EwoutH:
> Allowing tests to fail is like allowing condoms to have holes.

Like with ABMs, as the environment changes, behaviour changes. With tests that are allowed to fail, the use case is not to definitely say in binary if that's good or bad behaviour, but that it might be just interesting behaviour. It depends on why it's fails. And from the why something can be learned or prevented.

But yes, that takes a certain amount of time and maintenance effort. It buys you the fact that you discover an issue potentially before users encounter it, making the project more robust.

fbanning:
> With tests that are allowed to fail, the use case is not to definitely say in binary if that's good or bad behaviour, but that it might be just interesting behaviour.

But automated tests are binary, they either fail or succeed. Declaring a fail as a success in CI is just bad practice imho. If you want to find out *why* some interesting behaviour occurs in specific circumstances (e.g. when using a pre-release version of Julia), then manual local testing necessarily is the way to go. And that would indeed be very much encouraged if you would want to anticipate problems with a pending release! 💪🏻 

And although I find this discussion very interesting from my standpoint of a person who is not from CS, I think it deviated a bit from the point of this PR. If there's an unintended breaking bug in a new/future minor/patch Julia version, then it's not our responsibility to test for that and fix anything in our package. I would therefore, with all due respect for the discussion here, go ahead and close this PR if there are no objections from your side. 🙂

PR #710, opened 2022-11-21
New `randomwalk!` function

### LINKED ISSUE #644: New function `randomwalk!` to replace `walk!(agent, rand, model)`.
I think our `walk!(agent, rand, model)` is an unintuitive function, with also less power than possible. When I think of a random walk, I would think that I provide a radius to a function, and the agent takes a step with distance as much as the radius, but random direction.

Our function doesn't allow for this. So, I propose that we implement a `randomwalk!` function with specification:

```julia
randomwalk!(agent, model, r)
```
which moves the agent in distance `r`, respecting periodic boundary conditions, but random orientation, respecting the space metric. So, for continuous space this is literally moving in a random direction in the unit circle. For Discrete space this uses the spaces metric and finds all points that have distance `floor(r)`. For Chebyshev this is hypercube, and so on. Only for Manhattan and Chebyshev this function can utilize points that have exactly `floor(r)` distance. For euclidean metric this can only be approximate.

mastrof:
Closes #644.

Replaces the old `walk!(agent, random, model)` function with a new one `randomwalk!(agent, model, r)`.

The new function moves `agent` in a random direction, at distance `r` from its current location, respecting boundary conditions and space metric.

For walks in continuous space, extra keywords `polar` (both 2D and 3D) and `azimuthal` (only 3D) allow the user to specify the distribution of reorientation to generate correlated random walks.
Rotations are performed with `Rotations.jl`.

In order to provide spherically uniform reorientations in 3D, a new distribution is defined (`Arccos`); if someone knows an "official" name for this distribution (`acos(Uniform(-1,1))`) I'm happy to rename it.


src/Agents.jl
```diff
@@ -6,6 +6,9 @@ using DataStructures
 using Graphs
 using DataFrames
 using Random
+using Distributions
+using Rotations
+using StaticArrays
```
Datseris, line 11: Put these to the file they are atually used

src/spaces/utilities.jl
```diff
@@ -233,6 +233,75 @@ walk!(agent, ::typeof(rand), model::ABM{<:AbstractGridSpace{D}}; kwargs...) wher
 walk!(agent, ::typeof(rand), model::ABM{<:ContinuousSpace{D}}) where {D} =
     walk!(agent, Tuple(2.0 * rand(model.rng) - 1.0 for _ in 1:D), model)
 
+
+"""
+    rotate(w::SVector{2}, θ)
+Rotate two-dimensional vector `w` by an angle `θ`.
```
Datseris, line 239: Wht are the units the angle expected in?

src/spaces/utilities.jl
```diff
@@ -233,6 +233,75 @@ walk!(agent, ::typeof(rand), model::ABM{<:AbstractGridSpace{D}}; kwargs...) wher
...
+"""
+    rotate(w::SVector{3}, θ, ϕ)
+Rotate three-dimensional vector `w` by angles `θ` (polar) and `ϕ` (azimuthal).
+"""
+function rotate(w::SVector{3}, θ, ϕ)
+    # find a vector normal to w
+    m = findfirst(w .≠ 0)
+    n = m%3 + 1
+    u = zeros(3)
+    u[n], u[m] = w[m], -w[n]
```
Datseris, line 252: It seems rather expensive to me that you would allocate a normal `Vector` here. You should just copy `w` and use `StaticArrays.setindex` to keep the static vector and not alocate memory.

src/spaces/utilities.jl
```diff
@@ -233,6 +233,75 @@ walk!(agent, ::typeof(rand), model::ABM{<:AbstractGridSpace{D}}; kwargs...) wher
...
+    a = AngleAxis(θ, u...) * w
+    # rotate a around the original vector w by the azimuthal angle ϕ
+    AngleAxis(ϕ, w...) * a
+end # function
+
+"""
+    randomwalk!(agent, model, r)
+Move `agent` for a distance `r` in a random direction, respecting periodic 
+boundary conditions and space metric.
+"""
```
Datseris, line 263: This documentation string does not mention the extra keywords you mention in the pull request

Datseris (commented):
Thank you for the contribution!

The most important part here us that this function **must be tested extensively**. Normally  PRs without tests shouldn't be reviewed, as it doesn't matter what Ill say in the review if the code doesn't run. But I added some comments here to help you get started!

Please add tests to the test suite for every case:

continuous 2d, continuous 3d, discrete chebyshev, discrete manhattan. 

Make sure it errors for using with euclidean disrete space.

(I think this PR only implements rando walk for continuous space, in which case forget about testing the discrete case)

mastrof:
Indeed this is still WIP. I thought it was good to open the PR as a draft while I work on it, but 'm not sure if it's the standard way to do these things.

I'll implement all the comments and of course I will add tests.
I will also do the discrete case, was just waiting for the comment on how to deal with Euclidean metric.

mastrof:
Explaining some choices I made:
1. (3bf60e3) it makes more sense to index `offsets_at_radius` by `floor(Int,r)` rather than by `r`; will save some computations in case `r` is changed, since anyway the walk properties will only depend on `floor(Int,r)`
2. (655320a) if type is not enforced, one is able to specify both Continuous and Discrete distributions, and also "custom" distributions represented by vectors (e.g. distributions measured from experiments which don't have an exact analytical counterpart); in principle every object `x` for which `rand(x)` is defined is acceptable

I believe everything is implemented and preliminary checks tell me it seems to work as expected, we're only missing the test suite now. I'll try to push it within the next few days

src/spaces/utilities.jl
```diff
@@ -255,7 +255,9 @@ function randomwalk!(
         ))
     end
     offsets = offsets_at_radius(model, r)
-    walk!(agent, rand(offsets), model)
+    # if ifempty is not set to false, `walk!` will error for GridSpaceSingle
+    ifempty = typeof(model.space) <: GridSpaceSingle ? false : true
```
mastrof, line 259: Of course this must be changed otherwise `GridSpaceSingle` will not work as expected, but currently the `walk!` function does not seem to work for these spaces? It's totally unclear to me how this is supposed to work (I hardly ever use grid models)
https://github.com/JuliaDynamics/Agents.jl/blob/ac265f9cd094684c1768e4170da70ce1cc220771/src/spaces/utilities.jl#L209
it tells me that `ids_in_position` is not defined for GridSpaceSingle; a different function should be used instead, `id_in_position` ?

src/spaces/utilities.jl
```diff
@@ -264,9 +264,10 @@ function randomwalk!(
     ifempty::Bool=true
 )
     θ = rand(polar)
-    direction = Tuple(rotate(SVector{agent.vel}, θ)) # assume vel has unit norm
+    r₀ = LinearAlgebra.norm(agent.vel)
+    direction = Tuple(rotate(SVector{agent.vel}, θ)) .* (r / r₀)
```
mastrof: This is problematic if `r₀=0`, which happens if at some point `randomwalk!` is called with step size `r=0`.
I should add a check to differentiate between the two cases `r₀=0` and `r₀!=0`; or `r=0` could be set to be invalid for `ContinuousSpace` but I assume there are cases where it can be convenient to have this option.

Datseris: nah, I would dissallow this option.

mastrof: Ok, now randomwalk in continuous space throws an error if r<=0

src/spaces/utilities.jl
mastrof: Solved this by dispatching over GridSpaceSingle

src/spaces/utilities.jl
```diff
@@ -233,6 +235,158 @@ walk!(agent, ::typeof(rand), model::ABM{<:AbstractGridSpace{D}}; kwargs...) wher
...
+function randomwalk!(
+    agent::AbstractAgent,
+    model::ABM{<:AbstractGridSpace},
+    r::Real;
+    kwargs...
+)
+    if model.space.metric == :euclidean
+        throw(ArgumentError(
+            "Random walks on a GridSpace with Euclidean metric are not defined. " *
+            "You might want to use a ContinuousSpace or a different metric."
```
Datseris, line 260:
```suggestion
            "Random walks on a `GridSpace` with Euclidean metric are not defined. " *
            "You might want to use a `ContinuousSpace` or a different metric."
```
Error messages are like comments or like docstrings. They support markdown syntax. When referring to code objects, use backticks

src/spaces/utilities.jl
```diff
@@ -233,6 +235,158 @@ walk!(agent, ::typeof(rand), model::ABM{<:AbstractGridSpace{D}}; kwargs...) wher
...
+    end
+end
+
+
+"""
+    rotate(w::SVector{2}, θ::Real)
+Rotate two-dimensional vector `w` by an angle `θ`.
+The angle must be given in radians.
+"""
+rotate(w::SVector{2}, θ::Real) = Angle2d(θ) * w
```
Datseris, line 297: In my experience it makes for more readable code if the functions are defined after they are called. So I'd put these `rotate` functions` _after_ the first `randomwalk!` function that actually calls then.

test/continuous_space_tests.jl
```diff
@@ -334,6 +335,118 @@ using StableRNGs
...
+            @test !(norm(model.agents[1].pos .- x₀) ≃ norm(v₀))
+
+            # verify that reorientations obey the specified angles
+            space = ContinuousSpace((10,10), periodic=true)
+            model = ABM(RWAgent2D, space)
+            x₀ = (5.0, 5.0)
+            v₀ = (1.0, 0.0)
+            add_agent!(x₀, model, v₀)
+            r = 1.0
+            polar = [π/2] # degenerate distribution, only π/2 reorientations
```
Datseris, line 364: Let me say that this is an absolutely fantastic test set.

test/continuous_space_tests.jl
```diff
@@ -334,6 +335,118 @@ using StableRNGs
...
+            r = 1.1
+            polar = [0.0] # no reorientation, move straight
+            randomwalk!(model1.agents[1], model1, r; polar)
+            randomwalk!(model2.agents[1], model2, r; polar)
+            @test model1.agents[1].pos[1] ≈ 0.1
+            @test model2.agents[1].pos[1] ≈ 2.0
+        end
+
+        @testset "3D" begin
+            ≃(x,y) = isapprox(x,y,atol=1e-12) # \simeq
```
Datseris, line 403: Why not pull this outside the 2D / 3D sets

test/grid_space_tests.jl
```diff
@@ -262,6 +262,135 @@ end
...
+        # chebyshev distance after the random step should be 1
+        @test maximum(abs.(x₁ .- x₀)) == 1
+        # the list of available offsets at distance 1
+        # should be stored in space.offsets_at_radius[1]
+        @test haskey(space.offsets_at_radius, 1)
+        offsets_1 = Set((
+            (1,-1)  ,  (1,0)  ,  (1,1),
+            (0,-1)       ,       (0,1),
+            (-1,-1) ,  (-1,0) ,  (-1,1)
+        ))
```
Datseris, line 290: You should **not** test this. We only use the public API in the tests. Fields stored inside the spaces are not public api and hence not accessible / shouldn't be accessed. The tests here should be like in the continuous space by just testing the logical conclusions. (which you already do, so simply remove tests that don't use public api)

Datseris (commented):
Thank you @mastrof  this truly an incredible quality contribution. Besides the review comments, please also:
- Increment minor version in Project.toml
- Add a CHANGELOG.md entry for the new minor version
- Add a deprecation warning in the `walk!(agent, rand, model)` function, so that it says something like "don't use this, doesn't make sense".

Datseris:
oh also test fails

mastrof:
Oh, my bad; I only checked the tests for grid spaces and continuous spaces. Didn't expect other things to break since I'm only adding functionalities and not modifying preexisting ones. Will check

src/spaces/utilities.jl
```diff
@@ -226,12 +228,169 @@ end
...
+    kwargs...
+)
+    if model.space.metric == :euclidean
+        throw(ArgumentError(
+            "Random walks on a `GridSpace` with Euclidean metric are not defined. " *
+            "You might want to use a `ContinuousSpace` or a different metric."
+        ))
+    end
+    offsets = offsets_at_radius(model, r)
+    walk!(agent, rand(offsets), model; kwargs...)
```
Datseris, line 270: Ops @mastrof I just now saw this!!! Very important to fix: In all calls to `rand` you must use the model rng!!! Please replace all `rand(x)` with `rand(model.rng, x)` everywhere in the code you've added!!!

mastrof: should I force push or pull from main?

Datseris: first pull, as I added a merge commit for the version in project.toml. then edit, and then pull. or merge in your local, then push.

Datseris:
Okay I have to merge this in now, so I'll fix the remaining problems.

PR #711, opened 2022-11-24
id::Int field redundant: removed

gvdr:
I believe the id field in the definition of a NoSpaceAgent is not necessary (and indeed it gives an error when trying to define it explicitly). Removed that, the example runs smoothly.


Datseris:
correct, tank you! NoSpaceAgent already has the id field.

PR #712, opened 2022-11-24
Add try-catch to finding first agent instance in multi_agent_types!

tlow22:
When initializing the `adata` dataframe for multi-agents, code tries to look for the first object instance of the different agent types., so that it can be used to extract the desired property types for dataframe header. 

In the specific case where one agent type doesn't yet exist at model initialization (to be added later on while stepping the simulation), this will cause an error. 

The behavior now is to default the missing agent type's properties to `Missing`. If desired, I can also leverage `hasfield` and `fieldtype` when the desired property is a `Symbol`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - `add_node!` and `rem_node!` have been renamed to `add_vertex!` and `rem_vertex!` extending Graphs.jl homonymous methods to help standardise names across ecosystems. Therefore `add_node!` and `rem_node!` have been deprecated.  
 - The signature of `add_edge!` has been generalised with `args...` and `kwargs...` to be compatible with all the implementations the underlying graph supports. 
 - New function `rem_edge!` that removes an edge from the graph.
+- `multi_agents_type!` has been updated to handle edge case where agents of one (or more) type are absent at the beginning of the simulation. 
 
 # v5.5
 - The `@agent` macro has been re-written and is now more general and more safe.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
+# main
+- `multi_agents_type!` has been updated to handle edge case where agents of one (or more) type are absent at the beginning of the simulation. 
+
 # v5.6
 - `add_node!` and `rem_node!` have been renamed to `add_vertex!` and `rem_vertex!` extending Graphs.jl homonymous methods to help standardise names across ecosystems. Therefore `add_node!` and `rem_node!` have been deprecated.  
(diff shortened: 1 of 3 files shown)


Datseris:
How does `missing` come into play, given that the code returns `nothing` if the `try` fails?

tlow22:
@Datseris, ah I'm leveraging some of the native code a few more lines below the change, namely lines 250 - 259: 

``` 
if k isa Symbol
    current_type =
        hasproperty(a, k) ? typeof(get_data(a, k, identity)) : Missing
else
    current_type = try
        typeof(get_data(a, k, identity))
    catch
        Missing
    end
end
```

Even if an agent instance is not found, and `nothing` is returned, the code continues to get the object properties either via the `Symbol` or `function` getter method. I noticed that in both cases, a `try-catch` with `Missing` is being returned when the property getter was used. If the object is `nothing`, the property getters will fail and simply default to `Missing`.

Datseris:
Have you measured what impact on performance this may have? A typical `run!` call of one of the working examples of the repo. How much slower does it become on this branch?

Datseris:
Also, what happens when you actually use this in practice? The data frame is initialized with a Missing column, and once you try to add data once agents exist, what happens? 

This Pull Request does not have tests. Please add a test that test this functionalityt by making a minimal ABM that doesn't have agents and then goes through `run!` so that we have proof that the code works as exepected. The test must be added in the `test` subfolder. Have a look at the files already there and you'll get the idea of how it works.

tlow22:
@Datseris Took your comment into consideration and added functionality to determine the `fieldtype` when an object is not available. Added unit test to verify 

1) fieldtype determination 
2) successful addition to the dataframe when agents are introduced later on.  

Caveat - same trick can't be applied to functions or aggregators, as there are too many ways a function could fail. 

PR shouldn't affect performance, as it only affects the dataframe initialization right at the very beginning and not subsequent runs.

Datseris:
Thanks a lot @tlow22 . Can you please add a note in the changelog under the nextx minor version about this fix?

---

> PR shouldn't affect performance, as it only affects the dataframe initialization right at the very beginning and not subsequent runs

Sure, sounds reasonable, but can you please check explicitly?

tlow22:
@Datseris sure thing on changelog update! Here's a timing benchmark comparison for the PR: 

Main: 
image

PR: 
image

Datseris (approved):
Thank you very much for the contribution @tlow22  :)

tlow22:
@Datseris glad to help in a little way! Recently discovered DrWatson as well, very cool work!

PR #717, opened 2022-12-19
Add docstring support to `@agent`

### LINKED ISSUE #715: Add docstring support to @agent
Attaching a docstring to an `@agent` definition fails with an error: `'@agent' not documentable. See 'Base.@__doc__' docs for details.' This is unexpected and problematic, as documenting agent types is important.

The [@agent documentation](https://juliadynamics.github.io/Agents.jl/stable/tutorial/#.-The-agent-type(s)-1) states: "You can use the @doc macro from Julia to document the generated struct if you wish so." However, this statement (1) is easy to overlook, and (2) does not provide an example. (I did not get `@doc` to work in conjunction with `@agent`, and could not find a working example.)

My understanding is that docstring support can be added to a macro using [`@__doc__`](https://docs.julialang.org/en/v1/manual/documentation/#Core.@__doc__). Adding this to [`@agents`](https://github.com/JuliaDynamics/Agents.jl/blob/5b3f5125657515c8c9fead961f4e979369e4afe3/src/core/agents.jl#L170) would allow modellers to better document their code.

veddox:
Fixes #715

### DIFF
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -235,7 +235,7 @@ macro agent(new_name, base_type, super_type, extra_fields)
             end
             # @show expr # uncomment this to see that the final expression looks as desired
             # It is important to evaluate the macro in the module that it was called at
-            Core.eval($(__module__), Core.@__doc__(expr))
+            Core.eval($(__module__), expr)
         end
         # allow attaching docstrings to the new struct, issue #715
         Core.@__doc__($(esc(Docs.namify(new_name))))
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -192,13 +192,18 @@ macro agent(new_name, base_type, extra_fields)
                     $(base_fields...)
                     $(additional_fields...)
                 end
+                #@doc @doc(name) $name # $docstring $name
             end
-            # @show expr # uncomment this to see that the final expression looks as desired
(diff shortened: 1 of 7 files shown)


src/core/agents.jl
```diff
@@ -232,8 +235,11 @@ macro agent(new_name, base_type, super_type, extra_fields)
             end
             # @show expr # uncomment this to see that the final expression looks as desired
             # It is important to evaluate the macro in the module that it was called at
-            Core.eval($(__module__), expr)
+            Core.eval($(__module__), Core.@__doc__(expr))
```
MichaelHatherly, line 238: This extra `@__doc__` probably wasn't meant to be here? Probably won't end up doing the right thing.

veddox: Whoops, thanks for catching that. That was left over from earlier. I think it's time to call it a night, I'm getting tired :sweat_smile:

Datseris:
Can we add a test or two here to ensure that a docstring is captured when a new agent is defined? Honstly just check that the `ContinuousAgent` has a docstring, that's probably enough.

Datseris:
Fantastic! @veddox let's also increment the _patch_ version in the Project.toml file. This can be counted as a bugfix as far as I can tell!!! Then I merge and tag new release.

Datseris (approved):

PR #720, opened 2022-12-21
Allow using `NoSpaceAgent` in spatial models

### LINKED ISSUE #719: Use `NoSpaceAgent` alongside space agents
I have a model with multiple agent types, some of which require (grid-)space, and others for which a position argument doesn't really make sense. I would like to be able to use `NoSpaceAgent` as the base type for the latter, but this causes `agent_validator()` and `do_checks()` to throw an [ArgumentError](https://github.com/JuliaDynamics/Agents.jl/blob/6a8bbe54a63cff3bac7eb761cc792cf60c320562/src/core/model.jl#L327): "Second field of Agent struct must be `pos` when using a space."

**Describe the solution you'd like**
I'm wondering whether `agent_validator()` is still necessary, since creating one's own agents structs has been deprecated? It also gives me an annoying warning about a non-concrete Agent type, as I use a Union type for my multi-agent model.

**Describe alternatives you've considered**
If it's important to keep `agent_validator()`, it would be nice to include a check for whether the checked agent type extends `NoSpaceAgent`, and if so, exclude it from the spatial checks. However, I'm not sure how to do this, as agent types created with `@agent` only have `AbstractAgent` as a supertype, even if created with `NoSpaceAgent` as a base type.

veddox:
Fixes #719

### DIFF
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -235,7 +235,7 @@ macro agent(new_name, base_type, super_type, extra_fields)
             end
             # @show expr # uncomment this to see that the final expression looks as desired
             # It is important to evaluate the macro in the module that it was called at
-            Core.eval($(__module__), Core.@__doc__(expr))
+            Core.eval($(__module__), expr)
         end
         # allow attaching docstrings to the new struct, issue #715
         Core.@__doc__($(esc(Docs.namify(new_name))))
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -192,13 +192,18 @@ macro agent(new_name, base_type, extra_fields)
                     $(base_fields...)
                     $(additional_fields...)
                 end
+                #@doc @doc(name) $name # $docstring $name
             end
-            # @show expr # uncomment this to see that the final expression looks as desired
             # It is important to evaluate the macro in the module that it was called at
             Core.eval($(__module__), expr)
         end
+        # allow attaching docstrings to the new struct, issue #715
+        Core.@__doc__($(esc(Docs.namify(new_name))))
+        nothing
     end
 end
+
+
 # TODO: I do not know how to merge these two macros to remove code duplication.
 # There should be away that only the 4-argument version is used
 # and the 3-argument version just passes `AbstractAgent` to the 4-argument.
@@ -232,8 +237,11 @@ macro agent(new_name, base_type, super_type, extra_fields)
             end
             # @show expr # uncomment this to see that the final expression looks as desired
             # It is important to evaluate the macro in the module that it was called at
-            Core.eval($(__module__), expr)
+            Core.eval($(__module__), Core.@__doc__(expr))
         end
+        # allow attaching docstrings to the new struct, issue #715
+        Core.@__doc__($(esc(Docs.namify(new_name))))
+        nothing
     end
 end
 
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -190,8 +190,8 @@ macro agent(new_name, base_type, extra_fields)
                     $(base_fields...)
                     $(additional_fields...)
                 end
-                #@doc @doc(name) $name # $docstring $name
             end
+            # @show expr # uncomment this to see that the final expression looks as desired
             # It is important to evaluate the macro in the module that it was called at
             Core.eval($(__module__), expr)
         end
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -53,8 +53,6 @@ that the resulting new agent type will have.
 The `id` is an unchangable field (and in Julia versions ≥ v1.8 this is enforced).
 Use functions like [`move_agent!`](@ref) etc., to change the position.
 
-You can use the `@doc` macro from Julia to document the generated struct if you wish so.
-
 ## Examples
 ### Example without optional hierarchy
 Using
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -323,24 +323,26 @@ function do_checks(::Type{A}, space::S, warn::Bool) where {A<:AbstractAgent,S<:S
     fieldtype(A, :id) <: Integer ||
     throw(ArgumentError("`id` field in Agent struct must be of type `Int`."))
     if space !== nothing
-        (any(isequal(:pos), fieldnames(A)) && fieldnames(A)[2] == :pos) ||
-        throw(ArgumentError("Second field of Agent struct must be `pos` when using a space."))
-        # Check `pos` field in A has the correct type
-        pos_type = fieldtype(A, :pos)
         space_type = typeof(space)
-        if space_type <: GraphSpace && !(pos_type <: Integer)
-            throw(ArgumentError("`pos` field in Agent struct must be of type `Int` when using GraphSpace."))
-        elseif space_type <: GridSpace && !(pos_type <: NTuple{D,Integer} where {D})
-            throw(ArgumentError("`pos` field in Agent struct must be of type `NTuple{Int}` when using GridSpace."))
-        elseif space_type <: ContinuousSpace || space_type <: ContinuousSpace
-            if !(pos_type <: NTuple{D,<:AbstractFloat} where {D})
-                throw(ArgumentError("`pos` field in Agent struct must be of type `NTuple{<:AbstractFloat}` when using ContinuousSpace."))
+        if warn && !(any(isequal(:pos), fieldnames(A)) && fieldnames(A)[2] == :pos)
+            @warn "Second field of Agent struct should be `pos` when using a space."
+        elseif any(isequal(:pos), fieldnames(A))
+            # Check `pos` field in A has the correct type
+            pos_type = fieldtype(A, :pos)
+            if space_type <: GraphSpace && !(pos_type <: Integer)
+                throw(ArgumentError("`pos` field in Agent struct must be of type `Int` when using GraphSpace."))
+            elseif space_type <: GridSpace && !(pos_type <: NTuple{D,Integer} where {D})
+                throw(ArgumentError("`pos` field in Agent struct must be of type `NTuple{Int}` when using GridSpace."))
+            elseif space_type <: ContinuousSpace || space_type <: ContinuousSpace #FIXME this seems wrong?
+                if !(pos_type <: NTuple{D,<:AbstractFloat} where {D})
+                    throw(ArgumentError("`pos` field in Agent struct must be of type `NTuple{<:AbstractFloat}` when using ContinuousSpace."))
+                end
             end
-            if warn &&
-               any(isequal(:vel), fieldnames(A)) &&
-               !(fieldtype(A, :vel) <: NTuple{D,<:AbstractFloat} where {D})
-                @warn "`vel` field in Agent struct should be of type `NTuple{<:AbstractFloat}` when using ContinuousSpace."
-            end
+        end
+        if warn && space_type <: ContinuousSpace &&
+            any(isequal(:vel), fieldnames(A)) &&
+            !(fieldtype(A, :vel) <: NTuple{D,<:AbstractFloat} where {D})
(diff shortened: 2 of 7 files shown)


veddox:
Sorry, this still includes my commits from #717. I'm not sure how to get rid of them, and only show commits for this issue?

src/core/space_interaction_API.jl
```diff
@@ -285,8 +285,13 @@ function add_agent!(
     kwargs...,
 )
     id = nextid(model)
-    newagent = A(id, pos, properties...; kwargs...)
-    add_agent_pos!(newagent, model)
+    if any(isequal(:pos), fieldnames(A))
+        newagent = A(id, pos, properties...; kwargs...)
+        add_agent_pos!(newagent, model)
```
Datseris, line 290: Wait, isn't this a hit to performance? Please check if there is a more efficient way to test if a struct as a field `pos`.

Datseris:
> Sorry, this still includes my commits from https://github.com/JuliaDynamics/Agents.jl/pull/717. I'm not sure how to get rid of them, and only show commits for this issue?

You need to update your own `main` branch and sync it with the Agents one, and start a branch from the update. Alternatively, try merging the Agents `main` branch into your own local branch.

veddox:
Thanks for your comments. I'm on my Christmas break now, but will look into this again come January.

Merry Christmas to you :slightly_smiling_face: :christmas_tree:

Datseris:
to you too!!!

src/core/space_interaction_API.jl
jacobusmmsmit:
```suggestion
    if :pos in fieldnames(A)
        newagent = A(id, pos, properties...; kwargs...)
        add_agent_pos!(newagent, model)
```

I believe this is the idiomatic way of doing this.

Datseris:
bump

veddox:
Hi @Datseris, thanks for the reminder and sorry for the delay. Apart from having to take care of other tasks, I was a bit unsure about how to proceed with this issue.

On the one hand, I still think that it would be useful to allow non-spatial agents in spatial models, for example for institutions that act as agents but have no localisation in the landscape.

On the other hand, I realised that this is trickier than I thought, due to the underlying implementation. What I hadn't understood is that adding and removing agents is dispatched not based on the agent type, but on the model space type. 

This makes sense because the `@agent` macro doesn't (and cannot) create subtypes of `GridAgent` et al., but always of `AbstractAgent`. Therefore, spatial- and non-spatial-agents cannot be differentiated based on type.

However, dispatching based on model space type implicitly assumes that all agents in the model match the model space type. Circumventing this would require checking an agent's `fieldnames` for a `:pos` field every time an agent is added or removed, to see whether the model space needs to be changed along with the agent list. This is ugly and potentially a performance problem.

I'm not sure what you think about this? For the time being, I had more urgent things that needed dealing with, and so didn't pursue this further. But perhaps somebody has a good idea for a cleaner solution?

Datseris:
> On the one hand, I still think that it would be useful to allow non-spatial agents in spatial models, for example for institutions that act as agents but have no localisation in the landscape.

If that's your reason for allowing this, then we shouldn't allow this. Simply _do not use_ the position of institutions that act as agents but their position doesn't matter. Give them some random position. Is there any downside in that?

veddox:
That's what I'm doing at the moment. It means you have to do a bit more work when looking for nearby agents, because you then have to filter by agent type. Although that is something I have to do anyways with my multi-agent model, so actually for my particular use-case the issue of non-spatial agents is not as pressing as I thought at first. Though I still think that the current behaviour is unexpected from a UX perspective and may cause problems in other situations.

(On that note, it may be a sensible addition to the API to allow `nearby_agents` to filter by agent type, but that's another topic.)

Datseris:
I admit, I still do not full follow your process. Why are these "entities without a position" agents in the first place? Why aren't they a model property?

Datseris:
bump here. Is this PR actually necessary? Is there any reason to ever allow models with e.g., continuous space, to accept agents without a position? What's the reason for doing this? I argue above that whatever an agent without position is, it is more likely suited to be a model property instead. Are there any counter arguments to that?

If there aren't any, we can close both this PR and #719 . cc @AayushSabharwal @Tortar I would like some thoughts on the matter.

Tortar:
I think a counter-argument is related to scheduling the agents, suppose one wants some agents not on the space, but that they are activated randomly together with the spatial agents, this is enough in my opinion, and that means that currently a nearby search in this situation requires filtering of actually non-spatial agents which has to be implemented as spatial for now when using nearby function as @veddox said which is less performant and more clumsy

AayushSabharwal:
I actually think to the contrary. If some agents don't have any spatial meaning, can't they just be moved to `(0, 0)` or any other arbitrary fixed position? If the only issue is `nearby_agents`, we can improve the API for filtering through these agents. Instead of speculating over performance impacts, are there benchmarks we can do?

It's possible that if the non-spatial agents are localized to a corner of the space, they shouldn't come up in nearby searches very often. If they do, the cost could be considered as amortized over the cost of other such searches where they don't come up. Additionally, suppose there is a measurable performance penalty in doing these searches, does it outweigh the cost of dynamic dispatch during stepping since the agent type is not concrete any more? Inherently any such model would face dynamic dispatch unless the stepping is done manually through `model_step!`. If it is done manually, why not just have the non-spatial agents separate from the model altogether? Maybe we can generalize our scheduling API to not require models as input, and just sequences of IDs to aid in this.

Are we also sure that allowing this kind of mixture won't cause any bugs, complicated behavior, or edge cases later on? And if we do decide to go along this route, why not go the full distance and allow a mixture of spaces in a model? That feels like a generalization of this problem, and likely useful in more scenarios.

These are just my views, and if everyone else thinks differently I'm more than happy to continue in the agreed-upon direction. If there's something I'm missing, please point me in the right direction.

Datseris:
> Maybe we can generalize our scheduling API to not require models as input, and just sequences of IDs to aid in this.

Not really necessary, these "alternative beings" (not agents) can be a model property. There is no reason to have something within Agents.jl that is _not_ inside the ABM.

Tortar:
I need to say that now that I'm considering your points it seems also to me
better the "normal" way of putting agents in one spot, but the point you
made about multispace models is nonetheless valid anyway, I also considered
some time ago this extension as beneficial. I think that only a big and
generic change like this of multispace models is justified. So I think that
we all three agree about the matter in the end

Il Gio 8 Giu 2023, 11:35 George Datseris ***@***.***> ha
scritto:

veddox:
Hi everyone, thanks for weighing in. From the perspective of somebody working with multi-agent models, being able to combine spatial and non-spatial agents can make a lot of sense. Moving "non-spatial" agents to a dedicated position is a possible work-around, but it is quite a kludge. (Depending on the scenario, using model properties could be a clean solution.)

However, like I wrote above, I understand that implementing multi-space models is not trivial and would probably incur complexity and performance costs. I also don't have the capacity at the moment to really dig into this myself. So I would be fine with closing the issue and the PR.

jacobusmmsmit:
Is there a place in the docs which talks about multispatial models like
there is for multi agent-type models?

More people in the future might be interested in doing the same thing and
will run into the same issue. The answer should be easily visible in the
docs, probably in the Space section itself.

Datseris:
> Is there a place in the docs which talks about multispatial models like
there is for multi agent-type models?

Well, no, because multi-space models in Agent.jl are not possible while multi-agent models are.

jacobusmmsmit:
Right, but I'm just saying that we should write a little note about them in the docs in case someone wants to do multi-spaced stuff and then can't.

We should justify why not and what style workarounds are.

Datseris:
> Hi everyone, thanks for weighing in. From the perspective of somebody working with multi-agent models, being able to combine spatial and non-spatial agents can make a lot of sense. Moving "non-spatial" agents to a dedicated position is a possible work-around, but it is quite a kludge. (Depending on the scenario, using model properties could be a clean solution.)


@veddox 

You still haven't answer though the following:

> Why are these "entities without a position" agents in the first place? Why aren't they a model property? Why aren't they a vector of `struct` that is generated by `NoSpaceAgent` that is a model property and you can access independently of the normal agents that reside in the space? 

What is the downside of doing that?

Tortar:
Looking again at this proposal and I have to say there is probably a trivial solution if I understand it correctly: use two agent types and add the second agent type (the one without a position) with `add_agent_to_model!` instead of `add_agent!`, this way it will be not added to the space. However, I think `add_agent_to_model!`  is still not exported but we probably could

Datseris:
No, `add_agent_to_space` is an internal function and it should never be used by the user directly.

Tortar:
we could also not allow using add_agent_to_space directly, but add_agent_to_model, and its documented usage should be what is discussed in this issue i.e. to use NoSpaceAgents along with SpaceAgents. We can even call it in another way, but its effect should be this one. 

But there is a simpler way: just extend add_agent! for a NoSpaceAgent when the type of the space is is defined. And that should be enough.

Datseris:
I think we should just close this and just not allow this behavior. There has not been any convincing argument for why allowing this complex behavior, given that you can achieve the same effect with different, and much simpler, alternatives that are already possible such as:

- making these agents space properties instead
- using a dummy agent position very far away all agents (hence, artificially enlarging space)

Tortar:
but it's what is already attempted here I guess, I will then superseed this PR with a new one, which will finally solve the issue since I found the requirement here useful in some real situations.

Tortar:
sorry, didn't see your last message @Datseris before writing what I wrote, but I will add some more explaination to why this could be useful, and I find your solutions not really satisfactory.

Your first solution doesn't allow to use the Agents.jl API for those agents if represented as a property. Your second solution can be error prone, and it actually creates a `pos` field (and possibly some other internal fields) which only enlarge the memory consuption of the simulation and I have to say I find it hacky as @veddox said before. In any case the user could use this type of solution anyway, even if we allow another (in my opinion cleaner) way. All should be seen in relation to #841 where each type of agent will be contained in a different data structure where using the API will be even simpler since the filtering by type will be already implemented. 

In the end even if it is probably possible to circumvent the issue, if done rightly it should be a 10 lines of code change which I think is acceptable.

Tortar:
So to summarize the convincing arguments to me are:

1) making these agents space properties instead -> can't use the agents.jl API for things which can be conceived as agents
2) using a dummy agent position very far away all agents (hence, artificially enlarging space) -> more memory consumption for both the additional fields and since e.g. more you have a big radius more you have to enlarge the space + seems error prone + you can't use a dispatch on the agent types but if anyway these are different kind of agents you will need to add an if else somewhere for different behaviours (notice that the real problem with multi-agent types for performace reasons is not the dispatch of the step! but the abstract containers-general type instability inside the step! itself)

Datseris:
Well, based on the discussion it appears to me that _simply not using the position of these special agents is the way to go._ These "alternate agent beings" would anyways **must have** a different type. Hence, you anyways **must filter** by agent type when searching for nearby agents or doing any other operation. The filtering operation happens anyways, hence the argument "oh they are found as nearby neighbors" is not strong enough, as you would anyways have to filter. 

I conclude with what @AayushSabharwal has originally argued: the simplest solution is by far to put all these special agents at a special position, (1,1) or whatever. There is pretty much zero performance impact on this, when taking into consideration that a filter operation will necessarily have to happen at some point or another.

Tortar:
Actually I think that the filtering will never take place since if these agents are not added to the space, the filtering will not be required.

Datseris:
Should we close this now, and also #719, with the final claim: "feature can be achieved with other simpler means that are already readily available from existing API"?

jacobusmmsmit:
If we're not adding this, then we should definitely document well what these simpler means with existing API are. It's likely that there are others who will want to do the same. Perhaps we can add an "advanced" section which contains a few articles explaining how to do things like this, and then link to the relevant article in the "space" section through a note about putting non-space agents into a space model.

Datseris:
I'm closing this as unplanned, but happy to review and merge a documentation contribution.

Issue #668, opened 2022-08-02
Different versions of `AgentBasedModel`


AayushSabharwal:
@AayushSabharwal a bit off topic, but here the idea we've breafly chatted about once becomes useful: The "different versions of `AgentBasedModel` and a version for "unkillable" agents, which uses vector and hence much better performance. Would you please be so kind to open a new issue discussing this, before we forget it? There are benefits to implement this, it's clear, so let's have an issue for it in case someone has the time to do it.

_Originally posted by @Datseris in https://github.com/JuliaDynamics/Agents.jl/issues/659#issuecomment-1198490335_

We should document a required interface for `ABM`, similar to our spaces API. Adding an `AbstractABM` supertype would also be useful for dispatch.

Datseris:
This will also allow us to have a different type, with potentially different internals, for multi agent models. I've though of a couple of internal implementations with tuples of dictionaries, and a main dictionary mapping id to inner dictionary. But first let's write down the API we'd need for `AbstractABM`.

Datseris:
Okay, here is my list, and let's edit this list as we add more. `model` is an instance of a subtype of `AbstractABM` in the following.
- `model[id]` gives agent given id
- `model.property` gives model level property
- `model[id] = x` errors
- `model.property = x` calls `setindex!` on the property
- `model.rng`, which should be replaced to `rng(model)` gives the rng. The reason I am saying this should be replaced is because it allows separating the "get properties" interface from the "get 'internal' properties" like `rng` or `schedfuler`.
- similarly `model.scheduler` which should become `scheduler(model)` gives a default scheduler
- `agenttype(model)` returns the type(s) of agents
- `nagents(model)` returns the total number of agents
- `nextid(model)` which returns a valid unoccupied ID for making a new agent
- `seed!(modl)` is REMOVED in favor of `seed!(rng(model))`.
- `allagents(model) 

We also need the following new internal functions, as internally we currently just do `model.agents[agent.id] = agent`.
- `add_agent_to_model(agent, model)` which replaces `model.agents[agent.id] = agent`
- `remove_agent_from_model(agent, model)`
- `update_model_maxid(id, model)` which replaces `model.maxid[] < agent.id && (model.maxid[] = agent.id)`

AayushSabharwal:
Why does the `id` need to be passed in to `update_model_maxid`? Can't it be `increment_model_maxid(model)`?

Datseris:
[...]

PR #721, opened 2022-12-23
Allow agents to be stored in vectors

jacobusmmsmit:
This PR allows users to specify that they want their agents of type `A` stored in a `Vector{A}` instead of the current default of a `Dict{Int, A}`. The main reason for doing so is that iteration over vectors is much faster than over dictionaries. In order to allow for this functionality, the following changes were made:
* `ABM`s are now additionally parameterised by their container type `C`, current valid containers are stored in the parametric union `ContainerType{A}` which could be expanded in the future.
* Functions that behave differently for different container types have been split into two implementations (`ABM`, `set_index!`, `kill_agent!`, and `genocide!`), others have simply been made type agnostic (`random_agent`, `allids`, and `step!`).

Finally, it adds some rudimentary tests demonstrating the functionality of creating and being unable to destroy agents.

One change I would suggest adding whether or not this PR is accepted is to change all references to `keys` or `haskey` to `eachindex`, as it has the benefit of being generic in the container and was faster in my testing while returning identical results. This affects the functions above that were made type agnostic.

You can test this PR by running the [Flocking model](https://juliadynamics.github.io/Agents.jl/stable/examples/flock/) example simply by adding `ABM(...;..., container=Vector{Bird})`. Doing so provides a nontrivial speedup taking the minimum from 68ms to 32ms and the median from 74ms to 60ms.

<details>
  <summary>Benchmarks</summary>

```julia
julia> @benchmark run!(model, agent_step!, 100)  # vector-container model
  BenchmarkTools.Trial: 75 samples with 1 evaluation.
 Range (min … max):  32.044 ms … 589.730 ms  ┊ GC (min … max):  0.00% … 86.48%
 Time  (median):     60.657 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   67.015 ms ±  64.200 ms  ┊ GC (mean ± σ):  12.55% ± 12.22%

  ▁  ▁      █▇▆                                                 
  █▅██▅▅▁██▅███▁▅▁▅▁▁▁▁▁▁▁▁▁▁▁▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅ ▁
  32 ms         Histogram: log(frequency) by time       194 ms <

 Memory estimate: 13.76 MiB, allocs estimate: 150241.
```

```julia

julia> @benchmark run!(model, agent_step!, 100) # dict-container model
BenchmarkTools.Trial: 60 samples with 1 evaluation.
 Range (min … max):  68.412 ms … 336.749 ms  ┊ GC (min … max): 0.00% … 77.39%
 Time  (median):     74.112 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   83.660 ms ±  42.664 ms  ┊ GC (mean ± σ):  5.19% ±  9.99%

  █▃▄                                                           
  ███▅▁▃▁▁▁▁▁▃▁▁▃▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃ ▁
  68.4 ms         Histogram: frequency by time          265 ms <

 Memory estimate: 13.76 MiB, allocs estimate: 150241.
```
  
</details>

One apology: I have absolutely no idea how to keep formatting consistent with the code that's already there! Apologies for the auto-format that my editor does on save.


test/api_tests.jl
```diff
@@ -196,16 +212,16 @@ end
     end
 
     space = GridSpace((10, 10))
-    model = ABM(Union{Daisy,Land}, space; warn = false)
+    model = ABM(Union{Daisy,Land}, space; warn=false)
```
Datseris, line 215: Please make this PR easy to review by commiting changes like these, i.e., syntactic style. We will apply an automated code linter in future PRs but until then no reason to spend dev effort reviewing linting changes.

Datseris:
@jacobusmmsmit thanks for this. It is great that the code already works without any change and already gives such a clear improvment on the Flocking example!!! What I am thinking is whether the current implementation really is ok or will come back and bite us later.

In https://github.com/JuliaDynamics/Agents.jl/issues/668 we discuss how we could be implementing different ABM types by defining an interface. At the moment your solution is simpler and requires much code, but it is less extendable and flexible for future changes where e.g., multi-agent models may use different storing structure that isn't as simple as a vector-like access datastructure. It is worth considering whether you would like to implement the solution of #668 . If you think it is a big task, no worries, we go with your current solution on the storage type being a type parameter of `ABM` that can be set by users.  I'll let you decide on this.

> One change I would suggest adding whether or not this PR is accepted is to change all references to keys or haskey to eachindex

Yes please go ahead and change `keys` to `eachindex`. I am not sure what do you mean with `haskey`? There is no function `hasindex`.

Datseris:
Oh super important!!! Please never write `eachindex(model.agents))`!!! Use instead `allids(model)`. The function `allids(model)` must become `eachindex(model.agents)`

Datseris:
By the way the name "unkillable" doesn't sound too correct. I propose `FixedMass`. 

We should also think whether we should allow users to add more agents. Would we get any performance by dissalowing the amount of agents to change? I.e., requiring in advanced a fixed number of agents?

src/core/model.jl
```diff
@@ -107,27 +107,34 @@ export random_agent, nagents, allagents, allids, nextid, seed!
...
     m.agents[id] = a
     m.maxid[] < id && (m.maxid[] += 1)
     return a
 end
 
+function Base.setindex!(m::ABM{S,A,Vector{A}}, a::AbstractAgent, id::Int) where {S,A}
+    a.id ≠ id &&
+        throw(ArgumentError("You are adding an agent to an ID not equal with the agent's ID!"))
+    id == nagents(m) + 1 || error("Cannot add agent of ID $(id) to a vector ABM of $(nagents(m)) agents. Expected ID == $(nagents(m)+1).")
+    m.maxid[] < id && (m.maxid[] += 1)
```
AayushSabharwal, line 133: Wouldn't this prevent calling `setindex!` on existing agent IDs? E.g. If there are 4 agents, with ids from 1 to 4, calling `abm[3] = MyAgent(...)` would error

jacobusmmsmit: Yeah I clearly didn't think this one through properly.

test/api_tests.jl
AayushSabharwal: Additionally, I think we're fairly consistent with including spaces around `=` in keyword arguments in the rest of the repository, so this change is making parts of the code inconsistent with everything else.

jacobusmmsmit: Do you have a JuliaFormatter.toml with the repo's preferred style?

src/core/model.jl
Datseris: `abm[3] = MyAgent(...)` must always error. It is not supported syntax!!!! This is not the API way to add agents!!! Only the `add_agent!` function is supported!!!!

jacobusmmsmit:
@Datseris
RE: `const UnkillableABM{A, S, ...} = ABM{A, S, ..., Vector{A}}`
In my current implementation, the containertype is the third type parameter of `ABM`. I think it makes more sense to put it there given that it's closely related to the agent type, but if it breaks code then I can put it last. Thoughts?

Datseris:
Sure third is fine

jacobusmmsmit:
Re: `haskey`, `step!` used to have `haskey(model.agents, index) || continue` which I changed to `index in eachindex(model.agents) || continue` and now to `index in allids(model) || continue` which are all syntactically equivalent for dictionaries and also work on vectors :)

src/core/model.jl
jacobusmmsmit: How does one edit agents in an ABM if `abm[id] = MyAgent(id, ...)` is disallowed?

Datseris: Only `add_agent!` and `kill_agent!` can be used to add or remove agents from the model. If you want to "edit" an agent simply edit its fields in place. `a = model[id]; a.field = value`. You can't just assign a new agent to an ID previouslyt occupied by another agent.

Datseris:
@jacobusmmsmit as I've commented before, please keep all linting changes **outside** this PR for a review to take place.

jacobusmmsmit:
Yes I'm so sorry I just had a massive "wait" moment. I'll undo the linting as far as I can...

jacobusmmsmit:
So... Happy holidays everyone. I finally fixed the linting issues. I apologise not only that it was an issue in the first place, but also that it took me 5 commits to get all of the whitespace issues.

Datseris:
> We should also think whether we should allow users to add more agents. Would we get any performance by dissalowing the amount of agents to change? I.e., requiring in advanced a fixed number of agents? If so, we could make a `SizedVector` the container instead of just `Vector`.

This is still not clear to me. In the overwhelming majority of cases, in ABMs where agents aren't removed, new ones aren't added either after the construction of the model. So let's think if it is useful for performance to actually enforce this, that the model should have fixed amount of agents all the time.

Datseris:
actually, it doesn't matter in the sense that it doesn't change anything for us: we can have all three versions by using `SizedArray` as the container for the third option of "FixedMass".

test/api_tests.jl
jacobusmmsmit: Linting changes un-committed, new issue RE JuliaFormatter

src/core/model.jl
```diff
@@ -84,21 +105,28 @@ warnings are thrown when appropriate.
...
+    return ABM{S,A,C,F,P,R}(agents, space, scheduler, properties, rng, Ref(0))
 end
 
 function AgentBasedModel(agent::AbstractAgent, args...; kwargs...)
     return ABM(typeof(agent), args...; kwargs...)
 end
 
+"""
+    UnkillableABM(AgentType [, space]; properties, kwargs...) → model
+Create an unkillable agent-based model from the given agent type and `space`. For more details see `ABM`.
```
Datseris, line 126: Let's say a bit more here. I propose:
"""
Same as [`ABM`](@ref), but agents cannot be removed during model stepping. Additionally, it is mandatory that the agent ID is exactly the same as the agent insertion order (i.e., the 5th agent added to the model **must** have ID 5). This allows the internal storage of agents to be in a standard `Vector`, accelerating a lot the performance of retrieving agents from the model.
"""

This of course means that in the function `add_agent!`  you must add a dispatch method for `UnkillableABM` which enforces that an added agent must have ID that is the amount of current agents + 1.

src/core/model.jl
```diff
@@ -107,27 +112,39 @@ export random_agent, nagents, allagents, allids, nextid, seed!
...
 """
     model[id] = agent
     setindex!(model::ABM, agent::AbstractAgent, id::Int)
-
 Add an `agent` to the `model` at a given index: `id`.
 Note this method will return an error if the `id` requested is not equal to `agent.id`.
 **Internal method, use [`add_agents!`](@ref) instead to actually add an agent.**
 """
-function Base.setindex!(m::ABM, a::AbstractAgent, id::Int)
+function Base.setindex!(m::ABM{S,A,Dict{Int,A}}, a::AbstractAgent, id::Int) where {S,A}
```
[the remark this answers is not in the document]
Datseris: `setindex!` is not the way agents are added to the model. We should have a dedicated internal function, not exposed to the public API, called `add_agent_to_model!`.

src/spaces/nothing.jl
```diff
@@ -3,7 +3,7 @@ This file implements the "agent-space interaction API" for `nothing`, ie
 no space type
 =#
 
-function kill_agent!(agent::A, model::ABM{Nothing,A}) where {A<:AbstractAgent}
+function kill_agent!(agent::A, model::ABM{Nothing,A,Dict{Int,A}}) where {A<:AbstractAgent}
     delete!(model.agents, agent.id)
 end
```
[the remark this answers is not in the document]
Datseris: The functions `add_agent_to_model!` and `remove_agent_from_model!` dispatch on `::ABM` and `::UnkillableABM`.

Datseris (commented):
Thank you very much for this contribution. This is really amazing. I added some things that you can add here to make this more generalizable. I already know that in the future multi-agent models will be a different type with a different storage system, so we might as well start with a nice interface from the get go. Besides other minor changes I am asking you to add two new functions and remove the `setindex!` completely.

src/spaces/nothing.jl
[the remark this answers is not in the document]
jacobusmmsmit: Shall I make a new PR to work on the internal API before continuing here? It seems that it would be very helpful to have these core functions as neat as possible before building on them.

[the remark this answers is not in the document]
Datseris: These functions are anyway best seen in the context of two or more variants of ABM, so it's fine for me to do this in this PR.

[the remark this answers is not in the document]
jacobusmmsmit: In the meantime what's the proper way to implement addition and removal?

src/core/model.jl
jacobusmmsmit: Done with d2dd559082a971ac8ea066c459f9a514f2a121e5.

src/core/model.jl
```diff
@@ -43,23 +52,21 @@ union_types(T::Union) = (union_types(T.a)..., union_types(T.b)...)
...
     AgentBasedModel(AgentType [, space]; properties, kwargs...) → model
+
 Create an agent-based model from the given agent type and `space`.
 You can provide an agent _instance_ instead of type, and the type will be deduced.
 `ABM` is equivalent with `AgentBasedModel`.
-
 The agents are stored in a dictionary that maps unique IDs (integers)
-to agents. Use `model[id]` to get the agent with the given `id`.
-
+to agents or a vector whose indices are IDs. Use `model[id]` to get the agent with the given `id`.
```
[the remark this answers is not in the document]
jacobusmmsmit: Done with d2dd559082a971ac8ea066c459f9a514f2a121e5.

Datseris:
@jacobusmmsmit I was a bit busy but now I am back in programming. Do you need any support here? What's the progress?

jacobusmmsmit:
Honestly I interpreted "it's fine for me to do this in this PR" as you taking this part so I left it for a while. Right now I'm getting a workshop submission ready for AAMAS, deadline is at the end of January. I'll definitely come back to this though.

Datseris:
I could do this yes; but I think it will be a great learning experience for you to learn how to write such a more general design. So I'll give you until the end of the middle of Jan to get this through, I hope this will be enough. if not I'll do it. I am giving a workshop on Agents.jl at the Max Planck Institute for Evolutionary Biology start of March, and I'd certainly want your feature to be in there by then!

jacobusmmsmit:
On it :) I didn't mean to imply that I expected you to do it. Just a
misunderstanding on my part. I'll have it done by mid Feb.

Datseris:
no worries, I know you didn't mean to imply this :) Good luck with your workshop submission

src/core/model.jl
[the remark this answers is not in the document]
jacobusmmsmit: Attempted to address this in [161ab64](https://github.com/JuliaDynamics/Agents.jl/pull/721/commits/161ab64367682712c359805750a862a1ce0dc2a3)

Datseris:
Fantastic, thanks a lot. I will start a review today!!!

src/core/space_interaction_API.jl
```diff
@@ -140,11 +165,15 @@ end
...
-    delete!(model.agents, a.id)
+function kill_agent!(a::AbstractAgent, model::ABM{S,A,Dict{Int,A}}) where {S,A}
+    remove_agent_from_model!(a, model)
     remove_agent_from_space!(a, model)
 end
 
+function kill_agent!(a::AbstractAgent, model::ABM{S,A,<:AbstractVector{A}}) where {S,A}
+    nokill()
+end
+
```
Datseris, line 176: Isn't this unecessary? Instead of having two call signatures you can just make the `remove_agent_from_model!(a, model)` function throw this error for SizedArray.

src/core/space_interaction_API.jl
```diff
@@ -169,6 +198,12 @@ function genocide!(model::ABM, n::Integer)
     model.maxid[] = n
 end
 
+# Must be implemented separately, otherwise genocide! would return an unhelpful
+# BoundsError error to the user.
+function genocide!(model::ABM{S,A,Vector{A},F,P,R}, n::Integer) where {S,A,F,P,R}
+    nokill()
+end
+
```
Datseris, line 206: Same as above; as long as `remove_agent_from_model!` does what it is supposed to be, all these downstream functions don't need special methods. The function `nokill()` itself has no reason to exist. `remove_agent_from_model!` with `SizedArray` should become `nokill()`.

src/core/space_interaction_API.jl
```diff
@@ -140,11 +165,15 @@ end
...
-function kill_agent!(a::AbstractAgent, model::ABM)
-    delete!(model.agents, a.id)
+function kill_agent!(a::AbstractAgent, model::ABM{S,A,Dict{Int,A}}) where {S,A}
+    remove_agent_from_model!(a, model)
     remove_agent_from_space!(a, model)
 end
 
+function kill_agent!(a::AbstractAgent, model::ABM{S,A,<:AbstractVector{A}}) where {S,A}
+    nokill()
+end
```
Datseris, line 175: Isn't this unecessary? Instead of having two call signatures you can just make the `remove_agent_from_model!(a, model)` function throw this error for SizedArray.

src/core/space_interaction_API.jl
```diff
@@ -192,12 +227,22 @@ end
...
+
 """
     add_agent_pos!(agent::AbstractAgent, model::ABM) → agent
 Add the agent to the `model` at the agent's own position.
 """
 function add_agent_pos!(agent::AbstractAgent, model::ABM)
-    model[agent.id] = agent
+    add_agent_to_model!(agent, model)
     model.maxid[] < agent.id && (model.maxid[] = agent.id)
     add_agent_to_space!(agent, model)
```
Datseris, line 247: The `model.maxid` alteration **must** happen in the `add_agent_to_model!` call. Not outside it. This makes for safer code.

src/spaces/nothing.jl
```diff
@@ -3,17 +3,18 @@ This file implements the "agent-space interaction API" for `nothing`, ie
...
 function add_agent!(agent::A, model::ABM{Nothing,A}) where {A<:AbstractAgent}
     add_agent_pos!(agent, model)
 end
 
 function add_agent_pos!(agent::A, model::ABM{Nothing,A}) where {A<:AbstractAgent}
-    model[agent.id] = agent
-    return model[agent.id]
+    add_agent_to_model!(agent, model)
+    model.maxid[] < agent.id && (model.maxid[] = agent.id)
+    return agent
```
Datseris, line 17: Same like before: the only functions that should need special methods for add/remove agents are the `add_agent_to_model!`. Everything downstream should "just worK" without caring about the model, by simply calling the `add_agent_to_model!` etc. functions.

Datseris (commented):
This is fantastic, thanks. I left some comments on how you can significantly simplify the code even more: all top level API functions like `add_agent!`, `genocide!` , etc. should not care about model internals. They should have only one method that calls the internal `add_agent_to_model!`, and this function takes care of business accordingly.

I didn't really get your problem with `FixedMassABM`: why is it bad to export? Seems to me rather straight forward. Since it is fixed mass, a user must provde a vector of agents already. That's fine. What other problem was there for not making it Public?

Datseris:
(once you address these small comments, I'll merge, write some dev docs on the interface for agent based models, and tag a new fantastic release :)

src/core/space_interaction_API.jl
jacobusmmsmit: That's true, this is a remnant from before I implemented `remove_agent_from_model!`.

jacobusmmsmit: I've changed `remove_agent_from_model!` to throw `nokill` by default so we can remove this line.

src/core/space_interaction_API.jl
jacobusmmsmit: Done.

src/core/space_interaction_API.jl
jacobusmmsmit: This one I implemented specifically because iteration in the above one uses `for (k, v) in model.agents` which would throw a bounds error for vector-stored agents rather than anything about not being able to remove agents. This implementation is not perfect, but it saves having to qualify the above `genocide!` function with `genocide!(model::ABM{..., Dict{Int, A}} where ...`.

src/spaces/nothing.jl
jacobusmmsmit: I'm not sure I understand your comment. Currently the `add_agent_pos!` functions differ in dispatch because one of them adds the agent to the space, and the other has no space to add it to, so doesn't. Are you saying this should happen at the `add_agent_to_model!` level? i.e.
`add_agent_to_model!` dispatched on `ABM{Nothing, A}` vs `ABM{AbstractSpace, A}` to conditionally include `add_agent_to_space!` and there should only be *one* location in source code where `add_agent_pos!` is defined?

src/core/space_interaction_API.jl
jacobusmmsmit: On a slightly wider note: I think all of the arguments to genocide besides `model` should be keyword only. This comes about because `genocide!(model, [5])` and `genocide!(model, 5)` do something so dramatically different, the prior of which I wouldn't even call "genocide".

Datseris:
@jacobusmmsmit perhaps you forgot to push commits here? In some comments you reply that you've addressed them but the code has remained the same.

jacobusmmsmit:
Whoops... Pushed.

jacobusmmsmit:
Well that's awkward, let me fix that...

Datseris (approved):
@jacobusmmsmit this is a fantastic contribution, thank you very much! I'll merge once tests pass!!!

jacobusmmsmit:
Thank you for helping me contribute :) I look forward to making more PRs in the future. In fact I can think of a few enhancements...

Datseris:
I'm sad that GitHub doesn't have a _fire_ reaction emoji!

PR #731, opened 2023-01-19
 Simplify three argument macro version

### LINKED ISSUE #729: Simplify code for `@agent`
https://github.com/JuliaDynamics/Agents.jl/blob/f05a17aca2242df26423df1dc890f10d150658b1/src/core/agents.jl#L205-L207

How about doing it like this:

```jl
julia> macro testmacro(numba)
           quote
               @testmacro($numba, 4)
           end
       end
@testmacro (macro with 1 method)

julia> macro testmacro(numba, poppa)
           quote
               Core.eval($(__module__), ($numba * $poppa))
           end
       end
@testmacro (macro with 2 methods)

julia> @testmacro 2
8

julia> @testmacro 2 4
8

julia> @testmacro 2 6
12

julia> @testmacro 3
12
```

This seems to work as expected? I didn't test it yet within our actual macro, just thought that maybe this would work and indeed it seems as if it does? Have you already tried this approach @Datseris and if so, what was the problem with it? If not, I can write a PR for this and see whether it actually works.

fbanning:
Closes #729

Let's see if the tests still work because I have no experience with macro programming tbh.

### DIFF
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -203,12 +203,15 @@ macro agent(new_name, base_type, super_type, extra_fields)
     end
 end
 
-# Shorter macro for convenience
 macro agent(new_name, base_type, extra_fields)
-    quote
-        # Call the actual macro with default `super_type = AbstractAgent`
+    # Here we nest one macro call into another because there is no way to provide 
+    # defaults for macro arguments. We proceed to call the actual macro with the default
+    # `super_type = AbstractAgent`. This requires us to disable 'macro hygiene', see here
+    # for a brief explanation of the potential issues with this: 
+    # https://discourse.julialang.org/t/calling-a-macro-from-within-a-macro-revisited/19680/16?u=fbanning
+    esc(quote
         @agent($new_name, $base_type, AbstractAgent, $extra_fields)
-    end
+    end)
 end
 
 """
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -210,7 +210,7 @@ macro agent(new_name, base_type, extra_fields)
     # for a brief explanation of the potential issues with this: 
     # https://discourse.julialang.org/t/calling-a-macro-from-within-a-macro-revisited/19680/16?u=fbanning
     esc(quote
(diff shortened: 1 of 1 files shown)


fbanning:
Documentation of the agent types fails due to the recently (#717) introduced changes to docstring assignment (can now be found in line 201). This is a bit annoying.

I think it might have to do with the fact that I nest a macro inside a macro now. Maybe there's a way around this somehow.

fbanning:
Hm. Metaprogramming is rough. I think the closest that works is to disable macro hygiene for the case of the three argument macro version. See [this post on discourse](https://discourse.julialang.org/t/calling-a-macro-from-within-a-macro-revisited/19680/16?u=fbanning) which explains the issue at hand a bit.

With #5dec2bf the tests now all pass again.

The question is though: Do we want to do it like this? @Datseris

Datseris:
I think this is totally fine. The post instructs that the problem with disabling macro hygiene is that

> In this case you’re assuming that the module which uses `@innerloop` ex also has the macro `@simd` available

However in our case both macros are `@agent` and `@agent`. They are the same. We have a guarantee that this name exists in scope since its called in the first place.

One case I would imagine it didn't work is if a user only imported agents. Can you test this:
```julia
import Agents
Agents.@agent ....
```
and see if it works or not?

fbanning:
When importing Agents, it requires me to change `@agent($new_name, $base_type, AbstractAgent, $extra_fields)` to `Agents.@agent($new_name, $base_type, Agents.AbstractAgent, $extra_fields)`. Then everything seems to work. The test suite also still passes with this change. I'll push it and you can have a final look, otherwise I think this is fine to merge.

Datseris:
Sorry, I don't understand what you mean. _Where_ did it require you to do this change...? As a user you'd do:
```julia
import Agents

Agents.@agent x y z # three argument version
```
Would this work?

Datseris:
Nevermind, now I understand: you meant that it required a change in the source code that you already pushed.

fbanning:
> it required a change in the source code that you already pushed

Exactly. :) Happy to see this merged.

PR #735, opened 2023-01-31
Fix premade SIR model

fbanning:
`Models.sir()` was broken. I fixed it.


fbanning:
@Datseris In this PR I fixed the premade SIR model in our package but it's a bit annoying since it requires to use the `Distributions` package for its internal logic. I know that you dislike adding dependencies, so here's the question for you: Are you OK with adding `Distributions` as a direct dependency? 

- If yes, then I can merge this PR as is.
- If not, there are two ways we could tackle this:
  1. Create an extra environment inside the `src/models/` folder that is used by the `Models` module.
  2. Move all the example models from `src/models/` over to AgentsExampleZoo.jl

What do you think?

Datseris:
Why do we need the special distributions? Couldn't we get away with using just `rand` and `randn`? In the end the examples aren't meant to be demonstrating a realistic scientific application scenario, but more like a usage of Agents.jl

fbanning:
I didn't change anything about the model logic, that's just how it always was. Surely I can change the model itself.

Datseris:
I see. I checked the example again, and the dependency tree. Distributions isn't in our dependencies even if you go deeper layers. Additionally, the migration rates are given by the user. Since StatsBase is in our dependencies, couldn't we just use the `sample` function with the migration rates as weights? There is no way to obtain the Poisson distribution, but I'd say we could just use `rate*abs(randn(rng))` instead. 

Are you okay wiwth doing these changes? (You'd also need to do them in the example file.)

fbanning:
OK, things should work

Datseris:
Fantastic! Please increment patch version in Project.toml so I can tag a release immediatelly after!

Datseris (approved):

Issue #739, opened 2023-02-14
Extremely high memory usage for continuous models


Tortar:
The memory usage for the `flockers` models is way too high, I didn't look carefully to see where this goes wrong (probably some unbounded cache) but it is something which surely requires work. For the code below it has a pick of more than 8 gb of memory. 

<details> <summary> Flocker model benchmark </summary>

```julia
using Agents, LinearAlgebra

@agent Bird ContinuousAgent{2} begin
    speed::Float64
    cohere_factor::Float64
    separation::Float64
    separate_factor::Float64
    match_factor::Float64
    visual_distance::Float64
end

function initialize_model(;
    n_birds = 3200,
    speed = 1.0,
    cohere_factor = 0.25,
    separation = 4.0,
    separate_factor = 0.25,
    match_factor = 0.01,
    visual_distance = 10.0,
    extent = (200, 200),
)
    space2d = ContinuousSpace(extent; spacing = 1)
    model = ABM(Bird, space2d, scheduler = Schedulers.Randomly())
    for _ in 1:n_birds
        vel = Tuple(rand(model.rng, 2) * 2 .- 1)
        add_agent!(
            model,
            vel,
            speed,
            cohere_factor,
            separation,
            separate_factor,
            match_factor,
            visual_distance,
        )
    end
    return model
end
nothing # hide

# ## Defining the agent_step!
# `agent_step!` is the primary function called for each step and computes velocity
# according to the three rules defined above.
function agent_step!(bird, model)
    ## Obtain the ids of neighbors within the bird's visual distance
    neighbor_ids = nearby_ids(bird, model, bird.visual_distance)
    N = 0
    match = separate = cohere = (0.0, 0.0)
    ## Calculate behaviour properties based on neighbors
    for id in neighbor_ids
        N += 1
        neighbor = model[id].pos
        heading = neighbor .- bird.pos

        ## `cohere` computes the average position of neighboring birds
        cohere = cohere .+ heading
        if euclidean_distance(bird.pos, neighbor, model) < bird.separation
            ## `separate` repels the bird away from neighboring birds
            separate = separate .- heading
        end
        ## `match` computes the average trajectory of neighboring birds
        match = match .+ model[id].vel
    end
    N = max(N, 1)
    ## Normalise results based on model input and neighbor count
    cohere = cohere ./ N .* bird.cohere_factor
    separate = separate ./ N .* bird.separate_factor
    match = match ./ N .* bird.match_factor
[...]

PR #740, opened 2023-02-14
Solve unbounded cache problem when r is not an integer

Tortar:
This (still to be tested and adapted well) solves the memory bug described in https://github.com/JuliaDynamics/Agents.jl/issues/739, the problem was that we saved a float value which means that for a continuous space (with a grid behind it) it will grow and grow over all the simulation, but this can be avoided saving just the floored/ceiled values. The patch can surely be improved a bit in terms of code used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # main
 - Two new variants of agent based models: `UnkillableABM` and `FixedMassABM`: they yield huge performance benefits (up to twice the speed!!!) on iterating over agents if the agents can't get killed, or even added, during model evolution!
+- Huge memory performance increase in continuous space by fixing a memory leak bug.
 - `multi_agents_type!` has been updated to handle edge case where agents of one (or more) type are absent at the beginning of the simulation.
 - New function `npositions` that returns the number of positions of a model with a discrete space.
 
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -198,10 +198,18 @@ function nearby_ids(pos::ValidPos, model::ABM{<:ContinuousSpace{D,A,T}}, r = 1;
     # Calculate maximum grid distance (distance + distance from cell center)
     δ = distance_from_cell_center(pos, cell_center(pos, model))
     grid_r = (r + δ) / model.space.spacing
+    metric = model.space.grid.metric
+    if metric == :euclidean
+        int_grid_r = ceil(Int, grid_r)
+    elseif metric == :manhattan
+        int_grid_r= floor(Int, grid_r)
+    elseif metric == :chebyshev
+        int_grid_r = floor(Int, grid_r)
+    end
     # Then return the ids within this distance, using the internal grid space
     # and iteration via `GridSpaceIdIterator`, see spaces/grid_multi.jl
     focal_cell = pos2cell(pos, model)
-    return nearby_ids(focal_cell, model.space.grid, grid_r)
+    return nearby_ids(focal_cell, model.space.grid, int_grid_r)
 end
 
 """
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -197,19 +197,13 @@ function nearby_ids(pos::ValidPos, model::ABM{<:ContinuousSpace{D,A,T}}, r = 1;
     end
     # Calculate maximum grid distance (distance + distance from cell center)
     δ = distance_from_cell_center(pos, cell_center(pos, model))
-    grid_r = (r + δ) / model.space.spacing
+    # Ceiling since the grid has euclidean metric
+    grid_r = ceil(Int, (r + δ) / model.space.spacing)
     metric = model.space.grid.metric
-    if metric == :euclidean
-        int_grid_r = ceil(Int, grid_r)
-    elseif metric == :manhattan
-        int_grid_r= floor(Int, grid_r)
-    elseif metric == :chebyshev
-        int_grid_r = floor(Int, grid_r)
-    end
     # Then return the ids within this distance, using the internal grid space
     # and iteration via `GridSpaceIdIterator`, see spaces/grid_multi.jl
     focal_cell = pos2cell(pos, model)
-    return nearby_ids(focal_cell, model.space.grid, int_grid_r)
+    return nearby_ids(focal_cell, model.space.grid, grid_r)
 end
 
 """
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -199,7 +199,6 @@ function nearby_ids(pos::ValidPos, model::ABM{<:ContinuousSpace{D,A,T}}, r = 1;
     δ = distance_from_cell_center(pos, cell_center(pos, model))
     # Ceiling since the grid has euclidean metric
     grid_r = ceil(Int, (r + δ) / model.space.spacing)
-    metric = model.space.grid.metric
     # Then return the ids within this distance, using the internal grid space
     # and iteration via `GridSpaceIdIterator`, see spaces/grid_multi.jl
     focal_cell = pos2cell(pos, model)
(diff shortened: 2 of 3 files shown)


Tortar:
Running this benchmark

<details> <summary> Flocker model benchmark </summary>

```julia
using Agents, LinearAlgebra

@agent Bird ContinuousAgent{2} begin
    speed::Float64
    cohere_factor::Float64
    separation::Float64
    separate_factor::Float64
    match_factor::Float64
    visual_distance::Float64
end

function initialize_model(;
    n_birds = 3200,
    speed = 1.0,
    cohere_factor = 0.25,
    separation = 4.0,
    separate_factor = 0.25,
    match_factor = 0.01,
    visual_distance = 10.0,
    extent = (200, 200),
)
    space2d = ContinuousSpace(extent; spacing = 1)
    model = ABM(Bird, space2d, scheduler = Schedulers.Randomly())
    for _ in 1:n_birds
        vel = Tuple(rand(model.rng, 2) * 2 .- 1)
        add_agent!(
            model,
            vel,
            speed,
            cohere_factor,
            separation,
            separate_factor,
            match_factor,
            visual_distance,
        )
    end
    return model
end
nothing # hide

# ## Defining the agent_step!
# `agent_step!` is the primary function called for each step and computes velocity
# according to the three rules defined above.
    ## Obtain the ids of neighbors within the bird's visual distance
    neighbor_ids = nearby_ids(bird, model, bird.visual_distance)
    N = 0
    match = separate = cohere = (0.0, 0.0)
    ## Calculate behaviour properties based on neighbors
    for id in neighbor_ids
        N += 1
        neighbor = model[id].pos
        heading = neighbor .- bird.pos

        ## `cohere` computes the average position of neighboring birds
        cohere = cohere .+ heading
        if euclidean_distance(bird.pos, neighbor, model) < bird.separation
            ## `separate` repels the bird away from neighboring birds
            separate = separate .- heading
        end
        ## `match` computes the average trajectory of neighboring birds
        match = match .+ model[id].vel
    end
    N = max(N, 1)
    ## Normalise results based on model input and neighbor count
    cohere = cohere ./ N .* bird.cohere_factor
    separate = separate ./ N .* bird.separate_factor
    match = match ./ N .* bird.match_factor
    ## Compute velocity based on rules defined above
    bird.vel = (bird.vel .+ cohere .+ separate .+ match) ./ 2
    bird.vel = bird.vel ./ norm(bird.vel)
    ## Move bird according to new velocity and speed

end


model = initialize_model()
step!(model, agent_step!, 200)
```
</details>

Memory used at the end of the simulation before this pr

Screenshot from 2023-02-14 16-07-40

Memory used at the end of the simulation after this pr

Screenshot from 2023-02-14 16-08-52

Tortar:
I'm confused too actually now :-) 

What I'm noticing is that having `spacing` lower than `visual_distance` in `flockers` (done also in the example model) creates an abnormous amount of memory increase.

Tortar:
I think that the problem is that with the current main each time an agent move `offsets_within_radius` adds a key with the value of `grid_r` defined in `nearby_ids(pos::ValidPos, model::ABM{<:ContinuousSpace{D,A,T}}, r = 1; exact = false)`, so for example for the simulation above with 200 steps and 3200 agents the `offsets_within_radius` at the end of the simulation has a length of 200*3200 = 640000 elements. My PR tries to address this problem, even if I don't know if it the best approach.

What's strange  to me though is that if I change the value of `spacing` from 1 (the case in the benchmark code above) to higher values (say 100) then the memory increase is much less relevant.

Tortar:
ok now it's clear why spacing have an effect, it depends on the fact that the grid is bigger when spacing is low so even if the number of keys in the dictionary remains equals when spacing gets lower the number of values for each key gets greater, to verify this run 

```julia
println(sum(map(x -> length(x), values(model.space.grid.offsets_within_radius))))
println(length(keys(model.space.grid.offsets_within_radius)))
```

at the end of the simulation above, changing the value of spacing

Tortar:
I'm measuring a 1.5x speed-up for not so big benchmark. 

For a big benchmark like this one instead

<details>

```julia
using Agents, LinearAlgebra, Random, BenchmarkTools

@agent Bird ContinuousAgent{2} begin
    speed::Float64
    cohere_factor::Float64
    separation::Float64
    separate_factor::Float64
    match_factor::Float64
    visual_distance::Float64
end

function initialize_model(;
    n_birds = 100000,
    speed = 1.0,
    cohere_factor = 0.03,
    separation = 4.0,
    separate_factor = 0.015,
    match_factor = 0.05,
    visual_distance = 5.0,
    extent = (1000, 1000),
)
    space2d = ContinuousSpace(extent, spacing=spacing)
    model = ABM(Bird, space2d, scheduler = Schedulers.randomly)
    for _ in 1:n_birds
        vel = Tuple(rand(model.rng, 2) * 2 .- 1)
        add_agent!(
            model,
            vel,
            speed,
            cohere_factor,
            separation,
            separate_factor,
            match_factor,
            visual_distance,
        )
    end
    return model
end

    # Obtain the ids of neighbors within the bird's visual distance
    neighbor_ids = nearby_ids(bird, model, bird.visual_distance)
    N = 0
    match = separate = cohere = (0.0, 0.0)
    # Calculate behaviour properties based on neighbors
    for id in neighbor_ids
        N += 1
        neighbor = model[id].pos
        heading = neighbor .- bird.pos

        # `cohere` computes the average position of neighboring birds
        cohere = cohere .+ heading
        if edistance(bird.pos, neighbor, model) < bird.separation
            # `separate` repels the bird away from neighboring birds
            separate = separate .- heading
        end
        # `match` computes the average trajectory of neighboring birds
        match = match .+ model[id].vel
    end
    N = max(N, 1)
    # Normalise results based on model input and neighbor count
    cohere = cohere ./ N .* bird.cohere_factor
    separate = separate ./ N .* bird.separate_factor
    match = match ./ N .* bird.match_factor
    # Compute velocity based on rules defined above
    bird.vel = (bird.vel .+ cohere .+ separate .+ match) ./ 2
    bird.vel = bird.vel ./ norm(bird.vel)
    # Move bird according to new velocity and speed
end

@btime (model = initialize_model(); step!(model, agent_step!, 200)) samples=1 evals=1
```

</details>

I can't measure it since with my 16 gb of ram the process gets actually killed with the current main xD. It takes instead 160 seconds with the patch!

Datseris (commented):
Oh shiiiiiiit now I understand what the _fuck_ we were doing wrong oh my goooooooooooooood. There was a huge refactoring of the gridspace and continuous space internals. And we forgot to round the fricking grid radious yeeeees. Thank you. 

Notice that the internal grid space of `ContinuousSpace` always has `:euclidean` metric. So you can remove this if clause and just get the ceiling immediatelly.

Thank you :) 

Please add a changelog entry as well, something like "huge performance increase in continuous space by fixing a memory leak bug"

Datseris (approved):

Tortar:
ops didn't realize it is always euclidean, now the fix is as simple as applying a ceiling :-)

Datseris:
I've invited you to the agents.jl team so now you can branch off the repo not your fork

Tortar:
thank you very much for the invitation!

Tortar:
Actually I'm still a bit confused about why we sometimes ceil and sometimes floor the value of radius, what is the reason of not going with just one way or the other?

I also noticed there is still a bug: if a user calls `offsets_within_radius` (or any other function which doesn't firstly ceil or floor the value of radius) with a non-integer radius the unbounded cache problem would come back again, I think there is a design problem here on how the radius is passed around, I think we should do the following thing to improve this:  ceil or floor inside `offsets_within_radius` (and `offsets_within_radius_0`) instead of doing it from the outside and then use `Int` for radius in keys in the underlying dict (this is optional, and I think shouldn't break any code anyway) and when calling `calculate_offsets`. 

What do you think @Datseris about this?

PR #741, opened 2023-02-14
Abstract `AgentBasedModel` interface

### LINKED ISSUE #668: Different versions of `AgentBasedModel`

Datseris:
This is a continuation of #721 and closes #668 . It makes `AgentBasedModel` an abstract type, and it defines a generic interface for it. It is extendable, documented, and will be the basis of future model types specialized for multi agent models.

Locally all tests pass, however I am not sure what modifications, if any, are necessary for the checkpointing functionality


Datseris:
cc @jacobusmmsmit

src/core/model_concrete.jl
```diff
@@ -0,0 +1,257 @@
...
+"""
+    StandardABM(AgentType [, space]; properties, kwargs...) → model
+
+
+The most standard concrete implementation of an [`AgentBasedModel`](@ref),
+as well as the default version of the generic [`AgentBasedModel`](@ref) constructor.
+`StandardABM` stores agents in a dictionary mapping unique `Int` IDs to agents.
+See also [`UnkillableABM`](@ref), [`FixedMassABM`](@ref).
+"""
+StandardABM(args...; kwargs...) = SingleContainerABM(args...; container=Dict)
```
jacobusmmsmit, line 79: Super minor suggestion:
```suggestion
StandardABM(args...; kwargs...) = SingleContainerABM(args...; container=Dict{Int})
```

src/core/model_concrete.jl
```diff
@@ -0,0 +1,257 @@
...
+
+"""
+    FixedMassABM(agent_vector [, space]; properties, kwargs...) → model
+
+Similar to [`UnkillableABM`](@ref), but agents cannot be removed nor added.
+Hence, all agents in the model must be provided in advance as a vector.
+This allows storing agents into a `SizedVector`, a special vector with statically typed
+size which is the same as the size of the input `agent_vector`.
+This version of agent based model has slightly better iteration and retrieval speed
+than [`UnkillableABM`](@ref).
```
jacobusmmsmit, line 102:
```suggestion
This version of agent based model has slightly better iteration and retrieval speed
than [`UnkillableABM`](@ref).
However, there may be an impact on the compilation speed of the model so it is
advisable to try both container types.
```

jacobusmmsmit: I wasn't sure exactly what to write here but I felt it important to let users know to at least try both and see what compiles/runs faster.

src/core/model_concrete.jl
```diff
@@ -0,0 +1,257 @@
...
+"""
+function FixedMassABM(
+    agents::AbstractVector{A},
+    space::S = nothing;
+    scheduler::F = Schedulers.fastest,
+    properties::P = nothing,
+    rng::R = Random.default_rng(),
+    warn = true
+) where {A<:AbstractAgent, S<:SpaceType,F,P,R<:AbstractRNG}
+    C = SizedVector{length(agents), A}
```
AayushSabharwal, line 115: If we have `construct_agent_container`, shouldn't we aim to use it consistently?

```julia
construct_agent_container(T::Type{<:SizedVector}, A) = T{A}
```

Then call it as `construct_agent_container(SizedVector{length(agents)}, A)`

src/core/model_concrete.jl
```diff
@@ -0,0 +1,257 @@
...
+    agent_validator(A, space, warn)
+    return SingleContainerABM{S,A,C,F,P,R}(fixed_agents, space, scheduler, properties, rng, Ref(0))
+end
+
+#######################################################################################
+# %% Model accessing api
+#######################################################################################
+nextid(model::SingleContainerABM{<:SpaceType,A,Dict{Int, A}}) where {A} = getfield(model, :maxid)[] + 1
+nextid(model::SingleContainerABM{<:SpaceType,A,Vector{A}}) where {A} = nagents(model) + 1
+nextid(::SingleContainerABM{<:SpaceType,A,SizedVector}) where {A} = error("There is no `nextid` in a `FixedMassABM`. Most likely an internal error.")
```
AayushSabharwal, line 130:
```suggestion
nextid(::SingleContainerABM{<:SpaceType,A,<:SizedVector}) where {A} = error("There is no `nextid` in a `FixedMassABM`. Most likely an internal error.")
```

This won't work without the `<:`:
```
julia> f(::Vector{Integer}) = 3

julia> f(x) = 2
f (generic function with 2 methods)

julia> f([])
2
julia> f([1])
2

julia> f(::Vector{<:Integer}) = 3
f (generic function with 3 methods)

julia> f([1])
3
```

src/core/model_concrete.jl
```diff
@@ -0,0 +1,257 @@
...
+    maxid = getfield(model, :maxid)
+    if maxid[] < agent.id; maxid[] = agent.id; end
+    return
+end
+
+function add_agent_to_model!(agent, model::SingleContainerABM{<:SpaceType,A,Vector{A}}) where {A}
+    agent.id == nagents(model) + 1 || error("Cannot add agent of ID $(agent.id) in a vector ABM of $(nagents(model)) agents. Expected ID == $(nagents(model)+1).")
+    push!(agent_container(model), agent)
+    return
+end
```
AayushSabharwal, line 149: Would it be better if this method was for `SingleContainerABM{...,<:AbstractVector{A}}` and we error specifically for `SizedVector`? Otherwise people using a different vector container won't be able to add agents and would need to add another method.

src/core/model_concrete.jl
```diff
@@ -0,0 +1,257 @@
...
+        elseif space_type <: ContinuousSpace || space_type <: ContinuousSpace
+            if !(pos_type <: NTuple{D,<:AbstractFloat} where {D})
+                throw(ArgumentError("`pos` field in agent type must be of type `NTuple{<:AbstractFloat}` when using ContinuousSpace."))
+            end
+            if warn &&
+               any(isequal(:vel), fieldnames(A)) &&
+               !(fieldtype(A, :vel) <: NTuple{D,<:AbstractFloat} where {D})
+                @warn "`vel` field in agent type should be of type `NTuple{<:AbstractFloat}` when using ContinuousSpace."
+            end
+        end
```
AayushSabharwal, line 224: Can we improve these checks, where each space type declares its position type through a method. E.g. `positiontype(::Type{<:GraphSpace}) = Integer`? That way the checks become automatically extensible to more space types and don't break if we change the position type.

src/core/model_concrete.jl
```diff
@@ -0,0 +1,257 @@
...
+    end
+end
+
+#######################################################################################
+# %% Pretty printing
+#######################################################################################
+modelname(abm::ABM) = modelname(agent_container(abm))
+modelname(::Dict) = "StandardABM"
+modelname(::Vector) = "UnkillableABM"
```
AayushSabharwal, line 233: `<:AbstractVector` instead of `::Vector`, so people can use custom vectors if they so choose?

AayushSabharwal (commented):
These changes look excellent for the future of the package, but now that we're doing them there's a lot of potential to make the API more easily extensible. The only other things missing are tests for the new functions.

AayushSabharwal:
Serialization and deserialization would need more stringent tests with all the new types, too. I'll look into that hopefully tomorrow or over the weekend.

On a sidenote, I haven't been visibly active for a while here, but still keep myself informed of all the changes. I'm happy to review any PRs, and will continue to make suggestions/write code wherever appropriate or necessary.

src/core/model_abstract.jl
```diff
@@ -0,0 +1,251 @@
...
+    nextid(model::ABM) → id
+Return a valid `id` for creating a new agent with it.
+"""
+nextid(model::ABM) = notimplemented(model)
+
+"""
+    random_agent(model) → agent
+Return a random agent from the model.
+"""
+random_agent(model) = model[rand(abmrng(model), allids(model))]
```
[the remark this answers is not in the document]
jacobusmmsmit: As far as I know `eachindex` doesn't allocate. When I was implementing `allids` I chose to use `eachindex` over `keys` as `keys` allocates an "efficient" vector of indices when used on a vector i.e. one allocation but `eachindex` doesn't allocate at all and it was marginally faster because of this.

jacobusmmsmit:
Thanks for the thoughtful design comments. I definitely agree that there's room for improvement with regards to containers.

src/core/model_abstract.jl
```diff
@@ -0,0 +1,251 @@
...
+abstract type DiscreteSpace <: AbstractSpace end
+SpaceType = Union{Nothing,AbstractSpace}
+
+# This is a collection of valid position types, sometimes used for ambiguity resolution
+ValidPos = Union{
+    Int, # graph
+    NTuple{N,Int}, # grid
+    NTuple{M,<:AbstractFloat}, # continuous
+    Tuple{Int,Int,Float64} # osm
+} where {N,M}
```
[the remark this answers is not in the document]
Datseris: Yeah the problem is I don't see of a way / I don't know of a way to make this generaziable, besides the (rather easy actually) adding a new entry there. I'll addd a note to the dev docs

src/core/model_abstract.jl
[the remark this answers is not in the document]
Datseris:
```
julia> using BenchmarkTools

julia> model
UnkillableABM with 320 agents of type SchellingAgent
 space: GridSpaceSingle with size (20, 20), metric=chebyshev, periodic=false
 properties: min_to_be_happy

julia> @btime random_agent($model)
  6.500 ns (0 allocations: 0 bytes)
SchellingAgent(36, (6, 19), false, 1)
```

src/core/model_abstract.jl
```diff
@@ -0,0 +1,251 @@
...
+    ids = shuffle!(abmrng(model), collect(allids(model)))
+    i, L = 1, length(ids)
+    a = model[ids[1]]
+    while !condition(a)
+        i += 1
+        i > L && return nothing
+        a = model[ids[i]]
+    end
+    return a
+end
```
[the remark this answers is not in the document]
Datseris: I think this is a separate topic that needs to be done in a separate PR. I didn't adjust these functions in this PR.

src/core/model_concrete.jl
```diff
@@ -0,0 +1,257 @@
+export ABM, StandardABM, UnkillableABM, FixedMassABM
+using StaticArraysCore: SizedVector
+
+ContainerType{A} = Union{AbstractDict{Int,A}, AbstractVector{A}}
```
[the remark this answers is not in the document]
Datseris: No need to extend this: this is specific to the currently implemented concrete types for `SingleContainerABM`.

src/core/model_concrete.jl
```diff
@@ -0,0 +1,257 @@
...
+
+function construct_agent_container(container, A)
+    if container <: Dict
+        return Dict{Int,A}
+    elseif container <: Vector
+        return Vector{A}
+    else
+        throw(ArgumentError("Unrecognised container $container, please specify either Dict or Vector."))
+    end
+end
```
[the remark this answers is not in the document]
Datseris: Yes on the dispatch. However, at the moment there is nothing else to be allowed, in the sense that there are two concrete implementations that happen to share the same source code. The user doesn't know anything about this. THey have no idea of the existence of `SingleContainerABM`. It is not public API.

src/core/model_concrete.jl
Datseris: The problem is that the Sized case is rather special and initializes a container that already has all agents instead of an emptu container

src/core/model_concrete.jl
Datseris: Yes, but there is no other vector container at the moment. None of this is public. Perhaps in a different PR we can make the type of vector generalizable, but at the moment I don't see a need for it.

src/core/model_concrete.jl
Datseris: I agree, but this is not related to this PR and should be done in some other PR.

src/core/model_concrete.jl
Datseris: Same reply, choosing the vector type is not a possibility in this PR, but could be done in a future PR if someone wants to extend this further.

Datseris:
Alright, so to recap: Yes, it is definitely possible to generalize the container into abstract types. However, I would argue that this should not be done in this PR , but another. ANother PR may generalize things a bit, and then expose `SingleContainerABM`, and allow uses to choose the container type directly. For now, I am happy with things the way they are, and I want to merge this soon as I give a workshop on Agents.jl first of March and I want this to be part of the workshop!

jacobusmmsmit:
When you say "it doesn't work?" is this something I could look into? What didn't work with it?

PR #745, opened 2023-02-17
Subsitute deleteat! with swap and pop method

Tortar:
This method is faster than deleteat! since it's O(1) while deleteat! is O(length(vector)), there should be other places where this can be done, but before going on I'd like to ask if  keeping the insertion order of agents is instead preferable, in principle the agents all occupy the same position so maybe it's not important? I tested also that the two methods costs the same with vectors with only 1 agent

This benchmark compares the two methods with vectors of different length:

```julia
using BenchmarkTools

function delat(b,i)
    deleteat!(b,i)
end

function swappop(b, i)
    b[i], b[end] = b[end], b[i]
    pop!(b)
end

for i in [1, 10, 100, 1000, 10000, 100000]
    a = @belapsed (for k in 1:m delat(b, m) end) setup=(n = $i; b=copy([1:n;]); m = Int(ceil(n/2))) evals=1
    b = @belapsed (for k in 1:m swappop(b, m) end) setup=(n = $i; b=copy([1:n;]); m = Int(ceil(n/2))) evals=1
    println("length of vector: ", i, " -> ", round(a/b, digits=2), " times faster") 
end
```

</details>

which are probably too high in the last iterations (100000 agents in one position is a bit unusual :-)) gives:

```julia
length of vector: 1 -> 1.0 times faster
length of vector: 10 -> 1.67 times faster
length of vector: 100 -> 2.44 times faster
length of vector: 1000 -> 9.82 times faster
length of vector: 10000 -> 82.36 times faster
length of vector: 100000 -> 900.43 times faster
```

It's also important to notice that in this case, in `remove_agent_from_space`, the time will be dominated by `findfirst` after this modification, so sometimes the increase in speed will be the one above, sometimes will be much less, depending on the position of the agent in the vector.


Tortar:
Did a little profiling for this change inside `remove_agent_from_space` considering also the other operations and it seems that this change doesn't give much of a speed-up, something like 10-20% with big vectors, so it's probably not worth it

Datseris:
Hm my question is: why isn't base Julia doing this already...? Like, why isn't the source code of `deleteat!` exactly this, if this is faster...?

Tortar:
since it changes the order in the vector. Note that to restore the original ordering you need many more swapping, at worst n/2 if n is the length of the vector. Do you think that for us it is worth to use the proposed method?

Datseris:
Lets leave it as is

Tortar:
I thought again about this, and I don't see actually any problem with the change, and clearly this is faster because it's O(1) instead of O(n)

src/spaces/grid_multi.jl
```diff
@@ -97,7 +97,8 @@ function remove_agent_from_space!(a::AbstractAgent, model::ABM{<:GridSpace})
     prev = abmspace(model).stored_ids[a.pos...]
     ai = findfirst(i -> i == a.id, prev)
     isnothing(ai) && error(lazy"Tried to remove $(a) from the space, but that agent is not on the space")
-    deleteat!(prev, ai)
+    prev[ai], prev[end] = prev[end], prev[ai]
```
Datseris, line 100: Perhaps you can add a comment here that points to this PR for future reference?

Datseris (approved):
You are right; the insertion order of the agent in the space container does not matter. The user cannot access this and we clearly don't use it because we call `findfirst` anyways. So this is fine change to do. It won't have any realistic impact truth be told but it doesn't have any downsides either.

src/spaces/grid_multi.jl
Tortar: added a small comment

Datseris:
Hm, I am coming back here: did you make sure that this doesn't _lower_ the performance for typical simulations like e.g., the sheep wolf model? When you opened this PR the tests were done in a very different Julia version.

Tortar:
yes, I runned it, and this makes it a little bit faster, like 5% more

PR #746, opened 2023-02-19
stardardize metric functions

Tortar:
I noticed that the metrics code was not standardized: different methods (for loop, element-wise operations, ...) were used. I tried to simplify the code while mantaining/improving performance. Runned this benchmark

```julia
using Agents, BenchmarkTools
using Agents: euclidean_distance, manhattan_distance, euclidean_distance2, manhattan_distance2

@agent Ag GridAgent{2} begin end

function model_setup()
    multigrid = GridSpace((10, 10), periodic=true)
    model = ABM(Ag, multigrid)
    return model
end

a, b = (2,3), (5,4)
@btime euclidean_distance($a, $b, model) setup=(model=model_setup())
@btime manhattan_distance($a, $b, model) setup=(model=model_setup())
@btime euclidean_distance2($a, $b, model) setup=(model=model_setup())
@btime manhattan_distance2($a, $b, model) setup=(model=model_setup())
```
</details>

where `euclidean_distance2`, `manhattan_distance2` had the original code inside, which resulted in

```julia
@btime euclidean_distance($a, $b, model) setup=(model=model_setup())
  2.665 ns (0 allocations: 0 bytes)
@btime manhattan_distance($a, $b, model) setup=(model=model_setup())
  2.314 ns (0 allocations: 0 bytes)
@btime euclidean_distance2($a, $b, model) setup=(model=model_setup())
  6.141 ns (0 allocations: 0 bytes)
@btime manhattan_distance2($a, $b, model) setup=(model=model_setup())
  2.364 ns (0 allocations: 0 bytes)
```

Read some discussions saying that broadcasting can be slower sometimes, but doesn't seem the case here, it is actually two times faster in one case and roughly the same in the other. Alternatively, we can use for loops in all dispatched methods instead of broadcasting (also on the functions used when periodic=false).

### DIFF
--- a/src/spaces/utilities.jl
+++ b/src/spaces/utilities.jl
@@ -29,31 +29,10 @@ end
 function euclidean_distance(
     p1::ValidPos,
     p2::ValidPos,
-    model::ABM{<:ContinuousSpace{D,true}},
+    model::ABM{<:Union{ContinuousSpace{D,true},AbstractGridSpace{D,true}}},
 ) where {D}
-    total = 0.0
-    for (a, b, d) in zip(p1, p2, spacesize(model))
-        delta = abs(b - a)
-        if delta > d - delta
-            delta = d - delta
-        end
-        total += delta^2
-    end
-    sqrt(total)
-end
-
-function euclidean_distance(
-        p1::ValidPos, p2::ValidPos, model::ABM{<:AbstractGridSpace{D,true}}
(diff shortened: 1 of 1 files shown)


Tortar:
Tried other possibilities for this metrics and they are all slower than the methods above!

Datseris:
LGTM!

PR #749, opened 2023-02-23
Refactor some utilities functions

Tortar:
this is mostly a refactor where I added a dispatch to space instead to pass directly the model to some utilities function, but there is also an improvement in performance for normalize_position in the non periodic grid case, I calculated that this new function is 50x faster than the previous one since calling `ones(Int, D)` is avoided now!


Datseris (approved):
Fantastic. Add an entry under 5.8 for the performance increase and go ahead and merge!

Issue #750, opened 2023-02-28
Radius r not automatically casted to integer when inserted in grid dictionary can cause memory bugs


Tortar:
I noticed there shoud still be a (less severe) memory bug with keys in the `offsets_within_radius` dictionary: if a user called with a non-integer radius  (e.g. 1.1 first time 1.4 second time, etc...) any function which in turn calls the `offsets_within_radius` function in `gridgeneral.jl`, the unbounded cache problem (or anyway a waste of memory and time even if just two non-integer value non casted to integer are passed) could come back again.
I think there is a design problem here on how the radius is passed around but I think we can do the following thing to improve this:  ceil or floor the radius inside `offsets_within_radius` (and `offsets_within_radius_0`) instead of doing it from the outside and then use `Int` for radius keys in the underlying dict (I think this shouldn't break any code) and when calling `calculate_offsets`. One possibility to simplify a bit the code involved is to create a utility function which transforms the value of radius to the correct integer in relation to the metric used, it could be called `normalize_radius` or something similar.

Besides this, I'm still actually a bit confused about why we sometimes ceil and sometimes floor the value of radius, what is the reason of not going with just one way or the other with all metrics? In my opinion, if this is done just for supporting some better approximation of nearby ids for continuous space then it should be taken into account just there, and then just floor (I think it would be better than ceil) in all cases inside the internal grid

Tortar:
I actually verified that now changing `offsets_within_radius` eliminating the floor and ceiling, like this:

```julia
function calculate_offsets(space::AbstractGridSpace{D}, r::Real) where {D}
    r0 = floor(Int, r)
    if space.metric == :euclidean
        hypercube = CartesianIndices((repeat([(-r0):r0], D)...,))
        # select subset which is in Hypersphere
        βs = [Tuple(β) for β ∈ hypercube if LinearAlgebra.norm(β.I) ≤ r]
    elseif space.metric == :manhattan
        hypercube = CartesianIndices((repeat([(-r0):r0], D)...,))
        βs = [Tuple(β) for β ∈ hypercube if sum(abs.(β.I)) ≤ r0]
    elseif space.metric == :chebyshev
        βs = vec([Tuple(a) for a in Iterators.product([(-r0):r0 for φ in 1:D]...)])
    else
        error("Unknown metric type")
    end
    length(βs) == 0 && push!(βs, ntuple(i -> 0, Val{D}())) # ensure 0 is there
[...]

PR #751, opened 2023-03-01
Solve caching problems in offsets_within_radius

Tortar:
This would solve #750, but it's still a bit clumsy, to improve the design we should change the keys type of `offsets_within_radius` from `float64` to `int` and accept an `int` instead of a `real` in `calculate_offsets`, is this fine?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # main
+- A memory bug about offsets calculation has been solved; besides, the `calculate_offsets` function has been sped-up by a significant amount.
 
 # v5.8
 - `random_agent` is now faster and has two options on how to find a random agent, each of which can offer a different performance benefit depending on the density of agents that satisfy the clause.
--- a/src/spaces/grid_general.jl
+++ b/src/spaces/grid_general.jl
@@ -48,7 +48,7 @@ The function does two things:
 offsets_within_radius(model::ABM, r::Real) = offsets_within_radius(model.space, r::Real)
 function offsets_within_radius(
     space::AbstractGridSpace{D}, r::Real)::Vector{NTuple{D, Int}} where {D}
-    r0 = float(floor(Int, r))
+    r0 = floor(Int, r)
     if haskey(space.offsets_within_radius, r0)
         βs = space.offsets_within_radius[r0]
     else
@@ -111,7 +111,7 @@ offsets_within_radius_no_0(model::ABM, r::Real) =
     offsets_within_radius_no_0(model.space, r::Real)
 function offsets_within_radius_no_0(
     space::AbstractGridSpace{D}, r::Real)::Vector{NTuple{D, Int}} where {D}
-    r0 = float(floor(Int, r))
+    r0 = floor(Int, r)
     if haskey(space.offsets_within_radius_no_0, r0)
         βs = space.offsets_within_radius_no_0[r0]
     else
--- a/src/spaces/grid_general.jl
+++ b/src/spaces/grid_general.jl
@@ -48,11 +48,12 @@ The function does two things:
 offsets_within_radius(model::ABM, r::Real) = offsets_within_radius(model.space, r::Real)
 function offsets_within_radius(
     space::AbstractGridSpace{D}, r::Real)::Vector{NTuple{D, Int}} where {D}
-    if haskey(space.offsets_within_radius, r)
-        βs = space.offsets_within_radius[r]
+    r0 = float(floor(Int, r))
+    if haskey(space.offsets_within_radius, r0)
+        βs = space.offsets_within_radius[r0]
     else
-        βs = calculate_offsets(space, r)
-        space.offsets_within_radius[float(floor(Int, r))] = βs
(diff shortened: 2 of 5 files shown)


Tortar:
something is failing with the latest commit, will look into that soon

Tortar:
something is failing with the latest commit, will look into that soon

Tortar:
something is failing with the latest commit, will look into that soon

Tortar:
ok, the problem is related to the previous ceiling with the `euclidean` metrics when the radius is not an integer for grid spaces (there are no problems anymore with continuous spaces since the ceiling is accounted for before calling these functions), so what to do about that? probably it's breaking to change this but I think that ceiling and flooring with different metrics would be a surprise for the user, so it would probably be an unexpected behaviour which is something we should avoid, what do you think @Datseris ?

Datseris:
Well, the problem is we need the Euclidean type metric for the continuous space. For grid space in my opinion it is nonsense to use euclidean metric anyways. But for continuous space it makes sense to have the boxes according to this euclidean metric...

What is currently happening with the boxes in the continuous space? You are still rouding the radious always to an integer multiple of the underlying compartment size, right? Doesn't this change the accuracy of the algorithm, i.e., make it less accurate? 

There is a plotting code in the continuous space test file that you can use to visualize your change.

Tortar:
I don't think there should be any change in the behaviour of continuous space, these lines in continuous.jl are untouched in this pr

```julia
    # Ceiling since the grid has euclidean metric (but probably I could write that ceiling instead of flooring is better for accuracy of `nearby_ids` instead of the current comment)
    grid_r = ceil(Int, (r + δ) / model.space.spacing)
```

or am I missing something?

Tortar:
This is I think the final version, unfortunately I have some problems running GLMakie so as a final verification I would test the changes gathering some statistics of the flocking model before and after this PR, so that to see if all is still the same (which I think it is) :-)

Tortar:
ok, everything works as expected, running

<details> <summary> Flocking model with statistics </summary> 
```julia
using Agents, LinearAlgebra, Random

@agent Bird ContinuousAgent{2} begin
    speed::Float64
    cohere_factor::Float64
    separation::Float64
    separate_factor::Float64
    match_factor::Float64
    visual_distance::Float64
end


function flocking(;
    n_birds = 100,
    speed = 1.0,
    cohere_factor = 0.25,
    separation = 4.0,
    separate_factor = 0.25,
    match_factor = 0.01,
    visual_distance = 5.0,
    extent = (100, 100),
    rng = rng
)
    space2d = ContinuousSpace(extent; spacing)
    model = StandardABM(Bird, space2d, scheduler = Schedulers.Randomly(), rng=rng)
    for _ in 1:n_birds
        vel = Tuple(rand(model.rng, 2) * 2 .- 1)
        add_agent!(
            model,
            vel,
            speed,
            cohere_factor,
            separation,
            separate_factor,
            match_factor,
            visual_distance,
        )
    end
    return model
end

    neighbor_ids = nearby_ids(bird, model, bird.visual_distance)
    N = 0
    match = separate = cohere = (0.0, 0.0)
    for id in neighbor_ids
        N += 1
        neighbor = model[id].pos
        heading = neighbor .- bird.pos
        cohere = cohere .+ heading
        if euclidean_distance(bird.pos, neighbor, model) < bird.separation
            separate = separate .- heading
        end
        match = match .+ model[id].vel
    end
    N = max(N, 1)
    cohere = cohere ./ N .* bird.cohere_factor
    separate = separate ./ N .* bird.separate_factor
    match = match ./ N .* bird.match_factor
    bird.vel = (bird.vel .+ cohere .+ separate .+ match) ./ 2
    bird.vel = bird.vel ./ norm(bird.vel)
end

rngs = [MersenneTwister(x) for x in 1:10]

for rng in rngs
    model = flocking(rng = rng)
    step!(model, agent_step!, 200)
    print(sum(map(a -> a.vel[1], collect(values(model.agents)))), " ")
    print(sum(map(a -> a.vel[2], collect(values(model.agents)))), "\n")
end
```
</details>

where I ran 10 flocking models each with a different rngs always return before and after this PR the same output, which is 

```julia
2.2525640873792314 16.08840559085231
-8.669039492814019 2.026938937148426
0.5832006320775777 9.126733961314178
-2.1906408823711 16.068520361089284
-17.656967292641074 -6.6928148110441645
-0.07816622407620577 4.751271031761771
-2.6621811314673165 -8.995219690701814
3.796578622360355 6.4012949051749
-5.479452157044974 3.427887346049989
6.671483265577846 -7.2814175590563535
```

Tortar:
Also simplified the `calculate_offsets` function which is now even faster with euclidean and manhattan metrics: e.g. for radius 1 it is 1.5 faster and radius 5 it is 4 times faster

Datseris (approved):
oof I finally udnerstand this PR. tjanks a lot for the effort! can yoi please add a changelog entry and then go ahead and merge?

Datseris:
haha it's amazing how badly i type when i type fast :D

Tortar:
I can't even type fast :D

Issue #752, opened 2023-03-04
Use OrderedDict for agents container in StandardABM for better reproducibility


Tortar:
Currently we are using a normal Dict as agents container in StandardABM, but what about avoiding reproducibility issues using OrderedDicts instead? I asked [a question on SO](https://stackoverflow.com/questions/75632333/are-julia-dictionaries-and-sets-stable-between-runs) and it seems that people agree that it's better for reproducibility to use them in Julia. A good reason I see mentioned in one of the answers is that the ordering of keys can change **across machines** (I think it is referred to the 32/64 bit architectures after looking at the implemantion of the hash functions in Base). For the differences across versions instead I don't think it is really important for us since also RNGs are not assured to be stable across them. From the performance standpoint there shouldn't be regressions I think, the ordered version seems to perform almost always at least as good as the unordered one even if the comparisons I found are a bit old, we could run some benchmarks to check

Datseris:
we don't do that because it is slower I think. If it isn't slower, then sure!

I am not sure whether "reproducibility" can be an argument here: use an `rng` with a given seed and the current dict is perfectly reproducible, or it isn't...?

Tortar:
on a single machine I guess it is but on different machines I think it can produce different results if the architecture matters: asked for clarifications on that matter on SO, but looking at https://github.com/JuliaLang/julia/blob/master/base/hashing.jl in Base it seems that the hashing is the same for all types of integers so dicts with integer keys should be reproducible even across machines, will try to benchmark them anyways

Datseris:
Yes, I understand this, but what I meant is: if you care about making a reproducible ABM then you don't use the fastest scheduler. You use e.g., Randomy with a given seed. Or by property.

Oh, I think you are saying e.g., even the by_propery scheduling would give different results in different machines even with the same seed, right? In this case then yes we should go to the ordered version unless there is some strong performance defecit.

Tortar:
exactly, but it seems that the implementation of the hashing of integers types is still consistent across machines but this is an implementation detail which can change (for other types e.g. it seems that this is instead not consistent)

Tortar:
experimenting a bit I'm finding these results but doing it very quick and dirty I could have made some mistakes: 2.2x faster access and iteration, 1.7x slower add new key, 1.4 faster remove key. Will try anyway to change Dict in OrderedDict in Agents.jl and see the speed-up/slowdown for some models

Tortar:
[...]

PR #753, opened 2023-03-04
Use OrderedDict instead of Dict for Agents container

Tortar:
This would improve reproducibility of results (at least if some implementations details in Base will change) as discussed in #752 , I'm still measuring some possible performance regressions in using OrderedDicts in our models, currently for the flocking model I have measured no difference whatsoever in performance but surely it needs more benchmarks to check more throughly the matter. Probably running the schedules.jl file in the benchmark folder can be good to test this change. Also, we can create a new alias (as default name) of `fastest` scheduler called e.g. `byinsertion`


Datseris:
I think you should measure performance in a model that has a lot of agent insertions and deletions. Like the sheep wolf model.

Tortar:
Measured, it is a bit slower:

<details> <summary> Wolf-sheep-grass benchmark </summary>

```julia

@agent Sheep GridAgent{2} begin
    energy::Float64
    reproduction_prob::Float64
    Δenergy::Float64
end

@agent Wolf GridAgent{2} begin
    energy::Float64
    reproduction_prob::Float64
    Δenergy::Float64
end

# The function `initialize_model` returns a new model containing sheep, wolves, and grass
# using a set of pre-defined values (which can be overwritten). The environment is a two
# dimensional grid space, which enables animals to walk in all
# directions.

function initialize_model(;
        n_sheep = 100,
        n_wolves = 50,
        dims = (20, 20),
        regrowth_time = 30,
        Δenergy_sheep = 4,
        Δenergy_wolf = 20,
        sheep_reproduce = 0.04,
        wolf_reproduce = 0.05,
        seed = 23182,
    )

    rng = MersenneTwister(seed)
    space = GridSpace(dims, periodic = true)
    ## Model properties contain the grass as two arrays: whether it is fully grown
    ## and the time to regrow. Also have static parameter `regrowth_time`.
    ## Notice how the properties are a `NamedTuple` to ensure type stability.
    properties = (
        fully_grown = falses(dims),
        countdown = zeros(Int, dims),
        regrowth_time = regrowth_time,
    )
    model = ABM(Union{Sheep, Wolf}, space;
        properties, rng, scheduler = Schedulers.randomly, warn = false
    )
    ## Add agents
    for _ in 1:n_sheep
        energy = rand(model.rng, 1:(Δenergy_sheep*2)) - 1
        add_agent!(Sheep, model, energy, sheep_reproduce, Δenergy_sheep)
    end
    for _ in 1:n_wolves
        energy = rand(model.rng, 1:(Δenergy_wolf*2)) - 1
        add_agent!(Wolf, model, energy, wolf_reproduce, Δenergy_wolf)
    end
    ## Add grass with random initial growth
    for p in positions(model)
        fully_grown = rand(model.rng, Bool)
        countdown = fully_grown ? regrowth_time : rand(model.rng, 1:regrowth_time) - 1
        model.countdown[p...] = countdown
        model.fully_grown[p...] = fully_grown
    end
    return model
end

sheepwolfgrass = initialize_model()

# ## Defining the stepping functions
# Sheep and wolves behave similarly:
# both lose 1 energy unit by moving to an adjacent position and both consume
# a food source if available. If their energy level is below zero, they die.
# Otherwise, they live and reproduce with some probability.
# They move to a random adjacent position with the [`randomwalk!`](@ref) function.

# Notice how the function `sheepwolf_step!`, which is our `agent_step!`,
# is dispatched to the appropriate agent type via Julia's Multiple Dispatch system.
function sheepwolf_step!(sheep::Sheep, model)
    randomwalk!(sheep, model, 1)
    sheep.energy -= 1
    if sheep.energy < 0
        kill_agent!(sheep, model)
        return
    end
    eat!(sheep, model)
    if rand(model.rng) ≤ sheep.reproduction_prob
        reproduce!(sheep, model)
    end
end

function sheepwolf_step!(wolf::Wolf, model)
    randomwalk!(wolf, model, 1; ifempty=false)
    wolf.energy -= 1
    if wolf.energy < 0
        kill_agent!(wolf, model)
        return
    end
    ## If there is any sheep on this grid cell, it's dinner time!
    dinner = first_sheep_in_position(wolf.pos, model)
    !isnothing(dinner) && eat!(wolf, dinner, model)
    if rand(model.rng) ≤ wolf.reproduction_prob
        reproduce!(wolf, model)
    end
end

function first_sheep_in_position(pos, model)
    ids = ids_in_position(pos, model)
    j = findfirst(id -> model[id] isa Sheep, ids)
    isnothing(j) ? nothing : model[ids[j]]::Sheep
end



# Sheep and wolves have separate `eat!` functions. If a sheep eats grass, it will acquire
# additional energy and the grass will not be available for consumption until regrowth time
# has elapsed. If a wolf eats a sheep, the sheep dies and the wolf acquires more energy.
function eat!(sheep::Sheep, model)
    if model.fully_grown[sheep.pos...]
        sheep.energy += sheep.Δenergy
        model.fully_grown[sheep.pos...] = false
    end
    return
end

function eat!(wolf::Wolf, sheep::Sheep, model)
    kill_agent!(sheep, model)
    wolf.energy += wolf.Δenergy
    return
end

# Sheep and wolves share a common reproduction method. Reproduction has a cost of 1/2 the
# current energy level of the parent. The offspring is an exact copy of the parent, with
# exception of `id`.
function reproduce!(agent::A, model) where {A}
    agent.energy /= 2
    id = nextid(model)
    offspring = A(id, agent.pos, agent.energy, agent.reproduction_prob, agent.Δenergy)
    add_agent_pos!(offspring, model)
    return
end

# The behavior of grass function differently. If it is fully grown, it is consumable.
# Otherwise, it cannot be consumed until it regrows after a delay specified by
# `regrowth_time`. The dynamics of the grass is our `model_step!` function.
function grass_step!(model)
    @inbounds for p in positions(model) # we don't have to enable bound checking
        if !(model.fully_grown[p...])
            if model.countdown[p...] ≤ 0
                model.fully_grown[p...] = true
                model.countdown[p...] = model.regrowth_time
            else
                model.countdown[p...] -= 1
            end
        end
    end
end

sheep(a) = a isa Sheep
wolf(a) = a isa Wolf
count_grass(model) = count(model.fully_grown)
# Run simulation:

stable_params = (;
    n_sheep = 140,
    n_wolves = 20,
    dims = (30, 30),
    Δenergy_sheep = 5,
    sheep_reproduce = 0.31,
    wolf_reproduce = 0.06,
    Δenergy_wolf = 30,
)


@benchmark (model = initialize_model(;$stable_params...);
            run!(model, sheepwolf_step!, grass_step!, 2000)) samples=100
```

</details>

which resulted in, before and after this PR:

```julia
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  30.664 ms … 36.818 ms  ┊ GC (min … max): 0.00% … 13.22%
 Time  (median):     31.042 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   31.553 ms ±  1.442 ms  ┊ GC (mean ± σ):  1.05% ±  3.34%

    █▃                                                         
  ▅▇███▄▄▃▄▃▄▁▃▄▃▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▃▁▃▁▁▃ ▃
  30.7 ms         Histogram: frequency by time        36.7 ms <

BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  36.169 ms … 42.524 ms  ┊ GC (min … max): 0.00% … 5.78%
 Time  (median):     38.592 ms              ┊ GC (median):    5.64%
 Time  (mean ± σ):   38.078 ms ±  1.592 ms  ┊ GC (mean ± σ):  3.19% ± 3.05%

  ▂▂▆█ ▅                                ▂                      
  ████▅██▁█▄▁▁▁▄▁▁▁▁▄▁▁▄▁▁▁▁▁▁▄▄▅▅▇▇█▇█▇█▄▇▁▁▁▁▁▁▄▅▄▅▇▅▄▁▅▄▁▄ ▄
  36.2 ms         Histogram: frequency by time        40.9 ms <
```

so it is actually a bit slower, but also really strangely if I change the scheduler to `fastest` it fails with the modifications in this PR, don't know if there is some bug in OrderedCollections since it seems strange since this PR passes all tests here, and I also runned the benchmark of `schedulers.jl` and It worked.

So I'm inclined to close this PR

Datseris:
If insertion and deletion isn't faster then I think we can stay with Dict. For models with only insertions we have the Vector version anyways.

Tortar:
I agree!

PR #758, opened 2023-03-06
Check if position is already occupied

### LINKED ISSUE #757: Unbounded increase of agents in a GridSpaceSingle model
Below i am attaching an implementation of a spatial rock paper scissors model we used in a tutorial about Agents.jl. The model is a grid space single, yet agents increase in numbers infinitely, even though we are only using the basic Agents.jl API. Something is wrong somewhere and there is a leakage. One of hte API functions probably needs an additional check...

```julia
using Agents, Random

@agent Strategy GridAgent{2} begin
    #:rock, :paper or :scissors
    type::Symbol
end

function initialize_model(;
        dims = (60, 60),
        n_rock = 800,
        n_paper = n_rock,
        n_scissors = n_rock,
        selection_rate = 1.0,
        reproduction_rate = 1.0,
        seed = 23182,
    )

    rng = MersenneTwister(seed)
    space = GridSpaceSingle(dims, periodic = true, metric = :manhattan)
    #model properties
    properties = (
        selection_rate = selection_rate,
        reproduction_rate = reproduction_rate,
    )
    model = AgentBasedModel(Strategy, space;
        properties, rng, scheduler = Schedulers.randomly
    )
    # Add agents to the model to a random but unoccupied position
    # using the `add_agent_single!` function
    for (strategy, N) in zip((:rock, :paper, :scissors), (n_rock, n_paper, n_scissors))
        for _ in 1:N
            add_agent_single!(model, strategy)
        end
    end
    return model
end

# Time evolution rules:
function strategy_step!(strategy, model)
    # propensities
    a1 = model.selection_rate * nagents(model)
    a2 = model.reproduction_rate * nagents(model)
    # total propensities
    a0 = a1 + a2 
    p = rand(abmrng(model))
    if 0 <= p <= a1/a0
        selection!(strategy, model)
    elseif a1/a0 <= p < (a1+a2)/a0
        reproduce!(strategy, model)
    end
end

function selection!(strategy, model)
    contender = random_nearby_agent(strategy, model)
    if !isnothing(contender)
        if strategy.type == :rock && contender.type == :scissors
            kill_agent!(contender, model)
        elseif strategy.type == :scissors && contender.type == :paper
            kill_agent!(contender, model)
        elseif strategy.type == :paper && contender.type == :rock
            kill_agent!(contender, model)
        end
    end
    return
end

function reproduce!(strategy, model)
    for pos in nearby_positions(strategy, model)
        # empty position to put an offspring in
        if isempty(pos, model)
            add_agent!(strategy.pos, model, strategy.type)
        end
        break
    end
    return
end

function model_step!(model)
    if nagents(model) > length(positions(model))
        error("model has more agents than positions")
    end
end

model = initialize_model()

step!(model, strategy_step!, model_step!)
model

step!(model, strategy_step!, model_step!, 100)
model
```

Tortar:
FIx #757 but don't know if something else can be considered better than this, but I don't think so, and if the error message is a good one :D

### DIFF
--- a/src/spaces/grid_single.jl
+++ b/src/spaces/grid_single.jl
@@ -39,8 +39,9 @@ end
 function add_agent_to_space!(a::A, model::ABM{<:GridSpaceSingle,A}) where {A<:AbstractAgent}
     pos = a.pos
     !isempty(pos, model) && error("Attempt to add Agent $(a.id) to an occupied position")
-    model.space.stored_ids[pos...] = a.id  
+    model.space.stored_ids[pos...] = a.id
     return a
+end
 
 function remove_agent_from_space!(a::A, model::ABM{<:GridSpaceSingle,A}) where {A<:AbstractAgent}
     model.space.stored_ids[a.pos...] = 0
--- a/src/spaces/grid_single.jl
+++ b/src/spaces/grid_single.jl
@@ -38,7 +38,7 @@ end
 # Implementation of space API
 function add_agent_to_space!(a::A, model::ABM{<:GridSpaceSingle,A}) where {A<:AbstractAgent}
     pos = a.pos
-    !isempty(pos, model) && error("Attempted to add agent $(a) to occupied position $(pos)")
+    !isempty(pos, model) && error("Cannot add agent $(a) to occupied position $(pos)")
     model.space.stored_ids[pos...] = a.id
     return a
 end
--- a/src/spaces/grid_single.jl
+++ b/src/spaces/grid_single.jl
@@ -37,9 +37,11 @@ function GridSpaceSingle(d::NTuple{D,Int}; periodic = true, metric = :chebyshev)
 end
 # Implementation of space API
 function add_agent_to_space!(a::A, model::ABM{<:GridSpaceSingle,A}) where {A<:AbstractAgent}
-    model.space.stored_ids[a.pos...] = a.id
+    pos = a.pos
+    !isempty(pos, model) && error("Attempt to add Agent $(a.id) to an occupied position")
+    model.space.stored_ids[pos...] = a.id  
     return a
-end
+
 function remove_agent_from_space!(a::A, model::ABM{<:GridSpaceSingle,A}) where {A<:AbstractAgent}
     model.space.stored_ids[a.pos...] = 0
     return a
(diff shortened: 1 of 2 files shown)


jacobusmmsmit:
There are models where multiple agents can occupy a space. Given that `add_agent_to_space!` is a super low-level function, I don't think it should be the place where an empty check is made.

jacobusmmsmit:
Or perhaps I'm mistaken because `remove_agent_from_space!` has the line `model.space.stored_ids[a.pos...] = 0`, implying that *all* agents in a space are removed when one is, which doesn't really make sense. Would love to hear others thoughts...

Tortar:
I think there are no problems about that since this method is dispatched for just models with a spaceGridsingle, not for all models

jacobusmmsmit:
Oops, I missed that. LGTM.

src/spaces/grid_single.jl
```diff
@@ -37,9 +37,12 @@ function GridSpaceSingle(d::NTuple{D,Int}; periodic = true, metric = :chebyshev)
 end
 # Implementation of space API
 function add_agent_to_space!(a::A, model::ABM{<:GridSpaceSingle,A}) where {A<:AbstractAgent}
-    model.space.stored_ids[a.pos...] = a.id
+    pos = a.pos
+    !isempty(pos, model) && error("Attempt to add agent $(a) to occupied position $(pos)")
```
jacobusmmsmit, line 41:
```suggestion
    !isempty(pos, model) && error("Attempted to add agent $(a) to occupied position $(pos)")
```

src/spaces/grid_single.jl
```diff
@@ -37,9 +37,12 @@ function GridSpaceSingle(d::NTuple{D,Int}; periodic = true, metric = :chebyshev)
 end
 # Implementation of space API
 function add_agent_to_space!(a::A, model::ABM{<:GridSpaceSingle,A}) where {A<:AbstractAgent}
-    model.space.stored_ids[a.pos...] = a.id
+    pos = a.pos
+    !isempty(pos, model) && error("Attempted to add agent $(a) to occupied position $(pos)")
```
jacobusmmsmit, line 41:
```suggestion
    !isempty(pos, model) && error("Cannot add agent $(a) to occupied position $(pos)")
```

jacobusmmsmit: It seems that all of the `error` messages in `Agents` are in the present tense and say "Cannot" or something similar. This is really minor bikeshedding, so feel free to ignore.

Tortar: This way seems better to me so I'm okay with the change!

PR #762, opened 2023-03-07
Implementation of the function random_nearby_position and its tests

MA-Ramirez:


Tortar:
Hy @MA-Ramirez, good that you already implemented some tests! But there are actually two problems:

- the radius should be arbitrary;
- this implementation is not efficient, you are materializing all nearby positions to choose just one random one, you should look at `random_nearby_agent` to implement this function efficiently

src/core/space_interaction_API.jl
```diff
@@ -371,3 +372,41 @@ function random_nearby_agent(a, model, r = 1; kwargs...)
...
+"""
+    random_nearby_position(position, model::ABM, r=1; kwargs...) → position
+
+Return a random position near the given `position`. Return `nothing` if the space doesn't allow for nearby positions.
+
+The function uses the same Reservoir Sampling algorithm as [`random_nearby_id`](@ref). 
+
+The value of the argument `r` and possible keywords operate identically to [`nearby_positions`](@ref).
+
+"""
```
Tortar, line 385:
```suggestion
"""
    random_nearby_position(position, model::ABM, r=1; kwargs...) → position

Return a random position near the given `position`. Return `nothing` if the space doesn't allow for nearby positions.

The value of the argument `r` and possible keywords operate identically to [`nearby_positions`](@ref).
"""
```

I wouldn't mentioned the Reservoir Sampling in the docstring, I think it's not relevant to the users and it isn't done in another similar docstrings, as a comment inside the function instead it seems good to me as you already did.

Tortar:
For the rest, I will let someone else judge the implementation from this point on, but I think it's fine!

src/core/space_interaction_API.jl
```diff
@@ -22,6 +22,7 @@ export move_agent!,
     nearby_agents,
     random_nearby_id,
     random_nearby_agent,
+    random_nearby_position,
```
Datseris, line 25: This is an exported function, so it should also be added into the api.md file in the docs. Similarly, it should have a changelog entry in CHANGELOG.md.

Datseris (approved):
Seems good to me. Just be sure that you always have changelog entries and documentation export for new functions.

src/core/space_interaction_API.jl
```diff
@@ -361,7 +362,7 @@ end
 """
     random_nearby_agent(agent, model::ABM, r = 1; kwargs...) → agent
 
-Return the a random agent near the position of the given `agent`. Return `nothing` if no agent
+Return a random agent near the position of the given `agent`. Return `nothing` if no agent
```
jacobusmmsmit, line 365:
```suggestion
Return a random agent near the position of the given `agent` or `nothing` if no agent
```

src/core/space_interaction_API.jl
```diff
@@ -371,3 +372,38 @@ function random_nearby_agent(a, model, r = 1; kwargs...)
...
+
+"""
+    random_nearby_position(position, model::ABM, r=1; kwargs...) → position
+
+Return a random position near the given `position`. Return `nothing` if the space doesn't allow for nearby positions.
+
+The value of the argument `r` and possible keywords operate identically to [`nearby_positions`](@ref).
+"""
+function random_nearby_position(pos, model, r=1; kwargs...)
+    # Uses the same Reservoir Sampling algorithm than nearby_ids
```
jacobusmmsmit, line 384:
```suggestion
    # Uses the same Reservoir Sampling algorithm as nearby_ids
```
I feel like such a pendant right now

src/core/space_interaction_API.jl
```diff
@@ -371,3 +372,38 @@ function random_nearby_agent(a, model, r = 1; kwargs...)
...
+    iter = nearby_positions(pos, model, r; kwargs...)
+
+    res = iterate(iter)
+    isnothing(res) && return    # `iterate` returns `nothing` when it ends
```
jacobusmmsmit, line 388:
```suggestion
    isnothing(res) && return nothing    # `iterate` returns `nothing` when it ends
```
Better explicit returns than implicit ones

jacobusmmsmit:
Lgtm, just some minor grammatical and clarity comments :)

src/core/space_interaction_API.jl
MA-Ramirez: Thanks for the information
I have added the the function to these files. Nonetheless, I don't know how the addition of a new function changes the number version of the package

Datseris (approved):
Awesome thanks!

Issue #754, opened 2023-03-05
randomwalk! error on Predator-Prey example


glanzkaiser:
I try to run the example on Julia REPL version 1.8.5 of the
 https://juliadynamics.github.io/Agents.jl/stable/examples/predator_prey/

But, there is an error of 
**no method matching randomwalk!**

![Capture d’écran_2023-03-05_13-26-43](https://user-images.githubusercontent.com/72222484/222945478-244e7f5f-0168-4b7a-9933-82b18fb7aca5.png)

this is my code / MWE:

```

# Making the model

using Agents, Random

@agent Sheep GridAgent{2} begin
    energy::Float64
    reproduction_prob::Float64
    Δenergy::Float64
end

@agent Wolf GridAgent{2} begin
    energy::Float64
    reproduction_prob::Float64
    Δenergy::Float64
end

function initialize_model(;
        n_sheep = 100,
        n_wolves = 50,
        dims = (20, 20),
        regrowth_time = 30,
        Δenergy_sheep = 4,
        Δenergy_wolf = 20,
        sheep_reproduce = 0.04,
        wolf_reproduce = 0.05,
        seed = 23182,
    )

    rng = MersenneTwister(seed)
    space = GridSpace(dims, periodic = true)
    # Model properties contain the grass as two arrays: whether it is fully grown
    # and the time to regrow. Also have static parameter `regrowth_time`.
    # Notice how the properties are a `NamedTuple` to ensure type stability.
    properties = (
        fully_grown = falses(dims),
        countdown = zeros(Int, dims),
        regrowth_time = regrowth_time,
    )
    model = ABM(Union{Sheep, Wolf}, space;
        properties, rng, scheduler = Schedulers.randomly, warn = false
    )
    # Add agents
    for _ in 1:n_sheep
        energy = rand(model.rng, 1:(Δenergy_sheep*2)) - 1
        add_agent!(Sheep, model, energy, sheep_reproduce, Δenergy_sheep)
    end
    for _ in 1:n_wolves
        energy = rand(model.rng, 1:(Δenergy_wolf*2)) - 1
        add_agent!(Wolf, model, energy, wolf_reproduce, Δenergy_wolf)
    end
    # Add grass with random initial growth
    for p in positions(model)
        fully_grown = rand(model.rng, Bool)
        countdown = fully_grown ? regrowth_time : rand(model.rng, 1:regrowth_time) - 1
        model.countdown[p...] = countdown
[...]

PR #763, opened 2023-03-07
Set default value of `r` to `r=1` for `randomwalk!`

MA-Ramirez:
Pull request regarding issue #754

### DIFF
--- a/src/spaces/walk.jl
+++ b/src/spaces/walk.jl
@@ -80,7 +80,7 @@ end
 
 
 """
-    randomwalk!(agent, model::ABM{<:AbstractGridSpace}, r; kwargs...)
+    randomwalk!(agent, model::ABM{<:AbstractGridSpace}, r = 1; kwargs...)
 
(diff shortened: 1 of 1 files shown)


Datseris (changes requested):
Thanks, but have a look at the continuous code a bit closer. You will see that for continuous space the implementation is different. If `r` is not given, the velocity is used instead. Not `1`. So, this PR should not change the continuous space version.

Datseris (approved):

Tortar:
the types should be kept, otherwise this can cause performance issues, I added them

Datseris:
adding this type decleration does not affect perforamance. Julia will anyways compile a dedicated version for each input type whether it is declared or not. The only thing this does is throw a method error if someone tried to use the function with a non-real input argument, which I guess is useful.

Tortar:
ops @Datseris, at least there is the motivation you mentioned :-)

Tortar (approved):

PR #766, opened 2023-03-08
Rename functions like `genocide!`

### LINKED ISSUE #764: Renaming the `genocide!` function
Hello, 

I propose changing the name of the function `genocide!`.
This is because the word genocide is linked to heinous events in history and in our current times.
Therefore, to not normalise the word's meaning and the human suffering it signifies, I propose changing the name to something like `remove_all!`.

Similarly, the function `kill_agent!`could be renamed `remove_agent!`.

MA-Ramirez:
Closes #764  

genocide! --> remove_all!
kill_agent! --> remove_agent!
Unkillable --> Unremovable


jacobusmmsmit:
Thank you for raising this, it has a strong +1 from me. Firstly for the obvious reason of how (euphemistically) _unnecessary_ the current names are, but secondly the current wording goes against the naming convention that functions that do the opposite things should have opposite names. Hence, I believe that `remove` is the correct opposite of `add` and not `kill`.

examples/optim.jl
```diff
@@ -135,7 +135,7 @@
 # Unfortunately we've not given the optimiser information we probably needed to.
 # Notice that the death rate is 96%, with reinfection quite low.
 # When all the infected individuals die, infection doesn't transmit - the optimiser has
-# managed to reduce the infection rate by killing the infected.
+# managed to reduce the infection rate by removing the infected.
```
jacobusmmsmit, line 138:
```suggestion
# managed to reduce the infection rate by "removing" the infected.
```
Makes it sound more ominous and closer to the original joke.

src/spaces/graph.jl
```diff
@@ -130,15 +130,15 @@ export rem_node!, add_node!, rem_vertex!, add_vertex!, add_edge!, rem_edge!
 
 """
      rem_node!(model::ABM{<: GraphSpace}, n::Int)
-Remove node (i.e. position) `n` from the model's graph. All agents in that node are killed.
+Remove node (i.e. position) `n` from the model's graph. All agents in that node are removed.
```
jacobusmmsmit, line 133:
```suggestion
Remove node (i.e. position) `n` from the model's graph. All agents in that node are removed from the model.
```

src/spaces/graph.jl
```diff
@@ -161,7 +161,7 @@ end
 
 """
     rem_vertex!(model::ABM{<:GraphSpace}, n::Int)
-Remove node (i.e. position) `n` from the model's graph. All agents in that node are killed.
+Remove node (i.e. position) `n` from the model's graph. All agents in that node are removed.
```
jacobusmmsmit, line 164:
```suggestion
Remove node (i.e. position) `n` from the model's graph. All agents in that node are removed from the model.
```

jacobusmmsmit (changes requested):
There are a few mentions of kill left in the docs.

MA-Ramirez:
@jacobusmmsmit thank you for the additional modifications. I thought all the files in docs were automatically generated by the scripts, so I didn't check them. 
Is there something I can do to allow the push to my branch? Or should I make the changes in the branch myself?

jacobusmmsmit:
Honestly, I don't have the definitive answer to either of your comments.
Wrt docs, your comment makes sense and if that's the case then I guess it LGTM.
Wrt git, I don't think you need to do anything and the problem seems to be on my end. Maybe someone else can suggest why I can't push to her branch?

Datseris (changes requested):
fantastic, thank a lot. However, if I were to merge this as it is right now, it would break every single Agents.jl code out there because we are renaming functions that thousands of users use. You need to go to the src/deprecations.jl file and add deprecations for all changed API names. For example,

```
@deprecate UnkillableABM UnremovableABM
```
and so on and so forth. Also, please add a changelog entry for every API name that was changed.

Changes in the wording of docstrings or documentation pages do not need to be mentioned. Only changes in API names, i.e., names explicitly participating in `export` statements.

jacobusmmsmit (approved):
LGTM

jacobusmmsmit:
Wait hang on, shouldn't these changes go in v5.9 not v5.8 as it's already been released?

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 - A memory bug about offsets calculation has been solved; besides, the `calculate_offsets` function has been sped-up by a significant amount.
 
 # v5.8
+- New function `remove_all!` replaces `genocide!` (now deprecated).
```
jacobusmmsmit, line 6: Put these new changes in `main`

jacobusmmsmit (changes requested):

CHANGELOG.md
jacobusmmsmit: I hope my "request changes" doesn't seem too much like I'm bossing you around @MA-Ramirez! I don't like telling people what to do when I'm not super confident that it's the right thing, but I'm pretty sure `main` is the right place for these changes :')

Datseris (approved):

Datseris:
image

xlejx-rodsxn:
Ah, weird. how did I end up in that version? anyway, thanks for taking care of this issue and the time invested in explaining.

PR #768, opened 2023-03-08
Fix bug in `agents_in_position` + tests

### LINKED ISSUE #767: Bug in `agents_in_position`
Hello, 

**Describe the bug**
It seems that `agents_in_position` is not working properly.
Instead of returning the agents in the position, it returns Base.Generator{...}

**Minimal Working Example**
```Julia
using Agents
using Random: Xoshiro

space = GridSpace((25, 25))

@agent Rabbit GridAgent{2} begin
    energy::Float64
    Δenergy::Float64
    reproduction_prob::Float64
end

function init_rabbits(; 
        total_agents = 100, 
        griddims = (20, 20),
        energy = 10.0, 
        Δenergy = 5.0, 
        reproduction_prob = 0.1,
        seed = 1234,
    )
    
    space = GridSpace(griddims)
    
    rng = Xoshiro(seed)
    
    model = AgentBasedModel(Rabbit, space; rng)
    
    for n in 1:total_agents
        add_agent!(model, energy, Δenergy, reproduction_prob)
    end
    return model
end

model = init_rabbits()

a1 = agents_in_position(model[1],model)
```

**Agents.jl version**
Agents v5.8.0

**Solution to the bug**
I looked into the code. It seems that the problem arises because a `Generator` instead of an array outputs the desired `agent`(s).
Also, there was no test for this function.
Therefore, I created a PR to fix the bug with its corresponding tests.

MA-Ramirez:
Fixes #767

### DIFF
--- a/src/spaces/discrete.jl
+++ b/src/spaces/discrete.jl
@@ -57,14 +57,16 @@ ids_in_position(agent::A, model) where {A<:AbstractAgent} =
     ids_in_position(agent.pos, model)
 
 """
-    agents_in_position(position, model::ABM{<:DiscreteSpace})
-    agents_in_position(agent, model::ABM{<:DiscreteSpace})
+    agents_in_position(position, model::ABM{<:DiscreteSpace}) → Vector{agent}
+    agents_in_position(agent, model::ABM{<:DiscreteSpace}) → Vector{agent}
 
-Return the agents in the position corresponding to `position` or position of `agent`.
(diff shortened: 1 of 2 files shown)


Datseris:
This changes the non allocating version of the function to an allocating one which is less performant. This wont be merged. However, you can re-adjust the PR to not change the source code but still do the tests you do. You can do `agents = collect(agents_in_position(...))` and then `agents[1]` if you want to index the output of `agents_in_position`.

Datseris (approved):

Issue #770, opened 2023-03-09
New `empty_nearby_positions` function


MA-Ramirez:
Hello, 

I propose to add a function called `empty_nearby_positions`
While developing the spatial Rock-Paper-Scissors example I realised this could be a helpful predefined function
I created a PR, in case you think it is a nice function to have

Cheers!

PR #771, opened 2023-03-09
Add `empty_nearby_positions` + tests

MA-Ramirez:
Implements #770

### DIFF
--- a/src/spaces/discrete.jl
+++ b/src/spaces/discrete.jl
@@ -127,9 +127,13 @@ The value of r and possible keywords operate identically to `nearby_positions`.
 
 This function only exists for discrete spaces with a finite amount of positions.
 """
-empty_nearby_positions(agent::A, model, r = 1) where {A<:AbstractAgent} = 
-    empty_nearby_positions(agent.pos, model, r)
-empty_nearby_positions(pos, model, r = 1) = filter(x -> isempty(x, model), collect(nearby_positions(pos, model, r)))
+
+function empty_nearby_positions(agent::A, model, r = 1) where {A<:AbstractAgent}
+    return empty_nearby_positions(agent.pos, model, r)
+end
+function empty_nearby_positions(pos, model, r = 1)
+    return Iterators.filter(x -> isempty(x, model), nearby_positions(pos, model, r))
+end
 
 
 #######################################################################################
--- a/src/spaces/discrete.jl
+++ b/src/spaces/discrete.jl
@@ -118,21 +118,23 @@ function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
 end
 
 """
-    empty_nearby_positions(position, model::ABM{<:DiscreteSpace}, r = 1; kwargs)
-    empty_nearby_positions(agent, model::ABM{<:DiscreteSpace}, r = 1; kwargs)
+    empty_nearby_positions(position, model::ABM{<:DiscreteSpace}, r = 1; kwargs...)
+    empty_nearby_positions(agent, model::ABM{<:DiscreteSpace}, r = 1; kwargs...)
 
-Return an iterable of all empty positions within "radius" `r` of the given `position` or `agent`.
+Return an iterable of all empty positions within "radius" `r` of the given `position`
+or the position of an `agent::AbstractAgent`.
 
-The value of r and possible keywords operate identically to `nearby_positions`.
+The value of `r` and possible keywords operate identically to [`nearby_positions`](@ref).
 
 This function only exists for discrete spaces with a finite amount of positions.
 """
-
-function empty_nearby_positions(agent::A, model, r = 1) where {A<:AbstractAgent}
-    return empty_nearby_positions(agent.pos, model, r)
+function empty_nearby_positions(agent::AbstractAgent, model, r = 1; kwargs...)
+    return empty_nearby_positions(agent.pos, model, r; kwargs...)
 end
-function empty_nearby_positions(pos, model, r = 1)
-    return Iterators.filter(Base.Fix2(isempty, model), nearby_positions(pos, model, r))
+function empty_nearby_positions(pos, model, r = 1; kwargs...)
+    return Iterators.filter(
+        Base.Fix2(isempty, model), nearby_positions(pos, model, r; kwargs...)
+    )
 end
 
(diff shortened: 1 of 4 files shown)


src/spaces/discrete.jl
```diff
@@ -117,6 +117,21 @@ function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
...
+    empty_nearby_positions(position, model::ABM{<:DiscreteSpace}, r = 1; kwargs)
+    empty_nearby_positions(agent, model::ABM{<:DiscreteSpace}, r = 1; kwargs)
+
+Return an iterable of all empty positions within "radius" `r` of the given `position` or `agent`.
+
+The value of r and possible keywords operate identically to `nearby_positions`.
+
+This function only exists for discrete spaces with a finite amount of positions.
+"""
+empty_nearby_positions(agent::A, model, r = 1) where {A<:AbstractAgent} =
```
jacobusmmsmit, line 130: Please expand this to a full `function end` definition and probably the next one too. It's a nice one liner, but if it's on two lines then it's nicer to have it in the full function definition syntax.

src/spaces/discrete.jl
```diff
@@ -117,6 +117,21 @@ function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
...
+
+Return an iterable of all empty positions within "radius" `r` of the given `position` or `agent`.
+
+The value of r and possible keywords operate identically to `nearby_positions`.
+
+This function only exists for discrete spaces with a finite amount of positions.
+"""
+empty_nearby_positions(agent::A, model, r = 1) where {A<:AbstractAgent} = 
+    empty_nearby_positions(agent.pos, model, r)
+empty_nearby_positions(pos, model, r = 1) = filter(x -> isempty(x, model), collect(nearby_positions(pos, model, r)))
```
jacobusmmsmit, line 132: Is this `collect` needed?

Datseris: `collect` is not needed. However you need to replace `filter` with `Iterators.filter` to allow filtering an arbitrary iterable like the output of `nearby_positions`.

test/grid_space_tests.jl
```diff
@@ -63,6 +63,11 @@ using StableRNGs
         random_positions = positions(model, :random)
         @test all(n ∈ pos_map for n in random_positions)
```
jacobusmmsmit, line 64:
```suggestion
        @test all(n in pos_map for n in random_positions)
```
For consistency and simplicity, prefer `in` to `∈`.

test/grid_space_tests.jl
```diff
@@ -63,6 +63,11 @@ using StableRNGs
...
 
+        #test empty_nearby_positions
+        empty_near_positions =  empty_nearby_positions(model[5], model)
+        @test all(n ∈ nearby_positions(model[5], model) for n in empty_near_positions)
```
jacobusmmsmit, line 68:
```suggestion
        @test all(n in nearby_positions(model[5], model) for n in empty_near_positions)
```

test/grid_space_tests.jl
```diff
@@ -63,6 +63,11 @@ using StableRNGs
...
+        @test all(n ∈ [(1, 3), (2, 3)] for n in empty_near_positions)
```
jacobusmmsmit, line 69:
```suggestion
        @test all(n in [(1, 3), (2, 3)] for n in empty_near_positions)
```

src/spaces/discrete.jl
```diff
@@ -117,6 +117,25 @@ function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
...
+The value of r and possible keywords operate identically to `nearby_positions`.
+
+This function only exists for discrete spaces with a finite amount of positions.
+"""
+
+function empty_nearby_positions(agent::A, model, r = 1) where {A<:AbstractAgent}
+    return empty_nearby_positions(agent.pos, model, r)
+end
+function empty_nearby_positions(pos, model, r = 1)
+    return Iterators.filter(x -> isempty(x, model), nearby_positions(pos, model, r))
```
jacobusmmsmit, line 135:
```suggestion
    return Iterators.filter(Base.Fix2(isempty, model), nearby_positions(pos, model, r))
```
See https://github.com/SciML/SciMLStyle#closures-should-be-avoided-whenever-possible

src/spaces/discrete.jl
```diff
@@ -117,6 +117,25 @@ function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
     end
 end
 
+"""
+    empty_nearby_positions(position, model::ABM{<:DiscreteSpace}, r = 1; kwargs)
```
jacobusmmsmit, line 121: The type signature says `kwargs` but there aren't any kwargs.

src/spaces/discrete.jl
```diff
@@ -117,6 +117,25 @@ function random_empty(model::ABM{<:DiscreteSpace}, cutoff = 0.998)
...
+The value of r and possible keywords operate identically to `nearby_positions`.
+
+This function only exists for discrete spaces with a finite amount of positions.
+"""
+
+function empty_nearby_positions(agent::A, model, r = 1) where {A<:AbstractAgent}
+    return empty_nearby_positions(agent.pos, model, r)
+end
+function empty_nearby_positions(pos, model, r = 1)
+    return Iterators.filter(Base.Fix2(isempty, model), nearby_positions(pos, model, r))
```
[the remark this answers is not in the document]
jacobusmmsmit: I don't have a strong argument beyond avoid closures because you never know when one will cause a type instability.

Beyond that, anonymous function syntax is simpler but has more uses so can mean more things. `Fix2` adds a little complexity but only has one use i.e. binding an argument to a function so what we are doing is clearer.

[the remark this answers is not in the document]
jacobusmmsmit: That said, feel free to revert. No hard feelings as this is a very unimportant change in the grand scheme, especially if we're not following SciML style guide.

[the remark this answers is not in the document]
Datseris: I did some research on Fix2, and it is indeed better!

Datseris (approved):

PR #772, opened 2023-03-09
Use UnkillableABM for some models in the model folder

Tortar:
Models in this folder aren't still using `UnkillableABM`, but changing this would create conflicts with #766, but they should be really easy to solve anyway. Or we can even wait to change this after #766 is merged.

### DIFF
--- a/src/models/flocking.jl
+++ b/src/models/flocking.jl
@@ -37,7 +37,7 @@ function flocking(;
     spacing = visual_distance / 1.5,
 )
     space2d = ContinuousSpace(extent; spacing)
-    model = ABM(Bird, space2d, scheduler = Schedulers.Randomly())
+    model = UnkillableABM(Bird, space2d, scheduler = Schedulers.Randomly())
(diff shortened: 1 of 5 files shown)


Tortar:
ok, I was expecting that all tests would have passed but this is not the case, so this PR is officially a work in progress, probably UnkillableABM doesn't still support all the features a StandardABM supports, so I think this will require a bit of time to work out

Datseris:
this PR unveals another problem: these models are provided purely as convenience. They should not be used in the test suite. Everything in the test suite must create models/agents/stepping functions from scratch.

Tortar:
Now all tests pass, but I think I will try to rewrite the incriminated test suites so that they have the models inside the test file as you suggested :-) I think that we can still use some smaller schelling and flocking models anyway right?

Tortar:
Hopefully this is a better way to test the checkpointed models even if I think some more revisions of tests is still needed :-)

PR #776, opened 2023-03-11
Not check presence of id for UnremovableABM models in step!

Tortar:
The presence of the id can be spared to be checked when using an UnremovableABM. I also had to change the definition of the aliases in `model_concrete.jl` because previously referring to different ABM implementation wasn't possible, but I'm not sure if this is the right way to do it even if it works. Consider also that for an UnremovableABM checking if an id is present is much more costly than for a StandardABM since the vector has to be scanned.


Tortar:
ops, actually `allids(model)` returns a Base.OneTo object so the checking is fast, so this will not give much of a speed-up

Datseris:
yes

Datseris (approved):

PR #777, opened 2023-03-12
simplify code creating resorvoir sampling function

Tortar:
Simplified the code eliminating the repetition of the reservoir sampling function in the two methods


Datseris (approved):
ah thankis, i wanted to do this as well

PR #779, opened 2023-03-13
Use aliases in model accessing api

Tortar:
Used the aliases instead of the specifications in `SingleContainerABM` so that methods are easier to read, I also made some hopefully good modification to error messages


Datseris:
fantastic clarification

Datseris (approved):

Issue #773, opened 2023-03-10
Bug in serialization for UnkillableABM


Tortar:
Working on #772 I probably found a bug which precluded the changes in that PR. The problem is that the `jld2_integration.jl` file wasn't adapted for UnkillableABM models, I think that the problem is in this function

```julia
function from_serializable(t::SerializableABM{S,A}; kwargs...) where {S,A}
    abm = ABM(
        A,
        from_serializable(t.space; kwargs...);
        scheduler = get(kwargs, :scheduler, Schedulers.fastest),
        properties = from_serializable(t.properties; kwargs...),
        rng = t.rng,
        warn = get(kwargs, :warn, true)
    )
    abm.maxid[] = t.maxid

    for a in t.agents
        add_agent_pos!(a, abm)
    end
    return abm
end
```

as you can see it mutates **any** checkpointed ABM to a StandardABM.

Tortar:
Solved in #780

PR #780, opened 2023-03-14
Make model serialization more general

Tortar:
This solves #773 since now the serialization includes the type of the agents_container so that we can reinstantiate the model with the correct container in `from_serializable`, it should now be possible to use it with all models implementations

### DIFF
--- a/src/simulations/collect.jl
+++ b/src/simulations/collect.jl
@@ -219,7 +219,7 @@ function single_agent_types!(
     model::ABM,
     properties::AbstractArray,
 )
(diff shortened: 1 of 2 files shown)


Tortar:
Not sure why a macos test failed :Q

Tortar:
Also fixed another bug in collect.jl related to #772 , now all the bugs found that to that PR should be solved, this one was also related to the fact of assuming a Dict as container for agents

Datseris (approved):

PR #783, opened 2023-03-17
Improve the speed of GridSpaceIdIterator

Tortar:
Found some little optimizations for the iterator, running some benchmark it is more or less 2x faster to iterate over results now!

<details>

```julia
using Agents, BenchmarkTools

@agent A GridAgent{2} begin
end

function periodic_model(; numagents = 1000, griddims = (20, 20))
    space = GridSpace(griddims, periodic = true)
    model = ABM(A, space)
    for _ in 1:numagents
        add_agent!(A, model)
    end
    return model
end

function aperiodic_model(; numagents = 1000, griddims = (20, 20))
    space = GridSpace(griddims, periodic = false)
    model = ABM(A, space)
    for _ in 1:numagents
        add_agent!(A, model)
    end
    return model
end

@benchmark collect(nearby_ids($(10,10), model, 1)) setup=(model=periodic_model())
@benchmark collect(nearby_ids($(10,10), model, 1)) setup=(model=aperiodic_model())
@benchmark collect(nearby_ids($(10,10), model, 10)) setup=(model=periodic_model())
@benchmark collect(nearby_ids($(10,10), model, 10)) setup=(model=aperiodic_model())
```

</details>

These are the results for medians

```julia
radius 1 nonperiodic -> before pr 260.688 ns after pr 189.850 ns
radius 1 periodic -> before pr 669.392 ns after pr 378.476 ns
radius 10 nonperiodic -> before pr 16.481 μs after pr 8.131 μs
radius 10 periodic -> before pr 37.990 μs after pr μs 21.661 μs
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # main
 
+# v5.11
+
 - Iterating in neighborhood searches (with `nearby_ids` and derivative functions) is now about 2 times faster than before.
 
 # v5.10
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # main
 
-- Iterating over ids in a `GridSpace` is now 2 times faster than before.
+- Iterating in neighborhood searches (with `nearby_ids` and derivative functions) is now about 2 times faster than before.
 
 # v5.10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 # v5.11
 
-- Iterating in neighborhood searches (with `nearby_ids` and derivative functions) is now about 2 times faster than before.
+- Iterating in neighborhood searches (with `nearby_ids` and derivative functions) in `GridSpace` and `ContinuousSpace` is now about 2 times faster than before.
 
 # v5.10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # main
 
+- Iterating over ids in a `GridSpace` is now 2 times faster than before.
+
 # v5.10
 
 - `FixedMassABM` is now deprecated and will be removed in future versions. Turns out, there is no performance benefit of using it over `UnremovableABM`. In fact, there is a performance **deficit** in doing so.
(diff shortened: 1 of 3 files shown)


Tortar:
Actually it has some good impact on some example models: the flocking model in the comparison repo it passed fro 16.8 to 12.0 ms!

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # main
 
+- Iterating over ids in a `GridSpace` is now 2 times faster than before.
```
Datseris, line 3: Wait, this improves nearest neighbor searches right? The way it is phrased now is missleading; makes you think it improves iterating over `allids(model)`. You can say "Iterating over ids in neighborhood searches (`nearby_ids` and derivative functions)...".

Tortar: ops, sure, thought it was clear but it wasn't!

src/spaces/grid_multi.jl
```diff
@@ -133,7 +133,9 @@ Base.IteratorSize(::Type{<:GridSpaceIdIterator}) = Base.SizeUnknown()
 # Instructs how to combine two positions. Just to avoid code duplication for periodic
 combine_positions(pos, origin, ::GridSpaceIdIterator{false}) = pos .+ origin
 function combine_positions(pos, origin, iter::GridSpaceIdIterator{true})
-    mod1.(pos .+ origin, iter.space_size)
+    off_pos = pos .+ origin
+    checkbounds(Bool, iter.stored_ids, off_pos...) && return off_pos
```
Datseris, line 137: can we add a documentation comment here? Something like "it is faster to check for bounds and only if they fail call the `mod` function. Because, most positions do not need the `mod` function, and the `mod` function is expensive. So overall this leads to better perfroamnce.

Datseris (approved):
This is a very important change. Up minor version in project./toml and adjust changelog accordingly. this will be tagged immediatelly after merge.

src/spaces/grid_multi.jl
Tortar: Yes, that's something would be useful for someone reading the code!

Tortar:
Hopefully everything should be ok now! :-)

Tortar:
there is anyway something strange here which requires a mention I think:

As you can see the nonperiodic version is more than two times faster than the periodic version, but this is in my opinion a bit strange since as of the code after this PR the two iterators performs almost exactly the same operations (at least when we are inside the grid with all positions), there is only one difference, which one? this line 

`checkbounds(Bool, iter.stored_ids, off_pos...) && return off_pos`

not all this line though, just the `&&`!! All the rest is still present in the nonperiodic case, but is it possible that a  `&&` (or a if which causes a bit more slowdown that's why I used  `&&`) as a so big impact?

Datseris:
Its because the rest of the operations are also very fast, so indeed the branch switche of the `if-else` statement is costly as the rest of teh function. that makes sense. (`&&` here substitutes an if elkse)

PR #784, opened 2023-03-22
Faster random_nearby_position for GridSpaces

Tortar:
This functions are faster and faster than using a reservoir sampling technique, since this way the operations are O(1), benchmark code used 

<details> 

```julia
using Agents, BenchmarkTools

@agent A GridAgent{2} begin
end

function aperiodic_model(; griddims = (100, 100))
    space = GridSpaceSingle(griddims, periodic = false)
    model = ABM(A, space)
    return model
end

function periodic_model(; griddims = (100, 100))
    space = GridSpaceSingle(griddims, periodic = true)
    model = ABM(A, space)
    return model
end

@benchmark random_nearby_position($(1,1), model, 10) setup=(model=periodic_model())
@benchmark random_nearby_position($(50,50), model, 10) setup=(model=periodic_model())
@benchmark random_nearby_position($(1,1), model, 1) setup=(model=aperiodic_model())
@benchmark random_nearby_position($(50,50), model, 1) setup=(model=aperiodic_model())
```

</details> 

with r=1 it is from 2 to 10x faster but clearly almost always 10x, with r=10 it is more than 1000 times faster!

### DIFF
--- a/src/spaces/grid_general.jl
+++ b/src/spaces/grid_general.jl
@@ -148,6 +148,24 @@ function nearby_positions(
     return (mod1.(n .+ pos, space_size) for n in nindices)
 end
 
+function random_nearby_position(pos::ValidPos, model::ABM{<:AbstractGridSpace{D,false}}, r=1; kwargs...) where {D}
+    nindices = offsets_within_radius_no_0(model.space, r)
+    stored_ids = model.space.stored_ids
+    rng = abmrng(model)
+    while true
+        chosen_offset = rand(rng, nindices)
+        chosen_pos = pos .+ chosen_offset
+        checkbounds(Bool, stored_ids, chosen_pos...) && return chosen_pos
+    end
+end
+
+function random_nearby_position(pos::ValidPos, model::ABM{<:AbstractGridSpace{D,true}}, r=1; kwargs...) where {D}
+    nindices = offsets_within_radius_no_0(model.space, r)
+    chosen_offset = rand(abmrng(model), nindices)
+    chosen_pos = mod1.(pos .+ chosen_offset, spacesize(model))
+    return chosen_pos
+end
+  
 ###################################################################
 # pretty printing
 ###################################################################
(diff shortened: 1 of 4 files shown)


Tortar:
I updated the changelog, while I was doing it I noticed that the random_agent improvement was repeated in two releases, so I removed the latest mention

src/spaces/grid_general.jl
```diff
@@ -148,6 +148,24 @@ function nearby_positions(
...
+        chosen_offset = rand(rng, nindices)
+        chosen_pos = pos .+ chosen_offset
+        checkbounds(Bool, stored_ids, chosen_pos...) && return chosen_pos
+    end
+end
+
+function random_nearby_position(pos::ValidPos, model::ABM{<:AbstractGridSpace{D,true}}, r=1; kwargs...) where {D}
+    nindices = offsets_within_radius_no_0(model.space, r)
+    chosen_offset = rand(abmrng(model), nindices)
+    chosen_pos = mod1.(pos .+ chosen_offset, spacesize(model))
```
Datseris, line 165: Isn't it faster here to check for bounds, and only if out of bounds call the `mod1` function? I think you showed that in another PR.

Datseris (approved):
Fantastic, we should tag a minor release immediatelly, which means bump project.toml and adjust changelog. Do you know how to tag releases?

src/spaces/grid_general.jl
Tortar: right, I actually think that there are some more places where this applies but for now let's solve just the issue just for this case :-)

Tortar:
didn't tag any release yet, but I think I found the guide to do it here: https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository 

Probably I can follow it without doing mistakes :D

Datseris:
Its much simpler than that. After you merge and `main` has updated version register the github Agents url here: https://juliahub.com/ui/Registrator

Tortar:
Ohh, perfect!

PR #786, opened 2023-03-28
update comparison with all latest version of the frameworks

Tortar:
Updated the comparison table with the timing calculated with the script described in https://github.com/JuliaDynamics/ABM_Framework_Comparisons/pull/16. I want to build the docs before merging to see if the visualization is still ok, and I will merge after checking :-)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
         version:
           - '1'
         os:
-          - ubuntu-18.04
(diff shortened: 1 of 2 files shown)


Tortar:
I'm fighting with the documentation building :D It doesn't want to start, don't know why

Tortar:
I updated to ubuntu 20.04 but probably it is better ubuntu latest?

Tortar:
I think it's fine like so, I will then merge :-)

PR #787, opened 2023-03-28
Some small improvements to continuous space methods

Tortar:
Some little improvements to the methods, maybe most notably the test if the pos is inside the space in `move_agent` now tests also the lower bound as it was supposed to do.

edit: also changed netlogo version which I incorrectly reported as 6.4 but the lastest one, which I used to run the benchmark, it's actually the 6.3.0 :Q

### DIFF
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -106,6 +106,9 @@ function cell_center(pos::NTuple{D,<:AbstractFloat}, model) where {D}
     model.space.spacing .* (pos2cell(pos, model) .- 0.5)
 end
 
+distance_from_cell_center(pos, model::ABM) =
+    euclidean_distance(pos, cell_center(pos, model), model)
+
 function add_agent_to_space!(
     a::A, model::ABM{<:ContinuousSpace,A}, cell_index = pos2cell(a, model)) where {A<:AbstractAgent}
     push!(model.space.grid.stored_ids[cell_index...], a.id)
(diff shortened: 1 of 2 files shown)


Tortar:
ok, I probably understood it, this is the message on the CI/ Documentation:
`
This is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on April 1st, 2023. For more details, see https://github.com/actions/runner-images/issues/6002`

src/spaces/continuous.jl
```diff
@@ -134,9 +130,8 @@ end
 # move the agent in the GridSpace; only change its position field
 function move_agent!(agent::A, pos::ValidPos, model::ABM{<:ContinuousSpace{D},A}
     ) where {D, A<:AbstractAgent}
-    for i in 1:D
-        pos[i] > spacesize(model)[i] && error("position is outside space extent!")
-    end
+    space_size = spacesize(model)
+    all(i -> 0 <= pos[i] < space_size[i], 1:D) || error("position is outside space extent!")
```
Datseris, line 134: it is more efficient here to use `any` and `&&` instead of `all` and `||`. This will make the clause exit as soon as possible.

Tortar: I think it's actually the same, `all` will return false the first time a pos coordinate is not inside extent, with `any` it would return true the first time a pos coordinate is not inside extent, so with exchanging the logical operator everything would work the same

PR #788, opened 2023-03-30
Faster resorvoir_sampling_single function

Tortar:
This is a bit faster than before, something like 2x faster, this is for a r=10 on a grid using the same benchmark as the one in in #784 (even if the sampling method was changed in #784 for a better method I used the old one which used the function):

before:

```julia
 Range (min … max):  4.288 μs …   8.639 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     4.467 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.497 μs ± 131.984 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

               ▅▄█▇▇▇▄▅▂                                       
  ▁▁▁▁▁▁▂▂▂▄▅▇████████████▆▅▄▅▄▃▄▄▄▄▃▄▃▃▃▂▂▂▂▂▂▁▁▁▂▁▁▁▁▁▁▁▁▁▁ ▃
  4.29 μs         Histogram: frequency by time        4.84 μs <
```
 
now:

```julia
 Range (min … max):  2.096 μs …   6.704 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.228 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.251 μs ± 149.042 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

             ▁▂▆▆██▆▇▂▃                                        
  ▁▁▁▁▁▁▁▂▃▄▆███████████▇▆▄▅▄▃▃▃▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁ ▃
  2.1 μs          Histogram: frequency by time        2.54 μs <
```
This is better since it consumes the not to choose elements faster without the if-else branching! 

For r=1 I found a negligible worse time instead (1-2 ns), but already nearly 1.5x faster with r=2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # main
 
+# v5.13
+
+- `random_nearby_position`, `random_nearby_ids` and `random_nearby_agent` are up to 2 times faster thanks to a faster sampling function.
+
 # v5.12
 
 - The `random_nearby_position` function is now much faster for GridSpaces.
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati"]
-version = "5.12.0"
+version = "5.13.0"
 
 [deps]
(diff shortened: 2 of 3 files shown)


Datseris (approved):
Great, bump minor version, add changelog, merge and tag!

PR #792, opened 2023-04-04
Clean docstrings

pitmonticone:


Datseris (approved):
Great, thanks! I wonder how you find these things!

pitmonticone:
Hi @Datseris, you're welcome. Thank you very much for the maintenance of all these great research-quality packages!

Let me know if you believe it could be helpful if I join this GitHub organisation as a member. Happy to help when I find the time.

Datseris:
Sure, I've invited you, although I think this by itself doesn't change much, as due to the open-sourceness members of the org have practically the same priviliges as anyone else. You'd need to join a team for more priviliges but at the moment I don't see an obvious team to put you!

pitmonticone:
Thanks @Datseris, a practical difference might be that I can create direct branches in repositories without forking on my personal account.

Datseris:
Yeah this would need you having push access to a repo, which a regular member does not. It works with teams: a team is given push access to a set of repositories, and people become members of said team. At the moment you are not part of a particular team, because I don't know which team should that be.

PR #793, opened 2023-04-08
Add using StatsBase to sir.jl example

yunhwankim2:
In sir.jl example, ```sample()``` and ```Weights()``` functions were used without ```using StatsBase```.


Tortar:
I made it a bit more explicit, thank you!

Tortar (approved):

Issue #650, opened 2022-07-09
Filtering in `random_nearby_stuff` and `random_nearby_pos`.


Datseris:
**Is your feature request related to a problem? Please describe.**

I am simulating the Daisyworld (which exists in our examples) and I am noticing that at some step of the simulation, we need to find a random nearby position that is empty. At the moment the way to do this is to collect nearby positions and then filter them.

**Describe the solution you'd like**
It would be nice if we had a function `random_nearby_position(pos, model, r, f)` that uses filtering function `f`, as in standard `Base.filter!` to skip invalid positions. And of course, same filtering extension would be added to `random_nearby_id`. 

**Describe alternatives you've considered**
The alternative allocates memory. 

By the way @AayushSabharwal have we ever benchmarked whether `rand(collect(nearby_ids(pos, model)))` is faster than `random_nearby_id(pos, model))`?

Tortar:
I investigated why there is a difference, and the problem is that when the model is periodic the iterator for the multigrid  is not inlined, so using `@inline` before the multigrid iterator seemed to solve the problem (the timings for this benchmark becomes similar for periodic and aperiodic model) but there is a catch: trying to see if there was a performance benefit in the flocking model I noticed that there was actually a regression!! I read somewhere that forcing inlining can be indeed not good for performance, and this seems to corroborate, so, summarizing, there is nothing to do about this

Tortar:
Forgot to add measurements.jl in the pr but pushed to main the modified example!

PR #794, opened 2023-04-08
Filtering for random_nearby_*

Tortar:
This implements the filtering feature discussed in #650, in the process I created two new function `sampling_with_condition_single` and `sampling_with_condition_agents_single` so that to reuse the code. Some benchmarks using the code posted in #650 seem good:

```julia
julia> @btime random_nearby_id($(25,25), model, 10) setup=(model=periodic_model())
  171.405 μs (17884 allocations: 845.36 KiB)
54290

julia> f(x, y) = true

julia> @btime random_nearby_id($(25,25), model, 10, f) setup=(model=periodic_model())
  242.911 μs (17893 allocations: 1.14 MiB)
66712

julia> f(x, y) = false

julia> @btime random_nearby_id($(25,25), model, 10, f) setup=(model=periodic_model())
  354.953 μs (17893 allocations: 1.14 MiB)
```

But this PR still needs some tests for the new functionality, which I will implement soon --> test added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 # v5.14
 
-- New optional filtering functionality to restrict the sampling added to `random_nearby_position`, `random_nearby_ids` and `random_nearby_agent`.
+- New optional filtering functionality to restrict the sampling added to `random_nearby_position`, `random_nearby_id` and `random_nearby_agent`.
 
 # v5.13
 
-- `random_nearby_position`, `random_nearby_ids` and `random_nearby_agent` are up to 2 times faster thanks to a faster sampling function.
+- `random_nearby_position`, `random_nearby_id` and `random_nearby_agent` are up to 2 times faster thanks to a faster sampling function.
 
 # v5.12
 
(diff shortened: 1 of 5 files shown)


Tortar:
I implemented `sampling_with_condition_agents_single` so that we can use the filter also on `random_nearby_agent`, the cool thing is that we could use this function also in `random_agent`, but here I made the choice to require two arguments for the filter i.e. the element (`pos, id, agent`) and the `model` so that this method is not fully compatible with random_agent which instead requires only to pass the agent, I think that passing the element and the model will be better in terms of flexibility so we could break the current code for this new one, what do you think?

Datseris:
Changing the behavior of the `nearby_agent` is breaking, so we can't do that. besides, I'd argue we should also have the 1-argument version here as well. One can always create an anonymous closure
```
# inside `agent_step!`
f = a -> two_arg_version(a, model)
random_nearby...
```

Tortar:
I agree with you, everything should be ok now! 
Anyway Happy Easter!

Datseris (approved):
happy easter to you too! Thanks for the fantastic contributions! please bump minor version, add changelog, merge and tag!

PR #795, opened 2023-04-11
Small improvement in sampling_with_condition_single function

Tortar:
A small improvement in execution time

```julia
julia> f(x) = false

julia> @benchmark random_nearby_id($(25,25), model, 10, f) setup=(model=periodic_model()) #new
BenchmarkTools.Trial: 387 samples with 1 evaluation.
 Range (min … max):  305.611 μs …   4.827 ms  ┊ GC (min … max): 0.00% … 91.75%
 Time  (median):     332.832 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   350.827 μs ± 229.440 μs  ┊ GC (mean ± σ):  3.26% ±  4.66%

julia> @benchmark random_nearby_id($(25,25), model, 10, f) setup=(model=periodic_model()) #current
BenchmarkTools.Trial: 382 samples with 1 evaluation.
 Range (min … max):  346.306 μs …   4.814 ms  ┊ GC (min … max): 0.00% … 90.59%
 Time  (median):     413.123 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   454.338 μs ± 388.431 μs  ┊ GC (mean ± σ):  7.49% ±  8.04%
```

Not much, but I think worth it :-)

Also improved some testing

### DIFF
--- a/src/core/model_abstract.jl
+++ b/src/core/model_abstract.jl
@@ -168,9 +168,10 @@ end
 
 function optimistic_random_agent(model, condition; n_attempts = 3*nagents(model))
     rng = abmrng(model)
-    for _ in 1:n_attempts
-        idx = rand(rng, allids(model))
+    @inbounds while n_attempts != 0
+        idx = rand(rng, ids)
         condition(model[idx]) && return model[idx]
+        n_attempts -= 1
     end
     # Fallback after n_attempts tries to find an agent
     return sampling_with_condition_agents_single(allids(model), condition, model)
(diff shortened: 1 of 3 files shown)


Tortar:
I have actually thought about some further optimizations, hopefully the last ones :D but I need some time to figure out how to organize the code for applying them

Tortar:
experimenting seems that there is not much more to gain, so I think this is the final form of this PR

Tortar:
I'm thinking that maybe we could parallelize some of these algorithms in the future, this should be possible without any restriction for the user

Datseris (approved):

Datseris:
Parallelizing thinsg would be nice

PR #797, opened 2023-04-21
Simulation games

kavir1698:
Here is an example of game-like ABMs, where the user can control the behavior of an agent with the keyboard in real-time.


Datseris:
@kavir1698 I am honestly not sure whether this should be here or in the example zoo. I would argue that it should be in the example zoo because its more like an advanced use case of Agents.jl with Makie's interactivity than a genuinely new API functionality for Agents.jl.

Datseris:
But it is a great idea nevertheless!

PR #798, opened 2023-04-22
Refactor of some scheduler operations

Tortar:
I refactored some operations:

- Added a faster `get_ids!` function in the case of a `UnremovableABM`
- Changed the ByID for an `UnremovableABM` in the simple call of the id container, now it is much faster and equivalent to the fastest scheduler
- `ByProperty` now can accept any properties values, not just floats, but it reallocates each time the vector `properties`, it seems like reallocating is better since doing this 

```julia
    if isempty(sched.properties)
        sched.properties = [Agents.get_data(model[id], sched.p) for id in sched.ids]
    else
        resize!(sched.properties, length(sched.ids))
        for (i, id) in enumerate(sched.ids)
            sched.properties[i] = Agents.get_data(model[id], sched.p)
        end
    end
```

instead of  `sched.properties = [Agents.get_data(model[id], sched.p) for id in sched.ids]`

makes the code slower than before (anyway I will remove the properties field from the struct since it's unnecessary but this is used here just to show that a non allocating version was possible, but actually slower!). Also, I tested the new `ByProperty` on wolfsheep and the time is nearly the same with the used `by_property` scheduler, don't know if in other situation this `ByProperty` scheduler can have an advantage on the old one since it allocates less

### DIFF
--- a/src/submodules/schedulers.jl
+++ b/src/submodules/schedulers.jl
@@ -301,3 +301,11 @@ function (sched::ByType)(model::ABM)
 
     return Iterators.flatten(it for it in sched.ids)
 end
+
+@deprecate by_id ById
+@deprecate randomly Randomly
+@deprecate partially Partially
+@deprecate by_property ByProperty
+@deprecate by_type ByType
+
+end # Schedulers submodule
--- a/src/submodules/schedulers.jl
+++ b/src/submodules/schedulers.jl
@@ -158,11 +158,10 @@ function by_property(p)
     end
 end
 
-mutable struct ByProperty{P}
+struct ByProperty{P}
     p::P
     ids::Vector{Int}
     perm::Vector{Int}
-    properties::Vector
 end
 
 """
@@ -177,7 +176,7 @@ ByProperty(p::P) where {P} = ByProperty{P}(p, Int[], Int[], [])
 function (sched::ByProperty)(model::ABM)
     get_ids!(sched.ids, model)
     
-    sched.properties = [Agents.get_data(model[id], sched.p) for id in sched.ids]
+    properties = [Agents.get_data(model[id], sched.p) for id in sched.ids]
 
     initialized = true
     if length(sched.perm) != length(sched.ids)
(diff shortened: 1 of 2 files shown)


src/submodules/schedulers.jl
```diff
@@ -149,11 +158,11 @@ function by_property(p)
     end
 end
 
-struct ByProperty{P}
+mutable struct ByProperty{P}
     p::P
-    properties::Vector{Float64} # TODO: don't assume Float64
     ids::Vector{Int}
     perm::Vector{Int}
+    properties::Vector
```
Datseris, line 165: Doesn't this type instability bring a big performance hit...?

Tortar:
yes I mean every property which has an ordering, notice that string and symbols have them, and this way we can use this scheduler for example in the wolfsheep model

src/submodules/schedulers.jl
Tortar: this seems not to be the case, I calculated the performance with the scheduler benchmark and it seems fine, will post the result

Tortar: actually changed the implementation in its final form since saving properties n the struct is unnecessary, but anyway it didn't change the performance when I calculated it

Tortar: Let me know if you can think of a better method for handling `properties` (even if allocating doesn't seem to be problematic in the benchmarks)

Datseris:
Wait, is the current version just as fast as the previous one, even though it allocates a vector every single time...?

Tortar:
exactly, it is actually even a bit faster

Tortar:
Just calculated:

before pr: 	  "ByProperty" => Trial(411.094 ms)
after pr:	          "ByProperty" => Trial(399.422 ms)

running

```julia
for (model, name) in [
           (fake_model(; nagents = 80000, scheduler = Schedulers.ByProperty(:group)), "ByProperty"),
       ]
           SUITE["large_model"][name] = @benchmarkable step!($model, agent_step, dummystep, 100)
       end
```

in schedulers benchmark

Datseris (approved):
ok, go ahead and merge then

Issue #799, opened 2023-04-30
Uniformize `randomwalk!` in gridspaces


Tortar:
Currently the implementation of `randomwalk!` for gridspaces is like so:

```julia
function randomwalk!(
    agent::AbstractAgent,
    model::ABM{<:AbstractGridSpace},
    r::Real = 1;
    kwargs...
)
    if abmspace(model).metric == :euclidean
        throw(ArgumentError(
            "Random walks on a `GridSpace` with Euclidean metric are not defined. " *
            "You might want to use a `ContinuousSpace` or a different metric."
        ))
    end
    offsets = offsets_at_radius(model, r)
    walk!(agent, rand(abmrng(model), offsets), model; kwargs...)
end

function randomwalk!(
    agent::AbstractAgent,
    model::ABM{<:GridSpaceSingle},
    r::Real = 1;
    kwargs...
)
    if abmspace(model).metric == :euclidean
        throw(ArgumentError(
            "Random walks on a `GridSpace` with Euclidean metric are not defined. " *
            "You might want to use a `ContinuousSpace` or a different metric."
        ))
    end
    offsets = offsets_at_radius(model, r)
    target_positions = map(β -> normalize_position(agent.pos .+ β, model), offsets)
    idx = findall(pos -> id_in_position(pos, model) == 0, target_positions)
    available_offsets = @view offsets[idx]
    if isempty(available_offsets)
        # don't move, return agent only for type stability
        agent
    else
        walk!(agent, rand(abmrng(model), available_offsets), model; ifempty=false)
    end
end
```

which is odd since the first `randomwalk!` calls `walk!(agent, rand(abmrng(model), offsets), model; kwargs...)`, so that if `ifempty=true` is passed we can have that the agent actually **don't move** if the random position is occupied, in the case of a `SingleGridSpace` instead we always move the agent (provided that there is an unoccupied position), I don't think the docstrings suggests this difference in behaviour between the two, and also having two different behaviour for essentially the same space would be a good thing, so I propose to change the second function in

```julia
function randomwalk!(
    agent::AbstractAgent,
    model::ABM{<:GridSpaceSingle},
    r::Real = 1;
    kwargs...
)
    if abmspace(model).metric == :euclidean
        throw(ArgumentError(
            "Random walks on a `GridSpace` with Euclidean metric are not defined. " *
            "You might want to use a `ContinuousSpace` or a different metric."
        ))
    end
    offsets = offsets_at_radius(model, r)
[...]

PR #803, opened 2023-05-04
Create `uniform_randomwalk!`

Tortar:
This is a new function to support generic N-dimensional isotropic random walks, mentioned as interesting to have in #799 .

I benchmarked the new function against the old one with

<details> <summary>benchmark </summary>

```julia

@agent A ContinuousAgent{2} begin end
@agent B ContinuousAgent{3} begin end

function cont_model_2d(; numagents = 1000, dims = (20, 20))
           space = ContinuousSpace(dims, periodic = false)
           rng = MersenneTwister(23)
           model = ABM(A, space; rng = rng)
           for _ in 1:numagents
               add_agent!(A, model, (1,1))
           end
           return model
       end

function cont_model_3d(; numagents = 1000, dims = (20, 20, 20))
           space = ContinuousSpace(dims, periodic = false)
           rng = MersenneTwister(23)
           model = ABM(B, space; rng = rng)
           for _ in 1:numagents
               add_agent!(B, model, (1,1,1))
           end
           return model
       end

model_2d = cont_model_2d()
model_3d = cont_model_3d()

agent_2d = model_2d[1]
agent_3d = model_3d[1]

@benchmark randomwalk!($agent_2d, $model_2d, 1)
@benchmark randomwalk!($agent_3d, $model_3d, 1)
@benchmark uniform_randomwalk!($agent_2d, $model_2d, 1)
@benchmark uniform_randomwalk!($agent_3d, $model_3d, 1)
```
</details>

which shows some notable improvements (the greatest part of the time is now in the last call of `walk!`): 

```julia
julia> @benchmark randomwalk!($agent_2d, $model_2d, 1)
BenchmarkTools.Trial: 10000 samples with 956 evaluations.
 Range (min … max):  91.632 ns … 835.669 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     94.874 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   96.353 ns ±  14.382 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▃▆█▇▄▂                                                      ▂
  ▇██████▇▄▄▅▃▄▁▁▄▅▅▅▅▆▄▅▆▅▅▆▄▅▃▅▅▆▆▆▅▄▃▄▆▁▄▅▄▄▄▄▅▇▅▄▅▃▄▅▄▄▃▃▆ █
  91.6 ns       Histogram: log(frequency) by time       143 ns <

julia> @benchmark randomwalk!($agent_3d, $model_3d, 1)
BenchmarkTools.Trial: 10000 samples with 855 evaluations.
 Range (min … max):  142.807 ns … 947.602 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     147.251 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   149.270 ns ±  18.042 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▃▆██▆▃▁                                                      ▂
  ▆███████▅▄▄▁▄▄▄▄▄▅▇▇▇▆▆▆▆▄▄▃▄▆▆▅▅▃▁▁▁▁▃▃▄▆▆▆▆▆▅▃▅▅▄▄▄▄▆▅▅▅▆▄▄ █
  143 ns        Histogram: log(frequency) by time        209 ns <

julia> @benchmark uniform_randomwalk!($agent_2d, $model_2d, 1)
BenchmarkTools.Trial: 10000 samples with 976 evaluations.
 Range (min … max):  82.480 ns … 724.693 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     86.373 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   87.304 ns ±   9.864 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

      ▂█▁
  ▂▂▂▄███▃▂▂▂▂▁▂▁▁▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▁▂▂▂▂▂▂ ▂
  82.5 ns         Histogram: frequency by time          124 ns <

julia> @benchmark uniform_randomwalk!($agent_3d, $model_3d, 1)
BenchmarkTools.Trial: 10000 samples with 966 evaluations.
 Range (min … max):  84.886 ns … 734.576 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     89.027 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   90.114 ns ±  10.239 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

      ▂█▅
  ▂▂▃▅███▇▃▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂ ▂
  84.9 ns         Histogram: frequency by time          127 ns <

```

this version is also generic, it supports any number of dimensions.

I didn't choose to extend `randomwalk!` since it was not so easy to differentiate between the different functions via multiple dispatch due to the fact that we use polar and azimuthal as `kargs` currently. So that not to complicate the code much more, I chose to create a new function. If we make those kargs normal args then it will be easier to use multiple dispatch, but I'm fine also with the current version, let me know what you think. 

It still needs tests, but I want to know what is the opinion on all of this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # main
 
+- The `randomwalk!` function is now supported for any number of dimensions in ContinuousSpace when used to create isotropic/uniform random walks, besides, in this case random walks are up to 2 times faster than before.
 - The `ByProperty` scheduler can now accept any type of (ordered) properties, while before it was restricted to only floats. The `ByID` scheduler of an `UnremovableABM` is now as fast as the `Fastest` scheduler since in this case they are actually equivalent.
 
 # v5.14
--- a/src/spaces/walk.jl
+++ b/src/spaces/walk.jl
@@ -2,7 +2,7 @@ using Distributions: Distributions, Uniform, ContinuousUnivariateDistribution
 using Rotations
 using StaticArrays: SVector, setindex
 
-export walk!, randomwalk!, uniform_randomwalk!, normalize_position
+export walk!, randomwalk!, normalize_position
 export Arccos, Uniform
 
 #######################################################################################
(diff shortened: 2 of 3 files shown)


Datseris:
I think I'm a bit lost: isn't isotropic/uniform walks what happens by default in `randomwalk!`...?

Tortar:
yes, exactly, in practice I choose to do that since I found multiple dispatch in this particular case a bit difficult to deal with because we don't have enough args to differentiate between functions, but I'm willing to reformulate my approach to extend `randomwalk!` instead of creating a new function, if you have any suggestion on this front let know :-)

Datseris:
let me think about it. it is almost certain that we don't need two function names. I'llcome back to #802

Tortar:
Finally found a way through!! Everything works now

Tortar:
merging with #802

PR #805, opened 2023-05-13
Fix `true_pairs!`

### LINKED ISSUE #733: Bugfix in `symdiff` causes errors in Agents.jl tests
https://github.com/JuliaLang/julia/pull/46397 fixed a bug in `symdiff!`. This caused the test in https://github.com/JuliaDynamics/Agents.jl/blob/24fdc2ffc72221d6add4d03c4050eb8aa9f0ac4e/test/continuous_space_tests.jl#L134 to error.

At this point 

https://github.com/JuliaDynamics/Agents.jl/blob/24fdc2ffc72221d6add4d03c4050eb8aa9f0ac4e/src/spaces/continuous.jl#L454

the variable `pairs ==  [(1, 2), (2, 3), (3, 4)]`. Then, we call:

https://github.com/JuliaDynamics/Agents.jl/blob/24fdc2ffc72221d6add4d03c4050eb8aa9f0ac4e/src/spaces/continuous.jl#L455

```julia
julia> pairs = [(1, 2), (2, 3), (3, 4)];

julia> unique_pairs = unique(Iterators.flatten(pairs))
 1
 2
 3
 4

julia> collected_pairs = collect(Iterators.flatten(pairs))
 1
 2
 2
 3
 3
 4
```

which on 1.8 returned (the buggy)

```julia
julia> symdiff(unique_pairs, collected_pairs)
2-element Vector{Int64}:
 2
 3
```

but now returns

```julia
julia> symdiff(unique_pairs, collected_pairs)
Int64[]
```

I don't know why the test is written the way it is but something is strange here.

Tortar:
This should fix the failing of CI, and also fix #733. But I'm not completely sure it is the correct fix, review with care :-)

### DIFF
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -448,6 +448,7 @@ function true_pairs!(pairs::Vector{Tuple{Int,Int}}, model::ABM{<:ContinuousSpace
         end
     end
     to_remove = Int[]
+    # `counter` counts the number of occurencies for each item, it comes from DataStructure.jl
     for doubles in [k for (k,v) in counter(Iterators.flatten(pairs)) if v>1]
         # This list is the set of pairs that have two distances in the pair list.
         # The one with the largest distance value must be dropped.
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -448,7 +448,7 @@ function true_pairs!(pairs::Vector{Tuple{Int,Int}}, model::ABM{<:ContinuousSpace
         end
     end
     to_remove = Int[]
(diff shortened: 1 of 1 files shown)


Datseris (commented):
Where does `counter` does from?

Datseris:
Sometimes you can just preface functions with their module if they come from a somewhat "random" module, in the sense that is rarely used in our module, so write `DataStructres.counter` (here the comment makes it clear so no need)

Tortar:
Indeed I was thinking about this solution, but was too lazy to test it out :D

PR #806, opened 2023-05-13
Fix `manhattan_distance`

Tortar:
there was a huge mistake in `manhattan_distance` :O Probably my fault, sorry (new patch version/minor version?)

Also took the opportunity to make `get_direction` more performant, this is 2x faster

(failed test is the one solved in #805)


Datseris:
lets collect changes and release a new version before sending the reply letter

Tortar:
I agree!

PR #807, opened 2023-05-13
Fix small return type instability in `walk!`

Tortar:
This makes `walk!` a little bit more performant (something like 5%), there was no specification for return values in the docs so I think it should be fine to return the `agent`


Datseris:
p.s. @Tortar could you perhaps do a PR that fixes the 1.9 failures?

PR #815, opened 2023-06-03
New function `offline_run!` to write data during `run!`

mastrof:
Replaces #630


src/simulations/collect.jl
```diff
@@ -164,6 +165,121 @@ function run!(
     return df_agent, df_model
 end
 
+"""
+    offline_run!(model, agent_step! [, model_step!], n::Integer; kwargs...) → agent_df, model_df
+    offline_run!(model, agent_step!, model_step!, n::Function; kwargs...) → agent_df, model_df
+
+Run the model while dumping data to CSV files. See [`run!`](@ref) for details.
+
```
Datseris, line 173: I think you can say more here. A lengthier description can remove ambiguity. I suggest:
```
Do the same as [`run`](@ref), but write output directly to a CSV file instead of
collecting it into an in-memory dataframe. Useful in scenarios where the amount
of collected data is expected to exceed available memory.
```

src/simulations/collect.jl
```diff
@@ -164,6 +165,121 @@ function run!(
...
+## Keywords
+* `writing_interval=1` : write to file every `writing_interval` times the data is collected. Small values have larger effects on performance.
```
Datseris, line 175: I don't understand what this is from its documentation. How does this place with the `when` to collect keywords of `run!`? Also please keep lines of code within 92 characters max.

src/simulations/collect.jl
```diff
@@ -164,6 +165,121 @@ function run!(
     return df_agent, df_model
 end
 
+"""
+    offline_run!(model, agent_step! [, model_step!], n::Integer; kwargs...) → agent_df, model_df
+    offline_run!(model, agent_step!, model_step!, n::Function; kwargs...) → agent_df, model_df
```
Datseris, line 170: THe docstring says that the functions return something, but they return `nothing` instead.

Datseris (commented):
Is CSV the best thing to use here? Aren't there any other binary-based data formats that can be used instead, and loaded directly into a dataframe...?

src/simulations/collect.jl
```diff
@@ -164,6 +165,121 @@ function run!(
...
+            collect_model_data!(df_model, model, mdata, s; obtainer)
+            model_count_collections += 1
+            model_collected = true
+        end
+        step!(model, agent_step!, model_step!, 1, agents_first)
+        s += 1
+
+        if model_collected && model_count_collections % writing_interval == 0
+            AgentsIO.CSV.write(mdata_savefile, df_model, append=model_appendtocsv)
+            df_model = init_model_dataframe(model, mdata)
```
Datseris, line 248: We re-initialize the data frames all the time that makes this rather expensive. What we should be doing is instead emptying the data frame and adding a new column each time. Emptying also retains column names which do not change during the run.

Datseris (changes requested):

src/simulations/collect.jl
Datseris: Additionally, we should separate this using a function barrier. One function initializes the dataframes and the other adds 1 row and removes 1 row in place. So one function is just the loop itself.

mastrof:
> Is CSV the best thing to use here? Aren't there any other binary-based data formats that can be used instead, and loaded directly into a dataframe...?

Is the problem with CSV "only" that it is not binary, or something else?
At the moment I don't know an easy way to append to file with e.g. JLD2. I believe you would have to reload the dataframe to modify it in place and save it again, which would make the whole thing useless.. Maybe there is a way, but I don't know it. I will look into some of the larger simulation packages (e.g. Oceananigans comes to mind), maybe someone already found a solution to this.

Datseris:
I am surprised that DataFrames.jl does not provide a binary file format that can also be read and written to line-by-line. Seems like the data science people would have figured this out by now.

Datseris:
Yes the problem with CSV is "only" that it is non-binary, but this is a rather big problem. If you are using this functionality, you probably have _a lot_ of data, and to my knowledge text-based ways to save numerical data are the most inefficient in terms of memory. So the disk space you will fill with this will be very large.

Project.toml
```diff
@@ -45,6 +45,7 @@ julia = "1.5"
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
```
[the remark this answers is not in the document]
mastrof: Was a leftover from the original PR, CSV was already in the deps. Removed.

src/simulations/collect.jl
mastrof: Not sure I get what you mean here. I now replaced the reinitialization of the dataframes with a call to `empty!`, and the data collection is just performed exactly in the same way as it is performed in the `run!` function.
There is no "adding one row and removing one in place"; data is collected as usual, but when the dumping is triggered we write the current df to file and empty it.

mastrof:
Haven't taken time to look for an alternate solution to CSV yet.
But if we agree on the current logic then it will be just a matter of substituting `CSV.write` with a different write function once an appropriate candidate is found.

src/simulations/collect.jl
```diff
@@ -164,6 +165,128 @@ function run!(
...
+during execution.
+
+## Keywords
+* `adata_savefile="adata.csv"` : a file to write agent data on.
+* `mdata_savefile="mdata.csv"`: a file to write the model data on.
+* `writing_interval=1` : write to file every `writing_interval` times data collection \
+is triggered. If the `when` keyword is not set, this corresponds to writing to file \
+every `writing_interval` steps; otherwise, the data will be written every \
+`writing_interval` times the `when` condition is satisfied \
+(the same applies to `when_model`).
```
Datseris, line 184: Please remove the ending `\` and use 2 spaces of intentation for the next row as is done in all other docstrings.

src/simulations/collect.jl
```diff
@@ -164,6 +165,128 @@ function run!(
...
+        if should_we_collect(s, model, when_model)
+            collect_model_data!(df_model, model, mdata, s; obtainer)
+            model_count_collections += 1
+            model_collected = true
+        end
+        step!(model, agent_step!, model_step!, 1, agents_first)
+        s += 1
+
+        if model_collected && model_count_collections % writing_interval == 0
+            AgentsIO.CSV.write(mdata_savefile, df_model, append=model_appendtocsv)
```
Datseris, line 254: Please replace all `,` with `;` when going from positional arguments to keyword arguments.

mastrof:
> I propose to al;ready anticipate the possibility of different backends and provide a keyword argument that configures what backend should be used.

Would something like this work, at the beginning of `offline_run!`? Other backends could then be included by adding some `elseif`s, but I have a feeling this might be very bad performance-wise.
```
write_to_file(filename, data, append) = if lowercase(backend) == "csv"
    AgentsIO.CSV.write(filename, data; append=append)
else
    throw(ArgumentError("Backend $(backend) not supported."))
end
```
Would this be improved by defining a "maker" function?
```
function make_writer(backend)
    s = lowercase(backend)
    if s == "csv"
        return (filename, data, append) -> AgentsIO.CSV.write(filename, data; append=append)
    elseif s == "mybackend"
        return (filename, data, append) -> MyBackend.write(filename, data, append)
    else 
        throw(ArgumentError("Backend $(backend) not supported."))
    end
end
```
and then inside of `offline_run!` have a line `write_to_file = make_writer(backend)`?

Datseris:
creating a maker function is indeed a better idea. However you should also introduce a function barrier: after you initialize all data frames and the maker function, pass all of these into another function. This establishes type stability where it matters.

src/simulations/collect.jl
```diff
@@ -164,6 +167,152 @@ function run!(
...
+        end
+        if when_model == true
+            for c in eachcol(df_model)
+                sizehint!(c, n)
+            end
+        end
+    end
+
+    close(open(adata_savefile, "w"))
+    close(open(mdata_savefile, "w"))
```
Datseris, line 230: Exactly after this line you should introduce the function barrier

Datseris (commented):
This can't be merged unless conflicts are resolved (rebase wiht `main`). Please don't forget teh function barrier I am sure it will have a positive impact on performance.

fbanning:
> It would be great if we could have a more disk-efficient backend to save.

I propose [Arrow.jl](). I've used it quite a bit already for storing my research results and it plays very nicely with DataFrames.

If you are OK with this (read: if you don't object vehemently), I can add this backend to this PR.

Datseris:
@fbanning I think that is a great idea. Please do go ahead and add this here. This way we also see an immediate example case of whether the interface is straightforwardly extendable to other backends.

fbanning:
Arrow backend is implemented now. In the process I've condensed the functions a bit. Also added tests for Arrow functionality and they passed locally on my machine.

The workflow for adding a new `backend` would now be:
- Add `writer_$backend` function taking `filename, data, append` as inputs.
- Add backend to `get_writer` function.
- Done (hopefully). :)

fbanning:
https://github.com/JuliaDynamics/Agents.jl/actions/runs/5333264438/jobs/9663511968?pr=815#step:7:357

Can anyone test this on Windows? I don't have a Windows machine.

src/simulations/collect.jl
```diff
@@ -164,6 +162,157 @@ function run!(
...
+    backend, adata_filename, mdata_filename, writing_interval
+)
+    s = 0
+    p = if typeof(n) <: Int
+        ProgressMeter.Progress(n; enabled=showprogress, desc="run! progress: ")
+    else
+        ProgressMeter.ProgressUnknown(desc="run! steps done: ", enabled=showprogress)
+    end
+
+    writer = get_writer(backend)
```
Datseris, line 244: Sorry, I wasn't clear with respect to the function barrier. The function barrier has to come _after_ this `writer` object has been initialized, as its initialization is type unstable due to using strings to specify the type of writer.

Project.toml
```diff
@@ -1,11 +1,12 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati", "Adriano Meligrana"]
-version = "5.15.2"
+version = "5.16.0"
 
 [deps]
-CommonSolve = "38540f10-b2f7-11e9-35d8-d573e4eb0ff2"
+Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
```
Datseris, line 7: Now that we have moved in Julia 1.9 we should be using the extensions framework to make Arrow an extension instead of a formal dependency. We have already done this with Makie so it should be easy to apply here. The key is to make a function `write_arrow` without any arguments in the file that is inside agents, and then define a method for this function with arguments inside the `ext` folder.

If how extensions work isn't clear I'll do it later.

src/simulations/collect.jl
mastrof: change happened here @fbanning ab617f687cd1910d3095ed8d8f01b20a7c9d4174
just moving the writer initialization back into `offline_run!` and passing the initialized `writer` to `run_and_write!` should solve it

fbanning: Ah, didn't get that part with the function barrier, sorry. I've moved the writer initialisation back into `offline_run!`.

Project.toml
fbanning: I've had a stab at this. Tests pass locally but I still expect the CI tests on Windows to fail so somebody needs to look at it on their local Windows machine to find out what the problem is.

I haven't implemented any package extensions so far, so please @Datseris have a look at what I did and if I missed something.

Moved the `AgentsVisualizations` and `AgentsOSMVisualizations` extensions into their own subfolders under `ext/` to make sure that the extensions files don't get mixed up. (See [here](https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) that this is possible.)

fbanning: Oh, and I removed `arrow_integration.jl` in the AgentsIO submodule for now. It would only make sense to have this as a package extension as well then. (Same applies to the JLD2 integration but I won't touch that in this PR.)

Tortar: the problem on Windows is related to permissions: it fails when calling `rm(adata.arrow)` since the os refuses to remove the file, didn't still find any solution and I'm not really understanding why that doesn't happen with the .csv file created and removed some lines before this. Is it possible that the arrow file is still open when trying to remove it?

Tortar: Interestingly it doesn't fail with `rm(adata.arrow)`

fbanning (changes requested):
Sorry, this change doesn't make any sense. When the file doesn't exist or when we want to overwrite an existing file (i.e. `!append`), then we need to pass `file = false` to the `write` function.  Please revert it.

Project.toml
fbanning: > Is it possible that the arrow file is still open when trying to remove it?

Hm, shouldn't be the case but what do I know about Windows stuff... Worth a try at least?

Tortar:
yes, you are right, it was a last, really desperate, attempt to solve the problem, don't know why but my Windows didn't complain locally, anyway I tried to dig up a bit for an hour, no clue why Windows denies the possibility to remove the file

fbanning:
@Tortar Did you run the tests locally on your windows machine and they all passed? If so, then we can be sure that this is a CI windows problem.

Tortar:
no I didn't explain myself well, they passed when I changed the non sensical lines (but this is probably because I also changed something else), they don't pass locally on my Windows machine, the file `mdata.arrow` can't be removed, but I didn't find any fix

fbanning:
Ah sorry, misunderstood you then. Hm, really weird that it only fails on Windows but has no problems on Linux and MacOS. No idea how to approach this myself, as well, sorry. :/

Datseris:
Sorry for being away but I am overwhelmed with other projects and won't have much time for Agents.jl for the next semester or so... I'll be giving comments if you explicitly as for them by tagging me @Tortar or @fbanning but will have notifications off for the next semester :( 

Now, to finish this: if I understand correclty the status quo is that this works fine, but doesn't work on windows, yes? If so, can we then add an error to the first line of the function that checks the OS and prints and error, and open an issue that this function doesn't seem to work as expect in windows with Arrow? In the test suite we only run the test if OS is not windows. Then we can merge? Because everything else seems fine.

docs/src/api.md
```diff
@@ -328,5 +328,10 @@ AgentsIO.populate_from_csv!
 AgentsIO.dump_to_csv
 ```
 
+It is also possible to write data to file at predefined intervals while running your model, instead of storing it in memory:
+```@docs
+offline_run!
```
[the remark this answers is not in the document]
fbanning: Hm, possibly yes. But I think it also makes sense to keep only the stuff for (de-)serialising models themselves (and not data collected from model runs) in the AgentsIO submodule. I'd be happy with both, doesn't really matter much imho.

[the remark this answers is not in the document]
fbanning: @Datseris Are you OK with it if we just leave it as it is right now or would you prefer `offline_run!` to be in the AgentsIO submodule?

PR #817, opened 2023-06-03
Move to Julia v1.9 and other doc updates

### LINKED ISSUE #812: Docs: highlight the zoo a bit more?
I didn't realize that there was an "Agent Zoo" while reading the documentation for the first time. The first parts of the documentation don't mention it, nor does the main github page, and it's only later on in the documentation (the last subsection of Examples) that it gets a quick mention.

So it could be beneficial to future users if you added up front some kind of promo that links to the zoo docs - a picture of a few would probably be sufficient!

Documentation is pretty good, by the way!

Datseris:
This PR moves Agents.jl to Julia 1.9 and introduces several documentation updates.

- Package extensions are used now to load and export the plotting code
- Documentation mentions the ABM framework comparison and also uses it to make claims that Agents.jl is fastest etc.
- Highlights zoo more
- Updates latest news / update messages

Closes #812


Tortar:
Wow George, I feel very honored for having been included among the authors!! Thank you a lot!

AayushSabharwal:
Hi all! I have a suggestion here. The package extensions feature immediately reminded me of Agents. Would exposing the OSM module as a package extension requiring LightOSM be beneficial? It's likely someone using a open street map would also load LightOSM explicitly to manipulate/analyze the maps, and this way for the majority of models that don't require it we save startup time. Possibly OSM doesn't add too much to TTFX, but it might be worth a quick test with PrecompileTools (I'll run this in a few hours, my system is currently under significant load which would bias the results).

Datseris:
Sounds like a great idea, I will do this in this PR.

Datseris:
Hm. Actually, we have a problem @AayushSabharwal . Making this an extension is a breaking change, because code that used to work only with `using Agents` will no longer work.

AayushSabharwal:
Ah, that's too bad. I guess it's worth opening an issue marked breaking, for when the next breaking change comes out?

README.md
```diff
@@ -8,32 +6,8 @@
...
-  month = jan,
-  publisher = {{SAGE} Publications},
-  pages = {003754972110688},
-  author = {George Datseris and Ali R. Vahdati and Timothy C. DuBois},
-  title = {Agents.jl: a performant and feature-full agent-based modeling software of minimal code complexity},
-  journal = {{SIMULATION}},
-  volume = {0},
-  number = {0},
-}
-```
```
Tortar, line 39: I think these two sections are still very important to highlight in the readme, restoring them seems good enough to me

Datseris: Yes for the citation. No for contributing. Surely, someone that wants to contribute should at least visit the online documentation.

README.md
```diff
@@ -8,20 +6,23 @@
 [![codecov](https://codecov.io/gh/JuliaDynamics/Agents.jl/branch/main/graph/badge.svg)](https://codecov.io/gh/JuliaDynamics/Agents.jl)
 [![Package Downloads](https://shields.io/endpoint?url=https://pkgs.genieframework.com/api/v1/badge/Agents)](https://pkgs.genieframework.com?packages=Agents)
 
-Agents.jl is a [Julia](https://julialang.org/) framework for agent-based modeling (ABM). All further information about Agents.jl are provided in the docs.
+Agents.jl is a pure [Julia](https://julialang.org/) framework for agent-based modeling (ABM): a computational simulation of autonomous agents that react to their environment (including other agents) given a predefined set of rules.
```
Tortar, line 9:
```suggestion
Agents.jl is a pure [Julia](https://julialang.org/) framework for agent-based modeling (ABM): a computational simulation methodology where autonomous agents react to their environment (including other agents) given a predefined set of rules.
```

README.md
```diff
@@ -8,20 +6,23 @@
...
+Some major highlights of Agents.jl are:
 
-## Contributions
+1. It is fast (faster than MASON, NetLogo, or Mesa)
+2. It is simple: has a very short learning curve and requires writing minimal code
+3. Has an extensive interface of thousands of out-of-the box possible agent actions
+4. Straightforwardly allows simulations on Open Street Maps
 
-Any contribution to Agents.jl is welcome! For example you can:
+The simplicity of Agents.jl is due to the intuitive space-agnostic modelling approach we have implemented: agent actions are specified using generically named functions (such as "move agent" or "find nearby agents") that do not depend on the actual space the agents exist in, nor on the properties of the agents themselves. Overall this leads to ultra fast model prototyping where even changing the space the agents live in is matter of only a couple of lines of code.
```
Tortar, line 17: Don't want to sound too picky but the reference only to its simplicity seems kind of strange to me :D I like more the one here https://juliadynamics.github.io/Agents.jl/previews/PR817/:

"Besides its incredible performance, two more major highlights of Agents.jl are its simplicity and the total amount of features. Agents.jl is designed to require writing minimal code and being very simple to learn and use, even though it provides an extensive API with thousands of possibilities for agent actions. This simplicity is due to the intuitive space-agnostic modelling approach we have implemented: agent actions are specified using generically named functions that do not depend on the actual space the agents exist in or on the properties of the agents themselves. Overall this leads to ultra fast model prototyping where even changing the space of the agents is matter of only a couple of lines of code."

Tortar: But maybe it's a bit repetitive this way given that the bullet points already address those aspects at the same level of details...

Tortar:
I calculated how much time it takes for loading LightOSM

```julia
julia> @time using LightOSM
  0.027696 seconds (29.62 k allocations: 1.560 MiB, 79.18% compilation time: 100% of which was recompilation)
```

so very little, if it's breaking it's probably too little to change it?

README.md
Datseris: it is repetitive.

Datseris:
and then releasing v5.15.

PR #821, opened 2023-06-24
Fix performance of some versions of add_agent

Tortar:
Now I don't see any regression with these changes, running the benchmark in the related issue #820 

```julia
julia> @benchmark model_1()
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  125.087 μs …   2.810 ms  ┊ GC (min … max):  0.00% … 90.66%
 Time  (median):     134.490 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   158.740 μs ± 207.101 μs  ┊ GC (mean ± σ):  12.16% ±  8.70%

  ▁▆███▇▆▅▄▄▃▃▂▂▂▁▁▁▁▁▁▁  ▁▁                                    ▃
  ██████████████████████████▇█▇▆██▇▆▆▇▇▇▆▆▅▅▅▁▅▅▃▅▄▅▄▄▃▅▄▄▅▃▃▁▃ █
  125 μs        Histogram: log(frequency) by time        257 μs <

 Memory estimate: 434.19 KiB, allocs estimate: 4385.

julia> @benchmark model_2()
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  114.889 μs …   2.980 ms  ┊ GC (min … max):  0.00% … 88.82%
 Time  (median):     126.441 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   153.133 μs ± 237.755 μs  ┊ GC (mean ± σ):  14.52% ±  8.83%

   ▆███▇▅▄▄▃▂▂▂▁▁▁ ▁                                            ▂
  █████████████████████▆▆▇▇▆▇▆▆▆▆▅▅▆▄▆▅▄▅▄▃▅▅▁▁▃▄▄▁▅▁▃▁▁▁▄▁▃▄▃▃ █
  115 μs        Histogram: log(frequency) by time        293 μs <

 Memory estimate: 434.89 KiB, allocs estimate: 4394.
```

all of this was based on https://discourse.julialang.org/t/why-using-a-mutable-struct-type-argument-to-create-instances-creates-a-50x-slowdown/100764

PR #822, opened 2023-06-25
Declare arguments with Vararg

Tortar:
In #821 I didn't apply this technique in all places where could be useful, so I'm doing it here :-)

(Anyway given that now adding agents has the same performance in this shorter form we can use it in the ABM_Comparison when we release a new version)


Datseris (approved):
ok bump patch version and we immediately release new version

Tortar:
Released it 👍

Issue #818, opened 2023-06-05
Lazy-load `OSM` module using package extensions


AayushSabharwal:
The `OSM` module can be lazy-loaded when a user imports `LightOSM`, which would help cut down on TTFX. Unfortunately, this is a breaking change since existing functionality which worked with `using Agents` would no longer function correctly without `using LightOSM`

Tortar:
Probably this is more interesting than what I wrote in #817 :D 

This is what I get running @time_imports 

```julia
julia> @time_imports using Agents
      0.4 ms  Compat
      0.3 ms  Compat → CompatLinearAlgebraExt
     11.2 ms  OrderedCollections
     71.1 ms  DataStructures
     13.4 ms  MacroTools
      2.6 ms  SimpleTraits
      0.9 ms  Statistics
      3.5 ms  StaticArraysCore
    574.5 ms  StaticArrays
      3.6 ms  ArnoldiMethod
      0.8 ms  Inflate
     32.4 ms  Graphs
      0.5 ms  Reexport
      0.4 ms  SortingAlgorithms
      1.1 ms  DataAPI
     23.6 ms  PooledArrays
     13.6 ms  Missings
      3.4 ms  InvertedIndices
      0.3 ms  IteratorInterfaceExtensions
      0.2 ms  TableTraits
      0.9 ms  Formatting
      0.2 ms  DataValueInterfaces
     42.1 ms  Tables
     39.9 ms  StringManipulation
     30.5 ms  Crayons
      0.9 ms  LaTeXStrings
    141.4 ms  PrettyTables
     14.4 ms  Preferences
      0.4 ms  SnoopPrecompile
     42.5 ms  SentinelArrays
      0.4 ms  PrecompileTools
     49.6 ms  Parsers
     10.0 ms  InlineStrings
    863.9 ms  DataFrames
      4.9 ms  ProgressMeter
      0.3 ms  UnPack
      0.4 ms  Parameters
     32.2 ms  ArrayTools
     42.5 ms  QuickHeaps
      6.9 ms  TranscodingStreams
      0.6 ms  Requires
    140.4 ms  FileIO
    417.0 ms  JLD2 46.05% compilation time (61% recompilation)
      4.4 ms  StaticGraphs
      8.0 ms  SimpleWeightedGraphs
      5.6 ms  MetaGraphs
      0.3 ms  StatsAPI
      5.1 ms  Distances
     29.4 ms  NearestNeighbors
[...]

PR #829, opened 2023-07-12
Precompile some model examples reducing TTFX

Tortar:
This addresses a comment in https://github.com/JuliaDynamics/Agents.jl/issues/818 .

I think it's a very good improvement in TTFX with little side effects as I will illustrate:

Before pr:

`@time using Agents # -> 3.466778`

After pr

`@time using Agents # -> 3.798331`

so there is a little regression, however running this:

```julia
using Agents
models = [Models.flocking, Models.schelling, Models.zombies, Models.sir]

for m in models
    t = @time model, agent_step!, model_step! = m(); step!(model, agent_step!, model_step!, 10)
end
```

before:

```julia
  0.501026 seconds (1.27 M allocations: 87.205 MiB, 6.06% gc time, 99.95% compilation time)
  0.353901 seconds (1.15 M allocations: 76.176 MiB, 2.58% gc time, 99.96% compilation time)
  7.463354 seconds (13.40 M allocations: 879.895 MiB, 4.50% gc time, 99.23% compilation time: 4% of which was recompilation)
  1.325875 seconds (2.77 M allocations: 183.958 MiB, 3.19% gc time, 99.75% compilation time)
```

after:

```julia
  0.012826 seconds (2.43 k allocations: 129.844 KiB, 98.79% compilation time)
  0.011626 seconds (436 allocations: 33.000 KiB, 99.18% compilation time)
  0.200568 seconds (610.16 k allocations: 44.531 MiB, 10.97% gc time, 61.06% compilation time)
  0.007742 seconds (19.04 k allocations: 2.683 MiB, 61.97% compilation time)
```

So I think a win overall

Precompiling the models in the Models folder makes the examples models almost without latency which is good for users using the library for the first time


Tortar:
mmmmhhhh I'm actually not sure that this is as good as it was intended to be since running https://julialang.github.io/PrecompileTools.jl/stable/#Seeing-what-got-precompiled we get e.g.

`MethodInstance for Agents.add_agent!(::Agents.UnremovableABM{Agents.ContinuousSpace{2, true, Float64, typeof(Agents.no_vel_update)}, Main.Agents.Models.Bird, Agents.Schedulers.Randomly, Nothing, Random.TaskLocalRNG}, ::Tuple{Float64, Float64}, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64)`

which is actually still not reusable in other contexts, indeed for examples not precompiled (e.g. `predator_prey`) the time is very similar with or without precompilation.

The advantage of fast tutorial examples seems still something to consider, but don't know if worth the (even if little) regression in importing

Datseris:
Yeah it's because practically all functions in Agents.jl dispatch on the Agent type, which is almost certainly different in each application. So pre-compiling is very unlikely to lead to any difference in user developed models...

Datseris:
But thanks a lot for trying this out!

Issue #809, opened 2023-05-23
New function `replicate!(agent, model; different_kwargs…)`


franfram:
**Is your feature request related to a problem? Please describe.**

I'm working on a project involving an ABM for animal movement and disease ecology. Following the Predator-prey dynamics [example](https://juliadynamics.github.io/Agents.jl/stable/examples/predator_prey/), we use the `reproduce!` function to manage the reproduction of agents: 
```julia
function reproduce!(agent::A, model) where {A}
    agent.energy /= 2
    id = nextid(model)
    offspring = A(id, agent.pos, agent.energy, agent.reproduction_prob, agent.Δenergy)
    add_agent_pos!(offspring, model)
    return
end
```

The problem with this function is that it's not very general, it assumes that the agent has certain properties (e.g., energy, reproduction_prob, \Deltaenergy). Thus it can not be used in a different scenario with different agents. But if the function could somehow infer the properties of the agent, it would be more general and could be reused across different scenarios.

**Describe the solution you'd like**
A good solution would be a new function `replicate!(agent, model; different_kwargs...)` that given an agent, it creates a new agent at the same position and with the same fields, but with the possibility to specify certain fields that should be different.

Cheers!

Datseris:
```julia


function
replicate!(agent::AbstractAgent, model::ABM;
 kwargs...)

# Get the properties of the agent

properties =
fieldnames(typeof(agent))

# Create a dictionary to hold the new agent's properties

new_properties =
Dict{Symbol, Any}()



# Copy the properties from the old agent to the new agent

for property
in properties

if property
!= :id
&& property
!= :pos

new_properties[property] =
getfield(agent, property)

end

end



# Override any properties specified in kwargs

for (key, value)
in kwargs

new_properties[key] = value

end



# Create the new agent

add_agent!(agent.pos, model; new_properties...)

end


```

pasting here some reference code from Francis

Tortar:
I think this should be faster, even if probably less than manually creating the agent:

```
function replicate!(agent::AbstractAgent, model::ABM; kwargs...)

    newagent = deepcopy(agent)

    # Override any properties specified in kwargs
    for (key, value) in kwargs
        new_properties[key] = value
    end

    newagent.id = nextid(model)
    add_agent_pos!(newagent, model)
end
```

Tortar:
actually seems very slow, but anyway a deepcopy is necessary since we can't pass by reference all attributes (a vector for example)

[...]

PR #830, opened 2023-07-13
New function: replicate!

Tortar:
Implemented the function described in #809. 

I decided to include it in the sample.jl file since this redefinition of `deepcopy` should benefit also `sample!`.

This still needs some tests. Will also benchmark the benefit for `sample!`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # main
 
-- New function `replicate!` allows to create new instances of an agent given in input specyfing some fields with
+- New function `replicate!` allows to create a new instance of an agent given in input with the possibility to specify some fields with
   new values.
 - The `sample!` function is 3x faster than before.
   
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # main
 
-- New function `replicate!` allows to create a new instance of an agent given in input with the possibility to specify some fields with
-  new values.
+- New function `replicate!` allows to create a new instance of a given agent at the same position with the possibility to specify some
+  fields with new values.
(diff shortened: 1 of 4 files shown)


Tortar:
Running this:

using Agents, BenchmarkTools

mutable struct FakeAgent <: AbstractAgent
    id::Int
end

```julia
function fake_model(; nagents, scheduler)
    model = StandardABM(FakeAgent; scheduler)
    for i in 1:nagents
        add_agent!(model)
    end
    model
end

model = fake_model(; nagents = 1000, scheduler=Schedulers.fastest)

@benchmark sample!($model, 100000)
@benchmark sample!($model, 10000)
@benchmark sample!($model, 1000)
```

you can see that `sample!` is now 3x faster!

Datseris (approved):
seems good, i'd say up minor version and tag new release.

BTW, i was always thinking that some kind of model with `sample!` should have been in the ABM cvomparison framework since we have a simpler and more well thought out and almost surely faster API. Maybe you can open an issue in the ABGM comparison repo asking for a model with sample. Or maybe we just use the evolution model we have in the example zoo

Tortar:
yes, I think it is a good idea, but at the same time I'm not sure if an implementation of a similar model is ready-made in the other frameworks

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # main
 
+- New function `replicate!` allows to create a new instance of a given agent at the same position with the possibility to specify some
```
Datseris, line 3: Should have been a `# v5.17` here, you can push it to master directly

Tortar: ops sure :-)

Issue #709, opened 2022-11-17
Extend `@agent` macro to allow defining default values with `Base.@kwdef`


fbanning:
I make heavy use of `Base.@kwdef` when creating my ABMs because it
1. makes it clear which field's value is set during agent creation,
2. allows to set default values for each field, and
3. allows to skip those fields with existing default values.

A simple example:

```julia
julia> Base.@kwdef mutable struct Worker <: AbstractAgent
           id::Int
           age::Int = 16
           moneyz::Float64
       end
Worker

julia> Worker(id = 1, moneyz = 10)
Worker(1, 16, 10.0)
```

The new `@agent` macro, which we want to switch to completely at some point in the future, does not allow this:

```julia
julia> Base.@kwdef @agent Worker2 NoSpaceAgent begin
           age::Int = 16
           moneyz::Float64
       end
ERROR: Invalid usage of @kwdef
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] var"@kwdef"(__source__::LineNumberNode, __module__::Module, expr::Any)
    @ Base ./util.jl:478
  [3] eval
    @ ./boot.jl:368 [inlined]
  [4] eval
    @ ./Base.jl:65 [inlined]
  [5] repleval(m::Module, code::Expr, #unused#::String)
    @ VSCodeServer ~/.vscode/extensions/julialang.language-julia-1.38.2/scripts/packages/VSCodeServer/src/repl.jl:222
  [6] (::VSCodeServer.var"#107#109"{Module, Expr, REPL.LineEditREPL, REPL.LineEdit.Prompt})()
    @ VSCodeServer ~/.vscode/extensions/julialang.language-julia-1.38.2/scripts/packages/VSCodeServer/src/repl.jl:186
  [7] with_logstate(f::Function, logstate::Any)
    @ Base.CoreLogging ./logging.jl:511
  [8] with_logger
    @ ./logging.jl:623 [inlined]
  [9] (::VSCodeServer.var"#106#108"{Module, Expr, REPL.LineEditREPL, REPL.LineEdit.Prompt})()
    @ VSCodeServer ~/.vscode/extensions/julialang.language-julia-1.38.2/scripts/packages/VSCodeServer/src/repl.jl:187
 [10] #invokelatest#2
    @ ./essentials.jl:729 [inlined]
 [11] invokelatest(::Any)
    @ Base ./essentials.jl:726
 [12] macro expansion
[...]

Issue #728, opened 2023-01-15
allow `const` syntax in the `@agent` macro


thevolatilebit:
I use a variant of Agents.jl `@agent` macro in [AlgebraicAgents.jl](https://github.com/Merck/AlgebraicAgents.jl). @slwu89 recently suggested to me the following idea ([issue 4, AlgebraicAgents.jl](https://github.com/Merck/AlgebraicAgents.jl/issues/4)) which clearly applies also to Agents.jl:

*Any agent type being made from calling `@agent` is mutable, but with Julia versions >= 1.8 now we can specify particular fields as being const, which can help bring performance closer to immutable structs (see https://docs.julialang.org/en/v1/manual/types/#Mutable-Composite-Types). It would be nice to let users to specify fields as const if they like, to improve performance.*

To the best of my knowledge, `@agent` macro syntax is not compatible with `const field` declarations - at least I have been getting some parsing errors, as we need the `struct` wrap.

As a workaround, I proposed that the syntax of `@agent` could change to the following:

```julia
@agent AnotherAgentType [OptionalSupertype] struct YourAgentType{X}
        extra_property::X
        other_extra_property::Int
        const immutable_property::X
        # etc...
end
```

You may review a candidate implementation [there](https://github.com/Merck/AlgebraicAgents.jl/blob/8b7d05a462cd1ee8303fecbf77f01e7fa13b7d07/src/agent_macros.jl#L8).

Would it be possible to achieve the same functionality without the change of syntax?

Datseris:
Hm, unfortunately this syntax is a breaking change. But maybe there is a simple way to modify `@agent` so that it can parse `const` the way it is now. (I dont know this way at the moment). Have you asked for help in discourse? 

Allow the `cost` decleration is definitely desirable

thevolatilebit:
So far I asked in slack - the suggestion was to look at `Base.@kwdef`, which uses the `struct` syntax, so this does not really apply here. I will ask on discourse as well.

thevolatilebit:
https://discourse.julialang.org/t/const-syntax-outside-of-struct/93173

fbanning:
This would indeed be very nice. Also to be able to use `Base.@kwdef` (see #709) would be great. If we start to work on (or possibly rewrite) the `@agent` macro, then we should consider implementing that at the same time.

Datseris:
[...]

PR #831, opened 2023-07-14
Allow using const values in agent macro

Tortar:
<del>This should make it possible to use default values addressing https://github.com/JuliaDynamics/Agents.jl/issues/709 , let me know what do you think.  </del> Experimenting a bit, there shouldn't be any perf regression applying @kwdef as default

<del>(I also put in the api page of the documentation the replicate! function, forgot to do that in the other pr :D) </del>

edit: now it also includes the possibility to declare constants, which should help performance, but need some good tests


Tortar:
Implemented the first way described in #728 

```julia
using Agents

@agent A NoSpaceAgent begin
    f1::Int
    f2::Int
    consts = (:f1,)
end

agent = A(1, 3, 5)
agent.f1 = 10
```

returns

```julia
ERROR: setfield!: const field .f1 of type A cannot be changed
Stacktrace:
 [1] setproperty!(x::A, f::Symbol, v::Int64)
   @ Base .\Base.jl:38
 [2] top-level scope
   @ REPL[4]:1
```

Tortar:
this should be it, all new tests passes

Datseris:
does it allow inhereting from a type defined with constants? if we change `NoSpaceAgent` to have `const id::Int`, does `@agent` work with inhereting from `NoSpaceAgent`?

Tortar:
Yes it allows it, any const will be propagated to any inheriting new type, this is so because we actually check if there are any const field in the base type and then add it as a const also in the new type

Datseris:
ok great!

src/core/agents.jl
```diff
@@ -28,9 +28,7 @@ Define an agent struct which includes all fields that `AnotherAgentType` has,
 as well as any additional ones the user may provide via the `begin` block.
 See below for examples.
 
-Using `@agent` is the recommended way to create agent types for Agents.jl,
-however keep in mind that the macro (currently) doesn't work with `const` 
-declarations in individual fields (for Julia v1.8+).
+Using `@agent` is the recommended way to create agent types for Agents.jl.
```
Datseris, line 31:
```suggestion
Using `@agent` is the only supported way to create agent types for Agents.jl.
```

Datseris:
why isn't it possible to do things the native julia way and allow prefacing with `const` a constant field?

Tortar:
yes, a field name consts is disallowed, there is no way to have const inside the agent macro unfortunately due to parsing errors, see https://github.com/JuliaLang/julia/issues/45024

Tortar:
it's a bit hacky yes but it's a way not to redefine the agent macro as was done in https://github.com/JuliaDynamics/Agents.jl/issues/728, we could follow that path alternatevely but this means that every code will break :D

Datseris:
but the julia issue you cite is fixed...?

Datseris:
Okay, I see, there is no way to have `const x = 1`. Okay. I think your solution with the extra field is the only option then.

But, rename it to `constants`. and document it appropriately.

@Tortar this is a big change. please do not tag anything after mergfing the const and keyword improvements. i want to read the final docstring and changelog and give the ok for a new tag.

Tortar:
yes, but the fix is disallowing the construct :D

Il Sab 22 Lug 2023, 11:59 George Datseris ***@***.***> ha
scritto:

Tortar:
sure, I actually have another important change which I would like to finish up, we should probably tag them all together

Tortar:
@Datseris I completed the higlighting of the changes reverting some of the changed lines in https://github.com/JuliaDynamics/Agents.jl/pull/732/files , I also declared the `id` field of `NoSpaceAgent` as constant

src/core/agents.jl
```diff
@@ -103,6 +101,19 @@ Person(id = 1, pos = (1, 1), moneyz = 2000)
 # new age value
 Person(1, (1, 1), 40, 2000)
 ```
+It is also possible to specify that some fields are immutable
```
Datseris, line 104:
```suggestion
It is also possible to specify that some fields are immutable (constants)
```

test/runtests.jl
```diff
@@ -77,6 +77,26 @@ Agent8(id, pos; f1, f2) = Agent8(id, pos, f1, f2)
     f3::Float64 = 3.0
 end
 
+@agent Agent10 NoSpaceAgent begin
```
Datseris, line 80: please do not define agent types outside the test they are needed. I know this is what we do in the tests at the moment very often but it is very bad practice that we are slowly fixing.

Datseris: see https://github.com/JuliaDynamics/Agents.jl/issues/634

Datseris (approved):
Sounds great. Simple comments to address and then please merge (BUT NNOT TAG!!!)

Tortar:
Merging! (Not tagging :D)

PR #833, opened 2023-07-19
Allow using default values in agents macro

Tortar:
This tries to split in two parts #831 since the part with the consts variable is more hacky, but using default values is very easy and we can support it almost with no changes

### DIFF
--- a/test/model_creation_tests.jl
+++ b/test/model_creation_tests.jl
@@ -35,6 +35,13 @@ using Test, Agents, Random
     end
     @test Fisher <: AbstractHuman
     @test :fish_per_day ∈ fieldnames(Fisher)
+
+    agent_kwdef = Agent9(id = 1, f2 = 10)
+    values = (1, 40, 10, 3.0)
+    @test all(getfield(agent_kwdef, n) == v for (n, v) in zip(fieldnames(Agent9), values))
+    agent_kwdef = Agent9(1, 20, 10, 4.0)
+    values = (1, 20, 10, 4.0)
+    @test all(getfield(agent_kwdef, n) == v for (n, v) in zip(fieldnames(Agent9), values))
 end
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# main
+
+- The `@agents` macro now supports fields with default values.
+
 # v5.17
 
 - New function `replicate!` allows to create a new instance of a given agent at the same position with the possibility to specify some
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -91,6 +91,18 @@ mutable struct Baker{T} <: AbstractAgent
     breadz_per_day::T
 end
 ```
+Notice that you can also use default values for some fields, in this case you 
+will need to specify the field names with the non-default values
+```julia
+@agent Person{T} GridAgent{2} begin
+    age::Int = 30
+    moneyz::T
+end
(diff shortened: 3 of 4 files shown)


Tortar:
@fbanning I'd like to ask you what do you think about the pr since I saw you opened the linked issue :-)

fbanning:
To me this looks as clean and straightforward as possible.

fbanning:
Did you run any performance checks on the `@agent` macro with `@kwdef` added to it? I don't think it should result in any worse performance but would be good to check for that. I couldn't find any comments on the performance impact of using the `@kwdef` macro in a quick internet search.

fbanning (approved):

Tortar:
yes, tried to see if there were any regression with micro-benchmarks on agent creation and there wasn't any. Thanks for the review anyway!

Datseris:
wow what a big improvement! 🎉

Datseris:
does this impact in any way `add_agent!(pos, model, args...; kwargs....)`? No, right? 

Actually, I don't know. How does this work...?

test/model_creation_tests.jl
```diff
@@ -35,6 +35,13 @@ using Test, Agents, Random
     end
     @test Fisher <: AbstractHuman
     @test :fish_per_day ∈ fieldnames(Fisher)
+
```
Datseris, line 38: Here definitely more tests should have been added to see how this works with `add_agent!` that automatically creates agents.

Tortar: yes, it is actually something I didn't consider when doing the pr :D, it seems that it shouldn't have any effect on the version with `args`, I'm also pretty sure that currently if we pass anything in `kwargs` it will throw error (even before this pr)

Tortar: But we can support both ways with:

```julia
function add_agent!(
    pos::ValidPos,
    A::Type{<:AbstractAgent},
    model::ABM,
    properties::Vararg{Any, N};
    kwproperties...,
) where {N}
    id = nextid(model)
    if isempty(kwproperties)
        newagent = A(id, pos, properties...)
    else
        newagent = A(; id = id, pos = pos, kwproperties...)
    end
    add_agent_pos!(newagent, model)
end
```

It should be as fast as before since the branch prediction is pretty easy so the compiler should be able able to infer what to do

Tortar: indeed I don't see any meaningful difference in our model comparisons

```julia
# before this modification
Agents.jl WolfSheep-small (ms): 2.8282849999999997
Agents.jl WolfSheep-large (ms): 99.588342
Agents.jl Flocking-small (ms): 9.377578999999999
Agents.jl Flocking-large (ms): 67.284021
Agents.jl Schelling-small (ms): 0.453293
Agents.jl Schelling-large (ms): 4.035073
Agents.jl ForestFire-small (ms): 0.767832
Agents.jl ForestFire-large (ms): 16.152233

# after this modification
Agents.jl WolfSheep-small (ms): 2.82549
Agents.jl WolfSheep-large (ms): 99.25996699999999
Agents.jl Flocking-small (ms): 8.799429
Agents.jl Flocking-large (ms): 67.878292
Agents.jl Schelling-small (ms): 0.455848
Agents.jl Schelling-large (ms): 4.059518
Agents.jl ForestFire-small (ms): 0.776428
Agents.jl ForestFire-large (ms): 16.248959
```

PR #835, opened 2023-07-22
Support for keywords version of add_agent!

Tortar:
Thanks to a comment by @Datseris in #833 we identified an inconsistency, this doesn't show any slowdown (see the other pr) and we can then easily support it. Need to add some tests


Datseris:
I am trying to understand now. What does the `@agent` macro do? does it define a mixed version where some properties are positional and some keyword arguments? or are all keyword arguments? The original `@agent` macro had nothing to do with keywods. It made a struct where everything was positional. Now `@agent` **always** uses `base@kwdef`? if so, it always makes a version with positional and one version with all keyword arugments?

yes?

in any case, now that `@agent` supports `const` and keywords, it must be established as THE and ONLY way to make agents in Agents.jl, because at the moment your `if` clause simply doesn't allow for agents that have a mixed version: `add_agent!` takes both positional and keyword arguments.

Tortar:
yes you are right now it makes both versions: a positional and a keywords one. But the actual problem is: it doesn't seem possible to mix positional and keywords arguments in a `struct` (and so also in our macro), e.g.

```julia
julia> @kwdef struct A
           f1::Int
           f2::Int
       end
A

julia> A(1; f2=4)
ERROR: MethodError: no method matching A(::Int64; f2::Int64)

Closest candidates are:
  A(::Int64, ::Int64) got unsupported keyword argument "f2"
   @ Main REPL[9]:2
  A(::Any, ::Any) got unsupported keyword argument "f2"
   @ Main REPL[9]:2
  A(; f1, f2)
   @ Main util.jl:567

Stacktrace:
 [1] top-level scope
   @ REPL[10]:1

```

Am I wrong? It's probably worth opening an issue on this in the Julia repo? 

So for our problem, we can't actually support mixed versions if this is true.

Tortar:
But I'm actually not sure if it is useful either to extend `add_agent!` to the all keywords case inside the package, what do you think?

Datseris:
> Am I wrong? It's probably worth opening an issue on this in the Julia repo? But I think it's something one could define with extending the struct so probably not

no, that makes sense. A user however may explicitly define an additional method 
```julia
A(f1; f2 = 1) = A(f1, f2)
```
which would allow a mixed dispatch between positional and keyword arguments.

and this is exactly why we need to say that `@agent` is the only way to define agents, because we can't handle cases like above.

Datseris:
and i agree; if a user wants keywords, the only wayt is to make everything a keyword, not only select quantities

Tortar:
I think the work here should be finished as well if we want to support the feature

Datseris:
Shoujld we say expluicitly in the `add_agent!` docstring that only one of two ways ispossible?

Datseris:
in line 233 we should have two versions:
```julia
    add_agent!([pos,] A::Type, model::ABM, args...) → newagent
    add_agent!([pos,] A::Type, model::ABM; kwargs...) → newagent
```

as it is not possible to mix two versions, right/>

Tortar:
Now it should be clearer, good idea to specify both versions

Tortar:
this should be ok now

Datseris (approved):

PR #843, opened 2023-08-09
Agents macro allows parametric types and subtyping

Tortar:
this should solve #725 , can you @mastrof say to me if this is what you intended and that it works for your use case please?

Now this works:

```julia
using Agents
abstract type AbstractFoo{D} <: AbstractAgent where D end
@agent MyFoo2{D} ContinuousAgent{D} where D AbstractFoo{D} begin; end
```


Tortar:
with the latest commit

```julia
using Agents
abstract type AbstractFoo{D} <: AbstractAgent where D end
@agent MyFoo2{D} ContinuousAgent{D} AbstractFoo{D} begin end
```

works!

PR #846, opened 2023-08-09
Use SVector for ContinuousAgent

### LINKED ISSUE #672: Should we just go ahead and use `SVector` in position and velocity fields of ContinuousSpace...?

mastrof:
After I disgracefully deleted the branch I was using for #673, I restored and updated those changes. I could not reopen the original pr, sorry. Closes #672.

Still WIP, I'll try to improve this and then keep updating when needed

Main open points from #673:
- Pathfinder is messy due to multiple `SVector` <-> `NTuple` conversions
- Might be better to use `typeof` wherever possible rather than cast explicitly to `SVector`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -22,6 +22,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Rotations = "6038ab10-8711-5258-84ad-4b1120ba62dc"
 Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [weakdeps]
@@ -30,9 +31,9 @@ Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 OSMMakie = "76b6901f-8821-46bb-9129-841bc9cfe677"
 
 [extensions]
+AgentsArrow = "Arrow"
 AgentsOSMVisualizations = "OSMMakie"
 AgentsVisualizations = "Makie"
-AgentsArrow = "Arrow"
 
 [compat]
 Arrow = "2"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati", "Adriano Meligrana"]
-version = "5.17.1"
+version = "6.0.0"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
--- a/Project.toml
+++ b/Project.toml
@@ -22,7 +22,6 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Rotations = "6038ab10-8711-5258-84ad-4b1120ba62dc"
 Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
-StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [weakdeps]
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -12,7 +12,8 @@ using DataStructures
 using Graphs
 using DataFrames
 using Random
-using StaticArraysCore
+using StaticArrays: SVector
+export SVector
 import ProgressMeter
 import Base.length # TODO: This should not be imported!!!
 import LinearAlgebra
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -12,6 +12,7 @@ using DataStructures
 using Graphs
 using DataFrames
 using Random
+using StaticArrays
 import ProgressMeter
 import Base.length # TODO: This should not be imported!!!
 import LinearAlgebra
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -62,33 +62,17 @@ using Scratch
 
 function __init__()
 display_update = true
-version_number = "5.18"
+version_number = "6"
 update_name = "update_v$(version_number)"
 update_message = """
 Update message: Agents v$(version_number)
 Welcome to this new update of Agents.jl!
 
-Noteworthy changes:
+Breaking changes:
 
-- Agents.jl moved to Julia 1.9+, and now exports visualization
-  and interactive applications automatically once Makie (or Makie backends
-  such as GLMakie) come into scope, using the new package extension system.
-  The only downside of this is that now to visualize ABMs on open street
-  maps, the package OSMMakie.jl must be explicitly loaded as well.
-  InteractiveDynamics.jl is now obsolete.
-- Several performance improvements all across the board.
-- The `@agent` macro is now THE way to create agent types for Agents.jl simulations since
-  now supports declaring default and constant fields. Directly creating structs by hand is 
-  no longer mentioned in the documentation at all. This will allow us in the future to utilize
-  additional fields that the user does not have to know about, which may bring new features or
-  performance gains by being part of the agent structures.
-- DEI-motivated name change for all names that remove agents:
-    - `genocide! -> remove_all!`
-    - `kill_agent! -> remove_agent!`
-    - `UnkillableABM -> UnremovableABM`
-- We have created an objective fully automated framework for comparing open source
-  agent based modelling software. It shows that Agents.jl is much faster
-  than competing alternatives (MASON, NetLogo, Mesa).
+- Agents in `ContinuousSpace` now require `SVector` for their `pos`
+  and `vel` fields instead of `NTuple`.
+  Using `NTuple`s in `ContinuousSpace` is now deprecated.
 
 See the online documentation for more!
 """
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -12,7 +12,7 @@ using DataStructures
 using Graphs
 using DataFrames
 using Random
-using StaticArrays
+using StaticArraysCore
 import ProgressMeter
 import Base.length # TODO: This should not be imported!!!
 import LinearAlgebra
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -196,7 +196,11 @@ macro agent(new_name, base_type, super_type, extra_fields)
     # This macro was generated with the guidance of @rdeits on Discourse:
     # https://discourse.julialang.org/t/
     # metaprogramming-obtain-actual-type-from-symbol-for-field-inheritance/84912
-
+    if base_type isa Expr
+        if base_type.args[1] == :ContinuousAgent && length(base_type.args) == 2
+            base_type = Expr(base_type.head, base_type.args..., :Float64)
+        end
+    end
     # We start with a quote. All macros return a quote to be evaluated
     quote
         let
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -196,6 +196,8 @@ macro agent(new_name, base_type, super_type, extra_fields)
     # This macro was generated with the guidance of @rdeits on Discourse:
     # https://discourse.julialang.org/t/
     # metaprogramming-obtain-actual-type-from-symbol-for-field-inheritance/84912
+
+    # hack for backwards compatibility (PR #846)
     if base_type isa Expr
         if base_type.args[1] == :ContinuousAgent && length(base_type.args) == 2
             base_type = Expr(base_type.head, base_type.args..., :Float64)
--- a/src/core/model_abstract.jl
+++ b/src/core/model_abstract.jl
@@ -21,7 +21,7 @@ SpaceType = Union{Nothing,AbstractSpace}
 ValidPos = Union{
     Int, # graph
     NTuple{N,Int}, # grid
-    NTuple{M,<:AbstractFloat}, # continuous
+    SVector{M,<:AbstractFloat}, # continuous
     Tuple{Int,Int,Float64} # osm
 } where {N,M}
 
--- a/src/core/model_abstract.jl
+++ b/src/core/model_abstract.jl
@@ -21,7 +21,6 @@ SpaceType = Union{Nothing,AbstractSpace}
 ValidPos = Union{
     Int, # graph
     NTuple{N,Int}, # grid
-    NTuple{M,<:AbstractFloat}, # continuous
     SVector{M,<:AbstractFloat}, # continuous
     Tuple{Int,Int,Float64} # osm
 } where {N,M}
--- a/src/core/model_abstract.jl
+++ b/src/core/model_abstract.jl
@@ -21,6 +21,7 @@ SpaceType = Union{Nothing,AbstractSpace}
 ValidPos = Union{
     Int, # graph
     NTuple{N,Int}, # grid
+    NTuple{M,<:AbstractFloat}, # continuous
     SVector{M,<:AbstractFloat}, # continuous
     Tuple{Int,Int,Float64} # osm
 } where {N,M}
--- a/src/core/model_concrete.jl
+++ b/src/core/model_concrete.jl
@@ -174,11 +174,17 @@ function do_checks(::Type{A}, space::S, warn::Bool) where {A<:AbstractAgent, S<:
         elseif space_type <: GridSpace && !(pos_type <: NTuple{D,Integer} where {D})
             throw(ArgumentError("`pos` field in agent type must be of type `NTuple{Int}` when using GridSpace."))
         elseif space_type <: ContinuousSpace
-            if !(pos_type <: SVector{D,<:AbstractFloat} where {D} || (!isconcretetype(A) && pos_type <: SVector{D} where {D}))
+            if pos_type <: NTuple{D,<:AbstractFloat} where {D}
+                warn && @warn "Using `NTuple` for the `pos` and `vel` fields of agent types in ContinuousSpace is deprecated. Use `SVector` instead."
+            elseif !(pos_type <: SVector{D,<:AbstractFloat} where {D} || (!isconcretetype(A) && pos_type <: SVector{D} where {D}))
                 throw(ArgumentError("`pos` field in agent type must be of type `SVector{<:AbstractFloat}` when using ContinuousSpace."))
             end
             if any(isequal(:vel), fieldnames(A)) &&
-               !(fieldtype(A, :vel) <: SVector{D,<:AbstractFloat} where {D} || (!isconcretetype(A) && fieldtype(A, :vel) <: SVector{D} where {D}))
+               !(
+                    fieldtype(A, :vel) <: NTuple{D,<:AbstractFloat} where {D} ||
+                    fieldtype(A, :vel) <: SVector{D,<:AbstractFloat} where {D} ||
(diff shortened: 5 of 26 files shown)


src/Agents.jl
```diff
@@ -12,6 +12,7 @@ using DataStructures
 using Graphs
 using DataFrames
 using Random
+using StaticArrays
```
Datseris, line 15: StaticArraysCore is what we need here

Datseris:
for me the central point of concern is whether this is breaking or not. It is right? Code that was using continuous agents before this PR will not work with it, right? A trivial MWE would be making a new type of continuous agent inheriting from `ContinuousAgent` using `@agent`, and using `move_agent!` on it to move it to a _user specified_ position.

mastrof:
Yes, this is breaking right now. It can be surely made non-breaking for any code that was using `@agent`.
e.g. creation
```julia
@agent Foo ContinuousAgent{1} begin; end
Foo(; id=1, pos=(1.0,), vel=(1.0,))
```
works because `pos` and `vel` are promoted to `SVector`s.
Also arithmetics mixing `NTuples` and `SVector` everything gets promoted to `SVector`. So with just some minor modifications (like keeping `NTuples{D,Float64}` in the `ValidPos`) it should be alright.

The problem is for structs defined manually where `pos` and `vel` are hardcoded as `NTuple`.
`ContinuousAgent` has to explicitly support both `SVector` and `NTuple` if we want it to be non-breaking, and eventually we can drop `NTuple` later when/if you want to move to v6.

Datseris:
> ContinuousAgent has to explicitly support both SVector and NTuple if we want it to be non-breaking. 

`ContinuousAgent` itself isn't an issue here. It is not an abstarct type any ways. As you correctly pointed out `@agent`  will propagate the change to agents made with `@agent`. By the way, this kind of situation is **exactly** why we want to make `@agent`  the only supported way to make agents. It is just too messy to let users make agents manually.

> The problem is for structs defined manually where pos and vel are hardcoded as NTuple.

Yes, but why, or where, would this be a problem? What we need to do is identify which parts of the code will fail *provided that* `NTUple{Float}` is in the `ValidPos` union. You mention that most operations don't have any problem due to automatic promotion of tuples to SVectors. But I would still like to have a MWE in the test suite to show explicitly what fails and what not.

mastrof:
You are right, this works without problems (not pushed yet, will test properly and polish in coming days).
```julia
julia> mutable struct Foo <: AbstractAgent
           id::Int
           pos::NTuple{1,Float64}
           vel::NTuple{1,Float64}
       end

julia> model = StandardABM(Foo, ContinuousSpace((10,)))
┌ Warning: `vel` field in agent type should be of type `SVector{<:AbstractFloat}` when using ContinuousSpace.
└ @ Agents ~/.julia/dev/Agents/src/core/model_concrete.jl:183
StandardABM with 0 agents of type Foo
 space: periodic continuous space with (10.0,) extent and spacing=0.5
 scheduler: fastest

julia> add_agent!((2.0,), model, (-1.0,))
Foo(1, (2.0,), (-1.0,))

julia> move_agent!(model[1], (1.0,), model)
Foo(1, (1.0,), (-1.0,))
```
 Just had to disable the `ArgumentError` on model construction and allow `pos2cell` to accept any `ValidPos`. I guess we want to throw a deprecation warning for `NTuple`

mastrof:
Shall we also change `ContinuousSpace` to make `extent` a `SVector`?

Datseris:
yes and yes!

src/core/model_concrete.jl
```diff
@@ -174,7 +174,11 @@ function do_checks(::Type{A}, space::S, warn::Bool) where {A<:AbstractAgent, S<:
         elseif space_type <: GridSpace && !(pos_type <: NTuple{D,Integer} where {D})
             throw(ArgumentError("`pos` field in agent type must be of type `NTuple{Int}` when using GridSpace."))
         elseif space_type <: ContinuousSpace || space_type <: ContinuousSpace
-            if !(pos_type <: SVector{D,<:AbstractFloat} where {D})
+            if pos_type <: NTuple{D,<:AbstractFloat} where {D}
+                if warn
+                    @warn "Using `NTuple` for the `pos` field of agent types in `ContinuousSpace` is deprecated. Please consider using `SVector` instead."
```
Datseris, line 179:
```suggestion
                    @warn "Using `NTuple` for the `pos` field of agent types in `ContinuousSpace` is deprecated. Use `SVector` instead."
```
being too polite here can be misleading.

Project.toml
```diff
@@ -22,6 +22,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Rotations = "6038ab10-8711-5258-84ad-4b1120ba62dc"
 Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
```
Datseris, line 24: We can remove `StaticArrays`, now that we have the `Core` version.

mastrof:
Still need to change `ContinuousSpace` and remove some explicit `SVector` conversions here and there.
I'll also need to add more tests to make sure that everything works properly for pathfinder and random walks

Tortar:
I resolved the conflits due to #847 :-)

Project.toml
mastrof: We still need `StaticArrays` for `setindex` that we use in the `rotate` functions for the random walks

mastrof:
I think we should be there, or at least close. Most basic operations seem to work without breakages, both for SVectors and Tuples

Datseris (approved):
This seems good to me, however because it is a really large change, I think it could use a second pair of eyes. Would be wonderful if anyone (e.g., @Tortar @AayushSabharwal @fbanning @lmiq , or anyone really) could also provide a PR review.

---

@mastrof I see that this PR has not changed the `docs` folder at all. Shouldn't you adjust the Flocking example, which in principle should be made much simpler with this PR, as the SVectors can have dot products and everythong else out of the box?

lmiq:
One thing that pops up upon a quick view is if one wants to support any kind of float or only `Float64` for the position and velocity fields. That seems, at first sight, contradictory in different files (not that this was introduced in this PR, it was already like this with Tuples)

Datseris:
It doesn't cost us anything to define `ContinuousAgent{D,T}` that parameterizes `SVector{D,T}`. And then define `ContinuousAgent{D} = ContinuousAgent{D, Float64}`. Could you please add these lines @mastrof and add one very simple test for non FLoat64 numbers?

mastrof:
> Shouldn't you adjust the Flocking example, which in principle should be made much simpler with this PR, as the SVectors can have dot products and everythong else out of the box?

Hmm sorry I fail to see how SVectors change anything here. Also `dot` (for which I don't see any equivalent in the code) works on tuples too, so.. not sure

Datseris:
okay sure, but how about the change of parameterizing on the number type `T`?

ameligrana (changes requested):
I'm seeing a 3x slowdown in flocking-small and 2x slowdown in flocking-large in the [ABM_Comparisons Repository](https://github.com/JuliaDynamics/ABM_Framework_Comparisons) using this branch against `main`, I would say that this is not acceptable, can it be addressed somehow?

```julia
#before pr:
Agents.jl Flocking-small (ms): 11.547424
Agents.jl Flocking-large (ms): 65.103132
#after pr:
Agents.jl Flocking-small (ms): 35.399955
Agents.jl Flocking-large (ms): 136.164522
```

Tortar:
ok not a problem, changing the implementation to account for the new type regains the previous performance :-)

mastrof:
```julia
@agent ContinuousAgent{D,T} NoSpaceAgent begin
    pos::SVector{D,T}
    vel::SVector{D,T}
end
ContinuousAgent{D} = ContinuousAgent{D,Float64}
```
This gives `LoadError: invalid redefinition of constant ContinuousAgent`

<br><br><br>

```julia
@agent ContinuousAgent{D,T} NoSpaceAgent begin
    pos::SVector{D,T}
    vel::SVector{D,T}
end
ContinuousAgent{D}(args...; kwargs...) where D = ContinuousAgent{D,Float64}(args...; kwargs...)
```
This works, but 
```julia
julia> @agent Foo ContinuousAgent{2} begin; end

julia> (; zip(fieldnames(Foo), fieldtypes(Foo))...)
(id = Int64, pos = StaticArraysCore.SVector{2}, vel = StaticArraysCore.SVector{2}) 
```
which means that `pos` and `vel` become `SVector{2,Float64}` only after an agent is created
```julia
julia> Foo(; id=1, pos=SVector(2.0,2.0), vel=rand(SVector{2})).pos |> typeof
SVector{2, Float64} (alias for SArray{Tuple{2}, Float64, 1, 2})

julia> Foo(; id=1, pos=(2.0,2.0), vel=Tuple(rand(2))).pos |> typeof
SVector{2, Float64} (alias for SArray{Tuple{2}, Float64, 1, 2})
```
which I think it's not what we want.

I don't know if there are other options to try

Tortar:
very good point, I added a commit, now your second solution should work hopefully!

Tortar:
There is still a problem with how `do_checks` is implemented since e.g. `!(pos_type <: SVector{D,<:AbstractFloat} where {D})` throws a error when one uses a ContinuousAgent just specyfing the dimension, so it should be changed somehow to account for this.
Tests fail for this

Tortar:
Actually I noticed now that all of this is unfortunately **breaking**:

e.g.

```
using Agents
@agent Foo ContinuousAgent{2} begin end
a = Foo(1, (1.0, 1.0), (1.0, 1.0))
a.pos == (1.0, 1.0) # this was true, not it is false since pos is a SVector
```

Datseris:
Hm... I understand. I think unfortunately with this point @Tortar is right. This should yield a 6.0 release. So far everything in this PR is a-ok however, so it _should_ be merged.

Can someone please clarify what is going on right now in the **current** source code? If someone inherits 
```
@agent Foo ContinuousAgent{2} begin end
```

what does ` (; zip(fieldnames(Foo), fieldtypes(Foo))...)` return? Are the fields **concretely typed**?

Datseris:
Also, please re-export `SVector` in this PR.

Tortar:
yes, they are concretely typed, it now returns `(id = Int64, pos = Union{Tuple{Float64, Float64}, StaticArraysCore.SVector{2, Float64}}, vel = Union{Tuple{Float64, Float64}, StaticArraysCore.SVector{2, Float64}})`

Anyway all compatibility changes with `NTuple` should be dropped since they are not useful since even with them the code is breaking, so it is better not to have them.

src/spaces/continuous.jl
```diff
@@ -99,16 +100,24 @@ end
...
     euclidean_distance(pos, cell_center(pos, model), model)
 
+
+# required for backward compatibility with NTuples in ContinuousSpace
+function add_agent!(A::Type{<:AbstractAgent}, model::ABM{S}, properties::Vararg{Any, N};
+    kwargs...) where {N,S<:ContinuousSpace}
+    T = (; zip(fieldnames(A), fieldtypes(A))...)[:pos]
+    add_agent!(T(random_position(model)), A, model, properties...; kwargs...)
+end
+
```
Tortar, line 120:
```suggestion

```

src/spaces/walk.jl
```diff
@@ -72,14 +74,25 @@ spaces this clamps the position to the space extent.
...
+# for backward compatibility 
+function normalize_position(pos::NTuple{D}, space::ContinuousSpace{D,true}) where {D}
+    return Tuple(mod.(pos, spacesize(space)))
+end
+
+function normalize_position(pos::NTuple{D}, space::ContinuousSpace{D,false}) where {D}
+    return Tuple(clamp.(pos, 0.0, prevfloat.(spacesize(space))))
+end
+#----
+
```
Tortar, line 95:
```suggestion
```

src/core/model_abstract.jl
```diff
@@ -22,6 +22,7 @@ ValidPos = Union{
     Int, # graph
     NTuple{N,Int}, # grid
     NTuple{M,<:AbstractFloat}, # continuous
+    SVector{M,<:AbstractFloat}, # continuous
     Tuple{Int,Int,Float64} # osm
 } where {N,M}
```
Tortar, line 27:
```suggestion
    Int, # graph
    NTuple{N,Int}, # grid
    SVector{M,<:AbstractFloat}, # continuous
    Tuple{Int,Int,Float64} # osm
} where {N,M}
```

ameligrana (commented):
All of these + all transformations with `typeof(agent.pos)` which are not necessary now that we commit to `SVector`

mastrof:
> Also, please re-export `SVector` in this PR.

Since we have to keep `StaticArrays` anyway for `setindex`, should I drop `StaticArraysCore`?

Datseris:
> Since we have to keep `StaticArrays` anyway for `setindex`, should I drop `StaticArraysCore`?

Yes.




> This is on the last commit dropping the tuple support:
> 
> ```julia
> julia> @agent Foo ContinuousAgent{2} begin; end
> julia> (; zip(fieldnames(Foo), fieldtypes(Foo))...)
> (id = Int64, pos = SVector{2, Float64}, vel = SVector{2, Float64})
> ```

I don't understand. How can it be that inhereting form `ContinuousAgent{2}` leads to concretely typed fields, but `ContinuousAgent{2}` itself does not have concretely typed fields...?

mastrof:
Because `ContinuousAgent{D}` *is* parametric, atm we just made it so that its constructor generates a concrete type with the default `T=Float64`. In the macro we are just adding the `Float64` explicitly when it is not declared.

src/core/agents.jl
```diff
@@ -196,7 +196,11 @@ macro agent(new_name, base_type, super_type, extra_fields)
     # This macro was generated with the guidance of @rdeits on Discourse:
     # https://discourse.julialang.org/t/
     # metaprogramming-obtain-actual-type-from-symbol-for-field-inheritance/84912
-
+    if base_type isa Expr
+        if base_type.args[1] == :ContinuousAgent && length(base_type.args) == 2
+            base_type = Expr(base_type.head, base_type.args..., :Float64)
+        end
+    end
```
Datseris, line 203: Okay, now I see. This is a hack that we should not do. It is unintuitive. by reading the usage code it was super confusing to me how this could happen, but it turns out to be a macro magic hack.

Of course, the problem is that without the hack pretty much every continuous space code will break. While now, pretty much nothing will break, except when a user explicitly compares agent positions to a hand-coded position `a.pos == (0.5, 0.2)`. 

So, the conclusion is: we leave this hack in, but we do not use it in the documentation. The docstring of `ContinuousAgent` explicitly mentions the **two** type parameters `D, T`. In all examples that use `ContinuousAgent` we explicitly inherit from `ContinuousAgent{D, Float64}` with appropriate `D`.

Additionally, we should add one more check in `do_checks` that the continous space extent type is also of type T.

Datseris: And, we add a comment just above line 199 in the macro that says "this is a hack for backwrards compatibilty, see PR 846."

Tortar:
> > yes, they are concretely typed, it now returns `(id = Int64, pos = Union{Tuple{Float64, Float64}, StaticArraysCore.SVector{2, Float64}}, vel = Union{Tuple{Float64, Float64}, StaticArraysCore.SVector{2, Float64}})`
> 
> ????? This is **NOT** concretely typed. Union types are not concretely typed.
> 
> This cannot be allowed, if someone uses just `ContinuousAgent{2}` in a simulation, it will lead to type instability.

ops, sorry I was attempting a change locally where it was a Union which is of course not concretely typed :D, but here, it was concretely typed because of the macro hack I wrote myself precisely for this, sorry for the confusion

mastrof:
Just to be clear, given that we are moving to 6.0 anyway, are we completely dropping support for tuples (https://github.com/JuliaDynamics/Agents.jl/pull/846/commits/03e4a009f71b7f82b2e2d4d1f088b5372650f108) or should I revert that given that everything except manual comparison to tuples will work?

Tortar:
I would say removing the support is better since we are going to a new major version. 

A concern though is: manual comparison to tuples is very important, it is obviously something not happening often but it's probably something happened already (e.g. a set of user defined tuples with which compare the position of the agent).  The problem is that **this is a silent bug since it doesn't through any error**. So we should make the user aware that this doesn't work somehow, I thought about a possible strategy:

- When creating an agent inhering from `ContinuousAgent{D}` we warn the user with something along the lines: "ContinuousAgent{Int} is deprecated, use instead ContinuousAgent{Int, AbstractFloat}. Be aware that now the position and velocity fields of the agent are of type `SVector`, so any comparison between these fields and tuples will return false, possibly breaking your past code. Read about this change in version 6.0 at {link to docs}"

This is I think very important

Datseris:
**There is no reason to drop support for tuples**. This makes user experience better and allows them to use the package while changing minimal code. It makes the transition simpler.

We should just make sure that `ContinuousAgent` by itself does not include tuples and also that nowhere in the documentation we use tuples for continuous space.

I don't agree with displaying a message in `@agent`, The macro already has too much hocus pocus. We display an overall update message once the user installs version 6.0, as per usual we have done it like this always in Agents.jl. After all, updating to a new major version means it is up to the user to learn about the breaking changes.

Datseris:
And we should also never use `ContinuousAgent{D}` in the docs; only `ContinuousAgent{D, Float64}`. In the update message `ContinuousAgent` we will sya that for backwoards compatibility, there is currently a hack that allows agents inhereting from `ContinuousAgent{D}` to be type stable. In 6.1 this hack will be removed and these agents will be type unstable.

Datseris:
@mastrof isn't the support for tuples minimal anyways? Most of the code work, you only need to leave tuples in the ValidPos type and take them into account in `do_checks`, right?

Tortar:
I'm not particularly fond of the approach, but let's do it that way then! 

But I think that @mastrof implemented some more code for retro-compatibility in https://github.com/JuliaDynamics/Agents.jl/commit/03e4a009f71b7f82b2e2d4d1f088b5372650f108 which doesn't seem necessary, I think that @Datseris is right that to keep it we just need adding ntuples to `ValidPos` and take it into account in `do_checks`

mastrof:
With last commit, we should have everything we wanted:
- `ContinuousAgent{D,T}` is now used everywhere instead of `ContinuousAgent{D}`
- `ContinuousSpace` now has `extent` as an `SVector` (even if it is passed a `NTuple`)
- `NTuple` is accepted for the `pos`and `vel` fields but throws a deprecation warning so that old code using manually defined agent types should not break
- `SVector` is reexported

So we will only break code with agents defined by `@agent Foo ContinuousAgent` where the `pos` and `vel` fields are compared for in\equality against tuples.

As per the last comments, right now this backward support is not exactly "minimal", but its super-extensive either. It's just that there are a few places (e.g. when interfacing with the space extent or with the underlying grid space) where one needs to be careful with Tuple<->SVector conversions, because they happen automatically only one way.
It is possible that there are a few unnecessary conversions, but I'm not sure if they can be avoided *completely* just by spamming ValidPos in all the function signatures.

Tortar:
I removed some transformations but then realized that the consensus is that we should still support also agents manually defines as e.g.
```julia
mutable struct Foo <: AbstractAgent
           id::Int
           pos::NTuple{1,Float64}
           vel::NTuple{1,Float64}
       end
```

so rollbacked the changes, to me it seems good, thanks @mastrof to change back and forth between different versions without giving up :D 

I think that it should still be updated the changelog and the message in the `Agents.jl` file

ameligrana (approved):

Datseris:
@mastrof thanks a lot for your work on this huge pr!

mastrof:
thank you guys for the help and suggestions :)

PR #856, opened 2023-08-22
Fix model creation with ContinuousAgent{D}

### LINKED ISSUE #855: Unnoticed breaking change in #846: broken model creation with `ContinuousAgent{D}`
On main this now errors:
```julia
julia> ABM(ContinuousAgent{2}, space)
┌ Warning: Agent type is not concrete. If your agent is parametrically typed, you're probably
│ seeing this warning because you gave `Agent` instead of `Agent{Float64}`
│ (for example) to this function. You can also create an instance of your agent
│ and pass it to this function. If you want to use `Union` types for mixed agent
│ models, you can silence this warning.
└ @ Agents ~/.julia/dev/Agents/src/core/model_concrete.jl:139
ERROR: ArgumentError: `pos` field in agent type must be of the same type of the `extent` field in ContinuousSpace.
Stacktrace:
 [1] do_checks(#unused#::Type{ContinuousAgent{2}}, space::ContinuousSpace{2, true, Float64, typeof(Agents.no_vel_update)}, warn::Bool)
   @ Agents ~/.julia/dev/Agents/src/core/model_concrete.jl:191
 [2] agent_validator(#unused#::Type{ContinuousAgent{2}}, space::ContinuousSpace{2, true, Float64, typeof(Agents.no_vel_update)}, warn::Bool)
   @ Agents ~/.julia/dev/Agents/src/core/model_concrete.jl:147
 [3] Agents.SingleContainerABM(::Type{ContinuousAgent{2}}, space::ContinuousSpace{2, true, Float64, typeof(Agents.no_vel_update)}; container::Type, scheduler::typeof(Agents.Schedulers.fastest), properties::Nothing, rng::Random.TaskLocalRNG, warn::Bool)
   @ Agents ~/.julia/dev/Agents/src/core/model_concrete.jl:47
 [4] Agents.SingleContainerABM(::Type{ContinuousAgent{2}}, space::ContinuousSpace{2, true, Float64, typeof(Agents.no_vel_update)})
   @ Agents ~/.julia/dev/Agents/src/core/model_concrete.jl:38
 [5] AgentBasedModel(::Type, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Agents ~/.julia/dev/Agents/src/deprecations.jl:27
 [6] AgentBasedModel(::Type, ::Vararg{Any})
   @ Agents ~/.julia/dev/Agents/src/deprecations.jl:27
 [7] top-level scope
   @ REPL[24]:1
```
We had agreed to never use `ContinuousAgent{D}` without specifying the second type `T`, but with the back and forth between different versions and the additional checks on the space type at some point it has slipped through and seems we have broken compatibility for model creation. Now since `ContinuousAgent{2}` is not concrete, we fail the check against the space `extent`.

I guess we should just add an exception to the `do_checks` function

mastrof:
Closes #855.
I added an extra check to the `do_checks` function to allow the creation of models with calls of the form `ABM(ContinuousAgent{D}, ContinuousSpace((1,1))` for backward compatibility, even though this now results in model being not concretely typed.
This change is also mentioned in the welcome message and in the warning for agents with non-concrete types.

I used a regex for the check instead of subtyping because I did not find another easy way to *only* get the case where `ContinuousAgent{D}` is used with a Float64 space.

### DIFF
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -73,6 +73,11 @@ Breaking changes:
 - Agents in `ContinuousSpace` now require `SVector` for their `pos`
   and `vel` fields instead of `NTuple`.
   Using `NTuple`s in `ContinuousSpace` is now deprecated.
+- `ContinuousAgent{D}` is not a concrete type anymore.
+  The new interface requires two parameters `ContinuousAgent{D,T}`
+  where `T` is any `AbstractFloat` type.
+  If you want to use a type different from `Float64`, you will also
+  need to change the type of the `ContinuousSpace` extent accordingly.
 
 See the online documentation for more!
 """
--- a/src/core/model_concrete.jl
+++ b/src/core/model_concrete.jl
@@ -142,6 +142,9 @@ function agent_validator(
         (for example) to this function. You can also create an instance of your agent
         and pass it to this function. If you want to use `Union` types for mixed agent
         models, you can silence this warning.
+        If you are using `ContinuousAgent{D}` as agent type in version 6+, update
+        to the new two-parameter version `ContinuousAgent{D,Float64}` to obtain
+        the same behavior as previous Agents.jl versions.
         """
         for type in union_types(A)
             do_checks(type, space, warn)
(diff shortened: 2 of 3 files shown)


Tortar:
thanks!

Issue #853, opened 2023-08-22
Bug in periodic continuous spaces due to... floating point error?


mastrof:
`move_agent!` and `walk!` throw `position is outside of space extent` error, apparently due to floating point approximation

**Minimal Working Example**
```julia
using Agents
pos = (0.4936805014512577, 0.007750092779602313)
vel = (0.6319498548742291, -0.7750092779602327)
dt = 0.01
space = ContinuousSpace((1,1); periodic=true)
model = ABM(ContinuousAgent{2,Float64}, space)
add_agent!(pos, model, vel)
move_agent!(model[1], model, dt)
```

This particular MWE is on main, but I originally encountered the error on a code running 5.14.0.

What happens is the following (let `agent = model[1]`):
the direction of motion is evaluated as
```julia
julia> direction = agent.vel .* dt
2-element SVector{2, Float64} with indices SOneTo(2):
  0.006319498548742291
 -0.007750092779602327
```
then the "raw" target position would be
```julia
julia> target_raw = agent.pos .+ direction
2-element SVector{2, Float64} with indices SOneTo(2):
  0.5
 -1.3877787807814457e-17
```
which, when passed to `normalize_position`, becomes
```julia
julia> target = Agents.normalize_position(agent.pos .+ direction, model)
2-element SVector{2, Float64} with indices SOneTo(2):
 0.5
 1.0
```
which errors since 1.0 is not a valid position

mastrof:
I guess this is a very fringe case if it never came up but I have a set of fairly large simulations that simply cannot be run due to this approximation error.

Tortar:
Good catch. 

What about changing

`all(i -> 0 <= pos[i] < space_size[i], 1:D) || error("position is outside space extent!")`

in 

`all(i -> 0 <= pos[i] <= space_size[i], 1:D) || error("position is outside space extent!")`

does this have any unwanted side effects?

Tortar:
But changing also

`pos2cell(pos::ValidPos, model::ABM) = Tuple(floor.(Int, pos./abmspace(model).spacing) .+ 1)`

to 

`pos2cell(pos::ValidPos, model::ABM) = Tuple(ceil.(Int, pos./abmspace(model).spacing))`

it seems to work fine. 

The problem is that this is slightly breaking since for an integer position inside the grid, this will change the result of the position of the agent in the grid

Tortar:
[...]

PR #857, opened 2023-08-22
Fix bug in pos2cell

Tortar:
This should fix the bug described in #853.

Maybe it is good if we put the code in the issue with which this was found inside a test. 

@mastrof can you please confirm that this solves the problem with your code?

### DIFF
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -101,7 +101,7 @@ end
 
 "given position in continuous space, return cell coordinates in grid space."
 pos2cell(a::AbstractAgent, model::ABM) = pos2cell(a.pos, model)
-pos2cell(pos::ValidPos, model::ABM) = Tuple(floor.(Int, pos./abmspace(model).spacing) .+ 1)
+pos2cell(pos::ValidPos, model::ABM) = Tuple(max.(1, ceil.(Int, pos./abmspace(model).spacing)))
 
 "given position in continuous space, return continuous space coordinates of cell center."
 function cell_center(pos::ValidPos, model)
@@ -141,7 +141,7 @@ end
 function move_agent!(agent::A, pos::ValidPos, model::ABM{<:ContinuousSpace{D},A}
     ) where {D, A<:AbstractAgent}
     space_size = spacesize(model)
-    all(i -> 0 <= pos[i] < space_size[i], 1:D) || error("position is outside space extent!")
+    all(i -> 0 <= pos[i] <= space_size[i], 1:D) || error("position is outside space extent!")
     oldcell = pos2cell(agent, model)
     newcell = pos2cell(pos, model)
     if oldcell ≠ newcell
--- a/test/continuous_space_tests.jl
+++ b/test/continuous_space_tests.jl
@@ -48,6 +48,14 @@ using LinearAlgebra: norm, dot
         # remove
         remove_agent!(agent, model)
         @test nagents(model) == 0
+        # test edge case described in #853
+        space = ContinuousSpace(SVector(1,1); periodic=true)
+        model = ABM(ContinuousAgent{2,Float64}, space)
+        pos = SVector(0.4936805014512577, 0.007750092779602313)
+        vel = SVector(0.6319498548742291, -0.7750092779602327)
+        add_agent!(pos, model, vel)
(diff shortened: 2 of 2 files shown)


Tortar:
This is a little breaking though since I think that anyway if the agent is at an integer position inside the space, it will change the grid position it occupies which in turn would make the approximate nearby search (the one when calling `nearby_ids`) possibly return different agents, this is a very rare case though, and the function is an approximation anyway and we are going for 6.0 so...

alternatively we can use the previous way with a similar transformation, but I think it would be a little less nice and a little less fast, let me know your opinion

Tortar:
this is the same but without the little change in behaviour

```julia
pos2cell(pos::ValidPos, model::ABM) = Tuple(min.(floor.(Int, pos./abmspace(model).spacing) .+ 1, size(abmspace(model).grid)))
```

src/spaces/continuous.jl
```diff
@@ -141,7 +141,7 @@ end
 function move_agent!(agent::A, pos::ValidPos, model::ABM{<:ContinuousSpace{D},A}
     ) where {D, A<:AbstractAgent}
     space_size = spacesize(model)
-    all(i -> 0 <= pos[i] < space_size[i], 1:D) || error("position is outside space extent!")
+    all(i -> 0 <= pos[i] <= space_size[i], 1:D) || error("position is outside space extent!")
```
Datseris, line 144: why was equal added here? The extent end is not in the space, because it is no covered by the underlying grid space.

src/spaces/continuous.jl
```diff
@@ -101,7 +101,7 @@ end
 
 "given position in continuous space, return cell coordinates in grid space."
 pos2cell(a::AbstractAgent, model::ABM) = pos2cell(a.pos, model)
-pos2cell(pos::ValidPos, model::ABM) = Tuple(floor.(Int, pos./abmspace(model).spacing) .+ 1)
+pos2cell(pos::ValidPos, model::ABM) = Tuple(max.(1, ceil.(Int, pos./abmspace(model).spacing)))
```
Datseris, line 104: Is the `max(1, ...)` necessary since we do ceiling? is it even possible for the computation to return an integer 0? That would mean that the `pos ./ ... ` would return a negative number. Which isn't possible.

Datseris (commented):
The small change in behavior in edge cases is fine.

Datseris:
also, this returns a Tuple? Shouldnt it return an `SVector`.?

src/spaces/continuous.jl
Tortar: See https://github.com/JuliaDynamics/Agents.jl/issues/853#issuecomment-1688351895

src/spaces/continuous.jl
Tortar: since floating point approximation makes it possible to arrive at the extent unfortunately by https://github.com/JuliaDynamics/Agents.jl/issues/853

src/spaces/continuous.jl
Datseris: actually, `pos` can be 0 in continuous space. so I guess we do need the `max`. Oh man, this is so expensive though. We call the `max` function for any position, even though the probability of having exactrly 0 is so small. Maybe it leads to better performance to first do a check of `newpos = ceil...` and then `any(iszero, pos)` and if `true` then do the `max.(1, newpos)`?

Datseris:
we should measure the performance regression this has. If it is really impactful, we can honestly consider if simply not fixing #853 and accept that the edge case bug will forever exist.

src/spaces/continuous.jl
Tortar: I know, I'm also upset about this :D 
But it seems not to have actually any impact at all, consider that we have more costly operations inside the functions which call it, and also using `ceil` actually spares the `+1` we needed when we used it before

Tortar:
this could well even be just random variations 

```julia
#before
Agents.jl step just moves 200 agents (ms): 0.849313
#after
Agents.jl step just moves 200 agents (ms): 0.855405
```

we can live with them :D

mastrof:
My code seems to run alright with this fix

Datseris (approved):
oik

PR #858, opened 2023-08-23
Faster sample! function

Tortar:
This is up to 2x faster than before because it avoids to use `DataStructures.counter` which in turn use a dict to do the operations


Datseris (approved):
add a changelog entry as well please!

Issue #869, opened 2023-09-06
Design the new `@agent` macro


Tortar:
I'm designing the new macro (I will dispatch on the same macro `@agent`, with the old syntax deprecated in favor of the new one, but both usable with `@agent`)

Currently we agreed to something like:

```julia
@agent GridAgent{2} struct NewAgent <: Foo
    x::Int
    const y::Int
end
```

but I don't like it since it is not intuitive so I propose something like 

```julia
@agent struct NewAgent <: Foo
    GridAgent{2}
    x::Int
    const y::Int
end
```

which to me seems much better

Tortar:
@Datseris I would really like to hear your opinion on this :-)

Tortar:
I’m even inclined to export a utility function `fieldsof` to improve it further to 

```
@agent struct NewAgent <: Foo
    fieldsof(ContinuousAgent)
    x::Int
    const y::Int
end
```

Datseris:
I still think 
```
@agent GridAgent{2} struct NewAgent <: Foo
    x::Int
    const y::Int
end
```

is the way to go because it doesn't confuse a Type with the fields. I also think it is the most intuitive way to go.

Datseris:
how does default values to fields work here? can you also do 

```
@agent GridAgent{2} struct NewAgent <: Foo
    x::Int
    const y::Int
   z::Real = 0.5
end
```

Tortar:
Unfortunately this doesn't still work, but it could (talking about default values). But it's becoming a bit messy now in a sense since with #871 using default values would become almost useless. Indeed I think we should rediscuss #861 for this and other reasons which now are clear to me. 

However for what regards the macro I still really like the alternative with the other syntax though, since it makes it clear that the base_type is used only to paste its field in the new struct, notice that even

```julia
grid_agent_fields = fieldsof(GridAgent{2})
@agent struct NewAgent <: Foo 
    grid_agent_fields 
    x::Int
    const y::Int
    z::Real
end
```

works, which seems very clean and easy to grasp.

Tortar:
> But it's becoming a bit messy now in a sense since with #871 using default values would become almost useless. Indeed I think we should rediscuss #861 for this and other reasons which now are clear to me.

however I think we can hold everything together actually after some thought

Tortar:
Sorry for the back and forth, actually everything works in respect to default values

Tortar:
[...]

PR #870, opened 2023-09-06
New version of agent macro

Tortar:
Per #869 

e.g. this now works

```julia
abstract type Foo end
@agent struct NewAgent(GridAgent{2}) <: Foo
    x::Int = 1
    const y::Int
end
```


Tortar:
When https://github.com/JuliaLang/julia/issues/51224 is solved the user can choose to even do

```julia
ContinuousAgentFields = fieldsof(ContinuousAgent{2,Float64})
@agent struct NewAgent <: Foo
    ContinuousAgentFields
    x::Int
    const y::Int
end
```

which is cool (even if not really necessary) :-)

Issue #861, opened 2023-08-27
Disable manual assignment of ID in version 6.


Datseris:
Since we are going to a breaking 6.0 release due to the continuous space SVector change, we might as well ensure that there is no way for the user to manually add id to agents. cc @Tortar and discussions in #445 and #841

Datseris:
This means we need to deprecate the `add_agent!(a::Agent, ....)` function, as this is the only way to add user-created agents ot the model. Otherwise `add_agent!(model, ...)` does the trick.

Tortar:
I think this would be best! More we have control of the fields used internally the better

Tortar:
Implemented in #871

PR #871, opened 2023-09-08
Deprecate adding agent by add_agent!(a::Agent, ....)

Tortar:
By #861 

Still need to fix 4 tests and improve warnings messages


Datseris:
The way this was conveyd in the changelog is inappropriate. The crucial information to convey is the breaking change that one cannot manually mess with, or set, agent ids. This has to be its own dedicated entry that starts with this infiromation. 

Please create a bullet point that writes: 

Manually setting or altering the ids of agents is no longer allowed. The agent id is now considered a read-only field, and is set internally by Agents.jl to enable hidden optimizations in the future. As a consequence, `add_agent!(agent::AbstractAgent, pos::ValidPos, model::ABM)` and `add_agent!(agent::AbstractAgent, model::ABM)` have been deprecated. See issue #861 for more.


The fact that `add_agent!` now allows keyword propagation is a completely separate topic and hence should be its own bullet point.

Tortar:
you are totally right, will do that when I also solve the flocking example

Issue #872, opened 2023-09-08
Not install GraphMakie when installing the library


Tortar:
GraphMakie is an heavy dependency which means that even if it doesn't increase the load time of `Agents` since it is only used in the `AgentsVisualizations` extension, it does increase the time to install and precompile the library the first time. So it would be better to remove it from the dependencies in the `Project.toml`. But it's not clear how to do it without removing the possibility to use the package in the extensions.

Solutions:

- Make `AgentsVisualizations` its own package inside the main repo of `Agents`. This should work, but it's breaking.
- Adding a new extension loaded by `GraphMakie`, which extend the plotting of `AgentsVisualizations`. This should work also. And it is preferable since not breaking.

cc. @Datseris

PR #863, opened 2023-08-30
Add GraphMakie dependency

asgolovin:
Fixes #827 

The plotting recipe for graphs uses the `graphplot!` function from GraphMakie, but GraphMakie is not listed in dependencies and is never imported. This very short PR fixes this. 

I added GraphMakie as a weak dependency since this is how other Makie libraries are listed, hope that's okay. :)

PR #874, opened 2023-09-09
New extension AgentsGraphVisualizations

Tortar:
Solves #872

Now the code in #863 I posted works without needing `GraphMakie` as a dependency!


Tortar:
I'm almost sure it's ok so I merge :-)

Issue #837, opened 2023-07-25
Extend sampling function to more than one item


Tortar:
Currently we support only single versions of specialized sampling function e.g. `resorvoir_sampling_single`, `sampling_with_condition_single`, but we could support also the multi-versions so that the new function sample without replacement for more than one item

Tortar:
I think that here we can even create a new package specialized for these sampling function, the reason is that these sampling function aren't useful only for agent based modeling, they can be used with any iterator

Tortar:
created: https://github.com/Tortar/IteratorSampling.jl

it will be be easier to solve this issue with it

Tortar:
With `iteratorsampling.jl` it should be easy to create a fast version of multiple elements sampling functions, the problem is our current naming convetions which assume only one agent e.g. `random_agent`, `random_nearby_agent`, do you have any idea on how to handle this @Datseris ?

Datseris:
What does multiple elements mean here. what should `random_agents` mean for multiple elements?

Tortar:
say instead of a single random agent, between all nearby agents, you want two of them, currently you either need to repeat (and maybe reject) random_agent two times or collect and sample. Instead if we talk about `random_agents(model, n)` this would return `n` random agents in the model. Do you think that "duplicating" `random_agent`, ecc... with `random_agents`, ecc... could be acceptable?

Datseris:
yes, that would be acceptable. We cross ref the two functions in their docsttrings.

PR #875, opened 2023-09-10
new internal API for sampling

Tortar:
This will solve #837 when finished


Tortar:
This will actually be solved when I manage to find the time to finish the work in https://github.com/Tortar/IteratorSampling.jl

hopefully soon

Tortar:
Everything works now, so I think it should good to go

I needed to add some more Pkg tricks though because I noticed that if you didn't have internet connection the solution in #955 didn't work, I verified that instead works like this. This is only temporary, when we release v6 we will rapidly delete all of this.

I made our sampling depending from https://github.com/Tortar/IteratorSampling.jl where the sampling process is much more well tested than how it was in our library, so I guess it's okay to depend on it :-)

Anyway there we have also fast sampling for multiple items, but here I didn't include it because it requires some more work

Tortar:
actually there are some ensemble tests which didn't fail locally, not sure why at the moment

Tortar:
mmmmh, they seem actually totally unrelated

src/simulations/paramscan.jl
```diff
@@ -101,6 +101,11 @@ function paramscan(
 
     progress = ProgressMeter.Progress(length(combs); enabled = showprogress)
     mapfun = parallel ? pmap : map
+    println(combs)
+    println(mapfun)
+    println(output_params)
+    println(initialize)
+    println(kwargs)
```
Datseris, line 108: Mistake?

src/core/space_interaction_API.jl
```diff
@@ -380,69 +377,13 @@ is expensive since in this case the allocating version can be more performant.
 function random_nearby_position(pos, model, r=1, f = nothing, alloc = false; kwargs...)
     iter = nearby_positions(pos, model, r; kwargs...)
     if isnothing(f)
-        return resorvoir_sampling_single(iter, model)
+        return itsample(abmrng(model), iter)
```
Datseris, line 380: Does `itsample` come from your package? I would say, preface the functions that come from your package with `IteratorSampling.function_name`.

src/core/space_interaction_API.jl
```diff
@@ -304,8 +304,7 @@ nearby_agents(a, model, r = 1; kwargs...) =
 
 """
     random_nearby_id(agent, model::ABM, r = 1, f = nothing, alloc = false; kwargs...) → id
-Return the `id` of a random agent near the position of the given `agent` using an optimized
-algorithm from [Reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling#An_optimal_algorithm).
+Return the `id` of a random agent near the position of the given `agent`.
```
Datseris, line 307: It's fine to not say too many details, but i think it is still improtant to say that this function does not require collecting all the agents and randomly sampling one.

Datseris (changes requested):
Have you checked for performance regressions? The introduction of `Iteartors.Filter` e.g. could add some.

src/simulations/paramscan.jl
Tortar: yes, forgot about removing them

src/core/space_interaction_API.jl
Tortar: yes, good idea

src/core/space_interaction_API.jl
Tortar: reintroduced in a similar form

Tortar:
checked briefly with this code

```julia
using Agents, BenchmarkTools

@agent A GridAgent{2} begin
end

agent_step!(a,m) = nothing

function aperiodic_model(; griddims = (100, 100))
    space = GridSpace(griddims, periodic = false)
    model = ABM(A, space; agent_step!)
    fill_space!(model)
    return model
end

function periodic_model(; griddims = (100, 100))
    space = GridSpace(griddims, periodic = true)
    model = ABM(A, space; agent_step!)
    fill_space!(model)
    return model
end

@btime random_nearby_id($(1,1), model, 10) setup=(model=periodic_model());
@btime random_nearby_id($(50,50), model, 10) setup=(model=periodic_model());
@btime random_nearby_id($(1,1), model, 1) setup=(model=aperiodic_model());
@btime random_nearby_id($(50,50), model, 1) setup=(model=aperiodic_model());
```

no perf regression, same perf, in general I expect the same for all the other function apart from a little improvement where I introduced the new `Iterators.filter` in `random_empty_pos_in_offsets` because it is a fallback when there is a small number of empty cells, in this case using the other methodology  (`sampling_with_condition_single` function) is bad for performance.

So I think this PR should be good to go as it is. Let's see in the future if we want to support optimized multi sampling procedures or maybe give to the users some advice on how to perform them by themselves. Maybe we can discuss that in our call too

Datseris:
yes, this is good from me as well, but all tests fail!

Tortar:
I know but the tests that fail seem spurious, this doesn't happen locally, it happens also in #946 and #957

Tortar:
If you can try it out too locally it would be better to conclude about that something is strange just on github

Datseris (approved):
If tests pass locally I trust you. We would need to figure out how to solve this before v6. (if tests fail on `main` after merging this, we should also open an issue to keep track of things.

Tortar:
tests passed when you merged, so I think it alright

PR #879, opened 2023-09-11
Mixed boundary conditions

### LINKED ISSUE #828: Mixed boundary conditions
I'm working on a model with mixed boundary conditions, e.g. a 2D model which is periodic along the x direction, but has closed walls along the y direction.

What I'm doing right now is to use a periodic space and then re-define the behavior at the y boundary in the agent stepping function.
A bit clunky but it's ok.

I was wondering if there's any interest in properly implementing such behavior directly in the library?
"In principle" it doesn't seem too complicated, the main change would be to have `normalize_position` operate differently on each axis in a loop according to some `dims` argument.
I'm not yet sure, however, how one could deal with the `Space` for the dispatch.
Options I thought of:
- define new space types e.g. `ContinuousSpaceMixedBC`, `GridSpaceMixedBC` etc.. which by default have `periodic=true` and an extra field `dims` that defines which dimensions are periodic
- add the extra field `dims` to the already existing space types, taking default value `Tuple(1:D)` for a D-dimensional space so that if not specified everything would work exactly as it does now

The second option would be more elegant, but the first one would be immediate to add and doesn't require any change to the existing code.

Let me know if you think this could be useful

mastrof:
Closes #828.
WIP, there are still a lot of things to fix.
I tried to represent space periodicity with a new type `Periodic{D}` to only be used internally, which would just wrap a SVector with a Bool value for each spatial dimension. When given a `Bool` e.g. `Periodic{D}(true)` it is internally converted to a SVector as `fill(true, SVector{D})`.
I don't have a smart way to fall back to "full periodic" and "non periodic" via dispatch yet, so as it is now every time distances have to be evaluated, each dimensions needs to be checked for whether or not it is periodic. In low-dimensional systems (which I assume make up the vast majority of use cases) there is no performance difference; in high-dimensional ones it might start to become noticeable (although I don't know to what extent this is really an issue...)

Low-dimensional space:
```julia
julia> # mixed-boundary
       space = GridSpace(ntuple(i->5,2); periodic=true)
       model = ABM(GridAgent{2}, space)
       add_agent!(model)
       add_agent!(model)
       @btime euclidean_distance($(model[1]), $(model[2]), $(model))
  2.374 ns (0 allocations: 0 bytes)

julia> # main
       space = GridSpace(ntuple(i->5,2); periodic=true)
       model = ABM(GridAgent{2}, space)
       add_agent!(model)
       add_agent!(model)
       @btime euclidean_distance($(model[1]), $(model[2]), $(model))
  2.304 ns (0 allocations: 0 bytes)
```

High-dimensional space:
```julia
julia> # mixed-boundary
       space = GridSpace(ntuple(i->5,8); periodic=true)
       model = ABM(GridAgent{8}, space)
       add_agent!(model)
       add_agent!(model)
       @btime euclidean_distance($(model[1]), $(model[2]), $(model))
  8.389 ns (0 allocations: 0 bytes)

julia> # main
       space = GridSpace(ntuple(i->5,8); periodic=true)
       model = ABM(GridAgent{8}, space)
       add_agent!(model)
       add_agent!(model)
       @btime euclidean_distance($(model[1]), $(model[2]), $(model))
  5.168 ns (0 allocations: 0 bytes)
```

An alternative approach (probably better and possibly requiring less work) is to not have a `Periodic` type, just let spaces accept whatever is passed to them (be it a Bool, a tuple an svector etc...). If it's found to be a Bool, then they will just behave as usual, but if they are passed an iterable type they will do the checking.

### DIFF
--- a/src/spaces/grid_general.jl
+++ b/src/spaces/grid_general.jl
@@ -172,7 +172,7 @@ function nearby_positions(
             checkbounds(Bool, stored_ids, (n .+ pos)...) ?
             n .+ pos : mod1.(n .+ pos, space_size)
             for n in nindices
-            if all(checkbounds(Bool, axes(stored_ids,i), n[i]+pos[i]) || P[i] for i in 1:D)
+            if all(P[i] || checkbounds(Bool, axes(stored_ids,i), n[i]+pos[i]) for i in 1:D)
         )
     end
 end
--- a/src/spaces/grid_general.jl
+++ b/src/spaces/grid_general.jl
@@ -17,8 +17,6 @@ to vector of indices within each radius.
 """
 abstract type AbstractGridSpace{D,P} <: DiscreteSpace end
 
-isperiodic(::AbstractGridSpace{D,P}) where {D,P} = isperiodic(P)
-
 """
     GridAgent{D} <: AbstractAgent
 The minimal agent struct for usage with `D`-dimensional [`GridSpace`](@ref).
--- a/src/spaces/grid_general.jl
+++ b/src/spaces/grid_general.jl
@@ -157,6 +157,26 @@ function nearby_positions(
                 n .+ pos : mod1.(n .+ pos, space_size) for n in nindices)
     end
 end
+function nearby_positions(
+    pos::ValidPos, space::AbstractGridSpace{D,P}, r = 1,
+    get_indices_f = offsets_within_radius_no_0 # NOT PUBLIC API! For `ContinuousSpace`.
+) where {D,P}
+    stored_ids = space.stored_ids
+    nindices = get_indices_f(space, r)
+    space_size = size(space)
+    # check if we are far from the wall to skip bounds checks
+    if all(i -> r < pos[i] <= space_size[i] - r, 1:D)
+        return (n .+ pos for n in nindices)
+    else
+        return (
+            checkbounds(Bool, stored_ids, (n .+ pos)...) ?
+            n .+ pos : mod1.(n .+ pos, space_size)
+            for n in nindices
+            if all(checkbounds(Bool, axes(stored_ids,i), n[i]+pos[i]) || P[i] for i in 1:D)
+        )
+    end
+end
+
 
 function random_nearby_position(pos::ValidPos, model::ABM{<:AbstractGridSpace{D,false}}, r=1; kwargs...) where {D}
     nindices = offsets_within_radius_no_0(abmspace(model), r)
--- a/src/spaces/grid_general.jl
+++ b/src/spaces/grid_general.jl
@@ -17,6 +17,8 @@ to vector of indices within each radius.
 """
 abstract type AbstractGridSpace{D,P} <: DiscreteSpace end
 
+isperiodic(::AbstractGridSpace{D,P}) where {D,P} = isperiodic(P)
+
 """
     GridAgent{D} <: AbstractAgent
 The minimal agent struct for usage with `D`-dimensional [`GridSpace`](@ref).
--- a/src/spaces/grid_multi.jl
+++ b/src/spaces/grid_multi.jl
@@ -64,13 +64,15 @@ example for usage of this advanced specification of dimension-dependent distance
 where one dimension is used as a categorical one.
 """
 function GridSpace(
-        d::NTuple{D,Int}; periodic = true, metric::Symbol = :chebyshev
+        d::NTuple{D,Int};
+        periodic::Union{Bool,SVector{D,Bool},NTuple{D,Bool}} = true,
+        metric::Symbol = :chebyshev
     ) where {D}
     stored_ids = Array{Vector{Int},D}(undef, d)
     for i in eachindex(stored_ids)
         stored_ids[i] = Int[]
     end
-    return GridSpace{D,Periodic{D}(periodic)}(
+    return GridSpace{D,periodic}(
         stored_ids,
         metric,
         Dict{Int,Vector{NTuple{D,Int}}}(),
--- a/src/spaces/grid_multi.jl
+++ b/src/spaces/grid_multi.jl
@@ -192,13 +192,12 @@ combine_positions(pos, origin, ::GridSpaceIdIterator{false}) = pos .+ origin
     # checking for bounds is less expensive than calling mod1
     checkbounds(Bool, stored_ids, npos...) ? npos : mod1.(npos, space_size)
 end
-function combine_positions(pos, origin, iter::GridSpaceIdIterator{P,D}) where {P,D}
+@inline function combine_positions(pos, origin, iter::GridSpaceIdIterator{P,D}) where {P,D}
     npos = pos .+ origin
     space_size = iter.space_size
     stored_ids = iter.stored_ids
     ntuple(i ->
-        #(P[i] && checkbounds(Bool, stored_ids, npos[i])) ?
-        (!P[i] || checkbounds(Bool, stored_ids, npos[i])) ?
+        (!P[i] || checkbounds(Bool, axes(stored_ids, i), npos[i])) ?
         npos[i] : mod1(npos[i], space_size[i]),
         D
     )
@@ -211,8 +210,8 @@ combine_positions_nocheck(pos, origin, ::GridSpaceIdIterator) = pos .+ origin
     return !valid_bounds || @inbounds isempty(iter.stored_ids[pos_index...])
 end
 invalid_access(pos_index, iter::GridSpaceIdIterator{true}) = @inbounds isempty(iter.stored_ids[pos_index...])
-function invalid_access(pos_index, iter::GridSpaceIdIterator{P,D}) where {P,D}
-    valid_bounds = any((!P[i] && checkbounds(Bool, iter.stored_ids, pos_index[i]) for i in 1:D))
+@inline function invalid_access(pos_index, iter::GridSpaceIdIterator{P,D}) where {P,D}
+    valid_bounds = all(P[i] || checkbounds(Bool, axes(iter.stored_ids, i), pos_index[i]) for i in 1:D)
     return !valid_bounds || @inbounds isempty(iter.stored_ids[pos_index...])
 end
 invalid_access_nocheck(pos_index, iter::GridSpaceIdIterator) = @inbounds isempty(iter.stored_ids[pos_index...])
--- a/src/spaces/grid_multi.jl
+++ b/src/spaces/grid_multi.jl
@@ -65,7 +65,7 @@ where one dimension is used as a categorical one.
 """
 function GridSpace(
         d::NTuple{D,Int};
-        periodic::Union{Bool,SVector{D,Bool},NTuple{D,Bool}} = true,
+        periodic::Union{Bool,NTuple{D,Bool}} = true,
         metric::Symbol = :chebyshev
     ) where {D}
     stored_ids = Array{Vector{Int},D}(undef, d)
@@ -192,6 +192,17 @@ combine_positions(pos, origin, ::GridSpaceIdIterator{false}) = pos .+ origin
     # checking for bounds is less expensive than calling mod1
     checkbounds(Bool, stored_ids, npos...) ? npos : mod1.(npos, space_size)
 end
+function combine_positions(pos, origin, iter::GridSpaceIdIterator{P,D}) where {P,D}
+    npos = pos .+ origin
+    space_size = iter.space_size
+    stored_ids = iter.stored_ids
+    ntuple(i ->
+        #(P[i] && checkbounds(Bool, stored_ids, npos[i])) ?
+        (!P[i] || checkbounds(Bool, stored_ids, npos[i])) ?
+        npos[i] : mod1(npos[i], space_size[i]),
+        D
+    )
+end
 combine_positions_nocheck(pos, origin, ::GridSpaceIdIterator) = pos .+ origin
 
 # Must return `true` if the access is invalid
@@ -200,6 +211,10 @@ combine_positions_nocheck(pos, origin, ::GridSpaceIdIterator) = pos .+ origin
     return !valid_bounds || @inbounds isempty(iter.stored_ids[pos_index...])
 end
 invalid_access(pos_index, iter::GridSpaceIdIterator{true}) = @inbounds isempty(iter.stored_ids[pos_index...])
+function invalid_access(pos_index, iter::GridSpaceIdIterator{P,D}) where {P,D}
+    valid_bounds = any((!P[i] && checkbounds(Bool, iter.stored_ids, pos_index[i]) for i in 1:D))
+    return !valid_bounds || @inbounds isempty(iter.stored_ids[pos_index...])
+end
 invalid_access_nocheck(pos_index, iter::GridSpaceIdIterator) = @inbounds isempty(iter.stored_ids[pos_index...])
 
 ##########################################################################################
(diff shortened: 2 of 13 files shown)


mastrof:
> An alternative approach (probably better and possibly requiring less work) is to not have a Periodic type, just let spaces accept whatever is passed to them (be it a Bool, a tuple an svector etc...). If it's found to be a Bool, then they will just behave as usual, but if they are passed an iterable type they will do the checking.

Ok I think this is the way to go. With last commit there absolutely no changes for everything that uses `periodic=true` or `periodic=false`. Now I will just need to have some fun adding tests everywhere for the mixed boundary case.

For `ContinuousSpace` with a mixed boundary, I also reimplemented the display function to write "mixed-periodicity continuous space...". Maybe it's better to explicitly write which dimensions are periodic? I wonder why the display for continuous spaces is different from the more schematic one adopted for grid spaces (where now the display will show e.g. `periodic=(true,false)`)?

src/spaces/grid_multi.jl
```diff
@@ -64,7 +64,9 @@ example for usage of this advanced specification of dimension-dependent distance
 where one dimension is used as a categorical one.
 """
 function GridSpace(
-        d::NTuple{D,Int}; periodic::Bool = true, metric::Symbol = :chebyshev
+        d::NTuple{D,Int};
+        periodic::Union{Bool,SVector{D,Bool},NTuple{D,Bool}} = true,
```
Tortar, line 68:
```suggestion
        periodic::Union{Bool,SVector{D,Bool},NTuple{D,Bool}} = true,
```

why allowing also a `SVector`? In Gridspaces it seems better just to use tuples

mastrof: I allowed both `SVector` and `NTuple` because the `periodic` argument of `ContinuousSpace` goes through `GridSpace` anyway so thought we might as well allow both.
It's also true that there is no real reason to support `SVector` even for `ContinuousSpace`, since it's just extra keystrokes in this case, but there's no real reason to not support it either. Any iterable with fixed size is the same here.
If preferred we can just limit to `NTuple`s overall.

mastrof:
Is there anything I'm missing concerning places where periodicity is used to compute things? I tried grepping my way through the package but didn't see anything obvious.
Although I don't really know too well what goes on e.g. in the nearby id searches or in the pathfinder; I guess missing features will come up while I write tests, but if you know/notice something missing please point it out :)

src/spaces/grid_multi.jl
Tortar: In my opinion since there is not a good reason to support it, I would not support it

mastrof:
Can I get any help with this? I don't understand what's going on. 
How does this even work if `bound_range` is only defined for non-periodic spaces?

```julia
##########################################################################################
# nearby_stuff with special access r::Tuple
##########################################################################################
# TODO: We can re-write this to create its own `indices_within_radius_tuple`.
# This would also allow it to work for any metric, not just Chebyshev!

function nearby_ids(pos::ValidPos, model::ABM{<:GridSpace}, r::NTuple{D,Int}) where {D}
    # simply transform `r` to the Vector format expected by the below function
    newr = [(i, -r[i]:r[i]) for i in 1:D]
    nearby_ids(pos, model, newr)
end

function nearby_ids(
    pos::ValidPos,
    model::ABM{<:GridSpace},
    r::Vector{Tuple{Int64, UnitRange{Int64}}},
)
    @assert abmspace(model).metric == :chebyshev
    dims = first.(r)
    vidx = []
    for d in 1:ndims(abmspace(model).stored_ids)
        idx = findall(dim -> dim == d, dims)
        dim_range = isempty(idx) ? Colon() :
            bound_range(pos[d] .+ last(r[only(idx)]), d, abmspace(model))
        push!(vidx, dim_range)
    end
    s = view(abmspace(model).stored_ids, vidx...)
    Iterators.flatten(s)
end

function bound_range(unbound, d, space::GridSpace{D,false}) where {D}
    return range(max(unbound.start, 1), stop = min(unbound.stop, size(space)[d]))
end
```

mastrof:
I think I got a decent implementation for the nearby searches; with 6 dimensions the performance for a mixed-boundary model is pretty close to that of a full periodic model.
```julia
julia> bt_pure = @belapsed nearby_positions(
        $(1,5,1,5,1,5),
        $(ABM(GridAgent{6}, GridSpace((5,5,5,5,5,5); metric=:euclidean, periodic=true)))
    )

julia> bt_mix = @belapsed nearby_positions(
        $(1,5,1,5,1,5),
        $(ABM(GridAgent{6}, GridSpace((5,5,5,5,5,5); metric=:euclidean, periodic=(true,false,true,false,true,false))))
    )

pure: 1.483567134268537e-8
mix: 1.7302908726178538e-8
```

I guess AStar is the only missing thing now

src/spaces/grid_general.jl
```diff
@@ -157,6 +157,26 @@ function nearby_positions(
...
+    space_size = size(space)
+    # check if we are far from the wall to skip bounds checks
+    if all(i -> r < pos[i] <= space_size[i] - r, 1:D)
+        return (n .+ pos for n in nindices)
+    else
+        return (
+            checkbounds(Bool, stored_ids, (n .+ pos)...) ?
+            n .+ pos : mod1.(n .+ pos, space_size)
+            for n in nindices
+            if all(checkbounds(Bool, axes(stored_ids,i), n[i]+pos[i]) || P[i] for i in 1:D)
```
Tortar, line 175: If I remember correctly logical operator are evaluated left-to-right so this should be maybe more performant if the compiler doesn't reorder them automatically

```suggestion
            if all(P[i] || checkbounds(Bool, axes(stored_ids,i), n[i]+pos[i]) for i in 1:D)
```

src/spaces/grid_single.jl
```diff
@@ -117,6 +120,29 @@ function nearby_ids(pos::NTuple{D, Int}, model::ABM{<:GridSpaceSingle{D,false}},
...
+    if all(i -> r < pos[i] <= space_size[i] - r, 1:D)
+        ids_iterator = (stored_ids[p...] for p in position_iterator
+                        if stored_ids[p...] != 0)
+    else
+        ids_iterator = (
+            checkbounds(Bool, stored_ids, p...) ?
+            stored_ids[p...] : stored_ids[mod1.(p, space_size)...]
+            for p in position_iterator
+            if stored_ids[mod1.(p, space_size)...] != 0 &&
+            all(checkbounds(Bool, axes(stored_ids, i), p[i]) || P[i] for i in 1:D)
```
Tortar, line 140:
```suggestion
            all(P[i] || checkbounds(Bool, axes(stored_ids, i), p[i]) for i in 1:D)
```

src/spaces/grid_multi.jl
```diff
@@ -190,6 +192,17 @@ combine_positions(pos, origin, ::GridSpaceIdIterator{false}) = pos .+ origin
     # checking for bounds is less expensive than calling mod1
     checkbounds(Bool, stored_ids, npos...) ? npos : mod1.(npos, space_size)
 end
+function combine_positions(pos, origin, iter::GridSpaceIdIterator{P,D}) where {P,D}
+    npos = pos .+ origin
+    space_size = iter.space_size
+    stored_ids = iter.stored_ids
+    ntuple(i ->
+        #(P[i] && checkbounds(Bool, stored_ids, npos[i])) ?
+        (!P[i] || checkbounds(Bool, stored_ids, npos[i])) ?
```
Tortar, line 201:
```suggestion
        (!P[i] || checkbounds(Bool, axes(stored_ids,i), npos[i])) ?
```

mastrof:
There's one of the tests in the AStar (L176 in astar_tests.jl) which I marked as wrong, but after some further testing I realized that I don't understand what the expected behavior is supposed to be. I wonder if there is a bug somewhere else?
Consider this case, using a `periodic=true` space (so not affected by anything I implemented here):
```julia
julia> space = ContinuousSpace((5., 5.); periodic=true)
periodic continuous space with [5.0, 5.0] extent and spacing=0.25

julia> pf = AStar(space; walkmap=trues(10,10))
A* in 2 dimensions, periodic, diagonal, ϵ=0.0, metric=DirectDistance

julia> model = ABM(ContinuousAgent{2,Float64}, space; properties=(pf=pf,))
StandardABM with 0 agents of type ContinuousAgent
 space: periodic continuous space with [5.0, 5.0] extent and spacing=0.25
 scheduler: fastest
 properties: pf

julia> a = add_agent!((0.,0.), model, (0.,0.))
ContinuousAgent{2, Float64}(1, [0.0, 0.0], [0.0, 0.0])

julia> plan_best_route!(a, [SVector(0.0, 1.0)], model.pf)
2-element SVector{2, Float64} with indices SOneTo(2):
 0.0
 1.0

julia> move_along_route!(a, model, model.pf, 0.1)
ContinuousAgent{2, Float64}(1, [0.0316227766016838, 0.09486832980505139], [0.0, 0.0])
```
When running the last command, I was expecting the agent to move to `[0.0, 0.1]`, instead it moves also along the `x` axis, even if its target destination has `x=0`. So if you keep moving along the route, the agent first strays a bit to positive `x` and then corrects by going back and eventually gets to the correct destination. But why is it moving like this? Is this really the expected behavior?

If we take a different target destination then indeed the agent is moving straight towards it, without any seemingly arbitrary deroute:
```julia
julia> a = add_agent!((0.,0.), model, (0.,0.))
ContinuousAgent{2, Float64}(2, [0.0, 0.0], [0.0, 0.0])

julia> plan_best_route!(a, [SVector(1.0, 1.0)], model.pf)
2-element SVector{2, Float64} with indices SOneTo(2):
 1.0
 1.0

julia> move_along_route!(a, model, model.pf, 0.1)
ContinuousAgent{2, Float64}(2, [0.07071067811865477, 0.07071067811865477], [0.0, 0.0])
```

src/spaces/grid_multi.jl
```diff
@@ -198,6 +211,10 @@ combine_positions_nocheck(pos, origin, ::GridSpaceIdIterator) = pos .+ origin
     return !valid_bounds || @inbounds isempty(iter.stored_ids[pos_index...])
 end
 invalid_access(pos_index, iter::GridSpaceIdIterator{true}) = @inbounds isempty(iter.stored_ids[pos_index...])
+function invalid_access(pos_index, iter::GridSpaceIdIterator{P,D}) where {P,D}
+    valid_bounds = any((!P[i] && checkbounds(Bool, iter.stored_ids, pos_index[i]) for i in 1:D))
+    return !valid_bounds || @inbounds isempty(iter.stored_ids[pos_index...])
```
Tortar, line 216:
```suggestion
    invalid_bounds = any(!P[i] && !checkbounds(Bool, axes(iter.stored_ids,i), pos_index[i]) for i in 1:D)
    return invalid_bounds || @inbounds isempty(iter.stored_ids[pos_index...])
```

Tortar: this should be the correct version (I hope :D)

src/spaces/utilities.jl
```diff
@@ -60,16 +85,40 @@ function manhattan_distance(
     p2::ValidPos,
     space::Union{ContinuousSpace{D,false},AbstractGridSpace{D,false}},
 ) where {D}
-    sum(abs.(p1 .- p2))
+    sum(manhattan_distance_direct.(p1, p2))
```
Tortar, line 88: wait your versions assume real numbers

Tortar: I would just let the method as it is

src/spaces/utilities.jl
```diff
@@ -43,6 +43,31 @@ function euclidean_distance(
     sqrt(sum(min.(direct, spacesize(space) .- direct).^2))
 end
 
+function euclidean_distance(
+    p1::ValidPos,
+    p2::ValidPos,
+    space::Union{ContinuousSpace{D,P},AbstractGridSpace{D,P}}
+) where {D,P}
+    s = spacesize(space)
+    distance_squared = 0.0
```
Tortar, line 52: is this better?

```suggestion
    distance_squared = zero(eltype(p1))
```

mastrof: `zero(float(eltype(p1))` should be the right version I guess

Tortar: actually I don't think so, since if p1 is a tuple of ints then it is better to have 0 instead of 0.0

src/spaces/grid_multi.jl
```diff
@@ -190,6 +192,17 @@ combine_positions(pos, origin, ::GridSpaceIdIterator{false}) = pos .+ origin
     # checking for bounds is less expensive than calling mod1
     checkbounds(Bool, stored_ids, npos...) ? npos : mod1.(npos, space_size)
 end
+function combine_positions(pos, origin, iter::GridSpaceIdIterator{P,D}) where {P,D}
```
Tortar, line 195:
```suggestion
@inline function combine_positions(pos, origin, iter::GridSpaceIdIterator{P,D}) where {P,D}
```

src/spaces/grid_multi.jl
```diff
@@ -198,6 +211,10 @@ combine_positions_nocheck(pos, origin, ::GridSpaceIdIterator) = pos .+ origin
     return !valid_bounds || @inbounds isempty(iter.stored_ids[pos_index...])
 end
 invalid_access(pos_index, iter::GridSpaceIdIterator{true}) = @inbounds isempty(iter.stored_ids[pos_index...])
+function invalid_access(pos_index, iter::GridSpaceIdIterator{P,D}) where {P,D}
```
Tortar, line 214:
```suggestion
@inline function invalid_access(pos_index, iter::GridSpaceIdIterator{P,D}) where {P,D}
```

mastrof:
Regarding my last comment, atm I just made the pathfinder test consistent with what the library is currently doing. Whether that is correct or not is a different topic. I implemented all the review comments (thanks for that @Tortar), and I think we have everything now

Tortar:
> I think I got a decent implementation for the nearby searches; with 6 dimensions the performance for a mixed-boundary model is pretty close to that of a full periodic model.
> 
> ```julia
> julia> bt_pure = @belapsed nearby_positions(
>         $(1,5,1,5,1,5),
>         $(ABM(GridAgent{6}, GridSpace((5,5,5,5,5,5); metric=:euclidean, periodic=true)))
>     )
> 
> julia> bt_mix = @belapsed nearby_positions(
>         $(1,5,1,5,1,5),
>         $(ABM(GridAgent{6}, GridSpace((5,5,5,5,5,5); metric=:euclidean, periodic=(true,false,true,false,true,false))))
>     )
> 
> pure: 1.483567134268537e-8
> mix: 1.7302908726178538e-8
> ```
> 

note that you need to run collect otherwise you just compare the creation of the iterators

Tortar (approved):

Tortar:
Great work anyway :-)

mastrof:
You are right :s then there is some noticeable performance difference (1.5e-7 s and 5e-7 s on my machine with the above benchmark in 6 dimensions). Not as bad as I initially thought it would be, but not sure it can get much better.

Changelog was done btw: https://github.com/JuliaDynamics/Agents.jl/pull/879/commits/b2136fc2aafa8c51ed43de21513c5d057de76ad9

Tortar:
it's okay since in the end for the majority of points the time should be the same, on the boundary this perf difference seems legit

PR #882, opened 2023-09-14
Fix wraparound behaviour when r is big

### LINKED ISSUE #566: nearby_ids returns the same ID multiple times when r is bigger than the space size

Tortar:
This fixes #566. 

Need tests and the implementation should be simplified.

### DIFF
--- a/src/spaces/grid_general.jl
+++ b/src/spaces/grid_general.jl
@@ -37,7 +37,50 @@ npositions(space::AbstractGridSpace) = length(space.stored_ids)
 # We initialize a vector of tuples of indices within radius `r` from origin position.
 # We store this vector. When we have to loop over nearby_stuff, we call this vector
 # and add it to the given position. That is what the concrete implementations of
-# nearby_stuff do in the concrete spaces files.
+# nearby_stuff do in the concrete spaces files. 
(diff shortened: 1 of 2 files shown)


Tortar:
Actually the tests already hits almost all lines

Tortar:
the new tests made me realize that there is still some bugs in the implementation :P

Tortar:
there is still some sort of bug :D

Tortar:
I searched for it, but I didn't find it, maybe it was removed in some PR

PR #885, opened 2023-09-21
Agent macro field inheritance

### LINKED ISSUE #844: Inherit fields with default values when using `@agent`
```julia
julia> @agent Foo1 GridAgent{2} begin
           a = 1
       end

julia> @agent Foo2 Foo1 begin
           b = 2
       end

julia> Foo1(id=1, pos=(0,0))
Foo1(1, (0, 0), 1)

julia> Foo2(id=1, pos=(0,0))
ERROR: UndefKeywordError: keyword argument `a` not assigned
Stacktrace:
 [1] top-level scope
   @ REPL[19]:1
```
I think it would be nice to have `Foo2` inherit the default value of `a` from `Foo1`, is it problematic to implement?

(Sorry for the various requests concerning the `@agent` macro but I'm a bit out of my depth here and **really** appreciate all the work you guys put into this package)

mastrof:
Extends the `@agent` macro to inherit default values of base types.
Closes #844.


mastrof:
With https://github.com/JuliaDynamics/Agents.jl/pull/885/commits/2f5c0e0089582642578769f1b79d90e32359d1d6 I commented out the Models submodule because the macro is not yet working with calls of the form `@agent Foo(GridAgent{2})`.
The other base agent types are correctly generated on loading Agents.jl and their expressions stored as
```julia
julia> __AGENT_GENERATOR__
Dict{Symbol, JLKwStruct} with 5 entries:
  :GraphAgent      => begin mutable struct GraphAgent <: Agents.AbstractAgent; $(Expr(:const, id::…
  :GridAgent       => begin mutable struct GridAgent{D} <: Agents.AbstractAgent; $(Expr(:const, id…
  :OSMAgent        => begin mutable struct OSMAgent <: Agents.AbstractAgent; $(Expr(:const, id::In…
  :NoSpaceAgent    => begin mutable struct NoSpaceAgent;  $(Expr(:const, id::Int)); end; begin fun…
  :ContinuousAgent => begin mutable struct ContinuousAgent{D, T} <: Agents.AbstractAgent; $(Expr(:…
```

mastrof:
Also noticed that with the new form of the macro (#870) not even inheritance from parametric types is working currently, which instead was working trivially with the old macro version

src/Agents.jl
```diff
@@ -11,6 +11,8 @@ using Distributed
 using DataStructures
 using Graphs
 using DataFrames
+using ExproniconLite
+export @expr, JLKwStruct, JLKwField, codegen_ast
```
Tortar, line 15:
```suggestion
```

I don't think we need to export these functions, right?

Tortar:
Will clone the branch soon and try to solve the problems you are mentioning!

src/Agents.jl
mastrof: For the way it's written now I think it's needed because the expr has to be evaluated in main scope and these functions need to be available. There might be a way around it though

Tortar:
mmmh there is actually a problem we didn't identify when considering the alternative which I think it could be unsolvable with this methodology, parametric types can't work this way, take

```julia
abstract type Foo end
@agent struct NewAgent(GridAgent{2}) <: Foo
    x::Int = 1
    const y::Int
end
```

this is not in `__AGENT_GENERATOR__`, only `GridAgent{D}` is. This means that we can't simply do all the work to `ExproniconLite`, but we should first evaluate `GridAgent{2}` to obtain its specific fields and then rebuild the expression with the defaults, which is trickier...however it is not that simple since consider if patologically the user_defined agent is actually

```julia
abstract type Foo end
@agent struct NewAgent{D}(GridAgent{2}) <: Foo
    x::Int = D
    const y::Int
end

@agent struct AnotherAgent(NewAgent{2}) <: Foo
    z
end
```

then the default should not be only retrieved but even substituted back before returning the expression for new agent which is rather tricky. 

So I think that `ExproniconLite` is actually not helping us much in these regards, and maybe the simpler `@capture` from `MacroTools` could be enough, but the actual implementation is not obvious

Tortar:
I think a possible alternative strategy is to reconstruct only the function definition with default values which `@kwdef` creates macro-expanding it, this should be doable

mastrof:
I also eventually came to the conclusion that Expronicon is not the right tool for the job. `@macroexpand`ing kwdef to extract constructor sounds like a good option at this point, I guess the main problem then becomes to correctly parse and substitute the type parameters

Tortar:
ok, I think I got a good representation of what we need to achieve:

Trasform this

```julia
@kwdef mutable struct A{D}
    x::NTuple{D, Int} = Tuple(x for x in 1:D)
    y::Int
end

@kwdef mutable struct B{K}
    z::NTuple{K, Int} = Tuple(x for x in 1:K)
    u::Int
end
```


into

```julia
@kwdef mutable struct B{D, K}
    x::NTuple{D, Int} = Tuple(x for x in 1:D)
    y::Int
    z::NTuple{K, Int} = Tuple(x for x in 1:K)
    u::Int
end
B{K}(; kwargs...) where K = B{2, K}(; kwargs...)
```

this should the general problem

Tortar:
mmmhh I don't think what I proposed is optimal, it is probably better to go with the current way(even if I would actually avoid Expriconlite and go with MacroTools) and substitute the parameters when e.g. we saved in the dict GridAgent{2} but we have GridAgent{D} in the dict as described in https://stackoverflow.com/questions/51928959/how-to-find-and-replace-subexpression-of-ast-in-julia

I still didn't have a look at your latest commits, maybe you are already trying to do this :D

Tortar:
Good news: made it!!!

(But still very unpolished)

This should be actually even much better than #843 e.g. this now works:

```julia
using Agents
abstract type AbstractFoo{D} <: AbstractAgent where D end
@agent MyFoo2{D} ContinuousAgent{D, Float64} AbstractFoo{D} begin; end
```

No need for `where` inside the macro, and the parameters are passed better than before since before if you called `MyFoo2{2}` the types were still loosy like `pos::NTuple{D, Float64} where D` while now are like `pos::NTuple{2, Float64}`.

We even finally avoid `eval`!

Tortar:
it seems to me that the failures are actually unrelated, can you update the branch?

Tortar (approved):

src/core/agents.jl
```diff
@@ -17,6 +17,19 @@ and hence it is the **the only supported way to create agent types**.
 """
 abstract type AbstractAgent end
 
+__AGENT_GENERATOR__ = Dict{Symbol, Expr}()
+
+"""
+    NoSpaceAgent <: AbstractAgent
+The minimal agent struct for usage with `nothing` as space (i.e., no space).
```
Datseris, line 24: I am confused. Why is this docstring here? Shouldn't it be at No Space AGent...?

Datseris:
this was merged without a changelog entry even though it was a new feature and not a bugfix. In the future please ensure that merged PRs have some accompanying documentation if they introduce new features.

src/core/agents.jl
Tortar: Not to repeat again the same thing I evaluated the NoSpaceAgent expression in __AGENT_GENERATOR to create NoSpaceAgent

mastrof:
I am really amazed by your macro wizardry.

PR #886, opened 2023-09-22
Docs Updates

Tortar:
this fixes the Flocking example and improves the changelog for v6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 # v6
 - The `@agent` macro has been rewritten to support fields with default and const values. The old version still works but it's deprecated. Since now the macro supports these features, using `@agent` is the only supported way to create agent types for Agents.jl.
-- Manually setting or altering the ids of agents is no longer allowed. The agent id is now considered a read-only field, and is set internally by Agents.jl to enable hidden optimizations in the future. As a consequence, 
-`add_agent!(agent::AbstractAgent, pos::ValidPos, model::ABM)`  and `add_agent!(agent::AbstractAgent, model::ABM)`  have been deprecated. See issue #861 for more.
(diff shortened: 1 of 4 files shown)


Tortar:
removed also the comparison table in the docs

Datseris (commented):
> removed also the comparison table in the docs

Please keep PRs focused, so it is easier to backtrack. I see many changes here that are really totally unrelated to the title of the PR!

The whole comparison page needs to be overhauled to contain as little information as possible, and point to the comparison repo as much as possible. But this needs to be done in a different PR!

Tortar:
ok, let's do it this way even if we weren't that formal always e.g. here https://github.com/JuliaDynamics/Agents.jl/pull/591/files, not to do too many PRs I will just create a different pr for the comparison page, since I think this is an important change so maybe being more focused on that can be useful, and for this one I will just enlarge the scope of the PR

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # v6
 - The `@agent` macro has been rewritten to support fields with default and const values. The old version still works but it's deprecated. Since now the macro supports these features, using `@agent` is the only supported way to create agent types for Agents.jl.
-- The `add_agent!` function supports adding an agent propagating keywords arguments. `add_agent!(agent::AbstractAgent, pos::ValidPos, model::ABM)` and `add_agent!(agent::AbstractAgent, model::ABM)` have been deprecated to allow controlling the ids given to the agents, see issue #861.
+- Manually setting or altering the ids of agents is no longer allowed. The agent id is now considered a read-only field, and is set internally by Agents.jl to enable hidden optimizations in the future. As a consequence, 
+`add_agent!(agent::AbstractAgent, pos::ValidPos, model::ABM)`  and `add_agent!(agent::AbstractAgent, model::ABM)`  have been deprecated. See issue #861 for more.
```
Datseris, line 4: The formatting is off here. You are startng a new line without identation, i am not sure this will register as part of the bullet point. better to not start a new line alltogether.

Datseris (approved):
ok, address comment and merge!

PR #887, opened 2023-09-23
Remove comparison table and just direct to comparison repository

Tortar:
Probably still too many details are mentioned here.  Still the comparison repo doesn't have LOCs calculation, but will add them soon.

### DIFF
--- a/docs/src/comparison.md
+++ b/docs/src/comparison.md
@@ -2,13 +2,6 @@
 Many agent-based modeling frameworks have been constructed to ease the process of building and analyzing ABMs (see [here](http://dx.doi.org/10.1016/j.cosrev.2017.03.001) for a review).
 Notable examples are [NetLogo](https://ccl.northwestern.edu/netlogo/), [Repast](https://repast.github.io/index.html), [MASON](https://journals.sagepub.com/doi/10.1177/0037549705058073), and [Mesa](https://github.com/projectmesa/mesa).
 
-We used the following models for the comparison:
-
-- [Predator-prey dynamics](@ref) (Wolf Sheep Grass), a [`GridSpace`](@ref) model, which requires agents to be added, removed and moved; as well as identify properties of neighbouring positions.
-- The [Flocking model](@ref) (Flocking), a [`ContinuousSpace`](@ref) model, chosen over other models to include a MASON benchmark. Agents must move in accordance with social rules over the space.
-- The [Forest fire](@ref) model, provides comparisons for cellular automata type ABMs (i.e. when agents do not move). NOTE: The Agents.jl implementation of this model has been changed in v4.0 to be directly comparable to Mesa and NetLogo. As a consequence it no longer follows the [original rule-set](https://en.wikipedia.org/wiki/Forest-fire_model).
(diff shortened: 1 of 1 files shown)


docs/src/comparison.md
```diff
@@ -2,7 +2,6 @@
 Many agent-based modeling frameworks have been constructed to ease the process of building and analyzing ABMs (see [here](http://dx.doi.org/10.1016/j.cosrev.2017.03.001) for a review).
 Notable examples are [NetLogo](https://ccl.northwestern.edu/netlogo/), [Repast](https://repast.github.io/index.html), [MASON](https://journals.sagepub.com/doi/10.1177/0037549705058073), and [Mesa](https://github.com/projectmesa/mesa).
 
-In this page we compare Agents.jl with Mesa, Netlogo and Mason, to assess where Agents.jl excels and also may need some future improvement.
 We used the following models for the comparison:
```
Datseris, line 5: We also shouldn't say which models we used for comparison.

docs/src/comparison.md
```diff
@@ -12,24 +11,13 @@ We used the following models for the comparison:
...
-|Forest Fire|1|303.7x|222.2x|NA|
-|(LOC)|23|35|43 (545)|.|
-|Schelling|1|209.4x|166.4x|195.9x|
-|(LOC)|31|56|60 (743)|248|
+You can find the latest results in the `README.md` of the [ABM_Framework_Comparisons](https://github.com/JuliaDynamics/ABM_Framework_Comparisons) repository where you can find also the details on the parameters used for each comparison in the `DECLARATION.md` files inside each model subfolder. 
 
-ᕯ Netlogo has a different implementation to the other three frameworks here. It cheats a little by only choosing one nearest neighbor in some cases rather than considering all neighbors within vision. So a true comparison would ultimately see a slower result.
-
-The results clearly speak for themselves. Across all four models, Agents.jl's performance is exceptional whilst using the least amount of code. This removes many frustrating barriers-to-entry for new users, and streamlines the development process for established ones.
+Across all four models, Agents.jl's performance is exceptional whilst using the least amount of code. This removes many frustrating barriers-to-entry for new users, and streamlines the development process for established ones.
```
Datseris, line 20: Instead of saying "across all four models", we should write "In the majority of cases". This allows the possibility that there could be a case where this isn't true in the future. It also allows more models.

Datseris (changes requested):
I still think you can trim this done much more!

docs/src/comparison.md
Tortar: indeed also there are just three models now :P, but I agree on saying something less strict

docs/src/comparison.md
Tortar: should be ok now

Datseris (approved):
I am approving this, but we probably need to revisit the "tabled based comparison" as well; Firstly, move it to comparison repo and secondly make it markdown based. But this can happen elsewhere. 

(also, it is not clear to me how we would colorcode a markdown based table. Perhpas we could attach a unicode symbol to each cell denoting the "color")

Issue #864, opened 2023-09-04
Test visualization code


Tortar:
Currently no test is written for the visualization code, but there should be some: as a simple first implementation for the visualization part we can run some of the examples visualizations and see if everything works

PR #888, opened 2023-09-23
A first attempt to test visualization code

Tortar:
This is very rough, but at least we have something, so in my opinion it is nonetheless a good thing. Partially address #864 

Notice that I needed to change "Zombie" in "Zombiee" in `jld2tests.jl` because it conflicted with the "Zombie" in the examples folder. This is caused by the current design of the agent macro which always evaluates the quote at the module scope. We should fix this


Tortar:
As expected we got a big boost in coverage

Project.toml
Datseris, line 1: Ah, yeah. We need to create a dedicated `test/Project.toml` file and move all of this stuff there (see for example how we do this here: https://github.com/JuliaDynamics/ComplexityMeasures.jl ). 

The `[extras]` is a deprecated way to add test-related dependencies.

Datseris (commented):
Evaluating code coverage in the plots folder was a mistake. I didn't realize it. We need to remove this, irrespectiveily of whether we test the plotting or not.

See here for how to remove folders from the coverage: https://github.com/JuliaDynamics/Attractors.jl/blob/main/codecov.yml

Project.toml
Tortar: done!

Tortar:
I think everything should be fine now (if tests passes)

Tortar:
ok I probably need to test it locally instead :D

Tortar:
Actually now the boost is even better :D

Issue #883, opened 2023-09-15
Incorporate dynamic rule (stepping functions) into the AgentBasedModel types


Datseris:
The rest of the dynamic modelling ecosystem such as DifferentialEquations.jl and DynamicalSystems.jl incorporates the dynamic rule (or equations of motion or stepping functions) into the central type such as `ODEProblem` or `DynamicalSystem`. Agents.jl does not do this. 

We _should_ do this though. This has several benefits:

1. It makes a thematic connection to the rest of the modelling ecosystem, hence lowering the learning curve
2. It will allow us to have more flexibility when attempting to solve #841 etc.
3. It will allow us to implement #760 , as there it is clear that the current setup of sequentially activating agents once per step is simply not suitable

This is a breaking change, however, we may be able to do this change without breaking anything, and simply deprecating stuff. However, lots of care and testing should be done for this.

Crucially, this should be done before 6.0. Crucially, we should check for performance regressions when doing this.

Issue #760, opened 2023-03-07
Implementing the Gillespie algorithm


MA-Ramirez:
Implementing the Gillespie Algorithm in the agents.jl framework is more complex than I thought.
I realised this by coding a new example, the Spatial Rock-Paper-Scissors game. See pull request JuliaDynamics/AgentsExampleZoo.jl#17.

In this issue, I outline the algorithm, the attempted implementation, and the problem. Hopefully, someone comes up with an idea to implement the algorithm, given that it is a common stochastic algorithm used for modelling.

- **Gillespie algortihm - Direct Method**

1. Initialize the time $t=t_0$ and the system's state $x=x_0$
2. With the system in state $x$ at time $t$, evaluate all the propensities $a_i(x)$ and their sum $a_0(x)$ 
3. Generate a random value for $\tau$ using an exponential distribution with parameter $a_0(x)$
4. Generate a random value to draw action $i$ in a discrete distribution for which $P(X=i)= a_i/a_0$
5. Update $t \leftarrow t+ \tau$ and $x$ according to $i$
6. Save $(x,t)$ and return to step 1, or else end the simulation

- **Attempted implementation**

I will focus on steps 2, 3, 4, which are the complex implementation steps.

To take care of steps 2 and 4,  I included the following in the `agent_step!` function:

```julia
     #propensities
     a1 = model.selection_rate * nagents(model)
     a2 = model.reproduction_rate * nagents(model)
     a3 = model.exchange_rate * nagents(model)

     #total propensities
     a0 = a1 + a2 + a3

     p = rand(model.rng)
     if 0 <= p < a1/a0
         selection_RPS!(strategy, model)
     elseif a1/a0 <= p < (a1+a2)/a0
         reproduce!(strategy, model)
     elseif (a1+a2)/a0 <= p < 1
         swap!(strategy, model)
     end
```

For step 3, I define `steps` as a model property, which is initialized as `steps=0`.
For each model step, the steps increase by one, as follows:

```julia
function model_step!(model)
    model.steps += 1
end
```

The agents have a property called `action_step`. For each agent, the action step is determined by  `rand(Exponential(1/a0))`.
Therefore, when `model.steps` matches `agent.action_step`, a given action will occur according to step 4.

- **Problem** 

[...]

PR #889, opened 2023-09-25
New model with agent_step! and model_step!

Tortar:
Fixes https://github.com/JuliaDynamics/Agents.jl/issues/883.

Still need to be applied to all the library examples and tests and we need to deprecate the visualization functions which require to pass the stepping functions from the outside and create new ones. But this works.

Before finishing it, I'd like to ask @Datseris if this is the right fix.

After this PR we can release 6.0!


src/core/model_concrete.jl
```diff
@@ -3,9 +3,15 @@ using StaticArrays: SizedVector
...
-# And the three implementations here are just variants with different `C` type.
-struct SingleContainerABM{S<:SpaceType,A<:AbstractAgent,C<:ContainerType{A},F,P,R<:AbstractRNG} <: AgentBasedModel{S,A}
+# And the two implementations here are just variants with different `C` type.
+struct SingleContainerABM{
+    S<:SpaceType,
+    A<:AbstractAgent,
+    C<:ContainerType{A},
+    G,K,F,P,R<:AbstractRNG} <: AgentBasedModel{S,A}
     agents::C
+    agentstep!::G
```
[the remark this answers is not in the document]
fbanning: Hm, and it's also a bit uncommon to use the bang `!` in fieldnames, I think. Maybe we could just use `agent_step` and `model_step` instead? Cf. https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/blob/c85b4487f453422d1522b9dc1422a3fa7c000be3/src/core_systems/arbitrary_steppable.jl#L37 for reference implementation in DynamicalSystems.jl

src/simulations/step.jl
```diff
@@ -72,4 +68,4 @@ Use instead of `agent_step!` in [`step!`](@ref) if no function is useful to be d
 dummystep(agent, model) = nothing
 
 until(s, n::Int, model) = s < n
-until(s, n, model) = !n(model, s)
+until(s, f, model) = !f(model, s)
```
fbanning, line 71: Minor detail but good catch! :)

Datseris:
I will review this after we finalize the discussion in #891

src/simulations/step.jl
Tortar: thanks! :D

src/core/model_concrete.jl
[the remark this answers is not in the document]
Tortar: Right! But I stand with your new way so I will not address your concern for now ;)

src/core/model_concrete.jl
```diff
@@ -43,19 +43,24 @@ single container. Offers the variants:
...
+    agent_step!::G = dummystep,
+    model_step!::K = dummystep,
     container::Type = Dict{Int},
     scheduler::F = Schedulers.fastest,
     properties::P = nothing,
     rng::R = Random.default_rng(),
     warn = true
 ) where {A<:AbstractAgent,S<:SpaceType,G,K,F,P,R<:AbstractRNG}
+    if agent_step! == dummystep && model_step! == dummystep
+        @warn "From version 6.0 it is necessary to pass agent_step! and model_step! when defining
```
fbanning, line 56: It's not necessary to pass both. Suggestion: "From version 6.0 it is necessary to pass at least one of agent_step! or model_step! when defining"

src/simulations/ensemblerun.jl
```diff
@@ -2,11 +2,12 @@ export ensemblerun!
...
 
 """
-    ensemblerun!(models::Vector, agent_step!, model_step!, n; kwargs...)
+    ensemblerun!(models::Vector, n; kwargs...)
 Perform an ensemble simulation of [`run!`](@ref) for all `model ∈ models`.
 Each `model` should be a (different) instance of an [`AgentBasedModel`](@ref) but probably
 initialized with a different random seed or different initial agent distribution.
-All models obey the same rules `agent_step!, model_step!` and are evolved for `n`.
+All models obey the same rules `agent_step!, model_step!` contained in the model
+and are evolved for `n`.
```
Datseris, line 10: In this PR we need to generically shift away from explicitly mentioning `agent_step!, model_step!`. We should say "all models obey the same evolution rules contained in...". In the model documentation we explicitly say "the evolution rules are the functions `agent_step, model_step, scheduler`. If `agent_step == dummy`, the evolution rules is just the function `model_step`.

Datseris:
Please do not change any examples in this PR. All examples should work if our deprecations are correctly put in place.

Tortar:
I think we can actually change all the examples in this PR for the reason that we already know that all the deprecated code works since  all tests still pass, while I have not applied the new way to any examples. I will do commits that do no touch any code so that this is preserved

Datseris:
please first run the documentation locally and see that all examples build. before committing any changes to examples.

Datseris:
In fact, not only that. I argue that this PR in genertal should not alter any examples. The examples should be altered in a new PR after this one is merged in.

Datseris:
We need to be extra careful with critical changes such as this one.

Tortar:
ok then let's do it this way

Tortar:
this is the list of the affected functions signatures by this change:

ABM
StandardABM
UnremovableABM
step!
run!
offline_run!
ensemblerun!
ABMObservable
abmplot
abmplot!
abmexploration
abmvideo


since the list is very big, controlling the deprecation warnings is hard, so I'm thinking to **only** warn the users when they use one of ABM, StandardABM or UnremovableABM and in step!, and use a warning which actually says to them that all of these function don't need the passing of the stepping function, maybe in a general way as: "All of the functions definition which expected at least an agent_step! or a model_step! are deprecated, passing these arguments is not necessary anymore. Pass these functions as keywords arguments (agent_step! = your_agent_step!, model_step! = your_model_step!) inside your model definition (ABM(...), StandardABM(...), UnremovableABM(...))."

Tortar:
Actually I think I can manage to make it work, but anyway I think that saying that the change should be applied everywhere the stepping functions are passed seems a good idea to me

Tortar:
If we want to be extra-sure we should do this verification:

- running old tests with the above changes still works

Datseris:
Is this good to review? I'll take over the Tutorial.md file once this is ready.

Tortar:
still not good

Tortar:
The tests work even in the older version so I think this is ready for review

Datseris:
Okay. It will take me a while to review. Let's let this simmer here.

Datseris:
Okay, I am done here. You should also review the new tutorial and give feedback. I had to update schelling as well, because the schelling example is linked with the tutorial.

Tortar:
Removing the `agents_first` from `step!` broke some tests

docs/src/tutorial.md
```diff
@@ -31,62 +32,93 @@ An example of a space is [`OpenStreetMapSpace`](@ref). It is based on Open Stree
...
-However, for many models, this simplified approach offers the benefit of not having to write an explicit loop over existing agents inside the `model_step!`.
-Most of the examples in our documentation can be expressed using an independent `agent_step!` and `model_step!` function.
+## [3.3 The evolution rule - advanced (manual scheduling)](@id manual_scheduling)
 
-On the other hand, more advanced models require special handling for scheduling, or may need to schedule several times and act on different subsets of agents with different functions.
-In such a scenario, it is more sensible to provide only a `model_step!` function (and use `dummystep` as `agent_step!`), where all configuration is contained within.
-Notice that if you follow this road, the argument `scheduler` given to [`AgentBasedModel`](@ref) somewhat loses its meaning.
+Some advanced models may require special handling for scheduling, or may need to schedule agents several times and act on different subsets of agents with different functions during a single simulation step.
+In such a scenario, it is more sensible to provide only a model stepping function,  where all configuration is contained within.
```
Tortar, line 114:
```suggestion
In such a scenario, it is more sensible to provide only a model stepping function,  where all the dynamics is contained within.
```

docs/src/tutorial.md
```diff
@@ -31,62 +32,93 @@ An example of a space is [`OpenStreetMapSpace`](@ref). It is based on Open Stree
...
 
 Here is an example:
+
 ```julia
 function complex_step!(model)
+    scheduler1 = Schedulers.Randomly()
+    scheduler2 = user_defined_function
     for id in scheduler1(model)
         agent_step1!(model[id], model)
     end
```
Tortar, line 124:
```suggestion
    scheduler1 = Schedulers.Randomly()
    scheduler2 = user_defined_function_with_model_as_arg
    for id in schedule(model, scheduler1)
        agent_step1!(model[id], model)
    end
```

same for `scheduler2`

Datseris: Why complicate things? A scheduler is a function that takes in the model as an argument. `schedule(model, scheduler)` has no reason to exist. It should be `scheduler(model)` instead. The only thing that should exist is `schedule(model)` because it uses the default internal scheduler.

Datseris:
> Removing the `agents_first` from `step!` broke some tests

Sory about that... Could you please fix the tests?

docs/src/tutorial.md
Tortar: The problem this tries to address is that this way the check about removed agents is done automatically inside `schedule`, this seems to me a good safety reason to require that

Tortar: this is the code of `schedule(model, scheduler)`: https://github.com/JuliaDynamics/Agents.jl/pull/889/files#diff-8d4e4a89381c499f8748e691a16bd597edc7e55a035a72755fe69d4db4b0749bR13

Datseris: oooooooooooooh, now i understand. unfortunately this doesn't really help:
```julia
function schedule(model::ABM)
    scheduler = abmscheduler(model)(model)
    return Iterators.filter(id -> id in allids(model), scheduler)
end
schedule(model::UnremovableABM) = abmscheduler(model)(model)
schedule(model::ABM, scheduler) = Iterators.filter(id -> id in allids(model), scheduler(model))
```
You see, `allids(model)` is not dynamic. The only way to dynamically skip agents that are removed while the model evolves is the way we do it now:
```julia
for id in activation_order
        if model isa StandardABM # this if gets compiled out
            id in allids(model) || continue
        end
        agent_step!(model[id], model)
    end
```

Datseris: So `schedule` needs to be removed after all. It doens't add anything besides more `api` names.

`schedule` should be moved to the deprecations and be replaced by `abmscheduler(model)(model)`.

Datseris: Okay, forget what I said. `allids(model)` **is dynamic**.

Datseris: so over all your solution is fine, we keep `schedule` as is. OPlease adjust the source ccode accordingly.

Tortar: 👍

Tortar:
Tests are okay now, I looked at the tutorial and it seems fine to me, can we finally merge it? :-)

Datseris:
We aren't tagging v6 imediatelly after this goes in anyways, so there is no gain in rushing anything. I first need to do a draft for #760 to check if the new model interface makes sense.

Datseris:
The notice in the changelog of this absolutely massive change is too small. I'll update the changelog and then merge.

Datseris (approved):

PR #897, opened 2023-10-05
Improve speed of random_agent

Tortar:
It was actually very slow before, consider this silly model I was using to benchmark a different thing:

```julia

function init_model()
	model = StandardABM(GridAgent{2}, GridSpace((50, 50)), rng=Xoshiro(42))
	for _ in 1:1000
		add_agent!(model)
	end
	return model
end

function model_step!(model)
	add_agent!(model)
	remove_agent!(random_agent(model), model)
end

model = init_model()
@benchmark step!(model, dummystep, model_step!, 10^5)
```

before:

```julia
julia> @benchmark step!(model, dummystep, model_step!, 10^5)
BenchmarkTools.Trial: 27 samples with 1 evaluation.
 Range (min … max):  183.797 ms … 194.004 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     187.722 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   188.308 ms ±   2.707 ms  ┊ GC (mean ± σ):  0.13% ± 0.37%

                 ▃ █     ▃
  ▇▁▁▇▁▁▁▁▁▇▁▇▁▁▇█▁█▇▁▇▇▁█▁▇▁▁▁▁▇▇▁▇▇▇▁▇▁▁▁▁▇▁▁▁▁▁▁▁▁▇▁▁▁▁▇▇▁▁▇ ▁
  184 ms           Histogram: frequency by time          194 ms <

 Memory estimate: 5.19 MiB, allocs estimate: 100169.
```

after:

```julia
julia> @benchmark step!(model, dummystep, model_step!, 10^5)
BenchmarkTools.Trial: 306 samples with 1 evaluation.
 Range (min … max):  15.251 ms … 26.520 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     15.822 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.360 ms ±  1.440 ms  ┊ GC (mean ± σ):  1.50% ± 3.67%

    ▄█▆▁
  ▃▅████▇▆▃▃▄▅▄▄▃▃▁▃▃▃▃▄▃▃▃▃▃▂▂▂▃▂▃▁▁▂▁▁▁▁▁▃▂▁▁▁▁▂▁▂▁▁▁▁▁▂▂▁▂ ▃
  15.3 ms         Histogram: frequency by time        21.7 ms <

 Memory estimate: 5.19 MiB, allocs estimate: 100162.
```

A micro-benchmark tells that now random_agent for a StandardABM is 40x faster!

I exported a new function `random_id` in the process, which is in line to the other random functions which export also the id version


Tortar:
I have to say that this means that probably there is something to improve in Base, since that the previous method is so slow sounds strange

Tortar:
Actually it was a O(N) operation while this one is not, this means that the improvement in speed is arbitrary large!

Datseris (approved):
I don't think it is a bug in base. It is probably an inferrence issue since the agent container is complicated. The function barrier makes the inference work.

DOn't forge tto add changelog entry and then merge!

Tortar:
I think it should be https://github.com/JuliaLang/julia/issues/51605 :-)

PR #900, opened 2023-10-06
Further clarify mandatory API `ABM`

Datseris:
In preparation for the new Gillespie model, this PR closes #899 . 

I've also removed the agent type as a type parameter of the abstract `AgentBasedModel`, in foresight of future multi-agent models not being parameterized by a Union type. This change took me 2 hours to do.

PR #903, opened 2023-10-06
Remove unnecessary check in add_agent_pos!

Tortar:
This will fail with an easy to bisect error anyway since adding to the model will fail since the agent type will not be accepted. Forgot to do it when merging #900

### DIFF
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -20,7 +20,7 @@ import LinearAlgebra
 # Core structures of Agents.jl
 include("core/agents.jl")
 include("core/model_abstract.jl")
-include("core/for_free_extensions.jl")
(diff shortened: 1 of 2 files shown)


Tortar:
wops! sorry renamed one of the files before merging without changing the including statement!!

Tortar:
For some reason, I can't anymore bypass the protection rule otherwise I would have merged this one

Datseris:
@Tortar I've updated the rules of Agents.jl to ensure high quality of contributions.

1. Can't push directly to master, must do PR.
2. PRs must have one accepting review from someone other than the creator.

Things might get slower, but they are safer this way.

Datseris (approved):

PR #908, opened 2023-10-17
Update examples to v6.0

### LINKED ISSUE #898: Update examples into v6
Following https://github.com/JuliaDynamics/Agents.jl/pull/889  we need to update all other examples. This change is trivial and it simply means: remove passing `agent_step!` to `step!/run!/abmplot!` and instead give these functions to the model constructor.

Tortar:
Fixes #898

### DIFF
--- a/examples/celllistmap.jl
+++ b/examples/celllistmap.jl
@@ -105,7 +105,10 @@ function initialize_bouncing(;
         system=system,
     )
     model = StandardABM(Particle,
-        space2d,
+        space2d;
+        agent_step!,
+        model_step!,
+        agents_first = false,
         properties=properties
     )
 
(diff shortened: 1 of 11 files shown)


Tortar:
ok, seems that even if all examples run locally, they don't run on the docs since we need to first define the stepping functions since code is run in blocks in the docs

Tortar:
Actually, it was simpler than that, just move the call to initialization after the stepping definition. PR should be good to go now

Datseris (approved):

Datseris:
I haven' checked at the docs yet. If all videos are produced fine, then merge!

Tortar:
Seem fine to me 👍

PR #909, opened 2023-10-19
Enable arrow backend for offline_run! in Windows

Tortar:
Actually there are current problems with tempfile locking in Windows for Arrow.jl, indeed some tests are not enabled in Arrow.jl for Windows in that repo. I also remembered that offline_run! worked fine for me, there was just a problem in deleting the file during tests but not when using from the external of the package (now I understand that this is due to this locking problem). So I think that we can anyway enable the backend also in Windows so that users can benefit from it.

See https://github.com/apache/arrow-julia/pull/361

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,3 @@ src/update_*
 agents_visualizations.md
 .vscode
 celllistmap.mp4
-test/adata.arrow
-test/mdata.arrow
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ src/update_*
 agents_visualizations.md
 .vscode
 celllistmap.mp4
+test/adata.arrow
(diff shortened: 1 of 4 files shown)


Datseris:
@AayushSabharwal could you please be the one that reviews this PR? You have more knowledge on I/O in Agents.jl than me.

Tortar:
Actually I find IO errors when trying to remove .arrow files even if not in a tempdir, for the rest the backend seems to work fine on Windows, but it's probably better to understand it more deeply this problem before removing the throwing of the error

Tortar:
Ok, in the end I used a different approach so that now Arrow integration on Windows is actually tested except for the removal of the files, so it should be okay now

Datseris (approved):

PR #917, opened 2023-11-07
Implement EventQueueABM

### LINKED ISSUE #902: Design discussion for an Event-Based (Gillespie-like) agent based model in Agents.jl
The new `AgentBasedModel` API of the upcoming v6 allows us to make new model types where the "evolution rule" can be fundamentally different. This is one of them. This issue discusses a possible solution to #759 .

# Event-Queue Agent Based Model design

## Introduction

Goal of this is to provide an alternative type of ABM simulation.
It operates in **continuous time**, versus the currently available `StandardABM` which operates in discrete time. This is an "event queue" simulation, also known as "Gillespie" simulation or "discrete event simulation". There are Julia packages to perform such simulations (that are actually actively maintained), for example:

- JumpProcesses.jl
- ConcurrentSim.jl

However, for these two the _spatial_ aspect is either weak or non-existent, while for ABMs spatial dynamics are crucial.

I've spent considerable amount thinking whether it would possible to interface JumpProcesses.jl to make the current proposal work with it as a backend. But it didn't seem straight forward, or perhaps it would require writing a lot of code.
But maybe I am wrong, see my internals section at the bottom and let me know...

## How things will work in the new model type

Events are scheduled at some particular time, and once the model time reaches the event time, the event is executed. Events are arbitrary agent-stepping functions. They are very similar to the `agent_step!` function used currently in the standard ABM implementation. Hence, events have access to the full API of Agents.jl, including functions like `nearby_agents`, `move_agent`, `add_agent!`, `replicate!`, .... Whatever! This is **incredibly powerful** when it comes to performing spatial simulations with event based models.

Events that are occurring are chosen randomly during scheduling, based on the principle of "propensities". Each event has a possible propensity value and the probability for choosing a particular event is proportional to the propensity.
This probability is calculated automatically by an internal scheduler.

## User input

Besides a space instance, **for each agent type** a user provides:

- A tuple of actions. Actions are `agent_step!` functions.
  They operate exactly like an `agent_step!` does in `StandardABM`, it's just that
  the scheduling is different.
- A tuple of rates.
  - If the rate is a real number, the propensity is the rate times `nagents(model)`.
  - If the rate is a function, the propensity is `f(model)` times `nagents(model)`.

Note that unlike `StandardABM`, we can't have here the user providing the same input irrespectively of agent type and using multiple dispatch. That is because different agent types may have different _number_ of possible actions.

## Scheduling

We first have the `Event` type, which is
```julia
struct Event
    id::Int # id of the agent to be activated
    event_index::Int # index of the tuple of possible actions to act on the agent
end
```

(note: this has the same type instability problems that `StandardABM` has)

this event is stored in a `PriorityQueue{Event, Float64}()`.

During the model time evolution we go through the events in the Queue, pop! them from the queue and "activate" them.
When an event is activated, two things occur:

First, the relevant action is operated in the agent.
This is done by extracting the agent corresponding to the event ID,
and the action corresponding to the combination of agent type and event index.

Then, the scheduler is called upon the agent to generate a new event.
The scheduler utilizes the rates corresponding to the agent type,
generates the event, and adds it to the queue.

## Time evolution

time evolution is identical to `StandardABM`. `step!(model, time)` is used,
with the only difference here that `time` is Float64 instead of integer and steps the model until the current model time is equal or greater to the initial time plus `time`.

Time evolution skips events mapping to agents that have been removed from the model as is done generically already.

All other functions like `run!, ensemblerun!`, etc. work out of the box.

## Initial scheduling?

Okay we know how to add events in the que, but how does the que start in the first place? An empty que cannot generate events. So there needs to be some kind of way to initialize the queue?

## Minimal Working Example

Here is a full minimal working example of a spatial rock-paper-scissors example mentioned in https://github.com/JuliaDynamics/Agents.jl/issues/760 and in https://github.com/JuliaDynamics/AgentsExampleZoo.jl/pull/17 . On purpose I am making rock, paper, and scissor, three different types, to show how things would work for different type. (in this MWE the actual possible events are the same, but they don't have to be)

Naturally, the example isn't actually "working", as I haven't written the source code proposed here. But this is how it would look like!

```julia
using Agents

# define the three agent types
@agent struct Rock(GridAgent{2})
end
@agent struct Paper(GridAgent{2})
end
@agent struct Scissors(GridAgent{2})
end

# Define the possible events
function selection!(agent, model)
    contender = random_nearby_agent(agent, model)
    if !isnothing(contender)
        pos_contender = contender.pos
        if agent isa Rock && contended isa Scissors
            kill_agent!(contender, model)
        elseif agent isa Scissors && contender isa Paper
            kill_agent!(contender, model)
        elseif agent isa Paper && contender isa Rock
            kill_agent!(contender, model)
        end
    end
    return
end

function reproduce!(agent, model)
    pos = random_nearby_position(agent, model, 1, pos -> isempty(pos, model))
    isnothing(pos) && return
    add_agent!(pos, typeof(agent), model)
    return
end

function swap!(agent, model)
    rand_pos = random_nearby_position(agent.pos, model)
    if isempty(rand_pos, model)
        move_agent!(agent, rand_pos, model)
    else
        near = model[id_in_position[rand_pos]]
        swap_agents!(agent, near, model)
    end
    return
end

# Put agents and events in the format required
agent_types = (Rock, Paper, Scissors) # internally this will be made `Union`
# events
rock_events = (selection!, reproduce!, swap!)
scissor_events = (selection!, reproduce!, swap!)
# paper dynamics have no movement
paper_events = (selection!, reproduce!)
# same layout as agent types
allevents = (rock_events, scissor_events, paper_events)

# rates:
# the length of each of these needs to be the same as the length of event tuples
rock_rates = (0.5, 0.5, 0.2) # relative proportionality matters only
paper_rates = (0.5, 0.2, 0.1)
scissor_rates = (0.5, model -> count(a -> isa Rock, allagents(model)))
# same layout as agent types
allrates = (rock_rates, paper_rates, scissor_rates)

# Create model
space = GridSpaceSingle((100, 100))

abm = EventQueueABM(space, agent_types, allevents, allrates)

step!(abm, 1.32)
```

## Internals

The `step!` implementation is straight-forward:
```julia
function step!(model::EventQueueABM, t::Real)
    t0 = zero(t)
    while t0 < t
        event, dt = dequeue_pair!(model.event_queue)
        process_event!(event, model)
        t0 += dt
    end
end

function process_event!(event, model)
    (; id, event_idex) = event
    agent = model[id]
    agent_type = findfirst(isequal(typeof(agent)), model.alltypes)
    event_function! = model.allevents[agent_type][event_index]
    event_function!(event, model)
    # evaluate propensities:
    propensities = model.allrates[agent_type] .* nagents(model)
    total_propensity = sum(propensities)
    τ = rand(abmrng(model), Exp(total_propensity))
    new_event = select_event_based_on_propensities(propensities, model)
    enqueue!(model.event_queue, new_event => τ)
    return
end
```

Tortar:
Fixes #902 

Still very bad but the new example at least works :-)


Datseris:
I am confused about the existence of `model_step` and `agents_first`, as the whole point of this event based system is that there is no model step and there cannot be "before" or "after" the step, as events happen on a continuous time line and hence don't have an end; they go on until exhaustion or hitting a user-provided time limit.

I would argue that if a user wants "agent steps" and "model steps" they should just go ahead and use the StandardABM and hence here we should focus on just the quing system and simplify everything.,

Tortar:
I agree :+1:

src/simulations/step.jl
```diff
@@ -38,6 +38,44 @@ function CommonSolve.step!(model::ABM, n::Union{Function, Int} = 1)
...
+
+function process_event!(event, model)
+    id, event_idx = event.id, event.event_index
+    !haskey(agent_container(model), id) && return
+    agent = model[id]
+    agent_type = findfirst(isequal(typeof(agent)), tuple_agenttype(model))
+    event_function! = abmevents(model)[agent_type][event_idx]
+    event_function!(agent, model)
+    !haskey(agent_container(model), id) && return
+    propensities = abmrates(agent, model) .* nagents(model)
```
Tortar, line 65: It's not really clear to me why here we multiply by `nagents(model)`, shouldn't it be the number of agents of the related agent type?

Tortar: actually to me is not clear why applying a function such as `nagents` in general? shouldn't the user define his own rates? Given that we let choosing a function then every transformation of the propensities seems not useful. But it's not entirely clear to me how the algorithm should work, so I could be mistaken

Tortar: @MA-Ramirez @Datseris I'd need some help on this if you can give some :-)

docs/src/api.md
```diff
@@ -8,10 +8,11 @@ In this page we list the remaining API functions, which constitute the bulk of A
 Besides the generic interface of [`AgentBasedModel`](@ref), and the [`StandardABM`](@ref), highlighted in the tutorial, there is also:
 
 ```@docs
-UnremovableABM
+EventQueueABM
```
Datseris, line 11: I think we should move `AgentBasedModel` here as well, and only expand the docstring of `StandardABM` in the tutorial. In the tutorial we just mention explicitly part of the interface by showing how it works in terms of examples. E.g., do something like `model[2]`. I can do this after this PR is merged.

src/core/model_event_queue.jl
```diff
@@ -13,57 +13,117 @@ struct EventQueueABM{
...
+- the scheduler generates a new event utilizing the rates corresponding to the agent type,
+  and schedules it together with the same paired agent;
+
+Even if the scheduler generate automatically new events based on the rates, it should be
+initialized by adding some events at the start of the simulation through the [`add_event!`](@ref) 
+function.
+
+## Keywords
+
+- `all_events`:
```
Datseris, line 60: I think that `all_events` should not be keyword but positional mandatory argument. In contrast to StandardABM, this can't work without `all_events`.

src/simulations/step.jl
Tortar: If nobody has insights on this I will remove it, it is in any case possible to achieve through passing a function in `abmrates `

Datseris: Sorry for the late reply. Yes, you should remove this part and if the user needs to they can call `nagents` in `abmrates`.

src/core/model_event_queue.jl
```diff
@@ -0,0 +1,129 @@
...
+of the [Gillespie Algorithm](https://en.wikipedia.org/wiki/Gillespie_algorithm). Each event 
+has a possible propensity value and the probability for choosing a particular event is 
+proportional to the propensity. During the model time evolution, two things occur:
+
+- the model time reaches the time of the next scheduled event and the event action 
+  is performed on the paired agent;
+- the scheduler generates a new event utilizing the rates corresponding to the agent type,
+  and schedules it together with the same paired agent;
+
+Even if the scheduler generate automatically new events based on the rates, it should be
```
jacobusmmsmit, line 54:
```suggestion
Even if the scheduler automatically generates new events based on the rates, it should be
```

Datseris:
@Tortar what is remaining for this to be finished? It's the last PR before v6, right?

Tortar:
Mainly some docstrings need to be finished and we need to add some more tests for the new model type. We agreed with @fbanning to put also #934 in the v6 milestone. I'm a bit overwhelmed by other tasks at the moment, I think I can work on this monday. But if someone else wants to help feel free to add your commits to the branch

jacobusmmsmit:
Can I ask what happened with the ConcurrentSim implementation? Did anyone try doing a comparison? I looked into CS.jl and it was rather complicated. I think it raises the complexity barrier for contributers even if none of the internals are exposed.

Tortar:
I think that we can add the CS.jl integration later but I don't think that internal with CS.jl will become too complicated, maybe it is so but it is worth trying. I don't think it's a priority though

jacobusmmsmit:
At one point we can make a performance comparison with "from scratch continuous abm", "Agents-v6 continuous abm", and a "CS Agents continuous abm" and see if it's worth it, and if so how much.

Datseris:
I can take over this PR. I realized there are some things we need to improve. For example, if an agent is added to the model via e.g., `replicate!`, this agent will never get an event assigned to it. My proposal is that when an agent gets added to the model, we immediatelly trigger the event generatrion function for it. This also solves the "what's the starting event" problem. Because as soon as we populate the model with agents, there is also an event que.

I also want to do two more improvements: using the concept of probability mass to decide how events are chosen, and simplifying significantly the events. I will add a generic type `AgentEvent` that takes in an event function and the type of agent it applies over. In this way an event may be applicable to many agents by providing the `AbstractAgent` as the agent type, which will also be the deafult value.

Datseris:
I have had great progress in the `eventqueuv2` branch, but I am not finished yet. I take a break now but I will finish this soon!

Tortar:
Closing in favor of #940

Issue #918, opened 2023-11-07
Deprecate UnremovableABM in v6


Tortar:
Since we will have a new ABM type (`EventQueueABM`) which could work with both a Dict or Vector container for the Agents, I think it would be better to uniformize this feature: this means deprecating `UnremovableABM` (so actually not really breaking). Instead of it I would propose to allow the user to choose a `StandardABM(...; ..., agent_container=Vector)` which is actually already available, it is just not documented. It seems even more natural since this is just a little optimization, so it doesn't make much sense in my opinion to have a different type, we will just say that if you choose a Vector as container you won't be able to remove agents.

cc @Datseris

Datseris:
yes i agree

Tortar:
With #919

PR #919, opened 2023-11-08
deprecate UnremovableABM

Tortar:
First commit prove that everything still works with UnremovableABM deprecated.

So in subsequent commit I will just update everything outside source

Solves #918

### DIFF
--- a/src/core/model_standard.jl
+++ b/src/core/model_standard.jl
@@ -65,7 +65,7 @@ to all scheduled agents. Then, the `model_step!` function is called
 
 `StandardABM` stores by default agents in a dictionary mapping unique `Int` IDs to agents.
 For better performance, in case the number of agents can only increase during the model 
-evolution, a vector can be used instead.
+evolution, a vector can be used instead, see keyword `container`.
 
 ## Keywords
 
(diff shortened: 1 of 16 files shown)


src/core/model_standard.jl
```diff
@@ -118,12 +63,18 @@ first schedules agents by calling the scheduler. Then, it applies the `agent_ste
 to all scheduled agents. Then, the `model_step!` function is called
 (optionally, the `model_step!` function may be called before activating the agents).
 
+`StandardABM` stores by default agents in a dictionary mapping unique `Int` IDs to agents.
+For better performance, in case the number of agents can only increase during the model 
+evolution, a vector can be used instead.
```
Datseris, line 68:
```suggestion
evolution, a vector can be used instead, see keyword `container`.
```

src/core/model_standard.jl
```diff
@@ -118,12 +63,18 @@ first schedules agents by calling the scheduler. Then, it applies the `agent_ste
...
+
 ## Keywords
 
 - `agent_step! = dummystep`: the optional stepping function for each agent contained in the
   model. For complicated models, it could be more suitable to use only `model_step!` to evolve
   the model.
 - `model_step! = dummystep`: the optional stepping function for the model.
+- `container = Dict`: if agents cannot be removed, only added using instead `container = Vector`
+  allows storing agents more efficiently, yielding faster retrieval and iteration over agents.
```
Datseris, line 77:
```suggestion
- `container = Dict`: the type of container the agents are stored at. Use `Vector` if no agents are removed
  during the simulation. This allows storing agents more efficiently, yielding faster retrieval and 
  iteration over agents.
```

Datseris (approved):

Issue #910, opened 2023-10-21
Refactoring the access to space related functions to move towards multispace models


Tortar:
Take this function:

```julia
function id_in_position(pos, model::ABM{<:GridSpaceSingle})
    return abmspace(model).stored_ids[pos...]
end
```

we could refactor this code so that to have:

```julia
id_in_position(pos, model::ABM) = id_in_position(pos, abmspace(model), model) # this can be actually done only one time, probably
                                                                                                                                  # to put in space_interaction_API.jl
function id_in_position(pos, space::GridSpaceSingle, ::ABM)
    return space.stored_ids[pos...]
end
```

What does this allow? It gets us nearer to possible implementationa of multispace models since we give the possibility to the user to access the space-related function dispatching on the space, so that to allow to refer to different spaces inside his code. 

Note that this is a costless abstraction.

cc: @Datseris

Datseris:
> It gets us nearer to possible implementation of multispace models

For me this needs to be justified in the first place... How general is the need for this functionality?

But yes, in any case, the refactor outlined here is very easy to do and also non breaking.

Datseris:
However, the proposal here doesn't work for two spaces of the same kind, two different grid spaces. Which I would assume would be the most common case of this already extremely uncommon feature.

Tortar:
It works even for two space of the same kind, take `id_in_position` now its signature is `id_in_position(pos, model)`, we will change it in `id_in_position(pos, [, space], model)` so that the user can pass `id_in_position(pos, grid1, model)`, `id_in_position(pos, grid2, model)`,...

> For me this needs to be justified in the first place... How general is the need for this functionality?

We will solve #719 for example. More generally we will allow for separating types in different spaces which can be useful sometimes for performance e.g. the wolfsheep model would be faster this way. Also some mixture of space types could be actually interesting e.g. the one suggested in #719 is already an example NoSpace + Any other Space

Datseris:
Well, yes, but as I also said in #719 I am not at all convinced about the usefulness or necessity of this feature given the arguments of #719 as it is arguably simpler to achieve the same result without mixing spaces.


> More generally we will allow for separating types in different spaces which can be useful sometimes for performance e.g. the wolfsheep model would be faster this way

There is no proof yet of whether this will have measurable performance boost, because there is no way for multiple dispatch to deduce the type. GridSpaces do not store agent types nor have any knowledge of them. So type instability will remain the main problem.

[...]

PR #921, opened 2023-11-13
Add quality code tests: Aqua.jl + method ambiguities test

Tortar:
This resolved the 22 method ambiguities in Agents.jl. Notice that this resolved a bug with nearby_ids in a no space model which was ambiguous before, all other ambiguities are less important


src/core/space_interaction_API.jl
```diff
@@ -56,7 +56,7 @@ remove_agent_from_space!(agent, model) = notimplemented(model)
 # %% IMPLEMENT: Neighbors and stuff
 #######################################################################################
 """
-    nearby_ids(position, model::ABM, r = 1; kwargs...) → ids
+    nearby_ids(pos, model::ABM, r = 1; kwargs...) → ids
```
Datseris, line 59: Why is such a change made? Surely we can agree that `position` is more readable than `pos`. Besides, the docstring continues to refer to `position` which no longer exists.

src/spaces/discrete.jl
```diff
@@ -75,10 +75,10 @@ function empty_positions(model::ABM{<:DiscreteSpace})
 end
 
 """
-    isempty(position, model::ABM{<:DiscreteSpace})
+    isempty(pos, model::ABM{<:DiscreteSpace})
 Return `true` if there are no agents in `position`.
 """
-Base.isempty(pos, model::ABM) = isempty(ids_in_position(pos, model))
+Base.isempty(pos::ValidPos, model::ABM{<:DiscreteSpace}) = isempty(ids_in_position(pos, model))
```
Datseris, line 81: All these functions that use space are probably better done if they dispatch on the space instance and then the space instance implements `isempty`. This way we don't have to re-code twice all functions that ultimately dispatch on space.

Datseris (changes requested):

src/spaces/discrete.jl
Datseris: By that I mean that `isempty` dispatches to `isempty(pos, abmspace(model))`.

src/core/space_interaction_API.jl
Tortar: All the other functions use the `pos` convention

src/spaces/discrete.jl
Tortar: yes indeed this is an instance of #910

src/core/space_interaction_API.jl
Tortar: ok, actually not all, but most

src/spaces/discrete.jl
Tortar: actually it is not possible, the implementation use also the model (see e.g. `ids_in_position(pos, model)`) , but I actually would like to use the "trick" described in #910, but given it is a general refactoring maybe we can wait till that issue is solved

src/core/space_interaction_API.jl
Datseris: ok we can stay with `pos` but you need to update the docstrings to also refer to `pos` and not to `position`

src/spaces/discrete.jl
Datseris: How does `ids_in_position` use the model? It really shouldn't!

Tortar: yes, I know :D but there are many functions which should really belong to the space but need to keep a reference to the model, e.g. `random_id` since it needs the rng of the model.                                                                                                  

If we want to let the user access the space eventually for the multi-space models we should really follow a unique convention, namely passing the model as argument after the space, namely in this case `ids_in_position(pos, space, ::ABM)`. So my idea is that in the end this will be needed to be changed again, but probably it is ok to start dispatching now like this even if in the future it will be modified

Tortar:
I will also add Aqua.jl to the quality tests so expanding a bit the breadth of this PR :-)

Tortar:
- Aqua.jl ambiguities tests goes too far in the call stack: see https://discourse.julialang.org/t/aqua-jl-finds-many-ambiguities-in-core-base/103963 so disabled

test/package_sanity_tests.jl
```diff
@@ -0,0 +1,6 @@
+using Aqua
+
+@testset "Code quality" begin
+	Aqua.test_all(Agents, ambiguities = false, unbound_args = false)
```
Tortar, line 4: Still need to understand how to solve the `unbound_args` problems (but they seem innocuous nonetheless)

Tortar: Mmmh not an easy task, since they are innocuous I think it is okay like this

src/spaces/discrete.jl
Tortar: > All these functions that use space are probably better done if they dispatch on the space instance and then the space instance implements `isempty`. This way we don't have to re-code twice all functions that ultimately dispatch on space.

Actually it is not true we should then implement `ids_in_position(pos, space)` for all discrete spaces and keep `ids_in_position(pos, model)` for users. So I think not worth right now.

If you agree this PR should be good to go!

Datseris (approved):

Issue #930, opened 2022-11-14
Add support for plotting of ABMs with `nothing` space


fbanning:
Hey @saumil-sh,

> However, `abmplot` doesn't support `nothing` space, yet! I will open an issue similar to [this](https://github.com/JuliaDynamics/InteractiveDynamics.jl/pull/114) in `InteractiveDynamics.jl`.

Awesome that you'd like to add this functionality. It's also way up there on my priority list and it would allow the recipe to be used a bit more flexibly instead of using the predefined plot types. I first wanted to get the more specialised case of `GraphSpace` finished before tackling the more general case of `nothing` space.

My preferred way of doing it would be to just create a special case for the recipe which doesn't create any plot (i.e. doesn't call `_abmplot!`) but only provides an `Axis` or grid position or something similar for the user to plot into with a self-chosen in-place plot recipe (e.g. `scatter!` or whatever else they want to do).
It would be nice if a call to `abmplot(model)` where `model.space == nothing` provided a figure with the model control buttons in position `[1,2]` and an empty grid position at `[1,1]` (or maybe a placeholder for the user to erase again if it's not needed). The problem with just creating an empty `Axis` (which would be the easiest solution) is that users may instead want to have a 3D scene with `Axis3` or `LScene` or whatever. 

Feel free to play around with it a bit and if you have any question about how the `abmplot` recipe works, don't hesitate to ask. :)

Issue #925, opened 2022-12-12
Invalid argument `agents_step!` uncaught by `abmplot`.


iago-lito:
Hello, the minor issue I'm opening here follows a few hours that we've just spent screening all our code to eventually figure out that the error was in the following line:
```julia
fig, _ = abmplot(model; agents_step! = my_agent_step_function!, model_step! = my_model_step_function!, ...)
```
The error above is that the true argument name must be `agent_step!` and not `agents_step!` (with a spurious `s`). Our bad ^ ^

This being said, the behaviour of the above erroneous invocation of `abmplot` was **not** to yield an error complaining about the typo. This is unfortunate. Instead, `abmplot` ran almost-fine and was just behaving as if no `agent_step!` function had been provided at all, which has been very confusing.

Could some sort of check be added within `abmplot` to not pass typos silently?

fbanning:
I've also added the documentation label. In the PR following #914 I would like to include a short notice in the docs that tell people about this behaviour in the recipe system. After that has been done, we can close this issue.

PR #934, opened 2023-11-15
Refactor visualisation extensions

### LINKED ISSUE #914: Refactor `AgentsVisualizations` extension
https://github.com/JuliaDynamics/Agents.jl/tree/main/ext/AgentsVisualizations

The visualisation code works but it's of course far from perfect. Most notably, it relies on some hardcoded pieces (e.g. regarding the supported spaces) and cannot easily be extended by users to work with their custom spaces.[^1] I'm willing to take up this task and try to improve the visualisation code. But I would like to hear your ideas on this topic and maybe get some support from you as well.

## Background

I think it makes sense to look at a bit of history to explain why it is how it is: During the time of writing, Agents.jl was very different from now (most notably a lot smaller) and the plotting code was not intended to capture all use cases that the users might have. Indeed, what we have now is more or less a big refactor that I once contributed to InteractiveDynamics.jl which has been copy-pasted over to Agents.jl at some point in the form of an extension. The visualisation code as we have it now was only (i) a reproduction of the previously existing functionality in the form of a Makie recipe and (ii) a more convenient way to quickly plot and inspect our ABM. Accounting for other spaces outside the currently supported ones wasn't really a goal, even though I was very much aware of that possibility back when I wrote the recipe (there are explanations in the docs regarding [custom plots and data visualisation](https://juliadynamics.github.io/Agents.jl/stable/examples/agents_visualizations/#Creating-custom-ABM-plots-1)).

Yesterday evening the visualisation code was again called out by @Datseris over on Slack:
> in all honesty the plotting code is a big quality dropdown from the normal src code of Agents.jl, so we need to re-format it at some point....

So it seems as if it's a piece of code in the Agents.jl package that he really dislikes. It's a bit hurtful to read this in public and directed towards someone else because the "big quality drop" wasn't communicated to me when I added the code in the first place (or later). Still, I personally like the visualisation code overall because there's a lot of functionality in there (buttons, parameter sliders, automatic lifting and updates, mouse-over inspection, etc.). Scrapping it completely and starting from scratch doesn't seem sensible to me. It's just, as I mentioned above, that I never had the intention to code a highly modular and abstract interface for lifting and inspection. It was only meant to allow easy plotting with a bit of functional styling and customisation.

## Refactor

We're all aware that users want to create plots with a simple call to `abmplot`. So what needs to be done **at the very least** to improve the extension and make this and other custom plots possible?

1. Decouple the building blocks inside the recipe from one another. Simply put, there has to be an opportunity for users to create an `ABMObservable`, create a custom plot (without using `abmplot`!), add controls (i.e. buttons and sliders), and add some plots for the collected data to the same figure. To be honest, I think we're mostly there but especially the controls block is not accessible by the users.
2. Implement a user-facing API for extending the `abmplot` functionality with proper support for custom spaces. There are likely three crucial aspects we need to deal with:
    1. Should there be any pre-plotting?
    3. How to extract space size/dimensionality?
    4. How to map agent positions to the points of a scatter plot?
2. Document how to create own plots using this API.

That's of course not all and we need to ask ourselves another question: What needs to be done **at the very best** to improve the extension?

Well, that depends on how abysmal you think the visualisation code really is. I'm looking forward to your constructive ideas in this thread. :)

[^1]: There's a very clear tendency of users to misunderstand custom spaces. They often incorporate model properties into their custom space structs and most often those custom spaces are just a `GridSpace`/`ContinuousSpace` with a bit of sugar on top. Ultimately I think that the arising issues from custom space plotting are perfectly solvable by just telling people to not overload the functionality of their model's `space` field and stick to existing spaces. But we cannot solve this often-occurring misunderstanding, except for explaining over and over again what exactly a space is intended to do and what not. So this issue - and the PR that will follow from it - are there to allow for this because, honestly, who are we to tell people what to do and what not? ;)

fbanning:
Closes #914 , #925, #930, #951

## Preface

This PR is relatively big, I know. But this refactor was *really* necessary and kind of long overdue. So, sorry but not really all too sorry, I suppose? ¯\\\_(ツ)_/¯

## Description

> There is now a more general interface for plotting. The extension files have been reorganised according to space types. The necessary plotting methods for the different spaces are now gathered in separate files and extend the base plotting functionality for abstract spaces. There is no more arbitrary requirement of defining a set of supported spaces like we did before.
> 
> There is now also a convenient checking function that triggers when somebody uses a custom space. It tells the user which new methods for the API functions are necessary or optional and if the method is usable in its current state. If the necessary methods aren't there, it throws an error and points towards our documentation.
> 
> All of what I've described here is already done and I've tested it with the examples provided in our repository's example folder. Schelling, flocking, and zombies models all work fine.
> -- quoted from https://github.com/JuliaDynamics/Agents.jl/issues/914#issuecomment-1808964944

The space visualisation API is described in [space-visualization-API.jl](ext/AgentsVisualizations/src/spaces/space-visualization-API.jl). Users can easily copy this file and extend the functions to work with their custom space.

**This whole PR is supposed to be not breaking.** Where I've had to make some changes that could be seen as breaking, I've tried to handle them gracefully so that the user gets notified of any deprecations while their code still continues to work as it did before.

## Things left to do

- [x] Separation of the functions adding the control panels, e.g. buttons and sliders (`add_interaction!`)
- Refactoring of inspection code
    - [x] Put all the space-dependent functions into separate files, both for the abstract visualisation interface as well as per space
    - [x] Define missing inspection functions for `GraphSpace` and `OpenStreetMapSpace`
- [x] Create convenient alias for space dependent `_ABMPlot`
- [x] Remove unneeded type annotations
- Merge plot kwargs and handle deprecations
    - [x] `scatterkwargs` and `graphplotkwargs` become `agentsplotkwargs`
    - [x] `osmplotkwargs` becomes `spaceplotkwargs`
- [x] Add `abmplot` kwarg to disable custom space checks
- [x] Finalize custom space checks function once the public API is fixed
- More extensive testing
    - [x] custom space
    - [x] nothing space
    - [x] 3D plots
    - [x] flocking example with polygons
- [x] Handle agent position `offset` differently so that inspection continues to work
- [x] ~~Extend visualizations test suite to a very basic implementation matching the models used to test this PR~~
- [x] Documentation of the public API, i.e. which functions require a method extension for user defined spaces and which method extensions are optional depending on the needs for the space
- Makie v0.20 related things:
    - [ ] Fix 3D `ContinuousSpace` inspection
    - [ ] Polygon plots are wonky/broken and run very slowly
 - [x] Unify function arguments for `agentsplot!` and `spaceplot!`


Tortar:
This is awesome...will take the time to review soon!

fbanning:
Thanks for giving a review a shot, @Tortar , much appreciated.

Indeed, it's not that important that you look at all the details of the implementation. I would say that I'm already relatively deep into the Makie plotting rabbit hole, so I hope that you can put some trust into my plotting code.

Instead it would be especially nice if you can have a look at the structural aspects of the PR. For instance, the definition of the API, which things would need special attention when creating the documentation, if I wrongly associated any type annotations, or whatever. :)

fbanning:
Oh, and here are the two files that I've used to test the PR:
[CustomSpace.txt](https://github.com/JuliaDynamics/Agents.jl/files/13376641/CustomSpace.txt)
[test-AgentsVisualizations.txt](https://github.com/JuliaDynamics/Agents.jl/files/13376642/test-AgentsVisualizations.txt)

You need to rename them to `.jl` because GitHub is just being GitHub again and only allows `.txt` files to be directly uploaded... 🙄 

Put them into a separate folder and create a new Julia environment. For the setup of the env you should run:

```julia
]dev Agents # with the correct branch checked out, of course
]add GLMakie, GraphMakie, OSMMakie # standard released versions
```

Tortar:
I'm sure you are much more experienced than me on that, I will focus on the structural aspects you specified. 

Thanks for the .txt-actually-.jl files, indeed I was asking myself how to try to visualize everything

Tortar:
That before plotting this:

`julia> custom_space_schelling()`

prints

```
┌ Info: Checking for methods that have to be defined to plot an ABM with custom space type
│ 	CustomSpace{2, false}
│ === Required
│ ✔	agents_space_dimensionality(space::CustomSpace{2, false})
│ ✔	get_axis_limits!(model::ABM{CustomSpace{2, false}})
│ ✔	agentsplot!(ax, model::ABM{CustomSpace{2, false}})
│ === Preplots (optional)
│ ✔	spaceplot!(ax, model::ABM{CustomSpace{2, false}}; preplotkwargs...)
│ ✔	static_preplot!(ax, model::ABM{CustomSpace{2, false}}, p::ABMPlot)
│ === Lifting (optional)
│ placeholder
│ === Inspection (optional)
│ ✔	ids_to_inspect(model::ABM{CustomSpace{2, false}}, pos)
```

is very nice and we should probably add something similar for the other customizations in Agents.jl, but maybe a keyword to disable the checks (even if enabled by default)?

Tortar:
nothing space:

Screenshot from 2023-11-16 19-02-06


What about removing the blank plot instead? 

And also maybe put just two columns of charts, followed by at the bottom sliders + buttons? I don't know how much code changes this requires, maybe another PR outside the 6.0 release scope could improve this

ext/AgentsVisualizations/src/spaces/continuous.jl
```diff
@@ -0,0 +1,45 @@
...
+    return o, e
+end
+
+## Preplots
+
+"""
+Plot heatmap according to given `heatarray`.
+Special method for models with `ContinuousSpace`.
+"""
+function heatmap!(ax, p::_ABMPlot{<:Tuple{<:ABMObservable{<:Observable{<:ABM{<:S}}}}}
```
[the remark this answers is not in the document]
Tortar: Actually I'm not sure how this is done due to the nesting, but maybe it is possible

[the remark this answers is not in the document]
Tortar: I think it should be `alias = _ABMPlot{<:Tuple{<:ABMObservable{<:Observable{<:ABM{<:S}}}} where {S}`

ext/AgentsGraphVisualizations/AgentsGraphVisualizations.jl
```diff
@@ -1,20 +1,21 @@
...
-    node_color, 
-    node_marker, 
-    node_size,
-    graphplotkwargs,
-    edge_color, 
-    edge_width)
-    graphplot!(abmplot, graph; node_color=node_color, node_marker=node_marker, 
-               node_size=node_size, graphplotkwargs..., # must come first to not overwrite lifted kwargs
-               edge_color, edge_width)
+include("src/spaces/graph.jl")
```
Tortar, line 7:
```suggestion
```

ext/AgentsOSMVisualizations/AgentsOSMVisualizations.jl
```diff
@@ -1,15 +1,17 @@
 module AgentsOSMVisualizations
 
-using Agents, OSMMakie
-using OSMMakie.Makie: Axis, @colorant_str
+using Agents, Makie, OSMMakie
+const _ABMPlot = Agents.get_ABMPlot_type()
+
+include("src/spaces/openstreetmap.jl")
```
Tortar, line 6:
```suggestion
```

Tortar (commented):
Shouldn't be necessary (also tested)

ext/AgentsGraphVisualizations/AgentsGraphVisualizations.jl
fbanning: Why should this not be necessary? The files have to be included to load the plotting methods for GraphSpace and OpenStreetMapSpace.

When you tested it, did you maybe still have those methods loaded in a running Julia instance? Maybe you forgot to restart Julia to reload the extensions and test whether it works without the include statement?

fbanning:
> What about removing the blank plot instead?

I've wrestled with this for multiple hours straight and may have gotten a few more gray hairs because of it. A few problems with this: The axis is already created outside of the actual recipe, so we would need to circumvent this. However, without any axis, we don't have access to the scene created by the recipe and all the nice things it entails, like easy access to lifted plot properties, the abmobs, etc. Also Makie threw nasty errors at me when I just tried to leave the plot completely empty or when I even tried to delete any placeholder plots from inside the recipe. It was really quite a hassle and in the end I opted for this simple version.

Sure, it would really be nicer to just not show an empty plot for Nothing spaces. So I would prefer to postpone this change for a later release. It's really not that big of a deal right now because whoever wants to plot their Nothing space models in a specific way, has to deal with the agents visualizations API anyways. :)

fbanning:
> maybe a keyword to disable the checks (even if enabled by default)?

That's a good idea that I also had at some point along the way. Eventually I forgot to implement it. Will do. :)

ext/AgentsVisualizations/src/spaces/continuous.jl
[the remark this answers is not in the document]
fbanning: Yeah for readability this would be great. I'm not sure if we can easily use `where` inside an alias and then properly access `S` from inside the functions like we normally do. Need to test that out.

ext/AgentsGraphVisualizations/AgentsGraphVisualizations.jl
Tortar: never mind, I now noticed that the package didn't trigger precompilation when I made the change and so I assumed it worked

Tortar:
It is also true removing the blank plot is even somewhat unrelated to the scope of this PR, so let's postpone it. I'll open an issue about this

ext/AgentsVisualizations/src/spaces/abstract.jl
```diff
@@ -0,0 +1,100 @@
+Agents.agents_space_dimensionality(model::ABM{<:Agents.SpaceType}) =
```
Tortar, line 1:
```suggestion
Agents.agents_space_dimensionality(model::ABM) = 
```

Tortar: I think that if you aren't interested in qualifying the space type then you can also remove it

ext/AgentsVisualizations/src/spaces/continuous.jl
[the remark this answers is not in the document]
Tortar: I'm also asking myself if you need to qualify it this much: why not just using `_ABMPlot`, it doesn't seem the code uses `S` most of the times, and maybe when you use it, you can just create a function which extract the spacetype from the `_ABMPlot` type e.g. 

`abmspace(::_ABMPlot{<:Tuple{<:ABMObservable{<:Observable{<:ABM{<:S}}}}) where {S} = S`

ext/AgentsVisualizations/src/spaces/abstract.jl
fbanning: That's true. At some point earlier in the process I had `AbstractSpace` here but once I've added the `Nothing` space option, I've changed it to `SpaceType`. I will remove the space type annotations in this file.

ext/AgentsVisualizations/src/spaces/continuous.jl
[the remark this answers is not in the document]
fbanning: If I understood your last suggestion correctly, then this would only save us the `where {S}` part in the function arguments. In this case here, we would still need the whole `_ABMPlot{<:Tuple{<:ABMObservable{<:Observable{<:ABM{<:ContinuousSpace}}}}}` to properly dispatch to the correct method of the `heatmap!` function. I would argue that the extra step of introducing a `abmspacetype` function isn't really all that helpful. All these functions are internal and since all of us know about the functioning of `where`, it's just easier like it currently is. But that's only my opinion. If you disagree, I can definitely make the suggested change.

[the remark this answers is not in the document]
Tortar: if that's the case I agree with you, but I would nonetheless try the aliasing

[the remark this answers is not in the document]
fbanning: The aliasing works even with something like `Alias{S} = T1{T2{T3{<:S}}}` so it's all good. I've added an alias called `ABMP{S}` that I've now used wherever appropriate. Hope this makes things a bit more legible.

Tortar:
I created sometime ago a test file for visualizations: https://github.com/JuliaDynamics/Agents.jl/blob/main/test/visualization_tests.jl, but it is just a scratch. Maybe we can substitute the tests there with your different tests for this PR? The tests could be just if the commands works for now, we will maybe add some introspections later, what do you think?

fbanning:
Sure, I would be happy with something like that. It would be a great first attempt at a mini testsuite for visualizations.

We could also try to build a test suite similar to how GraphMakie does it. They first create persistent reference images that are then compared to the output generated by the plotting functions. Maybe we can do something similar?

Tests for interactivity are somewhat complicated to write. Although we could programmatically change the observable values for clicks of each button and then check whether the model state correctly changes. Bigger topic though, so maybe something for another PR?

Tortar:
Yes, the good extensions you are proposing are better for another PR, in this one you could just slap your .txt-actually-jl content in the test file in my opinion :D

Datseris:
I'll also review this, let me know when you'd like that!

fbanning:
OK, today's commits brought back all the core functionality that we had before. This means that inspection for `GraphSpace` and `OSMSpace` are working again. As a nice addition, their implementations serve as nice examples of what users can do when they have a bit more complex custom spaces and how they can handle things inside their userspace code. :)

Feel free to give a review a shot now, @Datseris and @Tortar. As can be seen in the todos listed in the first post of this PR, there's mostly the user-facing documentation left to do. I will write this once I have received your feedback on the code so that I know its state is somewhat final.

src/spaces/openstreetmap.jl
```diff
@@ -951,6 +951,26 @@ function Agents.nearby_ids(
...
+    # the Float32 conversion is needed for inspection in visualization extension
+    # TODO maybe track agents' lonlat positions as Float32 inside the space itself?
+    ids_lonlat32 = (Pair(id, Float32.(lonlat(model[id], model))) for id in allids(model))
+    for (id, apos) in ids_lonlat32
+        if sqrt(sum(abs2.(pos .- apos))) <= distance # simple euclidean distance
+            push!(nearby, id)
+        end
+    end
+    return nearby
+end
```
Tortar, line 972:
```suggestion
    # need to create an iterator for all agents' lonlat positions as Float32
    # the Float32 conversion is needed for inspection in visualization extension
    # TODO maybe track agents' lonlat positions as Float32 inside the space itself?
    ids_lonlat32 = ((id, Float32.(lonlat(model[id], model)) for id in allids(model))
    nearby = (id for (id, apos) in ids_lonlat32 if euclidean_distance(pos, apos, model) <= distance)
    return nearby
end
```

Tortar: I'm also not sure why you need the float32 conversion if you only return nearby ids from this function?

fbanning: Our `lonlat` function returns Float64 values and the positions in Makie plots are `Float32`. We have to convert this here because of the lower precision that comes from the mouse position. Otherwise we would need to run some `isapprox` on the `Float64` `lonlat` positions and to be honest, converting them to `Float32` seemed like the way simpler approach to me.

fbanning: Your suggestion to use the `euclidean_distance` method is totally fine with me but we would need to define that first for `OSMSpace`. Do we need this method somewhere else as well (or maybe in the future)? If so, I would also prefer to have it as a specialised `euclidean_distance` method.

fbanning:
I've tested the flocking model with polygons and it works. I've also found a way to deal with the weird polygon inspection that was bothering me for a long time. It's not perfect but it's a workable hack for now. Maybe someone can come up with a better version at some point.

fbanning:
And after adapting the [3D predator prey model](https://juliadynamics.github.io/Agents.jl/stable/examples/rabbit_fox_hawk/) to work with Agents 6.0, I can also confirm that 3D visualizations work fine.

Sidenote: Once 6.0 is published, I will push my adapted version of the model code to the Agents example zoo repo.

Screenshot_20231124_165142

src/visualizations.jl
```diff
@@ -37,7 +37,8 @@ See also [`abmvideo`](@ref) and [`abmexploration`](@ref).
 * `offset = nothing` : If not `nothing`, it must be a function taking as an input an
   agent and outputting an offset position tuple to be added to the agent's position
   (which matters only if there is overlap).
-* `scatterkwargs = ()` : Additional keyword arguments propagated to the `scatter!` call.
+* `agentsplotkwargs = ()` : Additional keyword arguments propagated to the `agentsplot!` 
+  call.
```
Datseris, line 41: Here we refer to an `agentsplot!` call, but this is not explained in the docstring. A user that has not familiarized themselves with the developer internals of the plotting won't know what this means. We could say something like "Additional keyword arguments propagated to the function that plots the agents (typically `scatter!`).".

src/visualizations.jl
```diff
@@ -58,12 +59,13 @@ See also [`abmvideo`](@ref) and [`abmexploration`](@ref).
   placed naturally.
 * `static_preplot!` : A function `f(ax, model)` that plots something after the heatmap
   but before the agents.
-* `osmkwargs = NamedTuple()` : keywords directly passed to `OSMMakie.osmplot!`
-  if model space is `OpenStreetMapSpace`.
-* `graphplotkwargs = NamedTuple()` : keywords directly passed to
-  [`GraphMakie.graphplot!`](https://graph.makie.org/stable/#GraphMakie.graphplot)
+* `spaceplotkwargs = NamedTuple()` : keywords used in `spaceplot!`. Directly passed to
```
Datseris, line 62: Similarly to the above comment, `spaceplot!` is referred to here but is not explained or defined anywhere above. So we can say something like "keywords utilized when plotting the space.". The next sentences explain where they are used in more detail.

src/visualizations.jl
```diff
@@ -236,8 +260,100 @@ Scale given polygon by `s`, assuming polygon's center of reference is the origin
...
+
+"""
+    agents_space_dimensionality(space::S) where {S<:Agents.AbstractSpace}
+
+Return dimensionality of given model space.
+"""
+function agents_space_dimensionality end
+
+"""
+    get_axis_limits!(model::ABM{S}) where {S<:Agents.AbstractSpace}
```
Datseris, line 286: Why does this end in `!`?

src/visualizations.jl
```diff
@@ -236,8 +260,100 @@ Scale given polygon by `s`, assuming polygon's center of reference is the origin
...
+Plot agents at their positions.
+"""
+function agentsplot! end
+
+## Preplots
+
+"""
+    spaceplot!(ax, model::ABM{S}; spaceplotkwargs...) where {S<:Agents.AbstractSpace}
+
+Create a space-dependent preplot.
```
Datseris, line 305: I think I have not understood why `agentsplot!` requires the `ABMPlot` object but `spaceplot!` does not.

ext/AgentsVisualizations/src/spaces/space-visualization-API.jl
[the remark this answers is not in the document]
Datseris: In that scenario, we would get a `Methoderror` instead of a `"not implemented" error`.

ext/AgentsVisualizations/src/utils.jl
```diff
@@ -1,6 +1,89 @@
 export subscript, superscript
 export record_interaction
 
+function Agents.check_space_visualization_API(::ABM{S}) where {S}
```
Datseris, line 4: I am not sure about this function. It is lots of code that is hard to maintain and does not really offer much. If a method is not implemented, the developer will get a `MethodError` error message when trying to run the code. We have already instructed them to go through the file and implement the methods in the file `space-visualization-API.jl`. I understand that it took you a lot of effort to write this function, but perhaps we should not keep it. It's lots of lines of code that offer little and would incur maintainance burder each time something new is added to the plotting API.

src/spaces/openstreetmap.jl
```diff
@@ -951,6 +951,26 @@ function Agents.nearby_ids(
     return nearby
 end
 
+function Agents.nearby_ids(
```
Datseris, line 954: I haven't understood why this function exists. THere is already a version for this function, right?

src/spaces/openstreetmap.jl
```diff
@@ -951,6 +951,26 @@ function Agents.nearby_ids(
...
+        pos::Tuple{Float32, Float32},
+        model::ABM{<:OpenStreetMapSpace},
+        distance::Real,
+        args...;
+        kwargs...
+    )
+    nearby = []
```
Datseris, line 961: This is type unstable. It should be `Int[]`.

Datseris (commented):
I think the refactoring is very good! It is now rather clear how to add support for a new space in the visualization process. This should be mentioned also in the developer docs.

One thing that we should further check is that all plotting is model agnostic. I.e., it should work out of the box with either the current `StandardABM` or the new `EventQueueABM`. For this to occur, all methods that access model should use the model accessing API. E.g., use things such as `allagents(model)` to get the list of agents, etc. Also ensure that we don't type annotate `::StandardABM` but rather the abstract type `::ABM`. I haven't tested whether this works yet with `EventQueueABM`, but we can test this after merging both this and #940 .

src/spaces/openstreetmap.jl
Tortar: Related: https://github.com/JuliaDynamics/Agents.jl/pull/934#discussion_r1403801703

ext/AgentsVisualizations/src/spaces/space-visualization-API.jl
```diff
@@ -0,0 +1,84 @@
...
+=#
+
+## Required
+const ABMPlot = Agents.get_ABMPlot_type()
+
+Agents.agents_space_dimensionality(space::Agents.AbstractSpace) =
+    notimplemented(space)
+
+Agents.get_axis_limits!(model::ABM{<:Agents.AbstractSpace}) =
+    notimplemented(abmspace(model))
```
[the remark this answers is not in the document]
fbanning: That's correct, thanks for noticing. Fixed the naming.

src/visualizations.jl
fbanning: Because `agentsplot!` is dynamic which means that it updates over time whenever the model field inside the `ABMObservable` changes. `spaceplot!` on the other hand is static because (as far as I can anticipate right now) it only relies on the `space` field so we can simply pass `model` instead.

fbanning: If we want to unify the arguments of `agentsplot!` and `spaceplot!`, I would propose to use only `ABMPlot` because all the info is contained in there and remove the `model` argument. Would your prefer that or were you just confused about the differences in arguments taken by the two functions and wanted to clarify why?

src/visualizations.jl
fbanning: You're right. Fixed that.

src/visualizations.jl
fbanning: Fixed.

src/spaces/openstreetmap.jl
fbanning: Type instability is fixed.

src/spaces/openstreetmap.jl
fbanning: We talked about this in another comment thread (https://github.com/JuliaDynamics/Agents.jl/pull/934#discussion_r1403802913). Maybe you'd like to chime in over there how we would like to handle this case?

fbanning:
> One thing that we should further check is that all plotting is model agnostic. I.e., it should work out of the box with either the current `StandardABM` or the new `EventQueueABM`. For this to occur, all methods that access model should use the model accessing API. E.g., use things such as `allagents(model)` to get the list of agents, etc. 

Afaict there is no place in the plotting extensions that *doesn't* use `allagents`/`abmspace`/whatever API functions. 

> Also ensure that we don't type annotate `::StandardABM` but rather the abstract type `::ABM`.

I've only used `ABM` everywhere.

> I haven't tested whether this works yet with `EventQueueABM`, but we can test this after merging both this and #940 .

Yep, I think this can be done after merging this PR. If any problems occur, we can fix them separately because they're not directly related to the refactor but to an extension of model types.

ext/AgentsVisualizations/src/utils.jl
fbanning: Hm. I think this function is a big step forward in terms of developer experience. It provides an easy way to check if the own space type *can be plotted* and if something *optional is missing*.

I see your point though. It may prove a bit cumbersome in the long term to maintain this function. But I think the plotting API is pretty fixed for now and whenever something gets added to it, we just have to make it a requirement that the method check is added to this function. Not that hard to do imho but certainly a thing that has to be kept in mind.

ext/AgentsVisualizations/src/spaces/space-visualization-API.jl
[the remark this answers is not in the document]
fbanning: This file is there to improve the developer experience. It helps to quickly understand which methods *can* be defined and which args/kwargs they require. I took inspiration from the `src/core/space_interaction_API.jl` file in the main module.

As described in the long comment at the start of the `space-visualization-API.jl` file, the basic workflow for extending visualisations with a custom space would be: copy the file, replace the space to dispatch on with the custom space type, remove the unnecessary method definitions, and finally fill in the functions as desired. Like this, our users don't need to "gather" the method definitions from the docstrings in the `visualizations.jl` file.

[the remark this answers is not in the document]
fbanning: We could also move the whole content of this file into a multi-line code block inside the developer documentation. By that it wouldn't clutter the extension's source code but would still be easily accessible to our user. Would that be a feasible approach? We could do the same for the content of the `space_interaction_API.jl` file.

ext/AgentsVisualizations/src/utils.jl
Tortar: To me it seems a good idea instead

- not too hard to maintain
- cool
- what if a method is called rarely and the method error so happens unexpectedly? I think that knowing in advance the methods one should implement is a safeguard in this regard

Datseris: Okay, I accept, go ahead and resolve the thread!

ext/AgentsVisualizations/src/spaces/space-visualization-API.jl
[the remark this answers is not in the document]
Datseris: Sure, I get your point, but aren't we doing things twice? You have already created the function that explicitly checks if functions are implemented, but now you have here an entire extra file for also adding "not implemented" error messages to the functions not implemented. Aren't these two things essentially doing the same? Why do we need two files to do in essence the same job?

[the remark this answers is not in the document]
fbanning: I had the same thought that they're somewhat redundant. We have now agreed that the custom space API checker function will stay and that it is exposed to the users to check their space implementations. So it's easy for users to find out if they missed some important method.

(Quick sidenote in case this wasn't clear: `space-visualization-API.jl` is not included, just like `space_interaction_API.jl` isn't, so their purpose is really just for developer reference.)

I would then argue it would be best to just move the content of both `ext/AgentsVisualizations/src/spaces/space-visualization-API.jl` and `src/core/space_interaction_API.jl` into the developer docs. Right now the docs link to the `space_interaction_API.jl` file for those who are interested ([see here](https://juliadynamics.github.io/Agents.jl/stable/devdocs/#Creating-a-new-space-type-1)). The descriptions at the top of the files about the methods to implement are really helpful and the ability to just copy some of them and easily implement own methods is also nice. Would that be OK with you?

fbanning:
Just found out that how we handle `offset`s of agent positions is not so optimal for inspection. Will add that to the list of things that have to be fixed in this PR as well.

src/visualizations.jl
Tortar: I think that having more similar arguments if possible is a good idea :+1:

fbanning:
Inspection of plots with positions with offset works for `GridSpace` now but it's just a workaround. I need to have a closer look at how to handle this. The problem is that offsets are generated on the fly and we don't have a way to keep track of the offset values for each agent position so that we could correct from inside the inspection function. It's nothing crucial for now and it's also not a regression due to the refactor. I'm pretty sure that this has been a bug before this PR and it was just never used by anyone so it went unnoticed.

fbanning:
In other news: `Makie v0.20` was released a few days ago which is a great release (seriously!), however it introduced a few issues.
- Inspection tooltips are now drawn below the Axis spines which is not so nice to look at. This is a Makie issue ([see here](https://github.com/MakieOrg/Makie.jl/issues/3436)) and I don't know how we could fix it on our side. Let's wait for an answer there, I'm sure they'll have an idea how to sort it out. :)
- Agent inspection of 3D `ContinuousSpace` models (e.g. Predator-Prey-3D) is broken. Somehow the text of the tooltips always stays empty and I don't know yet what's the issue here. Need to dig a bit deeper.
- Models with polygon plots (e.g. the flocking example) are very slow to run and the polygons are also way bigger than they were with `Makie v0.19`. Setting `as = 5` fixes it in the short term but after running `abmplot` a few times, I get a weird atlas warning from Makie that I've never seen before. Need to find out how to properly replicate this behaviour and then report the bug over there.

Tortar:
Seems like they solved the problems in Makie if I'm not mistaken!

Related to #948, can we fix the version of Makie in the project.toml to 0.20? So that we can control better that everything works before updating

fbanning:
> Seems like they solved the problems in Makie if I'm not mistaken!

Yep, it's fixed. :)

> Related to #948, can we fix the version of Makie in the project.toml to 0.20?

Sure we can do that. Would make sense to do it in this PR, so I'll add that to the to-do list.

I've been on vacation the past days and haven't found any time to look at the other two issues that I mentioned after the switch to Makie 0.20. Will try to work on them *soonish* and if I'm unable to fix them, then it's maybe better to move them into separate issues that we can work on after this PR has been merged.

Datseris:
sorry, i've been away for too long. Do you guys need any support from me here?

Tortar:
Hi @fbanning do you happen to have some time to finish up this PR? Because we are really close to finalize everything for v6 I think, otherwise we could merge this as is, and improve the rest in the future

fbanning:
I don't see myself finding the time to finish this PR in the next days. If 6.0 should be released really soon, feel free to merge. The new plotting infrastructure should mostly work as far as I can tell from the stuff done two months ago. It won't be without bugs (e.g. some inspection issues I know of and also the open to-dos from the list above). But maybe it's also a good idea to start a clean PR after this has been merged to fix all the remaining issues with it (and new ones that will probably arise soon after this has been released).

Oh and I just saw there are two file conflicts that need to be resolved. Those haven't been there the last time I committed to this branch, but it's likely they can be fixed easily.

Tortar (approved):
Tests pass on my machine so I approve. There is one conversation left but it doesn't seem blocking. We are left with the problems stemming from the update of Makie.

Tortar:
thank you for all the past work @fbanning ! I think as you say that it is better to improve what needs to before v6 in separate PRs

PR #938, opened 2023-11-20
Remove ) causing parse error in example_zoo.jl

jacobusmmsmit:
Title


Tortar (approved):

Tortar:
I think because you created a branch on a fork instead that on main, but in any case it is problematic if it doesn't work on forks because you need privileges to create a branch on main

PR #939, opened 2023-11-24
Move `Models` submodule into test folder

fbanning:
This PR does three things:

1. It moves the `Models` submodule out of the main `Agents` package and into the `test` folder. This makes it available internally for us for the documentation and the test suites but doesn't expose it to the users anymore. The module is also renamed to `AgentsTestModels` to avoid any confusions and/or potential naming collisions.
1. It extends the predefined models that are available via the `Models` submodule. The `rabbit_fox_hawk` model (from our examples section) has been added to the set of available predefined models and can be accessed via the `AgentsTestModels.rabbit_fox_hawk()` function.
2. It makes sure that our model initialisation functions only ever return `model` and not `agent_step!` and `model_step!` as well because they are included in the `ABM` struct now.

~~Important note: Adding the rabbit_fox_hawk model in its current form requires us to add `FileIO` package as a dependency. Is that acceptable to you, @Datseris?~~


Tortar:
What about not downloading the png and make it available inside Agents.jl, it is 20 kb so it should be okay. It takes 0.1 second to load `FileIO` so I think it's better to just make the image available inside Agents

Tortar:
sorry but it was unrelated since it is for loading the image

Tortar:
A better solution in my opinion is to move the entire `models` in the `test` folder, in the end we use these models only for testing if I'm not mistaken

Datseris:
That is correct, we no longer mention `Models` in the documentation because providing pre-defined models is deprecated. We use them in the tests but what @Tortar says is the way to go I think.

fbanning:
I've changed the direction of this PR now to reflect your suggestions to fully remove the  `Models` submodule. It is now called `AgentsTestModels` and is located inside the `test` folder from now on. It is thus not exposed to or accessible by the users anymore.

> in the end we use these models only for testing if I'm not mistaken

We also reference it in the example section of our documentation but those references can easily be replaced or removed which I've done now.

Tortar (approved):
seems good to me (if tests pass :D)

examples/agents_visualizations.jl
```diff
@@ -228,7 +230,8 @@ fig
 # and plot it with [`abmplot`](@ref).
 using Graphs
 using ColorTypes
-sir_model = Models.sir()
+using .AgentsTestModels: sir #hide
+sir_model = sir()
```
Tortar, line 234: Actually I'm a bit concerned about this: does it mean that the user won't be able to copy-paste the code in the example? It will error right? Can we instead use a sort of relative path from the Agents module to import it so that it should also work for a user?

Tortar: Also `include("../test/AgentsTestModels/AgentsTestModels.jl") #hide` has the same problem, I think that adding `#hide` makes the problem worse

Tortar: or maybe another solution would be to put those models used in the documentation inside `AgentsExampleZoo` and then import them from there?

Tortar: mmmh, given the structure of `AgentsExampleZoo` which just uses docs, to me it seems better the other way

Tortar: ok, It should be `include(joinpath(splitpath(pathof(Agents))[1:end-2]..., "test/AgentsTestModels/AgentsTestModels.jl"))`, maybe we can call it something like `include_Agents_test_models()`

fbanning: > does it mean that the user won't be able to copy-paste the code in the example? 

Not with a simple click but if they just go the the link mentioned in the docs themselves, they can copy the model quite easily.

> maybe we can call it this way Agents.include_test_models()

Creating such a helper function is just a more complicated way to have the `Models` submodule in the package itself. I'm against such a half-baked solution. Either we don't expose those predefined models to the users at all or we put all of this in `AgentsExampleZoo`. I've actually thought about the latter quite a bit while doing this PR and I would be in favour of that. Give me your opinion and we'll decide what to do. :)

Tortar:
yes, in my opinion it is the cleanest solution to have them in the
AgentsExamplesZoo :-)

Il Mar 19 Dic 2023, 09:25 Frederik Banning ***@***.***> ha
scritto:

Tortar:
I updated this pr given that now models are present in the `AgentsExampleZoo` :-)

Datseris:
thanks for this guys!

PR #940, opened 2023-11-25
EventQueue ABM v3

### LINKED ISSUE #760: Implementing the Gillespie algorithm

### LINKED ISSUE #902: Design discussion for an Event-Based (Gillespie-like) agent based model in Agents.jl

Datseris:
Closes #760
Closes #902

Alright, this is pretty cool. To review this PR read the following three things:

1. The docstring of `StandardABM` in `model_standard.jl`
2. The docstring of `EventQueueABM` in `movel_event_queue.jl`
3. The example `event_rock_paper_scissors.jl` file for an example application.

So this PR superseeds #917 by changing a bit the model declaration.

We now have events, that include an action (`agent_step!`), can be applied to arbitrary agents. Event generation can happen in two ways:

1. Automatically, by sampling randomly by propensities. This automatic can happen in (a) when an agent is added and (b) when an event finishes its action.
2. A user may manually add events to the que via `add_event!` or `generate_event_to_queue!`.

This is the best of all worlds. It solves all problems. It allows the model to work without the user having to provide a list of initial events. But also allows them to take full control of event scheduling if they want.

Also, it allows going beyond the standard Gillespie simulations by allowing the event times to be arbitrary functions. No longer limited to an exponentially distributed time.

By default however we do have an exponentially distributed time.

Please review by reading the three files I mention at the start! The PR is not finished of course, there is more docmentation to be added and tests. However, I would argue, if we agree with the interface we should merge and we should ask for tests in another PR by opening an issue.

cc @MA-Ramirez


jacobusmmsmit:
Not a review (yet), but I was reading your branch this morning and I like the proposed changes. Good solution to the "initial events" problem.

examples/event_rock_paper_scissors.jl
```diff
@@ -0,0 +1,221 @@
...
+# which is `Union{Scissors, Paper}`.
+# Additionally, we would like to change how the timing of the movement events works.
+# We want to change it from an exponential distribution sample to something else.
+# This "something else" is once again an arbitrary Julia function,
+# and for here we will make:
+
+function movement_time(agent, model)
+    ## `agent` is the agent the event will be applied to,
+    ## which we do not use in this function!
+    t = 0.1randn(abmrng(model)) + 1
```
Tortar, line 146: small thing, usually we use this convention:

```suggestion
    t = 0.1 * randn(abmrng(model)) + 1
```

examples/event_rock_paper_scissors.jl
```diff
@@ -0,0 +1,221 @@
...
+# so we need to specify the agent super type that it applies to,
+# which is `Union{Scissors, Paper}`.
+# Additionally, we would like to change how the timing of the movement events works.
+# We want to change it from an exponential distribution sample to something else.
+# This "something else" is once again an arbitrary Julia function,
+# and for here we will make:
+
+function movement_time(agent, model)
+    ## `agent` is the agent the event will be applied to,
+    ## which we do not use in this function!
```
Tortar, line 145:
```suggestion
    # `agent` is the agent the event will be applied to,
    # which we do not use in this function!
```

src/core/model_event_queue.jl
```diff
@@ -0,0 +1,278 @@
...
+    if getfield(model, :autogenerate_on_add)
+        generate_event_in_queue!(agent, model)
+    end
+end
+
+# This is the main function that is called in `step!`
+# after the other main function `process_event!` is
+# done
+"""
+    generate_event_in_queue!(agent, model)
```
Tortar, line 200: I propose to dispatch this to `add_event!`

```suggestion
    add_event!(agent, model)
```

src/simulations/step.jl
```diff
@@ -3,19 +3,27 @@ using CommonSolve: step!
 export step!, dummystep
 
 """
-    step!(model::ABM [, n::Int = 1])
+    step!(model::AgentBasedModel)
```
Tortar, line 6:
```suggestion
    step!(model::ABM)
```

src/simulations/step.jl
```diff
@@ -36,19 +44,43 @@ function CommonSolve.step!(model::ABM, n::Union{Function, Int} = 1)
...
+function CommonSolve.step!(model::EventQueueABM)
+    next_event_t = first(abmqueue(model))[2] # second entry of pair is the time
+    dt = next_event_t - abmtime(model)
+    step!(model, dt)
+end
 
-Used instead of `agent_step!` in [`step!`](@ref) if no function is useful to be defined.
-"""
-dummystep(model) = nothing
-dummystep(agent, model) = nothing
```
Tortar, line 51: Removing `dummystep` breaks backward compatibilty

Tortar: ops, you just moved them

Tortar:
Small review, will take a closer look tomorrow!

Datseris:
I won't have the time to work on this until next weekend again. If anyone wants to contribute (by e.g., making `run!` work) go ahead.

This PR has a major problem: **its performance is abysmal**. Go to the rock example and then do
```
@btime step!(model)
```
which will perform a single event and schedule one more event. Well, this takes 1 μs and allocates a million times. That's crazy. We need to do some sort of performance optimization.

It could very well be that the API I have here is simply bad for performance and one needs to think of different ways for users to specify the events... I am not sure.

On the other hand, we could say that "well, this way of modelling is just much slower than the discrete time version."

I am sure the majority of the performance drop comes from type instabilities...

Datseris:
Hm. Okay. If you can have a look at my code and see if you can bring in your code to increase performance, that would be great! If we can't fix the performance problems, then we go back to your code and add my ideas from here for creating new events on agent addition.

It could be that my decision to allow for arbitrary timing functions rather than exponential sampling is also part of the problem.

Tortar:
will try to fix it! Hopefully shouldn't be too hard

Tortar:
I think it is mostly due to how AgentEvent is conceived, it is actually okay to give it to users, but it shouldn't be used internally since it has abstract containers and increases the number of branch in the code. Instead the previous tuple of tuples way should be used which is fast

Datseris:
I hope that's the solution. I doubt it though because `AgentEvent` is concretely typed and immutable so it should be equivalent with a tuple. But I hope I am wrong and there is an easy solution.

Tortar:
AbstractAgent is not concrete

Datseris:
> AbstractAgent is not concrete

This doesn't matter. The field `types` is concretely typed. It contains a `Type` instance.

Tortar:
Yes, you are right @Datseris 

Profiling says that

```
function obtain_propensity(event::AgentEvent, agent, model)
    if event.propensity isa Real
        return event.propensity
    else
        p = event.propensity(agent, model)
        return p
    end
end
```

is the cause, what's the error? :D

jacobusmmsmit:
Nooo, all of the type instabilities will be fixed by the time I wake up tomorrow at this rate :/

jacobusmmsmit:
Is the return type of the propensity function known at call time? Is it the return type of `Union{Real,FunctionReturn}` causing the instability?

Tortar:
ahh nothing, this is the problem

```julia
function reproduction_propensity(agent, model)
    same = count(a -> a isa typeof(agent), allagents(model))
    return 2*same/nagents(model)
end
```

is this, there is no big problem in the actual code

Tortar:
Before 
```

 Range (min … max):  421.000 ns …   3.147 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     116.504 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   126.073 μs ± 123.855 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅                     ▄█▇▂                                     
  █▂▂▂▁▁▁▁▁▁▁▁▂▁▁▁▁▁▁▂▃▇████▆▄▄▄▃▃▃▂▂▂▂▂▂▂▂▂▃▃▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂ ▃
  421 ns           Histogram: frequency by time          293 μs <

 Memory estimate: 160 bytes, allocs estimate: 1.
```


After (with reproduction_propensity(agent, model) = 1/2 * nagents(model)):

```julia
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  674.300 ns … 211.674 μs  ┊ GC (min … max): 0.00% … 98.00%
 Time  (median):     905.700 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   996.727 ns ±   2.948 μs  ┊ GC (mean ± σ):  4.09% ±  1.39%

           ▁▃▅▆██▇█▇▅▃▃▁                                         
  ▁▁▁▂▂▃▄▅▇██████████████▆▅▅▄▄▄▄▄▅▅▆▄▅▄▄▃▃▃▃▃▂▂▂▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁ ▄
  674 ns           Histogram: frequency by time         1.47 μs <

 Memory estimate: 808 bytes, allocs estimate: 12.

```

src/simulations/step.jl
```diff
@@ -71,11 +60,40 @@ function CommonSolve.step!(model::EventQueueABM)
...
+        model_t[] = stop_time
+        return
+    end
+    event_tuple, t_event = dequeue_pair!(queue)
+    if t_event > stop_time
+        enqueue!(queue, event_tuple => t_event)
+        model_t[] = stop_time
+    else
+        process_event!(event_tuple, model)
+        model_t[] = t_event
```
Datseris, line 88: I think here we first need to update the model time and then process the event. Because, the event occurs at the even time. So that if the event reads the model time, the correct time to read is the time the event is supposed to occur at.

Tortar: right :+1:

jacobusmmsmit:
It's lame of me to just come in here and stop the merge for such a seemingly small thing _but_ I believe it would be good to change the name of the `timing` argument as it's part of the public API and is harder to change once fixed.

I propose we use a longer, less ambiguous name. I think `interevent_time` or `interevent_distribution` may be better. Currently, the name `timing` is explained in the docstring every time it's mentioned, someone reading the code may not know exactly what it means when seeing it written down. I think having Agents.jl code easy to read is a top priority.

jacobusmmsmit:
In that case let's come back to it when more important things are resolved
:)

Tortar:
Optimized a bit:

```julia
#before
julia> @time step!(model, 0.01)
  0.420770 seconds (6.30 M allocations: 349.013 MiB, 4.37% gc time)
#after 
julia> @time step!(model, 0.01)
  0.243629 seconds (2.88 M allocations: 207.045 MiB, 6.63% gc time)
```

changing a small assumption, basically now the defaults of `AgentEvent.timing` is `exp_propensity` but it means that we add the propensity as an argument, then I changed also the API requiring to pass propensity even for the user, which I think it should be okay.

Still the code should be cleaned up and docstrings updated

Tortar:
I think there are still some bugs in behaviour on this implementation, but
hopefully fixing them shouldn't change the perf improvement

Il Mar 28 Nov 2023, 18:11 Martin Smit ***@***.***> ha scritto:

> In that case let's come back to it when more important things are resolved
> :)
>

Datseris:
I've been thinking a lot about the performance of this, and in general multi-agent models. What if, for all multi agent models we use metaprogramming to create a single special agent that has all fields that all agents have? And this is used everywhere. For this specially generated agent the function `typeof(...)` would be overloaded to return a special field `.type` that contains the type. This is a bit off topic of this PR and in general about addressing multi-agent model performance.

(I am thinking about this because increasing the performance here by 30% is nothing. It would still take microseconds for a single agent step, while it should be taking nanoseconds)

Datseris:
Yeah okay. In any case I think my suggestion can be also be done by creating a third type of agent container. a "multi-agent" container that will generate this agent type.

Datseris:
And we can also leave the multi-agent stuff for v7 if it's breaking and it\s not possible to do it non breaking

Tortar:
> if using a single agent type will surely improve a bit the performance

anyway I'm interested in seeing how much, I will rapidly test this :-)

Datseris:
> anyway I'm interested in seeing how much, I will rapidly test this :-)

Well in this example it is quite trivial to have only 1 agent, you just add a named field for the type. No need for creating extra code!

Tortar:
exactly :+1:

Tortar:
1 type RockPaperScissors

```julia
julia> @benchmark step!($model) seconds=1 evals=1 samples=10^6
BenchmarkTools.Trial: 471880 samples with 1 evaluation.
 Range (min … max):  271.000 ns … 2.536 ms  ┊ GC (min … max): 0.00% … 98.72%
 Time  (median):     551.000 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   597.350 ns ± 3.697 μs  ┊ GC (mean ± σ):  0.89% ±  0.14%

     ▁█▁▂     ▄█▁▆▃    ▁▁                                      
  ▁▂▃████▆▃▅▆▅█████▅██▅██▅█▇▄▆▆▃▅▅▃▄▄▂▃▃▂▃▃▂▂▂▂▂▂▁▂▂▁▂▁▁▁▁▁▁▁ ▃
  271 ns          Histogram: frequency by time        1.25 μs <

 Memory estimate: 224 bytes, allocs estimate: 6.
```

3 types Rock, Paper, Scissors

```julia
julia> @benchmark step!($model) seconds=1 evals=1 samples=10^6
BenchmarkTools.Trial: 397554 samples with 1 evaluation.
 Range (min … max):  160.000 ns … 2.875 ms  ┊ GC (min … max): 0.00% … 98.36%
 Time  (median):     852.000 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   925.216 ns ± 9.526 μs  ┊ GC (mean ± σ):  3.92% ±  0.38%

                        ▁▆▃▃▄█▃▃▃█▃▃▂▅   ▁                     
  ▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▃▃▃▄▇██████████████████▇█▆▆▅▆▄▄▄▄▃▃▃▃▃▂▂▂▂▂▂ ▄
  160 ns          Histogram: frequency by time        1.49 μs <

```

it helps, but not that much as expected

Datseris:
It's almost a 2x performance increase, isn't this huge? I would say so.

Tortar:
> I've been thinking a lot about the performance of this, and in general multi-agent models. What if, for all multi agent models we use metaprogramming to create a single special agent that has all fields that all agents have? And this is used everywhere

In general this is bad for memory usage and doesn't allow multiple dispatch on the agent type, so I think this should be a optional feature: I discovered some time ago this package: https://github.com/MasonProtter/SumTypes.jl, which could be the solution. This doesn't require specifying default values which is really cool. But still I don't totally understand how the pachage works :D

Tortar:
> It's almost a 2x performance increase, isn't this huge? I would say so.

in my gergo it is a cut of 40% of the time :D

Tortar:
Actually I also understood now that it's harder to replicate the current behaviour with this model type:

how do you implement this?

```julia
movement_event = AgentEvent(
    action! = move!, propensity = movement_propensity,
    types = Union{Scissors, Paper}, timing = movement_time
)
```

removing the `types` here to make a fair benchmark gets

```julia
# 2 type
julia> @time step!(model, 10000.0)
  0.258423 seconds (2.78 M allocations: 166.209 MiB, 12.64% gc time)
# 1 type
julia> @time step!(model, 10000.0)
  0.182622 seconds (2.42 M allocations: 100.065 MiB, 12.93% gc time)
```

Datseris:
Sure. Let's finish this and merge it in. After all, my plan is to add a clear information in the docs that the event queue is experimental and may change in the future without triggering a new major version. We need users to test it I feel.

Tortar:
I don't have the time these days to finish it, if someone has please go on!

Tortar:
ok, it seems to me that it is hard to make everything work all at once, so I marked this new model type as experimental both in the docstring and when creating an instance of it, I also took care to list some limitations of the current implementation. I think that we should open an issue listing which functions are still not usable with an `EventQueueABM` and just merge this pr in the current form, because I think this is already quite usable anyway, even if it lacks some of the good features. We can then implement them one by one. Besides, we have already some important visualization changes happening in #934 so we should wait for the changes in there to make the visualization functions compatible with `EventQueueABM`. What do you think?

Datseris (commented):
I can't approve my own PR, even if you have put in more work in it... So you can approve and merge instead. 

Yes I agree, we can merge and finalize before releasing v6, after the visualizations PR is finished. Sorry, I am overwhelemd with both work and personal matters so i've been away.

Datseris (commented):
I can't approve my own PR, even if you have put in more work in it... So you can approve and merge instead. 

Yes I agree, we can merge and finalize before releasing v6, after the visualizations PR is finished. Sorry, I am overwhelemd with both work and personal matters so i've been away.

Tortar (approved):
don't worry, loved to work on it (even if it was harder than expected :Q)

Tortar:
I will open an issue listing functions we still need to implement for this model type soon

PR #942, opened 2023-12-10
Implement abmtime for StandardABM

### LINKED ISSUE #941: Create abmtime function for StandardABM
This is practically costless, and actually somewhat needed: run! and offline_run! restart from 0 in the step column each time they are called. Apart from that, this is a function some people would implement by themselves anyway to track the time of the ABM, and implementing it would make StandardABM more homogeneous with the new EventQueueABM which has one.

This means that this is a breaking change though if we change those two functions to instead use the time of the model

cc @Datseris

Tortar:
Fix #941 

This just changes the behaviour of the step data collection which now uses the correct time of the model


Datseris (approved):
Thanks a lot for your effort, this looks great to me!

Datseris:
Ah, I now realize that we don't mention this function in `AgentBasedModel` which describes the accessor API to ABMs.

Tortar:
Also, I actually noticed later that I slightly broke the `abmexploration`, when resetting the model, it returns to 0 step in data collection, which is correct, but doesn't play well with how data are plotted currently, will do a new PR fixing both of this issues eventually

PR #943, opened 2023-12-11
Optimize sample!

Tortar:
This is possible, and even much faster than when using a Dict (wrong)


Tortar:
Actually I don't really like that the same id will be given to different agents when using `sample!` with a vector

Tortar:
As you can see, the speed-up is very good: 

```
# this pr:

@btime sample!(model, 10^6) setup=(model = fake_model(Xoshiro(42), 10^7)) evals=1 samples=10 seconds=10
276.573 ms (1000010 allocations: 114.44 MiB)

# before:

@btime sample!(model, 10^6) setup=(model = fake_model(Xoshiro(42), 10^7)) evals=1 samples=10 seconds=10
  1.947 s (48424 allocations: 168.59 MiB)
```

(I changed the scope of this pr to be just an optimization of the current function)

Tortar:
Actually I used a model with no space to benchmark, the new method falls behind when there is a space instead. Need to change the code to acknowledge this

src/core/space_interaction_API.jl
```diff
@@ -149,8 +149,9 @@ Remove all the agents of the model.
 """
 function remove_all!(model::ABM)
     for a in allagents(model)
-        remove_agent!(a, model)
+        remove_agent_from_space!(a, model)
     end
+    empty!(agent_container(model))
```
Datseris, line 154: I don't like this change because `empty!(agent_container())` isn't the API way to remove an agent. The API way to remove agents is with `remove_agent!` which calls the from model and from space methods. Later down the line (multi-agent container) this `empty!` call may not work.

This optimization is not necessary. `remove_all` is not only called incredibly unfrequently, but its performance weight is miniscule compared to eg a single `step!` call.

Tortar: yes, clearly not used that much but I would use it in this new version of sample!. But I agree it is not right in the current form, but we can easily adapt it: we creare a `remove_all_from_model` with the `empty!` call for a `StandardABM`, and the fallback of looping over agents in other cases

Datseris: I still would argue the increase in code complexity is simply not worth it, because in Agents.jl you want a minimal API more than you want a performant API. This optimization isn't particularly impactful. I would strongly favor keeping the code simpler rather than micro optimizing this part.

Unless you can convince me of course that his optimization is really impactful. How much xx performance gain we get with using `empty!`?

Tortar: Benchmark:

```julia
using Agents, BenchmarkTools, Random

model_step!(model) = nothing

function fake_model(rng, nagents)
    model = StandardABM(NoSpaceAgent; model_step!, rng)
    for i in 1:nagents
        add_agent!(model)
    end
    model
end
```

with the optimization:

```julia
julia> @benchmark sample!(model, 10^6) setup=(model = fake_model(Xoshiro(42), 10^7)) evals=1 samples=10 seconds=10
BenchmarkTools.Trial: 5 samples with 1 evaluation.
 Range (min … max):  244.839 ms … 615.717 ms  ┊ GC (min … max):  0.00% … 50.54%
 Time  (median):     305.240 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   348.143 ms ± 151.834 ms  ┊ GC (mean ± σ):  17.88% ± 22.60%

  █  █     ██                                                 █  
  █▁▁█▁▁▁▁▁██▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  245 ms           Histogram: frequency by time          616 ms <

 Memory estimate: 114.44 MiB, allocs estimate: 1000010.
```


without:

```julia
julia> @benchmark sample!(model, 10^6) setup=(model = fake_model(Xoshiro(42), 10^7)) evals=1 samples=10 seconds=10
BenchmarkTools.Trial: 4 samples with 1 evaluation.
 Range (min … max):  1.024 s …    1.655 s  ┊ GC (min … max):  0.00% … 37.55%
 Time  (median):     1.209 s               ┊ GC (median):    12.97%
 Time  (mean ± σ):   1.274 s ± 292.504 ms  ┊ GC (mean ± σ):  18.34% ± 18.49%

  █  █                          █                          █  
  █▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  1.02 s         Histogram: frequency by time         1.66 s <

 Memory estimate: 114.44 MiB, allocs estimate: 1000010.
```

So it is significant, I would also argue that it's not particularly complex code, just more code :D

Datseris: Didn't you say that the new version is actually _slower_ for models with space? What's the slowdown there?

Tortar: > Didn't you say that the new version is actually _slower_ for models with space? What's the slowdown there?

yes if I remembered correctly it is like 2x slower in the worst case, in others it is already faster. But not worth it in this case until we don't have a faster removal of all agents.

Tortar:
should be okay now, this is an example benchmark:

```julia
using Agents, BenchmarkTools, Random

# no space model

model_step!(model) = nothing

function fake_model(rng, nagents)
    model = StandardABM(NoSpaceAgent; model_step!, rng)
    for i in 1:nagents
        add_agent!(model)
    end
    model
end

@btime sample!(model, 10^5) setup=(model = fake_model(Xoshiro(42), 10^6)) evals=1

# space model

function fake_model(rng, nagents)
    model = StandardABM(GridAgent{2}, GridSpace((100, 100)); model_step!, rng)
    for i in 1:nagents
        add_agent!(model)
    end
    model
end

@btime sample!(model, 10^5) setup=(model = fake_model(Xoshiro(42), 10^6)) evals=1
```

where we have that

```julia
before pr, no space: 87.246 ms (4769 allocations: 16.87 MiB)
after pr, no space: 17.157 ms (100010 allocations: 11.45 MiB)

before pr, no space: 162.883 ms (4789 allocations: 16.95 MiB) 
after pr, no space: 22.714 ms (100010 allocations: 12.98 MiB)
```

In general, the perf improvement can vary, but it should be always an improvement except for edge cases like 10 agent in a 1 million positions space.

Datseris (approved):
ooooookay ooookay. i cry a bit from adding all these new methods but the speedup is significant!!!

PR #946, opened 2023-12-29
Recreate add_agent!(agent,...) + new constructor for agents

### LINKED ISSUE #945: Reconsider the deprecation of adding agents with an agent instance
This post https://discourse.julialang.org/t/swap-agent-positions-in-agents-jl/107774/2 made me realize that we don't want to deprecate the adding of agents by passing an instance to `add_agent!`, if anything, we should instead deprecate the `add_agent!` function which passes `args` and `kwargs`.

The user in that thread wanted to remove the agent and then re-add it after some changes, the problem is that as of 6.0 this can't be done, since `add_agent!` requires passing all `args` except the id, this causes a much further complicated syntax, but not only that, it can't actually be done since we call `nextid` inside `add_agent!` effectively changing the agent. In general the first reference of the agent is lost (the one before `add_agent!`) and so it is a struggle to do this sort of things...

I think all of this problem can be reduced to the fact that `add_agent!` does more that what it should, it doesn't just add the agent to the model, but also creates it, which causes problems.

But there is a better solution without changing the fact that we handle the `nextid` field, we can just exploit the fact that we are using a macro to create the agent, adding an inner constructor like so:

```julia
@agent Foo(NoSpaceAgent)
    x::Int
end
```

creates

```julia
@kwdef mutable struct Foo <: AbstractAgent
    const id::Int
    x::Int
end

# methods added automatically
Foo(model::ABM, args...) = Foo(Agents.nextid(model), args...)
Foo(model::ABM; kwargs...) = Foo(; id = Agents.nextid(model), kwargs...)
```

Those definitions already work:

```julia
julia> model = StandardABM(Foo)

julia> Foo(model, 2)
Foo(1, 2)

julia> Foo(model; x = 2)
Foo(1, 2)
```

Notice that we need to pass the model as the first argument otherwise we have problems with dispatch, but I think it's okay to do so.

With this, I think that instead of deprecating `add_agent!(agent, model)` and `add_agent_pos!(agent, model)` we could deprecate the other versions of `add_agent! `.

cc @Datseris

Tortar:
Closes #945


Tortar:
ok, I realized we only need to reabilitate add_agent_pos!, the other usages are a bit superfluous with the other add_agent! methods in place. I also re-added add_agent_single! because it is a bit more useful as a function. 

In general, add_agent_pos! is an unfortunate name nonetheless because it could well be just add_agent!, which would be better for example with a nothing space, and it seems the correct name in any case. Unfortunately we used this version for adding the agent to a random position in the past, and so changing it to instead add it at the own position of the agent is a breaking change. What I'm thinking is that we can send to the direct users of add_agent_pos! a future warning saying that this function will be renamed add_agent! (maybe in 6.2?). Or keep it like this, even if it doesn't seem the right name. What do you think?

I also added the constructors accepting as the first argument the model instance so that it creates agents the way it should. I will document it in the agent macro docstring.

Datseris:
> In general, `add_agent_pos!` is an unfortunate name nonetheless because it could well be just `add_agent!`, which would be better for example with a `nothing` space, and it seems the correct name in any case. Unfortunately we used this version for adding the agent to a random position in the past, and so changing it to instead add it at the own position of the agent is a breaking change. What I'm thinking is that we can send to the direct users of `add_agent_pos!` a future warning saying that this function will be renamed `add_agent!` (maybe in 6.2?). Or keep it like this, even if it doesn't seem the right name. What do you think?

Ι agree with you. This is an unfortunate naming. My worry however is that this function is so fundamental, and so integrated into the package and all its usage, that changing would be just too much. 

What I think is the best solution is to find a better name alltogether for the function. How about `add_agent_own_pos`? This signifies that the agent is added to its own position. 

I believe that on average it is more common that the user wants agents added to random positions rather than a specified position hence I would argue that the less verbose function `add_agent` should stay as is and do the random position unless specified otherwise.

What do you think of that? 

This means that `add_agent_pos` is deprecated and throws a warning for the new function name.

Tortar:
Thought about this a bit, and I agree with you that `add_agent_own_pos!` is a little bit better so I guess it's okay to have both `add_agent!(agent, [, pos], model)`  and `add_agent_own_pos!(agent, model)`. Will update this PR in this direction.

Datseris:
@Tortar  what is the status here? is this PR ready for review?

Tortar:
Need still to finish this, I'll do it today

Tortar:
Should be okay now!

Tortar:
ok very good, same tests failing as in #875, locally everything works as in the other PR so we can merge both, but something is strange on Github

src/core/space_interaction_API.jl
```diff
@@ -14,6 +14,7 @@ fields should be supplied. See the top of src/core/agents.jl for examples.
 export move_agent!,
     add_agent!,
     add_agent_pos!,
+    add_agent_own_pos!,
```
Datseris, line 17: If we have the `own` version, the `pos` version without `own` should be moved to the deprecation file.

src/core/space_interaction_API.jl
```diff
@@ -188,17 +189,37 @@ end
...
 """
-function add_agent_pos!(agent::AbstractAgent, model::ABM)
+function add_agent_own_pos!(agent::AbstractAgent, model::ABM)
     add_agent_to_model!(agent, model)
     add_agent_to_space!(agent, model)
     return agent
 end
 
+"""
+    add_agent!(agent::AbstractAgent [, pos], model::ABM) → agent
```
Datseris, line 205:
```suggestion
    add_agent!(agent::AbstractAgent [, pos], model::ABM) → agent
    
```

src/core/space_interaction_API.jl
Datseris: (I am referring to the export statement above the new name)

Datseris (approved):
Yes. Just make sure you add the changelog entry for the deprecation and then you can merge. (also couple of minor comments)

PR #955, opened 2024-01-14
Use development version of AgentsExampleZoo

Tortar:


Datseris:
I am a bit confused. Why do we need to use this package at all in the documentation? Where is it used? and why?

Tortar:
Because we need it for the models used in the `agents_visualizations` example. Apart from that, we use it in the tests

Datseris:
> Apart from that, we use it in the tests

This is definitely a mistake in my opinion. There is no reason for tests to be using models defined elsewhere. The tests should be as self-contained and as transparent as possible. Which model(s) do we use in the tests, and for what kind of tests?

For the visualizations page, it can also be debated. We use the daisyworld right? We could have a _delayed_ code snippet at the very end of the file that users can see if they need the definition.

image


I am arguing for this because in DynamicalSystems.jl I was using predefined models a lot, and after having years of experience on this i decided that the best course of action is to _not_ use premade examples in the docs. But I guess in Agents.jl the predefiend models are much more compelx and extensive, requiring 100s of lines of code instead of the typical 10 in DynamicalSystems.jl. So what I say can be debated as well.

Datseris:
another argument to not use this package in the test suite is because circular dependencies become a pain in the butt when there are major version increments. although, i hope that this will be the last major increment of agents.jl version!

Tortar:
> But I guess in Agents.jl the predefiend models are much more compelx and extensive, requiring 100s of lines of code instead of the typical 10 in DynamicalSystems.jl. So what I say can be debated as well.

this is the reason I would like to use predefined models, in general this would also mean less code to maintain because the code is just defined in `AgentsExampleZoo`. I think this is good for code redundancy.

And also for the tests is good to do so in my opinion because then we are forced to keep everything up to date (at least for those models) because we use them in both docs and tests, I see this as an advantage.

But I think that in any case this pr is really necessary because without this `Agents` tests will be broken when `AgentsExampleZoo` is released.

Datseris (approved):

Tortar:
> i hope that this will be the last major increment of agents.jl version!

I tend to like to break the code (with deprecation warnings) if some good new features require it :D

Tortar:
just a thought, I guess in a sense, we should just ensure a more slow path in the future, giving deprecation warning for some long time before removing the features

Datseris:
yes

PR #956, opened 2024-01-14
Handle compatibility with dev version of AgentsExampleZoo better

Tortar:
Since #875 has some problems I do this separately, see https://github.com/JuliaDynamics/Agents.jl/pull/875#issuecomment-1891054174 for why we need this

edit: actually #875 is fine because also #946 has the same errors, but both work locally. However, let's merge this separately anyway since it is more sensible anyway

### DIFF
--- a/examples/agents_visualizations.jl
+++ b/examples/agents_visualizations.jl
@@ -165,7 +165,7 @@ abmvideo("daisyworld.mp4", model; title = "Daisy World", frames = 150, plotkwarg
 # not familiar yet.
 
 # create a basic abmplot with controls and sliders
-model = daisyworld(; solar_luminosity = 1.0, solar_change = 0.0, scenario = :change)
+model = AgentsExampleZoo.daisyworld(; solar_luminosity = 1.0, solar_change = 0.0, scenario = :change)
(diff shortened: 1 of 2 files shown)


Tortar:
Strangely enough this works instead, github is crazy :S

Datseris:
The existence of this PR highlights why one should avoid circular dependencies in packages.

Tortar:
I think that it actually higlights this when both packages depends on *unreleased versions*. We are in this strange situation because of this. When both packages will be released, there will be no problem like this anymore, it will be as if the models where inside Agents.jl. We will just pin the major version on both packages to ensure everything works. 

I think the only problem which could arise is that if we make a major version of Agents without updating AgentsExampleZoo he will need a downgrade of Agents. But this will never happen if we don't completely forget about AgentsExampleZoo :D

Datseris:
> I think that it actually higlights this when both packages depends on _unreleased versions_.

The same problem will also always happen for every major version change in either package.

Datseris:
In any case, let's just get this in for now and we can consider more how to deal with it before releasing v6.

Datseris (approved):

Issue #963, opened 2024-02-03
Test suite sometimes does not pass in PRs


Datseris:
Very crucial to fix before any release and ensure that tests pass!

Tortar:
I would say this is not our fault, so I think we can remove it from v6 milestone

Datseris:
But can we at least understand why it is happening? For example, are the error messages consistent? The failure always comes with the same error message?

Tortar:
they seem somewhat consistent, but the data collection one is not always
failing, while the parallel ones always come togethere if my remembering is
correct. Sometimes all tests pass though in PR, and it is not clear what
triggers those failures in the first place

Il Sab 24 Feb 2024, 22:25 George Datseris ***@***.***> ha
scritto:

Tortar:
This is no longer happening

PR #969, opened 2024-02-04
Overhaul tutorial and `@docs` references and strings for core types

Datseris:
Would you believe that this PR took me more than 8 hours?!?!?!

This PR overhauls the introduction to Agents.jl for new users, or returning veterans alike. It does:

- Completely re-write the tutorial from scratch. Now the main tutorial and the schelling model are merged into one. I took particular care to introduce the new v6 Agents.jl by stressing the existence of different models and the `@multiagent` macro.
- Adds a short copy-pasteable example in the tutorial, inspired by ModelingToolkit.jl docs that also starts with copy pasteable stuff.
- Improves the docstrings of core types for the models as well as the agent constructors.
- Changes where the docstrings are. Now they are all expanded in the API page.
- [TODO:] makes sure the docbuild occurs and docs are pushed to gh-pages!


Tortar:
Great improvements. I will take the time to review it soon.

> Would you believe that this PR took me more than 8 hours?!?!?!

I quite believe it. I would have estimated the same time if I would have been done myself, but with a worse result probably :D

docs/src/tutorial.jl
```diff
@@ -0,0 +1,815 @@
...
+
+# however, only one of these names is an actual Julia type:
+
+fieldnames(MultiSchelling)
+
+# that contains all fields of all subtypes without duplication, while
+
+fieldnames(Civilian)
+
+# doesn't have any fields. Instead,
```
Tortar, line 737:
```suggestion
# doesn't have any field. Instead,
```

docs/src/tutorial.jl
```diff
@@ -0,0 +1,815 @@
...
+# you should think of `Civilian, Governor` as just convenience functions that have been
```
Tortar, line 738:
```suggestion
# you should think of `Civilian` and `Governor` as just convenience functions that have been
```

Tortar:
I think that this proves that #963 can be closed. Also, I think it is a good time to make sure the docbuild happens so that we can be sure that everything works properly, hope you find it easy enough to fix it

Datseris:
I will do my best! I anyways spend time the last weeks fixing docs in many other JuliaDynamics packages so I am sure I can get it to work

Tortar:
Great the new theme is cool but it seems the new zoom is a bit off:

Screenshot from 2024-02-11 18-12-00


(I checked this happens also with your new DynamicalSystems docs), does it happen to you too? Or the zoom is right on your browser? I'm using Firefox at the moment, maybe it is an issue in documenter though

---

Anyway I think you need to merge main into the branch to fix the test failing

Tortar (approved):

Tortar:
mmh I see you load some css, so I think the zoom problem should be there maybe

Datseris:
yeah with documenter v1 some things got broken, but i don't know how to fix them yet. I have asked online.

Datseris:
(we will of course try to fix everything before a final release, but it may be that some things will remain "somewhat broken")

PR #975, opened 2024-02-13
Fix some docs and deps issues

Tortar:


Tortar:
Fixed some docs issues, should be okay now. The test failing is non-sense and doesn't happen locally :(

Tortar:
ohh maybe I'm using a old version of Aqua, but in any case this fail is unrelated to these changes, it happened also on the last PR

Datseris (approved):

PR #976, opened 2024-02-17
Improve `Schedulers` docs

Datseris:
- All docstrings have been made "correct": they refer to ordering, and not scheduling, and they refer to the IDs, not the agents
- Deprecated using `schedule` with the non-default scheduler, as all schedulers are anyways functions of the model, there is no need to pipe them through `schedule`.
- Highlighted another usage for the schedulers
- Remove wording "once per step" from schedulers: this is clear from the `StandardABM` docstring. Furthermore, how many times scheduling happens in custom scheduling is unknown.

### DIFF
--- a/src/core/model_standard.jl
+++ b/src/core/model_standard.jl
@@ -100,6 +100,9 @@ functions during a single simulation step.
 In such a scenario, it is more sensible to provide only a model stepping function,
 where all the dynamics is contained within.
 
+Note that if you do not use the automated `agent_step!` option, you need to manually
+check for removed agents during evolution, using the [`hasid`](@ref) function.
+
 Here is an example:
 
 ```julia
@@ -112,6 +115,8 @@ function complex_model_step!(model)
     end
     intermediate_model_action!(model)
     for id in scheduler2(model)
+        # here `_step2!` may delete agents, so we check for it manually
+        hasid(model, id) || continue
         agent_step2!(model[id], model)
     end
     if model.step_counter % 100 == 0
--- a/src/core/model_standard.jl
+++ b/src/core/model_standard.jl
@@ -107,11 +107,11 @@ function complex_model_step!(model)
     # tip: these schedulers should be defined as properties of the model
     scheduler1 = Schedulers.Randomly()
     scheduler2 = user_defined_function_with_model_as_input
-    for id in schedule(model, scheduler1)
+    for id in scheduler1(model)
         agent_step1!(model[id], model)
     end
     intermediate_model_action!(model)
-    for id in schedule(model, scheduler2)
+    for id in scheduler2(model)
         agent_step2!(model[id], model)
     end
     if model.step_counter % 100 == 0
--- a/src/core/model_standard.jl
+++ b/src/core/model_standard.jl
@@ -115,7 +115,7 @@ function complex_model_step!(model)
     end
     intermediate_model_action!(model)
     for id in scheduler2(model)
-        # here `_step2!` may delete agents, so we check for it manually
+        # here `agent_step2!` may delete agents, so we check for it manually
         hasid(model, id) || continue
         agent_step2!(model[id], model)
     end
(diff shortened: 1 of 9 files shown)


Tortar:
I don't think that removing `schedule(model[, scheduler])` is a good idea, how do we manage to filter already removed agents from a custom scheduler? This was the motivation for this feature

src/core/model_standard.jl
```diff
@@ -107,11 +107,11 @@ function complex_model_step!(model)
     # tip: these schedulers should be defined as properties of the model
     scheduler1 = Schedulers.Randomly()
     scheduler2 = user_defined_function_with_model_as_input
-    for id in schedule(model, scheduler1)
+    for id in scheduler1(model)
         agent_step1!(model[id], model)
     end
     intermediate_model_action!(model)
-    for id in schedule(model, scheduler2)
+    for id in scheduler2(model)
```
Tortar, line 114: e.g. here now this safeguard is not in-place

Tortar, line 114: e.g. here now this safeguard is not in-place

Datseris:
ooooooh shit yeah I have completely forgotten about this....

I will revert everything related to removing `schedule`.

Datseris:
@Tortar the way we did it before was suboptimal. Now we check inside `step!` if an agent exists. it is more efficient than wrapping again the scheduled ids in another Iterators.filter. And now, `schedule(model, scheduler)` is indeed deprecated.

Datseris:
Alright so I introduced a new function `hasid` to make this process simpler.

Datseris:
tests passs locally

src/core/model_standard.jl
```diff
@@ -100,18 +100,23 @@ functions during a single simulation step.
...
+        # here `_step2!` may delete agents, so we check for it manually
```
Tortar, line 118:
```suggestion
        # here `agent_step2!` may delete agents, so we check for it manually
```

src/submodules/schedulers.jl
```diff
@@ -61,15 +59,16 @@ end
 
 """
     Schedulers.fastest
-A scheduler that activates all agents once per step in the order dictated by the
-agent's container, which is arbitrary (the keys sequence of a dictionary).
-This is the fastest way to activate all agents once per step.
+
+A scheduler that orders all agent IDs in the fastest way possible,
+which is the default order dictaed by the agent container.
```
Tortar, line 64:
```suggestion
which is the default order dictated by the agent container.
```

src/simulations/step.jl
```diff
@@ -37,6 +37,8 @@ function step_ahead!(model::ABM, agent_step!, model_step!, n, t)
     while until(t[], t0, n, model)
         !agents_first && model_step!(model)
         for id in schedule(model)
+            # ensure we don't act on agent that doesn't exist
+            hasid(model, id) || continue
```
Tortar, line 41: Note that this condition can be spared in the case of an ABM with a `Vector` container but we can also improve this in a separate PR

src/submodules/schedulers.jl
```diff
@@ -1,17 +1,11 @@
...
+    schedule(model::StandardABM) → ids
 
-If no `scheduler` is given, it returns an iterator over the scheduled IDs using the model's
-scheduler, otherwise it uses the given custom scheduler, which can be either a function which
-accepts `model` as argument or one of the already defined schedulers inside Agents.jl. See
-the [manual scheduling](@ref manual_scheduling) section for usage examples.
+Return an iterator over the scheduled IDs using the model's default scheduler.
 """
-schedule(model::ABM) = schedule(model, abmscheduler(model))
-schedule(model::ABM, scheduler) = Iterators.filter(id -> id in allids(model), scheduler(model))
```
Tortar, line 12: We would need to keep this line instead of `schedule(model::ABM, scheduler) = scheduler(model)` to deprecate this properly (also passing a deprecation warning probably)

Tortar:
Seems good anyway to me apart from those details

Tortar (approved):

PR #978, opened 2024-02-18
Faster sampling from small iterators

Tortar:
`:alg_R` is like 5x faster for small iterators than `:alg_L`, but slower with larger ones. So I applied one or the other depending on the probable scenario


Datseris (approved):
I have no idea what each does but also no time to look at the main repo so i trust you!

PR #979, opened 2024-02-18
Use ABM in step! definition

Tortar:
We need to still use `ABM` for user-defined `AgentBasedModel`s which have this kind of `step!` by default

### DIFF
--- a/src/simulations/step.jl
+++ b/src/simulations/step.jl
@@ -23,7 +23,7 @@ for, starting from the model's initial time.
 
 See also [Advanced stepping](@ref).
 """
-function CommonSolve.step!(model::StandardABM, n::Union{Real, Function} = 1)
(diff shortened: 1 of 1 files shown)


Datseris (changes requested):
I disagree. The dispatch on `ABM` should throw a MethodError or a custom "not implemented" error message. If a user makes a truly new ABM, they must extend `step!`.

If the user ABM has the same stepping nature as our `StandardABM`, then there is no reason for the user to define a new ABM in the first place. The only purpose for new ABMs is to change `step!`. I will describe this better in the "Developers Docs" new entry on new models that I will write soon.

Tortar:
you are very right, but as you say, we haven't ever highlighted that (and put in `model_abstract`). I will just then keep the other change of this PR about skipping the check on agent removal

Datseris (approved):

PR #984, opened 2024-02-23
Fix add_agent! with all default arguments

### LINKED ISSUE #983: "MethodError: no method matching..." with default agent attribute values.
Following the new 6.0.0 macro for defining agents, I'd like to be able to add agents with default values for all additional attributes. However, I get a `no matching method` error unless I supply at least one keyword argument for additional attributes, or positional arguments for all.  If I supply one karg, defaults are applied for other attributes.

Is this expected behavior?

Minimum working example mostly copied and pasted from source comments:

```
using Agents
using Graphs
@agent struct Agent(GraphAgent)
    w::Float64 = 0.1
    k::Bool = false
end
model = StandardABM(Agent, GraphSpace(complete_digraph(5)))

add_agent!(model; w = 0.5, k = true)
add_agent!(model; w = 0.5)
add_agent!(model; k = true)
add_agent!(model)
```

See my results:

image

Tortar:
Fixes #983

### DIFF
--- a/src/core/agents.jl
+++ b/src/core/agents.jl
@@ -33,7 +33,7 @@ The minimal agent struct for usage with `nothing` as space (i.e., no space).
 It has the field `id::Int`, and potentially other internal fields that
 are not documented as part of the public API. See also [`@agent`](@ref).
 """
-mutable struct NoSpaceAgent <: AbstractAgent
+@kwdef mutable struct NoSpaceAgent <: AbstractAgent
     const id::Int
 end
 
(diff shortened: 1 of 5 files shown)


Tortar:
Here we are again, tests are failing but with no real reason

Datseris:
Is this really "without real reason"? 

I see we get :
```
   UndefVarError: `#368#369` not defined


```
that is related with data collection.

Tortar:
yes, it is, I don't have an explaination for the failures, but they are always the same

Tortar:
just tested locally now, everything passed

Tortar:
I will report an issue to Github itself linking to this PR as soon as we merge this to let they see that tests pass when the PR is merged

Datseris (approved):

PR #985, opened 2024-02-24
Change agent in subagent inside multiagent

Tortar:
Locally tested, everything works


Datseris:
this is only a rename right?

Tortar:
yes exactly

Datseris (approved):

PR #990, opened 2024-02-27
Warn only once

Tortar:


Tortar:
Actually this still have some problems

Datseris:
Why warn only once? I would argue we should warn all the time to make sure people get it and update their code. If they don't want to update their code they shouldn't use agetns.jl v6.

Tortar:
This warns once in a session, but in all sessions it will restate the warning, it is a bit too annoying that you are flood with the same warning hundreds of times

Datseris (approved):

Datseris:
fine. i prefer to flood people and make them update the code but that is not a popular approach.

PR #991, opened 2024-02-29
Add precompilation workload for macros

Tortar:
This cuts down 2-3x for agent macro instantiation (e.g. from 0.2 s to 0.08 s), and 5x for multiagent (e.g. from 1 s to 0.2 s) e.g. now

```julia
julia> @time @eval @multiagent struct :opt_memory Animal{T,N,J}(GridAgent{2})
           @subagent struct Wolf{T,N}
               energy::T = 0.5
               ground_speed::N
               const fur_color::Symbol
           end
           @subagent struct Hawk{T,N,J}
               energy::T = 0.1
               ground_speed::N
               flight_speed::J
           end
       end
  0.143237 seconds (62.43 k allocations: 4.143 MiB, 25.01% gc time, 79.19% compilation time)
```

before 

```julia
julia> @time @eval @multiagent :opt_memory struct Animal{T,N,J}(GridAgent{2})
           @subagent struct Wolf{T,N}
               energy::T = 0.5
               ground_speed::N
               const fur_color::Symbol
           end
           @subagent struct Hawk{T,N,J}
               energy::T = 0.1
               ground_speed::N
               flight_speed::J
           end
       end
  1.064352 seconds (974.18 k allocations: 64.895 MiB, 2.83% gc time, 93.72% compilation time: 67% of which was recompilation)
```


Datseris (approved):
Fantastic idea. 

I hope in the future we can do more with precompile tools. It is not clear how much it can really help us because the models reference the functions, and so precompiling a model with a function doesn't help with other models with different stepping functions. (I guess?)

Issue #992, opened 2024-02-29
[DOCS] add multithreading discussion in performance tips


Datseris:
The points from this post as useful to have in the docs:

https://discourse.julialang.org/t/multi-threads-error-undefreferror-in-simple-abm-with-agents-jl/110900/19

mainly the locking of dictionaries, or the usage of a `:dead` field for later mass removal, or the usage of a thread safe dictionary for agent container.

(this means that we need to change the agent container types in the source code from `Dict` to `AbstrcactDict` and from `Vector` to `AbstractVector`)

Tortar:
In any case cool ideas !

Tortar:
I think that https://discourse.julialang.org/t/multi-threads-error-undefreferror-in-simple-abm-with-agents-jl/110900/21?u=tortar proves it!

PR #993, opened 2024-02-29
Enlarge DictABM and VecABM for dispatch

Tortar:
Just realized it is needed for #992 thread safe dict


Tortar:
lol I saw that you wrote

> (this means that we need to change the agent container types in the source code from Dict to AbstrcactDict and from Vector to AbstractVector)

but I jumped straight to the link and only read it now, I realized the same myself xD

Datseris (approved):

Datseris:
test did not pass but i kinda assumed this was teh "known issue". I am sorry if I was wrong!!!

Tortar:
yeah :D just did a PR now, testing it locally at the same time, I will let you know if it works

PR #994, opened 2024-02-29
Fix bug due to last commit

Tortar:


Tortar:
This passes tests locally, so it is safe to assume it does also here, if you want we can wait to see if this is the case for extra-safety @Datseris

Tortar:
Not all tests actually, still need some changes, I assumed it was very easy

Tortar:
Ok now everything really works locally

Tortar:
now failings are the usual suspects :D

Datseris (approved):

PR #997, opened 2024-03-05
Deprecate `as, am, ac` in favor of  `agent_size, agent_marker, agent_color`

### LINKED ISSUE #964: Deprecate `as, am, ac` keywords for longer, more descriptive names
I think we should deprecate the `as, am, ac` keywords of our main plotting function and call them instead `agent_size, agent_color, agent_marker`. Deprecation is easy for keyword arguments:
```
function f(; as = default, agent_size = as)
if as != default
@warn("as deprecated")
end
end
```

The rest of the keywords for the plotting function are not as cryptic.

Tortar:
Closes #964

### DIFF
--- a/examples/agents_visualizations.jl
+++ b/examples/agents_visualizations.jl
@@ -48,8 +48,8 @@ model
 # for the agents that depend on the agent properties, and
 # a size and marker style that are constants,
 daisycolor(a) = a.breed # agent color
-agent_size = 20    # agent size
-agent_marker = '✿'  # agent marker
+agent_size = 20
+agent_marker = '✿'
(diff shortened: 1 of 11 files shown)


examples/agents_visualizations.jl
```diff
@@ -48,8 +48,8 @@ model
 # for the agents that depend on the agent properties, and
 # a size and marker style that are constants,
 daisycolor(a) = a.breed # agent color
-as = 20    # agent size
-am = '✿'  # agent marker
+agent_size = 20    # agent size
+agent_marker = '✿'  # agent marker
```
Datseris, line 52:
```suggestion
agent_size = 20
agent_marker = '✿'
```

Datseris (approved):

PR #999, opened 2024-03-07
Fix reset button in abmexploration

### LINKED ISSUE #944: Fix reset button working in `abmexploration`
#942 broke that since when the model is reset the time returns to 0 as it should, but this means that the data graphs should take into account this new behavior to function properly

Tortar:
Fixes #944

### DIFF
--- a/examples/agents_visualizations.jl
+++ b/examples/agents_visualizations.jl
@@ -48,10 +48,10 @@ model
 # for the agents that depend on the agent properties, and
 # a size and marker style that are constants,
 daisycolor(a) = a.breed # agent color
-as = 20    # agent size
-am = '✿'  # agent marker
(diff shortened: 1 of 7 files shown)


Datseris (commented):
Can you please describe the PR? What's the purpose of the offset time?

Tortar:
It helps to track the shifting amount for new data points appearing when plotting the agent/model data, otherwise the new points would cover the old ones in the plots

Datseris (changes requested):
Something is wrong with git here. I see all commits of the PR #964 here as well.

Tortar:
Tested the branch at the latest commit locally, everything works, I think github is not working properly as usual, but let's see if everything works after merging

Datseris (approved):

PR #1001, opened 2024-03-10
Update `EventQueueABM` to multi-agent + make plotting and run model agnostic

Datseris:
This PR changes the `EventQueueABM` to use exclusively the `@multiagent` macro and the `kindof` shenanigans. That is because one cannot use multiple dispatch anyways in `EventQueueABM` to decide the event, so there is no obvious benefit in using `Union` type agents.

The PR also re-organizes the codebase.

Additionally, this PR makes the plotting and data collection codes model agnostic. Detaches them **entirely** from the internals of the `AgentBasedModel` type that is passed into the plotting recipe.

IT does the same for data collection and `run!`. 

TODOs:

- [x] Finish Rock Paper Scissors example, make sure it works (without plotting)
- [x] Update plotting (working on it now)
- [x] Update tests
- [x] Ensure all other examples and visualizations work as expected.


Tortar:
Should be available as soon as https://github.com/JuliaRegistries/General/pull/102631 is merged

Tortar:
I think replicate! should be already good as it is, you just need to pass `pos` as a kwarg: `replicate!(agent, model; pos=new_pos)`

Tortar:
Released a new version of MixedStructTypes with the new function `kindconstructor`, which seems to me a good name because it is specific and explicit enough. If you find any instances of MixedStructTypes.constructor inside Agents, you should change that to the new name

Tortar (approved):
Really great improvements!! I approve.

Tortar:
the test failure is no sense in the last commit, I just bumped the Agents.jl version

Tortar:
I try to merge if this resolves it, otherwise we make another PR

PR #1011, opened 2024-03-16
Add RPS example to docs

Tortar:


Tortar:
the new page looks good (https://juliadynamics.github.io/Agents.jl/previews/PR1011/examples/event_rock_paper_scissors/) but doesn't have a title in the left tab

Tortar:
Actually we should probably put this as a second tutorial

Datseris (approved):

PR #1016, opened 2024-03-24
Use Heap instead of PriorityQueue in EventQueueABM

Tortar:
This makes RPS more than 2x faster! in general a `Heap` is more suitable than a `PriorityQueue` for this task, also it actually makes everything better because a `PriorityQueue` denied the use of duplicate times, which is actually something which could happen

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati", "Adriano Meligrana"]
-version = "6.0.5"
+version = "6.0.6"
 
 [deps]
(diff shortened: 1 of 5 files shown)


Datseris:
I am having trouble understanding how this works. How can a binary heap be faster here? Do you have any idea why?

Tortar:
the motivation is that with a priority queue you can change the priority of a given key while here you can't, this is not something needed in our case so that is why it is faster I think

Tortar:
And I think it is a lot faster if the all time is cut in half

Datseris (approved):

Datseris:
fantastic, thanks for the great work. But how did you think of this? of changing the data structurte?

Tortar:
I was actually doing something else, and I needed something like a `PriorityQueue` but it was too slow and would make the algorithm unusable, so I looked for alternatives, then I remembered that in `Agents` we had a `PriorityQueue` and I also remembered it had a bad impact on overall performance :-)

PR #1018, opened 2024-04-04
Specify versions for Integration Libraries in Docs Project.toml

Tortar:
This way we signal (a bit) the version with which the integration surely works for users, because if a major version is made could be problematic if the docs are not updated

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati", "Adriano Meligrana"]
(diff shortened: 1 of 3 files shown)


Datseris:
Right, don't we also have a statement in the docs of the list of versions the docs where build with? We should have a section `## Reproducibility` at the end of the docs that states all versions in the project. Like here: https://docs.sciml.ai/DiffEqDocs/dev/index.html#Reproducibility

if you could copy the code that makes this nice list it would be fantastic :)

Tortar:
very good idea, adding it!

Tortar:
Let's see if it works

Tortar:
it does

Datseris (approved):
Perfect. Please put the reproducibility section _after_ the citation, and go ahead and merge!

PR #1022, opened 2024-04-10
Remove deprecated methods in version 5 rem_node!, add_node!

Tortar:
We forgot to do this during transition to version 6 :( But they were deprecated in v5 so it can be fixed this way


Tortar:
Actually there shouldn't be any problem because they were not part of the API in the docs fortunately

Datseris (approved):

PR #1023, opened 2024-04-10
Fix `lonlat` function

simsurace:
The `lonlat` function was assuming that the unit of `pos[3]` was an absolute distance, which made the points to be drawn outside of the way when the length of the way was smaller than unity.


simsurace:
Ok, after reading the source for `move_along_route!` more closely, I think this fix is wrong. There must be another reason why I'm seeing agents that are drawn outside of the graph. Closing for now.

Tortar:
yes, indeed I was expecting this now, and it seems `pos[3]` is actually an absolute distance :-), if you have a MWE this would help to understand the problem

simsurace:
It was my bad, I was planning a route with `pos[3] == 1` in the mistaken belief that this was the correct way to set a target.
Maybe the `plan_route!` function could do some validation that a correct position is passed?

PR #1024, opened 2024-04-11
Add OSM destination check

simsurace:
It is possible to plan a route to a destination `dest` for which `dest[3]` is not between zero and `road_length(dest[1], dest[2], model)`, which is not caught by the subsequent shortest path algorithm. When such a route is executed, the agent will snap to a position that is not on the graph after completing the route.

With this PR, we check that the destination is valid and if it is not, we exit the `plan_route!` function early and return `false`.

Discovered in #1023.


simsurace:
If this PR is welcome, I will add some unit tests.

Tortar:
thanks! This is surely useful to avoid the inconsistencies you described in #1023, I approve but I will wait till the unit tests being added before merging

Tortar (approved):

Tortar:
The tests are very good, thank you!

PR #1025, opened 2024-04-20
Move to StreamSampling for sampling agents

Tortar:
This is the new name for the package IteratorSampling


Datseris:
Context? please try to have descriptions in the top post of a PR!

Tortar:
yes sorry, I just released a new version of the package which I renamed this way and so I updated the code in Agents so that it uses the latest version

Datseris (changes requested):
`algL` and `algR` are short and cryptic names. You should replace them with `StreamSampling.algR`. You can also consider renaming them to something more appropriate, see e.g., guidlines in chapter 2 in : https://github.com/JuliaDynamics/GoodScientificCodeWorkshop 

To the best of my knowledge, in the Julia community it is recommended to use **type instances** and not type themselves as input arguments to functions to decide behavior. 
Instead of `algR` to pass in `algR()`.

Datseris:
`const algR = AlgR() ` this doesn't save you any effort overall as the words are the same length, but the reader of the source code then has to also find the extra line `const algR = AlgR()`. I missed it on the first read through. So, I'd say, just replace everything with `StreamSampling.AlgR()` etc.

Tortar:
should be okay now

Datseris (approved):
dont forget version bump

PR #1027, opened 2024-04-27
Update README.md

pricklypointer:
Fixed the download count badge using https://juliapkgstats.com/. The current badge is broken because the old API and PkgVisBoard is not maintained.

This PR:
- Replaces the broken badge with a new badge that uses the new API

Please review and provide any feedback. Thanks for considering this PR!


Datseris:
Thanks a lot! Similarly with DrWatson, and in general throughout JuliaDynamics the preferred statistic is the total downloads. Of course, we definitely appreciate the intermediate fix nevertheless!!!!

Datseris (approved):

PR #1028, opened 2024-05-02
Update celllistmap.jl example

lmiq:
This updates the CellListMap integration example to use the renamed `ParticleSystem` syntax, which was introduced in version 0.9.0 and will become the standard in a future 1.0 release. No functionality was changed.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -33,7 +33,7 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 BlackBoxOptim = "0.6"
-CellListMap = "0.8"
+CellListMap = "0.9"
 Documenter = "1"
 Graphs = "1"
 Measurements = "2"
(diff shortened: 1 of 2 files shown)


Datseris:
Thanks, but don't you also have to change the version in docs/project.toml: https://github.com/JuliaDynamics/Agents.jl/blob/main/docs/Project.toml#L36 ? since this will be the syntax in 1.0, maybe change it to `0.9, 1.0`?

lmiq:
> Thanks, but don't you also have to change the version in docs/project.toml: https://github.com/JuliaDynamics/Agents.jl/blob/main/docs/Project.toml#L36 ? since this will be the syntax in 1.0, maybe change it to `0.9, 1.0`?

Yes, thanks. I didn't know about having a Project.toml file that actually was used on deployment. Fixed.

I'll let the 1.0 for when it is available, after some testing of the new interface.

Datseris (approved):

Issue #1026, opened 2024-04-23
SIR video is completely white


KristofferC:
The video in https://juliadynamics.github.io/Agents.jl/dev/examples/sir/ does not show anything. It is just white.

PR #1029, opened 2024-05-06
fix SIR example

rakvat:
This fixes the SIR example for me. 

Might fix https://github.com/JuliaDynamics/Agents.jl/issues/1026

### DIFF
--- a/examples/sir.jl
+++ b/examples/sir.jl
@@ -96,7 +96,6 @@ function model_initiation(;
         detection_time,
         C,
         death_rate,
-        step
     )
     space = GraphSpace(complete_graph(C))
     model = StandardABM(PoorSoul, space; agent_step!, model_step!, properties, rng)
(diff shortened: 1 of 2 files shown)


Tortar:
thanks a lot!! We actually track the step of the simulation through `abmtime` so I update your fix with it :-)

Tortar:
Let's see if on the online docs everything works fine now after it builds

examples/sir.jl
```diff
@@ -94,10 +93,10 @@ function model_initiation(;
         reinfection_probability,
         detection_time,
         C,
-        death_rate
+        death_rate,
     )
     space = GraphSpace(complete_graph(C))
-    model = StandardABM(PoorSoul, space; agent_step!, properties, rng)
+    model = StandardABM(PoorSoul, space; agent_step!, model_step!, properties, rng)
```
rakvat, line 99: can't you remove this change as well now?

Tortar: sure, I assumed there was one before, that's what happens when you are multitasking xD

Tortar (approved):

PR #1031, opened 2024-05-08
Remove simplicity details from ReadMe

Tortar:
I feel like this is not something it should me mentioned in the ReadMe because we don't go into the details of the other bullet points. In the docs this is still explained in the main page anyway


Datseris (changes requested):
These sentences do not exist anywhere else in the docs, I just checked. You need to add them somewhere else. Perhaps in the design philosophy section of `index.md`.

Datseris (approved):

Issue #1014, opened 2024-03-20
Scheduler equivalent to `ByType` but for `@multiagent` models.


Datseris:
see https://discourse.julialang.org/t/agents-jl-v6-releases-announcement-post/111678/6?u=datseris would be nice to have a convenience scheduler that checks the _internal_ property `type` (is it called `type` @Tortar ?)

Datseris:
labelling as QoL as at the moment we can avhieve this with `ByProperty`.

PR #1033, opened 2024-05-08
Implement ByKind scheduler

Tortar:
Fixes https://github.com/JuliaDynamics/Agents.jl/issues/1014

### DIFF
--- a/src/submodules/schedulers.jl
+++ b/src/submodules/schedulers.jl
@@ -215,7 +215,7 @@ end
     Schedulers.ByKind(agent_kinds; shuffle_kinds = true, shuffle_agents = true)
 
 A scheduler useful only for mixed agent models using the [`@multiagent`](@ref) macro.
-- `agent_kinds` is a `Tuple` of all valid agent kinds e.g. (:B, :A, :C).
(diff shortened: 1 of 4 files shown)


Tortar:
Actually it is a needed feature because if you use ByProperty you just can't shuffle the agents

src/submodules/schedulers.jl
```diff
@@ -211,4 +211,63 @@ function (sched::ByType)(model::ABM)
     return Iterators.flatten(it for it in sched.ids)
 end
 
+"""
+    Schedulers.ByKind(agent_kinds; shuffle_kinds = true, shuffle_agents = true)
+
+A scheduler useful only for mixed agent models using the [`@multiagent`](@ref) macro.
+- `agent_kinds` is a `Tuple` of all valid agent kinds e.g. (:B, :A, :C).
```
Datseris, line 218:
```suggestion
- `agent_kinds` is a `Tuple` of all valid agent kinds e.g. `(:B, :A, :C)`.
```

when referring to **code** you must use code highlighting.

Datseris (approved):

src/submodules/schedulers.jl
Tortar: sure, it was just an oversight

PR #1036, opened 2024-05-20
Allow using Makie 0.21 with Agents

Tortar:


Tortar:
Seems like there is a problem with Flocking

Datseris (approved):

PR #1037, opened 2024-05-21
 Replace monthly download count badge with total downloads badge

pricklypointer:
Ref https://github.com/JuliaDynamics/Agents.jl/pull/1027


Tortar:
thanks!

Issue #1043, opened 2024-06-06
Conflict between nearest_neighbor() function and euclidean distance for agent pair.  


sahil-khan11:
**Issue**
The nearest neighbor for a given agent within the radius "r" doesn't align with the Euclidean distance between the agent and its nearest neighbor. The following example shows that the distance between neighbors is quite big compared to r used in the nearest_neighbor() function.  

**Minimal Working Example**
```using Agents, Random

@agent struct SocialAgent(ContinuousAgent{2, Float64})
    mass::Float64
end

function ball_model(; speed = 0.002)
    space2d = ContinuousSpace((1, 1); spacing = 0.02)
    model = StandardABM(SocialAgent, space2d; agent_step!, properties = Dict(:dt => 1.0),
                        rng = MersenneTwister(42))
    # And add some agents to the model
    for ind in 1:500
        pos = Tuple(rand(abmrng(model), 2))
        vel = sincos(2π * rand(abmrng(model))) .* speed
        add_agent!(pos, model, vel, 1.0)
    end
    return model
end

agent_step!(agent, model) = move_agent!(agent, model, model.dt)

function model_step!(model)
    for (a1, a2) in interacting_pairs(model, 0.01, :nearest)
        elastic_collision!(a1, a2, :mass)
    end
end

model = ball_model()

#### Checking euclidean distance for nearest pairs. 
r = 10^-100
euclidean_distance(model[3].pos, nearest_neighbor(model[3], model, r).pos, model)
```
*Output*
```
0.02637710352142322
```
**Agents v6.0.12**
**Julia v1.10.1**

sahil-khan11:
What do you think about this? I was thinking maybe replacing nearest_neighbor() in interaction_pair() with this logic. 

```
r = 0.7
agent = model[27]
partner  = model[27]
partner_dist = 10.0.          # some random big number, just to create a variable so we can replace 
                                           # new partners distance after comparison. 
z = partner_dist                # For return verification 

for id in allids(model)

    if id != agent.id
        x = euclidean_distance(agent.pos, model[id].pos, model)

        if x <= r && x <= partner_dist
            partner_dist = x
            partner = model[id]
        end
    end

    if partner_dist == z
        return nothing
    else
        return (partner, partner_dist)
    end
  
end
```

It doesn't look elegant at the moment but I think it does the work :) 

Some outputs here: 

with r = 1
```
[...]

PR #1044, opened 2024-06-06
Deprecate nearby_ids_exact in favour of nearby_ids with a search keyword

Tortar:
This makes the API cleaner

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -381,7 +381,6 @@ function schedule(model::ABM, scheduler)
     Iterators.filter(id -> id in allids(model), scheduler(model))
 end
 
-
 # Do the standard extensions for `_exact` as in space API
 function nearby_ids_exact(agent::AbstractAgent, model::ABM, r = 1)
     @warn "`nearby_ids_exact` is deprecated in favor of `nearby_ids(...; search=:exact)`." maxlog=1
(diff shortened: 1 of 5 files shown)


Tortar:
This fixes also #1043

Tortar:
Actually there is still something to take care because sometimes we pass `nearby_ids_exact` to some other functions

Tortar:
If tests pass, it should be okay now

src/deprecations.jl
```diff
@@ -379,4 +379,15 @@ until(s, f, model) = !f(model, s)
...
+
+# Do the standard extensions for `_exact` as in space API
+function nearby_ids_exact(agent::AbstractAgent, model::ABM, r = 1)
+    @warn "`nearby_ids_exact` is deprecated in favor of `nearby_ids(...; search=:exact)`." maxlog=1
+    all = nearby_ids_exact(agent.pos, model, r)
+    Iterators.filter(i -> i ≠ agent.id, all)
+end
+function nearby_agents_exact(a, model, r=1)
+    @warn "`nearby_agents_exact` is deprecated in favor of `nearby_agents(...; search=:exact)`." maxlog=1
+    return (model[id] for id in nearby_ids_exact(a, model, r))
```
Datseris, line 392: Shouldn't this source code just call the `nearby_ids(...; search = :exact)` instead of what it does now?

Datseris (approved):
Seems good, but bump minor version (and add changelog) and release new version after merge!

src/deprecations.jl
Tortar: yes, did it

Datseris:
I wonder if we can increase the performance of the `nearest_neighbor` function... We go thourhg all agents in the list.

On the other hand, we do near searches fundamentally differently from e.g., a tree structure, so it's not like we would ever have nearby agents sorted by distance. Anyways, merging and tagging.

Tortar:
I think it can be quite improved by the user, because decreasing the radius implies searching less neighbors, but it is model dependent, one can try little radius before big ones

PR #1046, opened 2024-06-07
Fix docstring of nearest_neighbor

Tortar:


Tortar:
Actually I want also to add a test for the bug which was present in the function, so let's wait a bit when I find the time to add it, but if in the meantime @Datseris you approve it, this would require less back and forth :-)

Datseris (approved):

PR #1050, opened 2024-06-22
Update DynamicSumTypes to 2.0

Tortar:
Still need to release it


Tortar:
The only breaking change here is that it is needed to use `@finalize_dispatch` if one uses `@multiagent` in a package. There is no way to avoid that, the previous methodology was actually worse, see here for more explainations on why: https://discourse.julialang.org/t/is-using-base-delete-method-on-user-defined-functions-safe/115194

src/core/agents.jl
```diff
@@ -442,3 +444,33 @@ For an illustration of its usage see the [Tutorial](@ref).
 macro dispatch(f_def)
     return esc(:(DynamicSumTypes.@pattern $f_def))
 end
+
+"""
+    @finalize_dispatch
```
Datseris, line 449: This method is not referenced anywhere. In the `@dispatch` docstring you need to say: "if you are creating your own module/package that uses Agents.jl, and you are using `@dispatch` inside it, then you need to put 
```
[`@finalize_dispatch`](@ref)`()`
```
before the module end (but after all `@dispatch` calls).

Tortar: yes, this is necessary, did it

Datseris (approved):

Datseris:
Merge and tag? This is done right?

Tortar:
yeah I tag now

PR #1055, opened 2024-07-07
Update to DynamicSumTypes 3

### LINKED ISSUE #1039: Does EventQueueABM require @multiagent?
**Describe the bug**
I'm trying out the new EventQueueABM functionality, and I get an error that my agent wasn't created with the `@multiagent` macro, which is true, but I just have a single agent type with no need for `Union` or `@multiagent`.

**Minimal Working Example**
```
using Agents
@agent struct Foo(GridAgent{2}) end

bar!(agent, model) = println(agent)

function init_foo()
    space = GridSpaceSingle((10, 10))
    events = (
        AgentEvent(action! = bar!, propensity = 0.1),
    )
    model = EventQueueABM(
        Foo,
        events, space)
    add_agent!((5, 5), model)
    model
end
init_foo()
```
ERROR: ArgumentError: Agent of type Foo has not been created via `@multiagent`.

**Agents.jl version**

Agents v6.0.13 `https://github.com/JuliaDynamics/Agents.jl.git#8c36b50`
Julia 1.10.3

Tortar:
DynamicSumTypes v3 makes it really really easy to work with sum types directly, with much more flexibility than before and a much cleaner approach, we need to deprecate `@multiagent` because it was backed by the previous approach. This time I think we can go instead by integration, listing DynamicSumTypes in the "Performance Tips" of the manual and giving an example because we actually need almost no work here for that.

This will also fix #1039


Datseris:
> we need to deprecate `@multiagent`

What?! You need to provide **much** more information for such a huge change. This is one of the big new things in v6 and you want to deprecate it in 6.1. This is not good experience for a user.

Datseris:
I don't understand this PR at all. `@agent` macro is completely unaffected and this PR replaces everything with `@agent`. How can this possibly not lead to the well known type instability performance problems we had pre v6?

Tortar:
This is still totally incomplete, I will add a lot more info when I finish the transition. In practice though we can replace all the work of `@multiagent` with

```julia
# how many agents you want
@agent struct FirstAgent end
@agent struct SecondAgent end
# the sum type which removes all type instabilities
@sumtype BothAgent(FirstAgent, SecondAgent) <: AbstractAgent
```

No need for a special macro for dispatch, no need to if-else branches. The sum type is actually made of real agents, and not by opaque representation like in `@multiagent`. Almost only plain Julia code.

Tortar:
If you take a brief look at the ReadMe repo you will see how simple the new methodology is: https://github.com/JuliaDynamics/DynamicSumTypes.jl

Datseris:
okay, when you update this please also write out, at least in brief, what was the new "magic" you discovered that made this possible at an internal level, as I am not sure I can figure it out by source diving.

Tortar:
In practice the "magic" is just to enclose all the types in a single one and then use if-else branches to type-stabilize the result of the various function one needs to operate on them. I somehow overlooked this simple methodology initially because I didn't expect that something this simple could work that well considering the complexity of other packages trying to do something similar. I tested the methodology from 1.9 on with our benchmarks and it always removes any dynamic dispatch there was with a `Union`, e.g look at https://github.com/JuliaDynamics/Agents.jl/pull/1055/files#diff-7f4ca59230554e8d12359a1e1e1a3b4bc6accee721ebda1f9b8742814b231d29R156. And it is incredibly simple, like 50 locs of core code.

There is actually no need to go further than this in my opinion, Julia is good enough to do all the rest for us :+1:

CHANGELOG.md
```diff
@@ -18,12 +18,7 @@ _We tried to deprecate every major change, resulting in practically no breakage
...
 - Logic of when to collect data in `run!` has been improved to accommodate both discrete and continuous time models. This is reflected in the new options for the keyword `when`.
   - A new keyword `init` is now available for `run!` to data collect from the model before evolving it. Whether data were collected at time 0 or not was not really obvious in the original version of `run!` due to the ambiguity of the previous handling of `when`.
-- A new `@multiagent` macro allows to run multi-agent simulations much more efficiently. It has
-  two version: In `:opt_speed` the created agents are optimized such as there is virtually
-  no performance difference between having 1 agent type at the cost of each agent occupying
-  more memory that in the `Union` case. In `:opt_memory` each agent is optimized to occupy practically
-  the same memory as the `Union` case, however this comes at a cost of performance versus having 1 type.
-- `@multiagent` kinds support multiple dispatch like syntax with the `@dispatch` macro.
+- Integration with `DynamicSumTypes.@sumtype` macro allows to run multi-agent simulations much more efficiently.
```
Datseris, line 21: You can't alter a previous changelog entry, that defeats the whole point of making a changelog. You need a new entry for the new version.

docs/src/index.md
```diff
@@ -16,7 +16,7 @@ using CairoMakie, Agents
     Please see the online [CHANGELOG](https://github.com/JuliaDynamics/Agents.jl/blob/main/CHANGELOG.md) for a full list of changes.
     The most noteworthy ones are:
 
-    - A new `@multiagent` macro allows to run multi-agent simulations much more efficiently.
+    - Integration with `DynamicSumTypes.@sumtype` allows to run multi-agent simulations much more efficiently.
```
Datseris, line 19: We shouldn't refer to any external packages here. Or anywhere in the docs, or in the changelog. See my main comment of the PR review.

Datseris: Additionally, can we have a name that is more intuitive for Agents.jl? Something that includes somehow "agent", like `@sumagents` or `@combine_agents`?

docs/src/performance_tips.md
```diff
@@ -138,5 +135,16 @@ include(t)
 ```
 
 We see that Unions of up to three different Agent types do not suffer much.
-Hence, if you have less than four agent types in your model, using different types is still a valid option and allows you to utilize multiple dispatch.
-For more agent types however we recommend using the [`@multiagent`](@ref) macro.
+Hence, if you have less than four agent types in your model, using different types is still a valid option.
+For more agent types however we recommend using the [`@sumtype`](@ref) macro.
+
+Finally, we also have a more realistic benchmark of the two approaches at `test/performance/sum_faster_than_multi.jl` where the
```
Datseris, line 141: What makes this benchmark more realistic?

docs/src/tutorial.jl
```diff
@@ -680,12 +680,11 @@ adf
...
 # In realistic modelling situations it is often the case the the ABM is composed
 # of different types of agents. Agents.jl supports two approaches for multi-agent ABMs.
-# The first uses the `Union` type, and the second
-# uses the [`@multiagent`](@ref) command. `@multiagent` is recommended
-# as default, because in many cases it will have performance advantages over the `Union` approach
-# without having tangible disadvantages. However, you should read through
-# the [comparison of the two approaches](@ref multi_vs_union) to
-# be better informed on which one to choose.
+# The first uses the `Union` type (this subsection), and the second uses the [`@sumtype`](@ref)
+# command (next subsection) available in [DynamicSumTypes.jl](https://github.com/JuliaDynamics/DynamicSumTypes.jl).
```
Datseris, line 684: I understand that you would like to promote your work on this package, and it is worth doing so because it is good work. But this shouldn't happen here. See main comment of PR review.

docs/src/tutorial.jl
```diff
@@ -680,12 +680,11 @@ adf
...
-# The first uses the `Union` type, and the second
-# uses the [`@multiagent`](@ref) command. `@multiagent` is recommended
-# as default, because in many cases it will have performance advantages over the `Union` approach
-# without having tangible disadvantages. However, you should read through
-# the [comparison of the two approaches](@ref multi_vs_union) to
-# be better informed on which one to choose.
+# The first uses the `Union` type (this subsection), and the second uses the [`@sumtype`](@ref)
+# command (next subsection) available in [DynamicSumTypes.jl](https://github.com/JuliaDynamics/DynamicSumTypes.jl). 
+# This approach is recommended as default, because in many cases it will have performance advantages 
+# over the `Union` approach without having tangible disadvantages. However, we strongly recommend you
```
Datseris, line 686:
```suggestion
# over the `Union` approach without having tangible disadvantages. However, we recommend you 
```
we shouldn't recommend this strongly. If for the typical user it doesn't matter to use the recommended way, there is no "strong" incentive to attempt to learn more about this advanced topic

docs/src/tutorial.jl
```diff
@@ -708,161 +707,87 @@ end
...
-# **types**.
+# ## Multiple agent types with `@sumtype`
 
-# E.g., you can initialize
+# By using `@sumtype` from `DynamicSumTypes.jl` it is possible to improve the 
+# computational performance of simulations requiring multiple types, while almost
+# everything works the same 
 
-civ = Civilian(; id = 2, pos = (2, 2), group = 2) # default `mood`
+using DynamicSumTypes
```
Datseris, line 740: This package should be a dependency of Agents.jl and its macro re-exported. Additionally the macro can be renamed or wrapped into a more Agents.jl friendly name like `sumagents`.

docs/src/tutorial.jl
```diff
@@ -708,161 +707,87 @@ end
...
 
-# or
+@sumtype MultiSchelling(Schelling, Politician) <: AbstractAgent
 
-gov = Governor(; id = 3 , pos = (2, 2), group = 2, influence = 0.5)
+# Now when you create instances you will need to enclose them in `MultiSchelling`
```
Datseris, line 744: This is a clear and definitive disadvantage over `@multiagent`. That you have to always create a type of this Multi Agent type, instead of the standard Agent type. It really adds one more step versus the single agent usage we had so far in both 1-type models, Union type, or `@multiagent` type.

Have you thought how this impacts things on the user side?

docs/src/tutorial.jl
```diff
@@ -708,161 +707,87 @@ end
...
+# agents are then all of type `MultiSchelling`
 
-# and hence you can't use `typeof` to differentiate them. But you can use
+typeof(p)
 
-kindof(gov)
+# and hence you can't use only `typeof` to differentiate them. But you can use
 
-# instead.
+typeof(variant(p))
```
Datseris, line 754: Right, okay, so we are in the situation that "You can't use `typeof` but you can use `kindof`." Now instead of `kind` of we have `typeof(variant`. Whether one or the other is more complex is up to debate. But this new approach doesn't appear to solve any problems or make things simpler.

docs/src/tutorial.jl
```diff
@@ -708,161 +707,87 @@ end
...
+# instead. Hence, the agent stepping function should become something like
 
-function multi_step!(agent, model)
-    if kindof(agent) == :Civilian
-        civilian_step!(agent, model)
-    elseif kindof(agent) == :Governor
-        politician_step!(agent, model)
-    end
-end
+agent_step!(agent, model) = agent_step!(agent, model, variant(agent))
```
Datseris, line 758: Same comment as above. We got rid of the `if` clause the dispatches using `kindof`, but we now fundamentally adjust the agent stepping function to have 3 input arguments. This is not harmonious with the rest of the library. I feel like so far this new approach *decreases* the overall harmony between the three scenarios: 1 agent, Union agent, and our macro/sumtype agent.

Datseris (changes requested):
I am stopping the review after reading through the tutorial. I've added lots of comments. Unfortunately @tortar, I don't see this as an upgrade of what we had before. I don't see it as a downgrade either. 

Please try to explain how this is a better version of what was happening with the `@multiagent`. Please keep in mind: **we dont care how complicated the source code of DynamicSumTypes.jl is. We care about how complicated it is to use Agents.jl.**
To my understanding, the main advantage over using `@multiagent` is that 1) `@multiagent` can lead to a rather large chunk of code, because all subagent types must be inside, while with this new approach the agent types are still kept separately. More over in this approach existing agent types can be reused. However, is this really such a big advantage? Agent types can only defined once regardless, as Julia can't redefine types. So does it matter if the agent declaration is inside `@multiagent` or not? With this new approach, you can also either use `MultiAgent` or `Union`. But does this really matter? As far as I can tell, one almost never has a reason to use `Union`, as the multi agent infrastructure is so good. 

Now, here are some points where I don't see any advantage:

- We replaced `@multiagent` with creating the other wrapper type `MultiAgent`. So we don't really gain anything here.
- We didn't get rid of needing a replacement for `typeof`. We went from `kindof` to `typeof(variant`. This is not an improvement and I would argue it is more complex over `kindof`.
- We didn't get rid of now having to alter the "standard Julia multiple dispatch" for the agent stepping function. We went from the `@dispatch` usage to the 3-argument `agent_step!` function, where one of the three input arguments isn't even used anywhere in the function. It is just used for dispatch. In my eyes this is a downgrade, because it is increase in complexity, without any obvious benefit. I like more the `@dispatch` and the 2-argument agent step, because it is more harmonious with the rest of the library.
- We have a clear loss of benefit as now you can't use `add_agent!` the typical way. In 99% of the cases I've used Agents.jl, I was adding agents to the model via the `add_agent!` automatic route, I wasn't first creating them and then adding them to the model.

If you think this new approach is a better route for Agents.jl, then please make it transparent why you think so. I am really sorry for the negative review, as you've put in some work on this. But it is important that such a big change overcomes sufficient adversity before getting in.

---

Main comment 2: it is totally fine for you to reference DynamicSumTypes.jl, but the typical user of Agents.jl does not need to hear about this package, or even `using` it in their code. Instead, you can mention this infrastructure at the end of the docstring of `@sumtype` or the wrapper macro for it with a more Agents.jl like name.

Tortar:
thanks for the review @Datseris, it is indeed a big change, but as I will explain really worthwhile in my opinion.

The problematic aspects of `@multiagent` which the new approach solves are:

- You can't reuse types. You already noticed this, but underestimate its importance in my opinion. This is a very good feature, because let's say you want to reuse an already defined type in another library. This allows to do that, `@multiagent` does not because the representation is opaque.

- The multiple dispatch it offers is incomplete and not totally correct. You can't for dispatch on 
  - the `@multiagent` type and at the same time variants of a type. This is a fundamental limitation of the previous approach, because you would need to write ambiguity resolution rules by yourself (very very hard) and at the same time you would need to anyway through when ambigous while Julia just warns you.

  - `Vararg` arguments can sometime (rarely actually) cause incorrect results, this could be solved but it would require a fair bit of work.

- You can actually create containers of only variants if you want, I'm not totally sure that this is very needed feature but for example you could do `Agent1[variant(a) for a in agents(model) if variant(a) isa Agent1]`. The previous approach is too opaque for this, you would end up breaking a lot of assumptions.

In practice the new approach is really the same of a `Union` of types but type stable...it is just a bit less ergonomic, but the way it works is very Julian in my opinion. It is easy to understand how to work with that, after some initial effort at least.

From the performance standspoint is actually amazing in my opinion. It is faster and more memory efficient than `:opt_memory` and only slower than `:opt_speed` when the types are very similar one from the other. This is because `:opt_speed` starts to lose its speed advantage when types becomes heterogeneous because they become fat memory wise which has effects also on the speed of them.

docs/src/tutorial.jl
Tortar: there is new also a `variantof` function also now

docs/src/performance_tips.md
Tortar: Types are more heterogeneous as I would expect in a multi agent simulation, also it is more complex to optimize away dynamic dispatch in this setting

docs/src/index.md
Tortar: Yes, better probably. Maybe `@sumagents`?

CHANGELOG.md
Tortar: yes sorry

docs/src/tutorial.jl
Tortar: actually you can even improve speed by utilize the variant sometimes, because if you modify the variant you modify the type, but faster. But this is a bit more advanced and so I didn't mention it :D 

But consider that anyway it doesn't require a new macro (`@dispatch`) and inside a module it doesn't require also the other (`@finalize_dispatch`) and the messy thing when you want to use the function before the end of the module

Tortar: This is another point, this is actually almost plain and simple. It doesn't require the fair bit of DSL we needed for `@multiagent`, which really increased more complexity in my opinion

Tortar:
Actually, I'm checking the performance of Union-splitting in Julia 1.11 and I'm finding that it works quite well, almost on par with `@sumtype`, seems like Julia is going to finally solve this issue without extra work!! I still need to see if it is robust enough, but maybe Unions are good enough in that? Then I would just report back some of the insights from DynamicSumTypes in Julia repo, and that's enough. So let's put this on stale for a bit

Datseris:
> In practice the "magic" is just to enclose all the types in a single one and then use if-else branches to type-stabilize the result of the various functions one needs to operate on them

This is a point I still don't understand. If we look at the code we have:
```
agent_step!(agent, model) = agent_step!(agent, model, variant(agent))
```
Isn't this type unstable? At a fundamental level, `variant(agent)` is type unstable, no? Now can you call dispatch on the outcome of `variant(agent)`? if you call dispatch on it, then it must be a type. But given that the input is always the type of `agent`, you cannot statically deduce what `variant(agent)` will be, since it has two possiblities, while `agent` only has one type.

I don't understand how this can be type stable. Are you sure this version of the code does not have performance regressions versus `@multiagent`? Maybe I just am not understanding something simple, but I don't understand how this new approach is faster, or different even, than `Union`. :S  I don't understand how this statement can be true:

> In practice the new approach is really the same of a Union of types but type stable...


---



> * but underestimate its importance in my opinion. This is a very good feature, because let's say you want to reuse an already defined type in another library

Right, that is a very good point I didn't think about.



> * The multiple dispatch it offers is incomplete and not totally correct

I accept that it is incomplete, and it is a valid point I also didn't think about. I do not understand when they can be not "totally correct", and what does this mean quantitatively: how much "incorrect" will they be, and what does "incorrect" even mean in this context? Code errors, code doesn't but the wrong method runs?



> * You can actually create containers of only variants if you want

I don't understand this. How is this code example you show there different from `[a if kindof(a) == :my_want for a in allagents(model)]`? Is it because the container type is not the general type (in the new case that would be "MultiAgent")? Well that's because you enforce it with the prefacing `Agent1[]`. The question would be wether the list comprehension `[...]` by itself would be type stable and yield `Agent1` type container, which I don't think it would.



> Actually, I'm checking the performance of Union-splitting in Julia 1.11 and I'm finding that it works quite well, almost on par with `@sumtype` in many cases, seems like Julia is going to finally solve this issue without extra work!!

Sure, that would be nice, although I doubt this would be the case for realistic ABM code. Perhaps we should use the current hawk rabbit fox example, and 3-4 more species, and use these dynamics as an additional benchmark. 

> Then I would just report back some of the insights from DynamicSumTypes in Julia repo, and that's enough. 

I didn't understand this. You will propose the main Julia repo to use DynamicSumTypes?

Datseris:
1.11 isn't too far away from release, but so far you have made a solid case for this PR. If we really want to get rid of `@multiagent` we want to do it earlier than later of course. The less people had the change to use it the better...

Tortar:
> 1.11 isn't too far away from release, but so far you have made a solid case for this PR. If we really want to get rid of `@multiagent` we want to do it earlier than later of course. The less people had the change to use it the better...

I will respond later to your other concerns, but yes I agree, at the same type given that Julia will get a lot better, I would just mention `DynamicSumTypes` in the performance tips and nowhere else. If we find that the performance of `Union` is solid enough of course.

Datseris:
From here: https://discourse.julialang.org/t/ann-dynamicsumtypes-jl-v3/116741/15?u=datseris

in Agents.jl you would typically care about 1-10 types. It would be very rare for real-world ABM simulations to have more types than that. And DynamicSumTypes.jl is better than union there also on v1.11. So perhaps it is worth doing a complex predator prey with 6-7 species after all to benchmark the case where agents are crreated and removed as well. But it seems that this is a solid candidate nevertheless.

Tortar:
I actually think that with removal/addition of agents `@sumtype` will be a bit more faster than a `Union` compared to the benchmarks we have now, some micro-benchmarks actually seem to confirm this but I didn't have the time to finish the exploration yet. Actually already in both macro-benchmark `@sumtype` is something like 50% faster than a `Union`. I will let you see this by building the docs with 1.11

Tortar:
Alright, I built with nightly and `@sumtype` still has the edge on both models: https://juliadynamics.github.io/Agents.jl/previews/PR1055/performance_tips/#sum_vs_union

docs/src/index.md
Datseris: Hm, I am not sure I like having `sum` inside because the agents aren't summed. I prefer combine, but perhaps `merge` is a common ground? `merge_agents`? or `@agentmerge` ?

Datseris:
Okay, then over the weekend I'll do a second review so we actively work to get this in.

Tortar:
Hi George, so how do you think we should proceed? In my opinion, we should deprecate `@multiagent` and link to `@sumtype` only in the performance tips (ensuring that it works fine with the library though, so some internal adaptation should be necessary)

Datseris:
What do you mean mention this only in the performance tips? So what would be the state of the main tutorial? At the moment the main tutorial discusses `Union` and `@multiagent`. What is your proposal?

Tortar:
My proposal is removing that section (or just show `Union`). Since the advantage of `@sumtype` is not that great and at the same type the macro is a bit less ergonimic, it probably just deserves a section in the performance tips in my opinion

Datseris:
No, I think this is too much of a breaking change. We can't just remove `@multiagent` and offer no replacement for it. What is the problem with replacing the `@multiagent` section with this new approach? 

And how are we handling the EventQueueABM which is currently hardcoded to use `@multiagent`?

Datseris:
Also keep in mind that the next LTS version of Julia is 1.10 not 1.11.

Tortar:
yes, changing that section would be also good, also in view that 1.10 is going to become LTS. We didn't make any guarantee on the working `EventQueueABM` because we marked it as experimental, indeed, I already made it working with single types and unions in this PR instead of multiagent.

Datseris:
ah, thanks, i didn't notice that. you are totally right.

Datseris:
Given the latest reply here https://discourse.julialang.org/t/ann-dynamicsumtypes-jl-v3/116741/27?u=datseris  it says that we should no matter what keep some custom , so this multi agent infrastructure is probably going to stay for the long run.

So shall we approach this PR with this in mind, and implement some wrapper macro with Agents.jl-specific name like I have suggest above? And also put this as a dedicated section in the tutorial?

docs/src/index.md
Tortar: maybe just use `@multiagent` as before? I would really prefer this name among all others, we just deprecate the old version

Tortar:
Should be okay now.

I didn't follow the suggestion to remove the 3-argument `agent_step!` because I actually think it is better to have that, it is cleaner and allows to understand that multiple-dispatch is available to be used, I don't think that the two argument reusability is a strong argument because even if you use only two, it can't be reused as is, if you use a `Union` you would need to annotate the functions and so change a bit the source code in any case

docs/src/api.md
```diff
@@ -31,11 +31,6 @@ AgentEvent
 ```@docs
 @agent
 AbstractAgent
-@multiagent
```
Datseris, line 34: This you have to add back, right? Because you used this name now for the "agents-like" version of `@sumtype`.

Datseris:
I think the tutorial should say how to add agents of a certain type in the model via `add_agent!` for both multi agent routes (which is teh recommended way to create new agents), rather how to just create a new agent and do nothing with it.

src/core/agents.jl
```diff
@@ -414,67 +253,75 @@ function compute_base_fields(base_type_spec)
...
-    @dispatch f(args...)
+    @multiagent YourAgentType(AgentTypesToMerge) [<: OptionalSupertype]
 
-A macro to enable multiple-dispatch-like behavior for the function `f`,
-for various agent kinds generated via the [`@multiagent`](@ref) macro.
-For an illustration of its usage see the [Tutorial](@ref).
+Define multiple agent "subtypes", which are variants of a unique type `YourAgentType`. 
+This means that all "subtypes" are enclosed in the overarching type. Then, You cannot
+distinguish them on the basis of `typeof`, but need to use instead the `variantof`
+function. The `allvariants` function for a convenient way to obtain all variants types.
```
Datseris, line 266: Can you please move the reference to DynamicSumTypes.jl here and say that it uses the `@sumtype` macro, instead of referencing DynamicSumTypes.jl in the main tutorial?

docs/src/tutorial.jl
```diff
@@ -680,12 +680,11 @@ adf
...
 # In realistic modelling situations it is often the case the the ABM is composed
 # of different types of agents. Agents.jl supports two approaches for multi-agent ABMs.
-# The first uses the `Union` type, and the second
-# uses the [`@multiagent`](@ref) command. `@multiagent` is recommended
-# as default, because in many cases it will have performance advantages over the `Union` approach
-# without having tangible disadvantages. However, you should read through
-# the [comparison of the two approaches](@ref multi_vs_union) to
-# be better informed on which one to choose.
+# The first uses the `Union` type (this subsection), and the second uses the [`@multiagent`](@ref)
+# command (next subsection) available in [DynamicSumTypes.jl](https://github.com/JuliaDynamics/DynamicSumTypes.jl).
```
Datseris, line 684:
```suggestion
# The first uses the `Union` type, and the second uses the [`@multiagent`](@ref)
# command. 
```
see comment below.

Datseris:
Thanks for finishing this @Tortar . Fine we can leave things as they are now. Let's try to get this in soon as I've seen several people use `@multiagent` already, so let's hope they won't complain too much :D

docs/src/api.md
Tortar: yes, thank you

Tortar:
yes let's hope that this is the case, unfortunately this was a bit of unknown territory for me so I made some overly complex implementation at the start :-)

I addressed your review, if you approve I will do a final check and then merge

docs/src/tutorial.jl
```diff
@@ -708,161 +707,81 @@ end
...
+# while almost everything works the same 
 
-fieldnames(MultiSchelling)
+@multiagent MultiSchelling(Schelling, Politician) <: AbstractAgent
 
-# that contains all fields of all subtypes without duplication, while
+# Now you can create instances with
 
-fieldnames(Civilian)
+p = MultiSchelling'.Politician(model; pos = random_position(model), preferred_demographic = 1)
```
Datseris, line 744: I noticed this before and I thought it was a typo. What's the point of the `'` here? That in linear algebra stands for the transpose operator? Why do we have `MultiSchelling'.Politician` instead of `MultiSchelling.Politician`?

Datseris: I think you can overload `getfield` to work withe subtypes properly without the `'`. I admit I don't really understand the syntax here, what does `'` achieve and why it is here.

Datseris: Actually, can we just use `Politician` directly? Why do we have to preface it with `MultiAgent` at all? That would be te best.

Tortar: well it's actually a trick, overloading `getproperty(::Type{...})` is actually a mistake because it created problems, see https://github.com/JuliaLang/julia/discussions/48883#discussioncomment-5362804

Tortar: also, we need to preface it because `Politician` is already taken from the type itself

Datseris: Can we perhaps make an actual function instead of this funny syntax? If you think about it it is really atypical to write `Someting'.something`. My guess is that using the "transpose" operator on `MultiAgent` transforms it to something else, from which you can then access the field `Politician`. 

For a front end user, this may fly over their heads. I can easily imagine users forgetting `'` and then getting errors (that may or may not be informative). It is very unfortunate that we can't use direct dot `.` syntax...

Can we briefly brainstorm alternatives that are more transparent? Or, better yet, how does the `add_agent!` version work for MultiAgent models? do you give `MultiAgent'.Politician` as first argument and that's it? Perhaps we could provide an alternative syntax at the level of `add_agent!`? 

I don't think we should be promoting making agents outside of `add_agent!`. We have already "forbidden" the user to set the ID, so I don't think making agents on their own is even compatible with the publically declared api.

Datseris: I also read https://github.com/JuliaLang/julia/discussions/48883#discussioncomment-5362804 but I don't understand the issue/problem as we have a macro that creates `MultiAgent` so in essence we can make the fields of this object whatever we want. can you perhaps explain in layman terms why `MultiAgent.Politician` isn't possible?

Tortar: > do you give `MultiAgent'.Politician` as first argument and that's it? Perhaps we could provide an alternative syntax at the level of `add_agent!`?

yes, that is how it works. I don't have good ideas to make it work better unfortunately.

I think we could provide also a function (which would return a closure) but this is not that required here and I'm not sure how I would call it. But this is in my opinion not a blocking aspect for this PR, this could be always improved in a new PR, the one based on adjoint is here to stay anyway because it is backed by the underlying package

Tortar: the issue is from what I get that for any type by default you have

```
julia> struct A end

julia> propertynames(A)
(:name, :super, :parameters, :types, :instance, :layout, :hash, :flags)
```

but if you overload `getproperty(::Type, ::Symbol)` for (only) your type you practically creates problems because I think Julia invalidates a lot of unrelated code because of this. I don't understand the reason either but that is true as attested here https://github.com/Roger-luo/Expronicon.jl/issues/77 (final comment of the thread I mentioned)

Datseris: Can you write a simple function `gettype(MultiAgent, Type) = MultiAgent'.Type` and use this function in the tutorial? Also adding a sentence saying "We can't use `Politician` directly, as this is an existing type that doesn't have any refenrce to `MultiAgent`. We have to extract an incorporated `Politician` subtype, which we can do with `gettype`."

I am happy to discuss better names for `gettype` :D

Datseris: > but if you overload getproperty(::Type, ::Symbol) for (only) your type you practically creates problems 

We have been doing this for many many years with `AgentBasedModel` though: https://github.com/JuliaDynamics/Agents.jl/blob/main/src/core/model_abstract.jl#L116-L132  . And I've never heard of these problems. Besides, teh Julia manual itself says that `getproperty` is the function you overload for the `.` syntax. So this is a point I don't understand.

Tortar: In `AgentBasedModel` is different, there you don't have a type, there you have an instance of a type (`ABM` instead of `Type{ABM}`), that is okay and very encouraged actually I think

Tortar: that would be okay, maybe `get_type`? I think it's the tersest we can get probably...maybe `getconstr`? let me know your preference and we will go with that

Datseris: > In `AgentBasedModel` is different, there you don't have a type, there you have an instance of a type (`ABM` instead of `Type{ABM}`), that is okay and very encouraged actually I think

ooooooh yes of course, completely missed that... `Type{}` changes everything.

Datseris: Let's go with just `constructor`? We don't need `get`, since the function is named after what it returns.

Tortar: that's good for me, I have implemented it

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v6.1
+
+- The version of `@multiagent` introduced in Agents.jl v6.0 has been deprecated because a better metodology to run multiagent simulation efficiently was found. For this reason,
```
Datseris, line 3:
```suggestion
The `@multiagent` macro introduced in Agents.jl v6.0 has been completely overhauled. We found a better methodology to create performant multi-agent types that can re-use existing agent types. Please read the docstring of the new `@multiagent` and consult the updated tutorial in v6.1 for more details.

All `@multiagent`-specific functions that were introduced in v6.0, such as `kindof`, have also been deprecated, see the main tutorial.

For `EventQueueABM`, that was an experimental feature that used to work only with `kindof`, it is also now changed to work with the standard `typeof` of the base Julia language.
```

Correct?

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+a new version was also introduced, its working is described in the tutorial and in the performance tips of the documentation.
```
Datseris, line 4:
```suggestion
```

src/core/agents.jl
```diff
@@ -414,67 +253,77 @@ function compute_base_fields(base_type_spec)
...
-@dispatch MethodSub1(::SubAgent2) = 1
+@multiagent Animal(Wolf, Hawk{Float64})
+```
+
+Then you can create `Wolf` and `Hawk` agents like so
+```
+hawk_1 = Animal'.Hawk(1, (1, 1), 1.0, 2.0, 3)
+hawk_2 = Animal'.Hawk(; id = 2, pos = (1, 2), ground_speed = 2.3, flight_speed = 2)
+wolf_1 = Animal'.Wolf(3, (2, 2), 2.0, 3.0, :black)
+wolf_2 = Animal'.Wolf(; id = 4, pos = (2, 1), ground_speed = 2.0, fur_color = :white)
```
Datseris, line 293: let's use `constructor` instead of `'`.

src/core/agents.jl
```diff
@@ -414,67 +253,77 @@ function compute_base_fields(base_type_spec)
...
+```
 
-end
+You can also access the enclosed variant instance with the variant `function`
+```
+variant(hawk_1) # Hawk(1, (1, 1), 1.0, 2.0, 3.0)
+variant(wolf_1) # Wolf(3, (2, 2), 2.0, 3.0, :black)
 ```
+
+See the [rabbit_fox_hawk](@ref) example to see how to use this macro in a model.
```
Datseris, line 308: The hyperlink here is broken.

examples/event_rock_paper_scissors.jl
Datseris, line 1: In this example we don't use `@multiagent`. Also it uses `DynamicSumTypes` in its start while it doens't have to.

Datseris (commented):
We are almost there, really.  Only minor stuff and the event que model!

src/core/agents.jl
Tortar: I'm not sure how to fix it quickly because I'm a bit rusty on Documenter things, could you fix it if you know how? I addressed the rest of the review

Datseris: I can't checkout the branch now but the fix is easy: first, change the title of the example to:
```
# [# 3D Mixed-Agent Ecosystem with Pathfinding](@id rabbit_fox_hawk)
```
in the example `.jl` file. Then refer to it as `[rabbit fox hawk example](@ref rabbit_fox_hawk)`.

Datseris:
Damn, using the multiagent instead of Union makes the event que example signigicantly more complicated.

Tortar:
not too much in my opinion, it is also a bit faster both in 1.10 and 1.11, 500ns per event instead of 560, not much but something :D

Datseris (approved):
WIll merge when docs build.

PR #1058, opened 2024-07-14
Fix some (innocuous) dynamic dispatches

Tortar:
These changes make JET a bit happier about the package

### DIFF
--- a/src/core/model_abstract.jl
+++ b/src/core/model_abstract.jl
@@ -72,7 +72,7 @@ const ABM = AgentBasedModel
 # internal methods or the dev docs.
 
 notimplemented(::A) where {A<:ABM{S}} where {S} =
-    error("Function not implemented for model of type $(nameof(A)) with space type $S.")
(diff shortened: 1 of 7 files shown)


Datseris:
What's `lazy""`?

Tortar:
```julia
help?> lazy""
  lazy"str"

  Create a LazyString using regular string interpolation syntax. Note that interpolations are
  evaluated at LazyString construction time, but printing is delayed until the first access to the
  string.

  See LazyString documentation for the safety properties for concurrent programs.

  Examples
  ≡≡≡≡≡≡≡≡

  julia> n = 5; str = lazy"n is $n"
  "n is 5"
  
  julia> typeof(str)
  LazyString

  │ Julia 1.8
  │
  │  lazy"str" requires Julia 1.8 or later.
```

this can be used to remove dynamic dispatch when interpolating variables in errors, I learnt the trick from Base Julia

Datseris (approved):

Issue #1045, opened 2024-06-07
Flocking model video doesn't work after update to Makie 0.21


Tortar:
Look at https://juliadynamics.github.io/Agents.jl/dev/examples/flock/

Datseris:
almost certainly it is because now makie checks for keyword arguments and throws errors for invalid kwargs. I've had similar problems in other packages. So we just have to be careful with the kwargs we give to makie.

amynang:
commenting out the agent_marker argument in abmvideo solves the issue (but agents are now plain circles)

> ERROR: Invalid attributes markersize and marker for plot type Poly.

Datseris:
Thanks @amynang , are you interested to create a PR for this? The issue isn't actually in `ambvideo`, but in `meshscatter` (or similar function) used internally deeper in the code.

amynang:
I'm learning Julia as I'm learning how to use Agents.jl so it might be a while before I can be useful...

PR #1059, opened 2024-07-15
Update abstract.jl solves #1045

amynang:
The culprit behind #1045 was the use of poly! for user defined polygons. As a first step, I replaced it with  `scatter!(p, p.pos; p.color, p.marker, p.markersize, p.agentsplotkwargs...)` and generating the plot with the custom polygon worked. In that case I think the whole if else statement is no longer necessary so I removed it.

```
if user_used_polygons(p.agent_marker[], p.marker[])
poly!(p, p.marker; p.color, p.marker, p.markersize, p.agentsplotkwargs...)
else
scatter!(p, p.pos; p.color, p.marker, p.markersize, p.agentsplotkwargs...)
end
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati", "Adriano Meligrana"]
-version = "6.0.16"
+version = "6.0.17"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"


Tortar:
I didn't test anything but I feel like removing `poly!` shouldn't be the solution, what does the error say when you try to create the video with the previous code?

Tortar:
ah also, thanks for your help on this :-)

Datseris:
unfortunately i have to agree, removing `poly!` also doesn't sound like a solution to me. Does the `scatter!` actually plot the polygons..? Does the flocking video looks like flocking, with triangles moving and rotating around? If not, then we haven't solved the problem.

I suspect that this code needs to stay as is, with `if`:
```julia
if user_used_polygons(p.agent_marker[], p.marker[])
poly!(p, p.marker; p.color, p.marker, p.markersize, p.agentsplotkwargs...)
else
scatter!(p, p.pos; p.color, p.marker, p.markersize, p.agentsplotkwargs...)
end
```
but what needs to be done is that we cannot blindly propagate `p.agentsplotkwargs` anymore. There are some keywords that `poly!` doesn't accept that are otherwise valid for `scatter!`, so we have to filter them out.

amynang:
@Tortar  [here](https://github.com/user-attachments/files/16238773/error.txt) is the full error message with the original code. [This](https://github.com/amynang/MSB/blob/main/Chapter2/ch2_0_particleworld.jl) is the model where I encountered the problem.

@Datseris scatter! does actually plot the polygons. The first picture is a screenshot with the PR code, the next one is with the agent_marker argument in abmvideo commented out. This is the strange behavior I described earlier, both cases have 100 agents but the first looks zoomed into a part of the space for some reason.
pr1
pr0

Datseris:
> agent_marker argument in abmvideo commented out

Don't do that, it doesn't make sense at the moment. So we don't have to spend resources figuring out what commenting this out does.

I didn't know that `scatter!` plots polygons. Then we can just use `scatter!` for everything. However, let's make sure about one more thing: if you increase the polygon size, for example by multiplying the polygon coordinates with some factor
```julia
const bird_polygon = Makie.Polygon(5 .* Point2f[(-1, -1), (2, 0), (-1, 1)])
function bird_marker(b::Bird)
    φ = atan(b.vel[2], b.vel[1]) #+ π/2 + π
    rotate_polygon(bird_polygon, φ)
end
```

(notice `5 .*`)


Are the polygons larger in size in `scatter!`? Does it respect the actual size of the polygons?

amynang:
Yep, polygons get larger
pr2

Tortar:
I tried your model, and with the last commit I pushed everything works fine, we don't propagate to `poly!` anymore `markersize`, but maybe it never worked? From the code of @datseris it seems so because you define the markersize when creating the polygon

Tortar (approved):

amynang:
Great! Thanks a lot!

Datseris (commented):
Alright, please go into top-level Project.toml and increment the _patch_ version (3rd number) by one. Then we can merge and immediatelly tag a new release.

PR #1064, opened 2024-08-09
Speed-up a bit multiagent benchmark against union

Tortar:
This gives a bit of speed-up of multiagent benchmark while not damaging the perf of Union


Datseris (approved):
sure, bump patch version so it's in the docs?

Tortar:
yes, actually this change makes a big difference in performance, more than I expected

PR #1065, opened 2024-08-10
Improve performance of EventQueueABM by better union splitting

Tortar:
before

```julia
julia> model = initialize_rps();

BenchmarkTools.Trial: 10000 samples with 197 evaluations.
 Range (min … max):  426.081 ns …  56.132 μs  ┊ GC (min … max): 0.00% … 98.59%
 Time  (median):     471.038 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   512.019 ns ± 745.854 ns  ┊ GC (mean ± σ):  5.40% ±  4.60%

         ▂▄▆▇█▇▆▄▃▁                                              
  ▁▁▂▃▄▅████████████▆▅▃▃▃▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁ ▃
  426 ns           Histogram: frequency by time          641 ns <

 Memory estimate: 378 bytes, allocs estimate: 7.
```

after

```julia
julia> model = initialize_rps();

BenchmarkTools.Trial: 10000 samples with 207 evaluations.
 Range (min … max):  343.237 ns …  46.423 μs  ┊ GC (min … max): 0.00% … 98.19%
 Time  (median):     390.423 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   409.583 ns ± 565.531 ns  ┊ GC (mean ± σ):  3.01% ±  2.53%

              ▂▃▅████▇▅▄▂▁                                       
  ▁▁▁▁▁▂▂▃▄▅▇█████████████▇▆▅▅▄▄▄▃▃▄▄▄▃▃▃▃▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  343 ns           Histogram: frequency by time          495 ns <

 Memory estimate: 157 bytes, allocs estimate: 2.
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Agents"
 uuid = "46ada45e-f475-11e8-01d0-f70cc89e6671"
 authors = ["George Datseris", "Tim DuBois", "Aayush Sabharwal", "Ali Vahdati", "Adriano Meligrana"]
-version = "6.1.1"
+version = "6.1.2"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
--- a/src/Agents.jl
+++ b/src/Agents.jl
@@ -17,7 +17,7 @@ using DynamicSumTypes
 export variant, variantof
 import ProgressMeter
 using Random
-using StaticArrays: SVector
+using StaticArrays: SVector, SizedVector
 export SVector
 import LinearAlgebra
 import StreamSampling
--- a/src/core/model_event_queue.jl
+++ b/src/core/model_event_queue.jl
@@ -141,7 +141,7 @@ can be evolved in time.
   event for an agent after an event affected said agent has been triggered.
 """
 function EventQueueABM(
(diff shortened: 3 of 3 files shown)


Datseris:
that's big!

can you give more info here? What are you doing? What is "union splitting" in this context?

Tortar:
The key idea here in my opinion is understanding that tuples are not made for this task (at least currently) because when the access to a tuple is random and there are multiple types inside it, the tuple will not union split but just use a dynamic dispatch mechanism. When the access is fixed a tuple would be instead be better. A SizedVector is instead a good one because as a Vector it will union-split, at the same time we know the size of it, so we can use a fixed-size one.

Datseris:
What does "union-split" mean?

Datseris (approved):

Tortar:
https://julialang.org/blog/2018/08/union-splitting/

PR #1067, opened 2024-08-12
Use a faster ordering for heap in EventQueueABM

Tortar:
Another 10% of time shaved with this, I think that's it for now :D


Datseris (approved):
For me a comment is about the benchmarks. You are measuring everything in Rock Paper Scissors, right? We are confident that this model is representative for a generic speedup? I guess so...

Tortar:
yes, I'm measuring it. I'm confident enough on the generality but surely we could improve the benchmark by increasing the number of types

Issue #841, opened 2023-08-04
Multi Agent model implementation proposal


Tortar:
It is known that Agents.jl suffers from dynamic dispatch when working with a high number of agents types. I think that now that we have an abstract interface to rely on and to implement we can try to solve the issue. Currently, we just have some workarounds: representing every type in only one struct, but this is unnatural, and wastes some RAM (and it is probably not as performant as the possible implementation described below)

I tried to use [Virtual.jl](https://github.com/thautwarm/Virtual.jl) and [ManualDispatch.jl](https://github.com/jlapeyre/ManualDispatch.jl) and custom macros to solve the problem, but to no avail, even though the first package seems actually suited to improve the matter it didn't work well (performance was worse) when I tried to make use of it.

So I think we have only the hard way to move forward on the issue: implement all the interface for a model where the main data structure is **a tuple where each type is inside its own Dict/Vector**. This should be even better in terms of performance than compressing in only one struct different agent types without the need to change anything.

Tortar:
I had an idea so that to make the access to the right container for a given id, namely how can `model[id]` work for a model with a tuple of vectors/dict. The problem is identyfing which is the container for a given id: the strategy I have thought is that we could try to set a new id with `nextid` such that if we have N types of agents we set them such that the remainder of the division by N gives the right value for the index of the container, this means that for example if we have 3 types and the id is 7 we can do `mod1(7, 3)`to identify the right container. 

However, this means that we can't use only `nextid(model)` to let the user set a new id since this is not enough to identify the right id to give to the new agent, we should extend it to `nextid(model, agent_type)`.

Tortar:
@Datseris I'd like to know what you think about all of this when you have the time :-)

Datseris:
Problem is that you are still using the same type (the integers) to get agents of different type, so your idea would also be type unstable.

Limiting the `id` assinged to agents is totally fine for me. There is reallh no reason for users in the first place to care about the id, having it auto assigned would have been simpler from a design view point anyways, especially now with the improvements of `@agent`. I just don't see how this would help performance though.

Datseris:
But then again, it may be much more performant to have a tuple of dicts, each with concrete type, rather than having a dict of Union type. But I truly do not know, we would need to test a minimal implementation (which really shouldn't be too hard).

Datseris:
BTW, although this is a different issue, I have been thinking that in version 6 we should rework the model interface so that the stepping functions are included in the model. 

[...]

PR #1073, opened 2024-09-06
Error if deletion of agent fails

Tortar:
This seems the only thing to improve in `remove_agent_from_space!` which could have been seen as a bug (but it is not), related to the issue in https://github.com/JuliaDynamics/Agents.jl/issues/841

### DIFF
--- a/src/spaces/continuous.jl
+++ b/src/spaces/continuous.jl
@@ -139,6 +139,7 @@ function remove_agent_from_space!(
 )
     prev = abmspace(model).grid.stored_ids[cell_index...]
     ai = findfirst(i -> i == a.id, prev)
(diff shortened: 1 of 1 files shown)


Datseris:
Interpolating the whole agent struct into the string does not seem like a good idea to me. Perhaps only the ID?

---

Beyond this however, what is the point of this error message? A user should never directly remove an agent "only from space" and this error message said that we can't remove "from the space", not that we generally can't remove the agent. Who is this error targetting?

Tortar:
The main motivation here is to remove (innocuous) dynamic dispatches found by JET, because guarantees not to throw if the error path is not followed.

Datseris:
Right, ok then.

Datseris (approved):

PR #1077, opened 2024-09-24
Change algorithm for reservoir sampling

Tortar:
I inadvertedly broke some code in Agents with my latest release of StreamSampling as can be seen from https://github.com/JuliaDynamics/Agents.jl/actions/runs/11009882272/job/30570524847, this fixes the failures, sorry


Tortar:
Actually I should probably fix the problem in the underlying package instead

PR #1086, opened 2024-10-06
DelaunayTriangulation example

DanielVandH:
As discussed @Datseris, this PR adds an example using DelaunayTriangulation.jl to simulate a cell model. The model has changed from what it used to be in DelaunayTriangulation.jl's, and now simulates interactions between three cell species and also includes cell death. This was my first time actually using the package so maybe some things are not done optimally.

Here's what the final video looks like.

https://github.com/user-attachments/assets/24f1b6fb-2e4d-46f9-8527-ed4934528379


docs/Project.toml
```diff
@@ -7,12 +7,13 @@ CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
...
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
+DelaunayTriangulation = "927a84f5-c5f4-47a5-9785-b46e178433df"
 DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
-DynamicSumTypes = "5fcdbb90-de43-509e-b9a6-c4d43f29cf26"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterTools = "35a29f4d-8980-5a13-9543-d66fff28ecb8"
 DrWatson = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
+DynamicSumTypes = "5fcdbb90-de43-509e-b9a6-c4d43f29cf26"
```
Tortar, line 16:
```suggestion
LightSumTypes = "f56206fc-af4c-5561-a72a-43fe2ca5a923"
```

This package is actually deprecated in favour of `LightSumTypes.jl`

Tortar:
Very cool model anyway!! My only concern is about plotting, is there a way to reuse more of the Agents.jl infrastructure for it?

Datseris:
I'd like to review this before merging. Bit busy at the moment but will find time for sure.

DanielVandH:
Maybe? I could have made the animation simpler and just not show the data graphs so that `abmvideo` works. But it seemed like a nice way to also demonstrate how that could be done.

If there's a way to change the recipe used in `abmplot` (e.g. something other than `scatter` or `poly`), then it will also be a lot simpler since I could use `voronoiplot`.

examples/delaunay.jl
```diff
@@ -0,0 +1,609 @@
...
+# sample according to the triangle areas. In the function below, there is some obvious room for improvement
+# to reduce allocations, but this is not a concern in this example.
+using StatsBase
+function random_triangle(tri::Triangulation)
+    triangles = collect(DT.each_solid_triangle(tri)) # collect since we need to know the ordering of the areas
+    areas = [DT.triangle_area(get_point(tri, triangle_vertices(T)...)...) for T in triangles]
+    weights = Weights(areas)
+    T = sample(triangles, weights)
+    return T
+end
```
Tortar, line 329:
```suggestion
# Next, we need to know how to select a random triangle from a triangulation. This is done by using a weighted 
# sample according to the triangle areas.
using StreamSampling
function random_triangle(tri::Triangulation)
    triangles = DT.each_solid_triangle(tri)
    area(T) = DT.triangle_area(get_point(tri, triangle_vertices(T)...)...)
    T = itsample(triangles, area)
    return T
end
```

Tortar: This package is already used in Agents, and it should reduce allocations, actually in this case maybe not, because I ran the example and I saw that `random_triangle` is used only a few times, but anyway this could be useful

Tortar: yes, that is the reason, indeed I see some notable time improvements with `@time step!(initialize_cell_model(radius=0.1, ninit=3), 2000);`

Tortar (approved):

Tortar:
To me, this seems mostly fine, just wait for the review of @Datseris to be merged :-)

Tortar:
@Datseris do you have the time ro review this? Otherwise we could merge it and push updates in the future

Datseris:
I'm sorry, I am in a difficult time in life now with no time to spare. I wanted to review because I am convinced we can simplify ploitting, but I don't have the time to do it now. GO ahead and merge and in the future maybe I can make a PR myself!

Tortar:
no problem, I'm also running a bit out of time these days, after graduation I have a bit more frenetic routine :D 

If I find the time I will try to update the plotting part, but I don't want to put this onhold because I think it's a great contribution

PR #1091, opened 2024-10-10
Update multiagent vs. union performance docs for Julia 1.11

Tortar:
The new timings are much more similar in 1.11, see https://juliadynamics.github.io/Agents.jl/dev/performance_tips/#multi_vs_union. But `@multiagent` has still the edge


Datseris:
I've been seeing performance regressions all across the board with 1.11. In compile time and run time... I suggest that we wait for 1.11.1 fix first before considering this seriously.

Tortar:
that seems good to me, but I'm skeptical about that making a big difference in compile times, I think it will take more than 1 release to improve this substantially, run times seems better to me on 1.11 instead

Tortar:
1.11.1 is out and it didn't improve much the state of Julia 1.11 release, I think it will take a fair amount of time to fix the compile issues of 1.11

Tortar:
we can merge this one now that 1.11.2 works fine

Datseris (approved):

PR #1105, opened 2024-11-15
Fix dispatch for `remove_all_from_space!`

### LINKED ISSUE #1104: Missing type signature in `remove_all_from_space!` method for discrete space?
Is this method supposed to be a generic fallback or not? If not (as I would guess) it should probably be restricted to `model::ABM{<:DiscreteSpace}`.

https://github.com/JuliaDynamics/Agents.jl/blob/24bd774f55e86cfc0b6b2b4f4a7cb35b0abe8592/src/spaces/discrete.jl#L304-L308

When defining a new space, and absent this fallback, it seems like the mandatory interface to be defined would also include `remove_all_from_space!`. However, this is currently not listed in [the docs](https://juliadynamics.github.io/Agents.jl/stable/devdocs/#[Creating-a-new-space-type](@ref-make_new_space)). Therefore, when I implemented a new space and did not define a method for this function, I hit the method above.

simsurace:
This fixes #1104.
- Restrict the method in discrete.jl to discrete spaces
- Add a test that defines a new space as described in the dev docs and checks that adding and removing agents works. The test fails without the next step
- Add a generic fallback that makes the new test pass
- Add a note in the dev docs that this is a candidate to be overloaded for performance

### DIFF
--- a/docs/src/devdocs.md
+++ b/docs/src/devdocs.md
@@ -61,7 +61,7 @@ In principle, the following should be done:
 6. Extend `nearby_ids(pos, model, r)`.
 7. Create a new "minimal" agent type to be used with [`@agent`](@ref) (see the source code of [`GraphAgent`](@ref) for an example).
 
-And that's it! Every function of the main API will now work. In some situations you might want to explicitly extend other functions such as `move_agent!` for performance reasons.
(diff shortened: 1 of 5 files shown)


Tortar:
Cool, thank you, unfortunately tests are broken on 1.11.1, but I suppose they run locally, right?

Tortar (approved):

simsurace:
They do pass on 1.10 locally. I don't understand the failure on 1.11 unfortunately. I'll bump.

PR #1109, opened 2024-12-09
simplify a type: `Tuple{<:T}` -> `Tuple{T}`

nsajko:
```julia-repl
julia> Tuple{<:Int, <:Int, <:AbstractFloat} == Tuple{Int, Int, AbstractFloat}
true
```


Datseris (changes requested):
This change has no functional impact on the code base. However, it does have readibility impact. To understand this change, one has to understand the internals of Tuples, the special Julia type whose type parameters are covariant: https://docs.julialang.org/en/v1/devdocs/types/#Tuple-types 

I didn't know about this, and in fact with the PR I got confused thinking "how can this work???". So, in conclusion, there is no reason to do this change, as it only requires the reader of the source code to have an even more advanced knowledge of Julia. The current version of the source code requires standard knowledge of parametric subtyping.

nsajko:
Yeah, this is just a stylistic change. Your choice.

PR #1111, opened 2024-12-10
Remove "experimental" from continuous time

Datseris:
I guess we should do this now?

### DIFF
--- a/src/core/model_event_queue.jl
+++ b/src/core/model_event_queue.jl
@@ -63,11 +63,6 @@ end
 A concrete implementation of an [`AgentBasedModel`](@ref) which operates in
 continuous time, in contrast with the discrete time nature of [`StandardABM`](@ref).
 
-This is still experimental which means that it is subject to breaking changes in the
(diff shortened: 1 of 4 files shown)


src/core/model_event_queue.jl
```diff
@@ -138,6 +132,11 @@ can be evolved in time.
   an agent when the agent is added to the model.
 - `autogenerate_after_action::Bool = true`: whether to automatically generate a new
   event for an agent after an event affected said agent has been triggered.
+
+
+!!! warn "No IO yet"
+    This model does not yet IO support such as [`save_checkpoint`](@ref).
```
Tortar, line 138:
```suggestion
    This model does not yet support IO operations such as [`save_checkpoint`](@ref).
```

Tortar:
If I remember correctly we promised to remove the experimental parts with 6.2. I would finish #1103 before realising though, I will do it soon

Datseris:
Sounds good, this can go in 6.2.

Tortar (approved):

PR #1120, opened 2024-12-21
line 603, 'schelling' instead of 'model'

raylv17:
line 603 in docs/src/tutorial.jl
changed from
```julia
figure, _ = abmplot(model; agent_color = groupcolor, agent_marker = groupmarker, as = 10) # not executing
```
to
```julia
figure, _ = abmplot(schelling; agent_color = groupcolor, agent_marker = groupmarker, agent_size = 10)
```

PR #1121, opened 2024-12-21
Update to 6.2.1

Tortar:
To release #1120


Datseris (approved):

Datseris:
(for reference good practice is the one who merges also does the tag; i'm doing it now!)

PR #1130, opened 2025-01-28
Update predator_prey.jl

maeckha:
I guess it should be 'hawk' instead of 'wolf' and also, @ref is not working (probably for that reason)


Datseris (approved):
Thanks! Do you mind bumping the _patch_ version (3rd number) of the version in the Project.toml file so that I can immediatelly tag a new release wit hthe fixes?

PR #1138, opened 2025-02-11
Remove deprecated keyword from tutorial

ultradian:
change deprecated keyword in tutorial


Datseris (approved):
Thank you. Can you please go to `Project.toml` file and increase the _third_ version number (patch version) so that I tag a new release after merging your PR?

ultradian:
I ran into a couple of more places in the docs where the `as` keyword needed updating, so I'm including those, too.

Datseris (approved):

PR #1152, opened 2025-04-05
Improve code for flocking example

bergio13:
The flocking example implementation had a very slow and inefficient code. I changed the code adapting it from [here](https://github.com/JuliaDynamics/ABMFrameworksComparison/blob/main/Flocking/Agents/Flocking.jl).

### DIFF
--- a/examples/flock.jl
+++ b/examples/flock.jl
@@ -19,12 +19,12 @@ using Agents
 using Random, LinearAlgebra
 
 @agent struct Bird(ContinuousAgent{2,Float64})
-    speed::Float64
-    cohere_factor::Float64
-    separation::Float64
-    separate_factor::Float64
-    match_factor::Float64
-    visual_distance::Float64
+    const speed::Float64
+    const cohere_factor::Float64
+    const separation::Float64
+    const separate_factor::Float64
+    const match_factor::Float64
+    const visual_distance::Float64
 end
 
 # The fields `id` and `pos`, which are required for agents on [`ContinuousSpace`](@ref),
@@ -44,7 +44,7 @@ end
 # a model object using default values.
 function initialize_model(;
     n_birds = 100,
-    speed = 1.5,
+    speed = 1.0,
     cohere_factor = 0.1,
     separation = 2.0,
     separate_factor = 0.25,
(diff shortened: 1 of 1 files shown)


examples/flock.jl
```diff
@@ -53,12 +53,12 @@ function initialize_model(;
...
 )
-    space2d = ContinuousSpace(extent; spacing = visual_distance/1.5)
+    space2d = ContinuousSpace(extent; spacing = visual_distance / 1.5)
     rng = Random.MersenneTwister(seed)
 
-    model = StandardABM(Bird, space2d; rng, agent_step!, scheduler = Schedulers.Randomly())
+    model = StandardABM(Bird, space2d; rng, agent_step!, container = Vector, scheduler = Schedulers.Randomly())
     for _ in 1:n_birds
-        vel = rand(abmrng(model), SVector{2}) * 2 .- 1
+        vel = SVector{2}(rand(abmrng(model)) * 2 - 1 for _ in 1:2)
```
Tortar, line 61:
```suggestion
        vel = rand(abmrng(model), SVector{2}) * 2 .- 1
```

the previous version was more readable, and shouldn't have any impact on performance

examples/flock.jl
```diff
@@ -44,7 +44,7 @@ end
 # a model object using default values.
 function initialize_model(;
     n_birds = 100,
-    speed = 1.5,
+    speed = 1.0,
```
Tortar, line 47:
```suggestion
    speed = 1.5,
```

is the speed on velocity intended?

bergio13: yes, I think it creates a better visualization but we can keep speed=1.5 if you prefer

examples/flock.jl
```diff
@@ -19,12 +19,12 @@ using Agents
...
-    separation::Float64
-    separate_factor::Float64
-    match_factor::Float64
-    visual_distance::Float64
+    const speed::Float64
+    const cohere_factor::Float64
+    const separation::Float64
+    const separate_factor::Float64
+    const match_factor::Float64
+    const visual_distance::Float64
```
Datseris, line 27: Does this have any actual performance impact? The compiler should be able to understand this I would wager.

bergio13: I didn't conduct extensive tests on this, you are probably correct

Tortar (approved):

Issue #1146, opened 2025-03-15
Optional Struct-of-Arrays layout for agents


Tortar:
While currently Agents.jl uses an Array-of-Structs (AoS) layout for Agents, it would be useful in some situations to have a Struct-of-Arrays (SoA), in particular when many operations are easily vectorizable, and the number of operations which are not are a small part of the total number of operations. I think it should be possible to use AoS or SoA just by building agents under the hood differently, but without changing any interface, or even any functions operating on agents inside the library. And we also have a precedent for this in https://github.com/JuliaDynamics/Agents.jl/pull/492. If @fbanning would like to share their opinion on this given that he worked in this albeit a lot of time ago. I think that SoA is bad when removal of agents are allowed, but even now we have a version of ABM which don't allow for removals and work well. And actually not having removals is fine for my use case for SoA which is mainly to refactor [BeforeIT.jl](https://github.com/bancaditalia/BeforeIT.jl) to be based on [Agents.jl](https://github.com/JuliaDynamics/Agents.jl) with an easier transition and without incurring any perf penalty.

Datseris:
> I think it should be possible to use AoS or SoA just by building agents under the hood differently, but without changing any interface, or even any functions operating on agents inside the library.

The only issue I see here is the user defined agent-stepping functions. Typically as a user you do something like `a = model[id]; a.property = x`. To make this non-breaking, you need to make `model[id]` return a special data structure, so that when doing `a.property = x` this actually updates the `id`-th entry in the `property`-th array in the struct-of-arrays. This doesn't sound hard at all actually, seems to me like a 10-20 lines of code helper data structure?

The same thing would happen when `add_agent` is called.

Worth doing a MWE that operates this way and extends the `AgentBasedModel` interface.

A downside I see: what happens with `@agent`? How does this (admidetly super duper nice) framework interplace with struct-of-arrays?

Tortar:
Yeah, the `@agent` part is a bit shaky, but we can probably have an option on the macro as

```julia
@agent layout=:SoA struct A
    x::Int
end
```
and generate something different for this case.

Tortar:
mmmh yes actually the most problematic aspects I see of `@agent` is that it should work differently from how it works now (we can use an option in the macro for this) and most importantly it should return the same object as `model[id]`

Datseris:
[...]

PR #1153, opened 2025-04-07
Support StructVector as a container (Issue 1146)

bergio13:
This is a draft for the fix of issue 1146, the code is adapted from [here](https://github.com/JuliaDynamics/Agents.jl/issues/1146#issuecomment-2780752740). The wrapper implementation adds some complexities which I was not able to resolve yet in a clean and efficient way and should be adressed before making the pr. I tested my implementation on the flocking example by changing the type of container to `StructVector` and it works if we apply some small tweaks to the code in particular:
- in `continuous.jl ` we change `agent.pos = pos` to `model[agent.id].pos = pos`
```
function move_agent!(agent::AbstractAgent, pos::ContinuousPos, model::ABM{<:ContinuousSpace})
    space_size = spacesize(model)
    D = length(space_size)
    all(i -> 0 <= pos[i] <= space_size[i], 1:D) || error("position is outside space extent!")
    oldcell = pos2cell(agent, model)
    newcell = pos2cell(pos, model)
    if oldcell ≠ newcell
        remove_agent_from_space!(agent, model, oldcell)
        add_agent_to_space!(agent, model, newcell)
    end
    model[agent.id].pos = pos
    return agent
end
```
- in `flock.jl` (remove the bird marker for plotting or change type of the input agent)
```
    ## Obtain the ids of neighbors within the bird's visual distance
    neighbor_agents = nearby_agents(bird.pos, model, bird.visual_distance)
    N = 0
    match = separate = cohere = SVector{2}(0.0, 0.0)
    ## Calculate behaviour properties based on neighbors
    for neighbor in neighbor_agents
        N += 1
        heading = get_direction(bird.pos, neighbor.pos, model)

        ## `cohere` computes the average position of neighboring birds
        cohere += heading
        ## `match` computes the average trajectory of neighboring birds
        match += neighbor.vel
        if sum(heading .^ 2) < bird.separation^2
            ## `separate` repels the bird away from neighboring birds
            separate -= heading
        end
    end

    ## Normalise results based on model input and neighbor count
    cohere *= bird.cohere_factor
    separate *= bird.separate_factor
    match *= bird.match_factor
    ## Compute velocity based on rules defined above
    bird.vel += (cohere + separate + match) / max(N, 1)
    bird.vel /= norm(bird.vel)
    ## Move bird according to new velocity and speed
    move_agent!(Agents.agent_container(model)[bird.id], model, bird.speed)
end
```
I will happily listen to advices on how to integrate this in the code in a smooth way.

Moreover, it is still necessary to adress the issue mentioned in this [comment](https://github.com/JuliaDynamics/Agents.jl/issues/1146#issuecomment-2781363742).


src/core/model_accessing_API.jl
```diff
@@ -2,10 +2,14 @@ const DictABM = Union{StandardABM{S,A,<:AbstractDict{<:Integer,A}} where {S,A},
...
 const VecABM = Union{StandardABM{S,A,<:AbstractVector{A}} where {S,A},
                      EventQueueABM{S,A,<:AbstractVector{A}} where {S,A}}
+const StructABM = Union{StandardABM{S,A,<:StructVector{A}} where {S,A},
+                         EventQueueABM{S,A,<:StructVector{A}} where {S,A}}
 
 nextid(model::DictABM) = getfield(model, :maxid)[] + 1
 nextid(model::VecABM) = nagents(model) + 1
 hasid(model::VecABM, id::Int) = id ≤ nagents(model)
+nextid(model::StructABM) = nagents(model) + 1
+hasid(model::StructABM, id::Int) = id ≤ nagents(model)
```
Tortar, line 12:
```suggestion
nextid(model::Union{VecABM, StructABM}) = nagents(model) + 1
hasid(model::Union{VecABM, StructABM}, id::Int) = id ≤ nagents(model)
```

src/core/model_accessing_API.jl
```diff
@@ -39,5 +51,9 @@ function remove_agent_from_model!(agent::AbstractAgent, model::VecABM)
     error("Cannot remove agents in a `StandardABM` with a vector container.")
 end
 
+function remove_agent_from_model!(agent::AbstractAgent, model::StructABM)
+    error("Cannot remove agents in a `StandardABM` with a vector container.")
+end
+
```
Tortar, line 57:
```suggestion
function remove_agent_from_model!(agent::AbstractAgent, model::StructABM)
    error("Cannot remove agents in a `StandardABM` with a vector container.")
end
```

same as before, just use `Union`

src/core/model_accessing_API.jl
```diff
@@ -27,6 +31,14 @@ function add_agent_to_model!(agent::AbstractAgent, model::VecABM)
     return
 end
 
+function add_agent_to_model!(agent::AbstractAgent, model::StructABM)
+    agent.id != nagents(model) + 1 && error(lazy"Cannot add agent of ID $(agent.id) in a vector ABM of $(nagents(model)) agents. Expected ID == $(nagents(model)+1).")
+    tuple_agent = NamedTuple(x => getfield(agent, x) for x in fieldnames(typeof(agent)))
+    push!(agent_container(model), tuple_agent)
+    extra_actions_after_add!(tuple_agent, model)
```
Tortar, line 38:
```suggestion
    extra_actions_after_add!(model[agent.id], model)
```

this should be like this (at least is already a bit better) because you pass the wrapper this way

src/core/model_accessing_API.jl
Tortar: I don't see anymore remove_agent_from_model! for `StructABM`, you should add `Union{VecABM, StructABM}` in the signature

Tortar:
There is still an inconsistency in all of this which causes problems. The flocking stepping problem is solved, but the marker problem in Flocking reveals this other problem, the marker is defined as so:

```julia
function bird_marker(b::Bird)
    φ = atan(b.vel[2], b.vel[1]) #+ π/2 + π
    rotate_polygon(bird_polygon, φ)
end
```

the problem with this implementation is that b is not `Bird` but is `AgentWrapperSoA`. This is easy to solve in this case, just use `AbstractAgent` in the signature, but this is in general problematic because the user can't just switch between containers because the signature are actually wrong with `StructVector`. So, this should be fixed.

To do so, maybe we can have a different macro variant for the `StructVector` for `@agent`. It would actually be cool to have e.g.

```julia
@agent array Bird(ContinuousAgent{2, Float64}) begin
    speed::Float64
    cohere_factor::Float64
    separation::Float64
    separate_factor::Float64
    match_factor::Float64
    visual_distance::Float64
end
```

and reject things created with the `struct` version in model creation.

Tortar:
or maybe even simpler as

```julia
@agent :SoA struct Bird(ContinuousAgent{2, Float64})
    speed::Float64
    cohere_factor::Float64
    separation::Float64
    separate_factor::Float64
    match_factor::Float64
    visual_distance::Float64
end
```

src/core/model_accessing_API.jl
```diff
@@ -27,6 +29,14 @@ function add_agent_to_model!(agent::AbstractAgent, model::VecABM)
     return
 end
 
+function add_agent_to_model!(agent::AbstractAgent, model::StructABM)
+    agent.id != nagents(model) + 1 && error(lazy"Cannot add agent of ID $(agent.id) in a vector ABM of $(nagents(model)) agents. Expected ID == $(nagents(model)+1).")
+    tuple_agent = NamedTuple(x => getfield(agent, x) for x in fieldnames(typeof(agent)))
+    push!(agent_container(model), tuple_agent)
```
Tortar, line 35: I have the feeling that pushing is slow this way, worth benchmarking

Tortar: If this is so, to solve this, one should go down the calling path, and have a specialized version of `add_agent!` which doesn't construct the agent (if this is not already passed, and instead only its argument are passed)

Tortar: but this last thing is clearly advantageous in any case

src/core/model_accessing_API.jl
```diff
@@ -2,10 +2,12 @@ const DictABM = Union{StandardABM{S,A,<:AbstractDict{<:Integer,A}} where {S,A},
                       EventQueueABM{S,A,<:AbstractDict{<:Integer,A}} where {S,A}}
 const VecABM = Union{StandardABM{S,A,<:AbstractVector{A}} where {S,A},
                      EventQueueABM{S,A,<:AbstractVector{A}} where {S,A}}
+const StructABM = Union{StandardABM{S,A,<:StructVector{A}} where {S,A},
```
Tortar, line 5:
```suggestion
const StructVecABM = Union{StandardABM{S,A,<:StructVector{A}} where {S,A},
```
seems slightly better

Datseris:
Why are we doing this as an additional container? I think this unecessarily complicates things because it requires us to adjust existing code to accommodate for something the code was never built for. 

Why are we not creating a brand new model type? To my understanding this type of functionality is only useful for rather special time evolution rules, where one broadcast the same operations across agents, rather than having per-agent instructions. I guess the simply `nearby_agents` iteration becomes significantly slower with this new container type. On the other hand, broadcasting across agents is fast.

So why are we not think about just implementing a `StructVectorABM` which can only allow for a struct-of-arrays container?

Datseris:
> in `continuous.jl ` we change `agent.pos = pos` to `model[agent.id].pos = pos`

There is no reason to do this. You just have to overload `setproperty!` for the new agent wrapper type. `agent.pos = x` must work also for the user-facing API.

Datseris:
> the problem with this implementation is that b is not `Bird` but is `AgentWrapperSoA`. This is easy to solve in this case, just use `AbstractAgent` in the signature, but this is in general problematic because the user can't just switch between containers because the signature are actually wrong with `StructVector`. So, this should be fixed.

Yes, this is really unfortunate. The way I see it, there is no real way to allow users to use exactly the same code for StandardABM and StructVectorABM. I don't have any ideas how to change this. I don't like making `@agent` even more complicated. It is already as complicated as it can get. Perhaps we can simply instruct the users to not use type declaration in their functions if they want to switch to also use StructVectorABM? The downside is of course that they won't be able to use multi-agent models this way.

Datseris:
Does BeforeIT.jl allow for multi-agent models? How do they handle it? I just realized now that actually I have no idea how multi-agent models would be handled at all with this struct of array approach.

Tortar:
you are right on the `EventQueueABM`, it actually doesn't make sense.

But actually it makes sense to have it as a container because it's actually not slower than a `Vector` container, I benchmarked flocking and it is the same, actually it is even slightly faster than a `Vector` in quick benchmarks, my question is then, instead, why don't we use a `StructVector` instead of a `Vector`? I'm not sure at the moment if there are cases where a `Vector` would still be advantegeous, and to be conservative I would add a `StructVector` as an option instead, and also because unlike a `Vector` it requires this tweak to the macro.

Tortar:
ah and also `Vector` is very easy to multiagent, while a multi-StructVector would need something along the lines of https://github.com/JuliaDynamics/Agents.jl/issues/841 which is not trivial.

src/core/model_standard.jl
```diff
@@ -154,7 +154,13 @@ function StandardABM(
...
+    if C <: StructVector
+        fields_with_types = zip(fieldnames(A), fieldtypes(A))
+	tuple_init = NamedTuple(x => T[] for (x, T) in fields_with_types)
+	agents = C(tuple_init)
+    else
+        agents = C()
+    end
     agents_types = union_types(A)
     T = typeof(agents_types)
     return StandardABM{S,A,C,T,G,K,F,P,R}(agents, agent_step!, model_step!, space, scheduler,
```
Tortar, line 166:
```suggestion
    return StandardABM{S,A,typeof(agents),T,G,K,F,P,R}(agents, agent_step!, model_step!, space, scheduler,
```

otherwise the type is abstract

Tortar:
so this needs to be totally overhauled with a new type, which I propose it could be called `VectorizedABM`. The fact that Flocking, even though it didn't have any vectorized operation is even slightly faster is cool though.

Tortar:
in BeforeIT.jl we use #841 multiagent style, we don't have problems of type instability because agents of a certain type always step before agents of another type (actually I don't think type instability would be problematic in any case if #841 is done right)

Tortar:
I want just to note another cool aspect of all of this: we are actually doing what https://github.com/projectmesa/mesa-frames does in 100 loc of code while they have more than 10000 loc of code :D

Tortar:
Actually rethinking a bit about the fact that `EventQueueABM` is not sensible with a `StructVector`, I'm actually not sure if this is so, in a sense a `StructVector` could be handy whenever you want to do vectorized operations inside an event of a single agent, even though this is rare, this is actually sensible to have vectorized because that could be the most expensive operation inside all the events

Datseris:
> benchmarked flocking and it is the same, actually it is even slightly faster than a `Vector` in quick&dirty benchmarks, my question is then, instead, why don't we use a `StructVector` instead of a `Vector`?

Where are these benchmarks? I find this hard to believe given the extensive testing that we have done for Struct of Arrays in the past...

Tortar:
so all in all, I'm actually opposed to have a new type, given that effectively we are not changing the rule of the games as in `EventQueueABM`, but we are just changing container

Datseris:
> The fact that #841 is needed for multi agent `VectorizedABM` (do you like this name?) is a point in favour of having a different type

I don't understand where the reference to #841 is coming from. Actually, why is #841 even open? This issue should be closed because your recent additions to `@multiagent` removed the dynamic dispatch.

The core issue of #841 remains, and will remain forever: as long as we are obtaining agents by ID, this fundamental type instability will remain. One needs to add an additional interface so that agents are obtained by type, instead of by ID. Then you would loose the power to obtain agents by ID but you could still iterate over all agents of a particular type. That's all fine. It just needs a different function, and hence, I would argue, a new model type. It makes it much easier to have this SoA model where you obtain agents by type instead of by ID.

Tortar:
> Where are these benchmarks? I find this hard to believe given the extensive testing that we have done for Struct of Arrays in the past...

I cook up a benchmark in a second and show this :D

Datseris:
> I want just to note another cool aspect of all of this: we are actually doing what https://github.com/projectmesa/mesa-frames does in 100 locs while they have more than 10000 locs :D

Yeah, I am not surprised...

Datseris:
> > I want just to note another cool aspect of all of this: we are actually doing what https://github.com/projectmesa/mesa-frames does in 100 locs while they have more than 10000 locs :D
> 
> Yeah, I am not surprised...

We should write a blog post about this or something when this PR is nice and done. The Julia community will appreciate it.

Tortar:
benchmark with the code in this pr:

```julia
using Agents, LinearAlgebra

@agent struct Bird(ContinuousAgent{2, Float64})
    const speed::Float64
    const cohere_factor::Float64
    const separation::Float64
    const separate_factor::Float64
    const match_factor::Float64
    const visual_distance::Float64
end

function flocking_model(rng, extent, n_birds, visual_distance, C;
        speed = 1.0, cohere_factor = 0.03, separation = 1.0, separate_factor = 0.015,
        match_factor = 0.05, spacing = visual_distance / 1.5,)
    space2d = ContinuousSpace(extent; spacing)
    model = StandardABM(Bird, space2d; agent_step!, rng, container = C,
        scheduler = Schedulers.Randomly())
    for n in 1:n_birds
        vel = SVector{2}(rand(abmrng(model)) * 2 - 1 for _ in 1:2)
        add_agent!(model, vel, speed, cohere_factor, separation,
            separate_factor, match_factor, visual_distance)
    end
    return model
end

    neighbor_agents = nearby_agents(bird, model, bird.visual_distance)
    N = 0
    match = separate = cohere = SVector{2}(0.0, 0.0)
    for neighbor in neighbor_agents
        N += 1
        heading = get_direction(bird.pos, neighbor.pos, model)
        cohere += heading
        match += neighbor.vel
        if sum(heading .^ 2) < bird.separation^2
            separate -= heading
        end
    end
    cohere *= bird.cohere_factor
    separate *= bird.separate_factor
    match *= bird.match_factor
    bird.vel += (cohere + separate + match) / max(N, 1)
    bird.vel /= norm(bird.vel)
end

using BenchmarkTools, Random

rng = Xoshiro(42)
model = flocking_model(rng, (100, 100), 5000, 3.0, Vector)
@benchmark step!($model)

rng = Xoshiro(42)
model = flocking_model(rng, (100, 100), 5000, 3.0, StructVector)
@benchmark step!($model)
```

we have

```julia
julia> model = flocking_model(rng, (100, 100), 5000, 3.0, Vector)
StandardABM with 5000 agents of type Bird
 agents container: Vector
 space: periodic continuous space with [100.0, 100.0] extent and spacing=2.0

BenchmarkTools.Trial: 168 samples with 1 evaluation per sample.
 Range (min … max):  22.428 ms … 45.217 ms  ┊ GC (min … max):  0.00% … 25.77%
 Time  (median):     29.986 ms              ┊ GC (median):    13.59%
 Time  (mean ± σ):   29.815 ms ±  3.007 ms  ┊ GC (mean ± σ):  13.16% ±  4.63%

                     █▄▃▃  ▃   ▆▅▂ ▄ ▂                         
  ▄▁▅▁▅▃▄▃▁▁▁▁▁▁▁▁▁▃▆█████▇█▇▆▆███▆████▃▇▄▄▃▁▁▁▁▃▃▄▃▃▁▁▁▁▃▁▁▃ ▃
  22.4 ms         Histogram: frequency by time          38 ms <

 Memory estimate: 36.14 MiB, allocs estimate: 1183942.

julia> rng = Xoshiro(42)
Xoshiro(0xa379de7eeeb2a4e8, 0x953dccb6b532b3af, 0xf597b8ff8cfd652a, 0xccd7337c571680d1, 0xc90c4a0730db3f7e)

julia> model = flocking_model(rng, (100, 100), 5000, 3.0, StructVector)
StandardABM with 5000 agents of type Bird
 agents container: StructVector
 space: periodic continuous space with [100.0, 100.0] extent and spacing=2.0

BenchmarkTools.Trial: 192 samples with 1 evaluation per sample.
 Range (min … max):  19.608 ms … 36.707 ms  ┊ GC (min … max):  0.00% … 0.00%
 Time  (median):     25.666 ms              ┊ GC (median):    14.69%
 Time  (mean ± σ):   26.063 ms ±  2.762 ms  ┊ GC (mean ± σ):  14.82% ± 5.18%

                  ▃   ▃▆▂▂       █                             
  ▄▅▄▁▁▁▁▁▁▁▃▁▁▁▄▅█▅▆▆█████▄▇█▆▆███▄▆▇▄▃▁▁▁▃▄▁▃▃▄▃▁▄▁▄▁▁▁▁▃▁▄ ▃
  19.6 ms         Histogram: frequency by time        34.1 ms <

 Memory estimate: 38.72 MiB, allocs estimate: 1268139.
```

Tortar:
in general, we have some pros and cons for a new type from my perspective, I'm not sure what's better actually

Datseris:
can you run the same with the agent fields being non-constants?

Datseris:
In general, we should also be benchmarking this with a model that has (a) agent field changing and (b) agents being added and removed. I think the wolf sheep model is a good test case for this.

Tortar:
that is impossible because we don't support either multiagent or removal of agents, the latter of which is probably impossible to support efficiently in any case

Datseris:
> that is impossible because we don't support either multiagent

For the removal part, okay, but if this cannot support multiagent, then it cannot be a generic container either right? ALl the more reason to make this a separate model.

I would still argue, it is better for start to make this a separate model type, and once it has seen a bit more testing, see if it can generalize as a container also for `StandardABM`. There is no reason to be greedy and try to achieve everything in one go.

Tortar:
ok, to be it is fine, it is sensible to track it as experimental and make it a type while we learn about its working. The only concern I have is that I see use cases for event based ABMs, e.g. they can be faster for single-agent event-based ABMs and allow some better vectorization, so maybe we could have two types: a `VectorizedABM` and a `VectorizedEventABM`? It is a bit cumbersome but this is the way to go if we want to add types instead of a container

Tortar:
it should'nt be off, if you want to be sure you just need to run the benchmarks and see if you can reproduce, on my computer those are the results when using this branch. The fact that there aren't any vectorized operation and that this is even a bit faster breaks my mental model too, but this is the truth (at least in this case).

Tortar:
The difference in flocking becomes even bigger and bigger more you add agents, e.g. same benchmark as above but with 10^5 agents, one step with the StructVector requires 1.5 seconds, while with a Vector it requires 2.6 seconds

Datseris:
I can confirm that the struct-vector example is faster, which confuses me greatly. It shouldn't be...? In any case, this means that we should prioritize merging this PR because this is a big positive change. Here are the important things to do:

1. Organize source code better. There should a dedicated file `core/agent_containers.jl` that has all the business of agent containers.
2. To simplify, the `construct_agent_container` function should give both the type `C` as well as the `agents` empty container. Then these `if` clauses are not necessary anymore. 
3. Add tests, documentation, changelog entry, and bump version. 

I'll then do a more detailed line-by-line review.

---

I don't know why this is such a surprise. Either newer Julia versions have improved this data layout, and the compiler code generation from it drastically. Or, the original SoAs implementation in #492 was so different to the one here that it simplu couldn't match the performance.

Datseris:
cc @bergio13

Datseris:
We also need to implement removal. @Tortar you said that agent removal and addition doens't work here, but why? The struct-vector object simply has fields that are normal `Vector`s; We can simply `pop!` all such fields at a given `id` location. We just need to first find the `id` from the `.id` field of the struct vector, because as you remove agents it is no longer true that agent id = location in the inner vectors. 

Am I missing something on why addintion and removal of agents shouldn't work with StructVector? I am not considering performance at all here.

Tortar:
This can work for removal, but the model will have abysmal performance after the removal for indexing so I don't think this is worth doing. To make it a bit better, we can instead not `pop!` it but using `removeat!`, so that you will still preserve the invariant of ordering, then you can at least use binary search for indexing, removal will be o(n) though. And this will add probably a non-zero cost for any model, because you still need to understand if `agent id = location in the inner vectors` is true even for the no-removal case, though the branch is perfectly predictable in this case so maybe the cost will be infinitesimal. In any case, this doesn't seem worth doing to me, because it has this kind of trade-offs which can only be undestood by benchmarking and the case of removal is probably better with a `Dict` in any case, so I think this PR should not try to chase something that is not clear if it is worth or not. Maybe a new PR could try it, we could open an issue on this to track it (All of this applies the same for a `Vector` by the way).

Also, what I'm not understanding from your words is if you think that `StructVector` should be a container in the end, or inside a new model type?

Tortar:
If we go with a container, we need also a new macro, or a specialization of the `@agent` macro.

Datseris:
> If we go with a container, we need also a new macro, or a specialization of the `@agent` macro.

What do you mean? There is no change at least in the flocking example we are running with `@agent`...?

Tortar:
yes, but the problem is that the `StructVector` implementation passes around the wrapper, which causes problems whenever you dispatch on the agent type because it is not what it is actually passed around, so we need a new variant of the macro/a new macro to make the type refer to the wrapper

Datseris:
> the model will have abysmal performance after the removal for indexing so I don't think this is worth doing.

Yes, you are right. At the moment we can just access the vector entries using the ID field. If we allow removal of agents, we then every time we access a property must first find the index corresponding to agent ID. This is not just a huge hit to performance - it will likely also prevent parallelization . 

Well, actually, this isn't a surprise, this is the fundamental reason ABMs are not parallelizable _in general_: the removal of agents.

Datseris:
> yes, but the problem is that the `StructVector` implementation passes around the wrapper, which causes problems whenever you dispatch on the agent type because it is not what it is actually passed around, so we need a new variant of the macro/a new macro to make the type refer to the wrapper

I udnerstand the problem but I don't fully undertstand the proposed solution. Perhaps you can leave comments on the PR for the contributor to implement.

Tortar:
There are two possible solutions:

1. Do not accept the type itself when doing `StandardABM(Type, ...; container = StructVector)` but instead accept something like  `StandardABM(SoA(Type), ...; container = StructVector)`. This means that you need to change any signature annotation to `Type` to `SoA(Type)`, which is a bit cumbersome.

2. A new macro e.g. `@agentsoa ...` or create a new variant `@agent :soa ...` and then the `Type` will already be a `SoA(Type)` so no need to change the signatures.

2. seems better to me, and the first version of 2. seems better than the second, so `@agentsoa` has my vote.

Tortar:
mmmh, thinking about it, 2. is very tricky and it is too opaque for the user e.g. `typeof` of an agent created inside the step won't have the same type as `model[id]`, so it will cause tricky bugs. Going with 1. instead the user is in full control of all the nuances. We just need to add to the docs that `model[id]` will be of type `SoA(Type)` instead of `Type` and so signatures should be adjusted accordingly.

Datseris:
Sound like to me both options don't fix all the problems,. Perhaps the real solution is to simply document clearly that when you do `model[id]` the agent type isn't what you expect. Actually, we should only worry about this if multi-agent models are even a viable option with strcut of arrays. So far we have only tested it with a single agent type and there it seems to be performant. We went to hell and back to make the multi agent normal case performant, let's see how easy it will be here. The `@multiagent` macro clearly doen't work for struct vectors.

Tortar:
yes, the first option I suggested doesn't do actually anything very sophisticated, you just give a convenience function to access the wrapper type of the model, and you force the user to acknowledge that this difference exist by rejecting normal types. This is important even in cases there is only one agent type because the agent constructor won't have the same type of the indexing in the model. So I think there is value in adding this even with single agent models.

In any case, everything is settled from a discussion standpoint in my opinion, and just the work needs to be done :D

src/core/model_accessing_API.jl
Tortar, line 35:
```suggestion
    push!(agent_container(model), agent)
```

this reduces indirection

src/core/model_accessing_API.jl
```diff
@@ -31,8 +31,7 @@ end
 
 function add_agent_to_model!(agent::AbstractAgent, model::StructABM)
     agent.id != nagents(model) + 1 && error(lazy"Cannot add agent of ID $(agent.id) in a vector ABM of $(nagents(model)) agents. Expected ID == $(nagents(model)+1).")
-    tuple_agent = NamedTuple(x => getfield(agent, x) for x in fieldnames(typeof(agent)))
-    push!(agent_container(model), tuple_agent)
+    push!(agent_container(model), agent)
```
Datseris, line 34: Rather, this should be a dedicated function `add_agent_to_container!`. And the function `add_agent_to_model!` shouldn't be named at all this way. We have `add_agent!`. This is the function that means "add agent to model", which entails more lower level processes: add to space, add to container, update next id, etc. etc. 

So all `add_agent!` functions propagate to the lowest `add_agent!(agent::AA, model::ABM)`. From there we call the internal functions `add_agent_to_container!`, to space, and the extra actions.

Tortar: yes, right, my change was from a performance standpoint

Datseris:
Jolting down some brief ideas for multi-agent types with this approach after the discussion in the JuliaDynamics meeting

- the container can be a tuple/named tuple of struct vectors , each corresponding to an agent type. 
- the "id" has to be come a composite type. Users would have to request an agent by type and id, such as `model[(Bank, 23)]`. 
- similarly, the `allagents` function can have an extra argument verfsion that returns all agents of a given type.

This approach has problems with space. What does `nearby_ids` do for such a data structure? What sorts of IDs would be stored in the space containers? At least for integration downstream in e.g., BeforeIT.jl, there is no space. So we don't have to worry about this. SO there is a real use case for a specialized struct-vector model that does not allow for spaces.

Datseris:
oh yeah we still have the problem that exists in this PR: the agent type is this wrapper type. But there your solution is fine for me at least:



> Going with 1. instead the user is in full control of all the nuances. We just need to add to the docs that `model[id]` will be of type `SoA(Type)` instead of `Type` and so signatures should be adjusted accordingly.

Tortar:
Documentation fails because of "This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101". So the runner should be updated.

src/core/model_accessing_API.jl
```diff
@@ -16,26 +18,29 @@ function add_agent_to_model!(agent::AbstractAgent, model::DictABM)
...
+    getfield(model, :autogenerate_on_add) && add_event!(model[agent.id], model)
+end
 
 function remove_agent_from_model!(agent::AbstractAgent, model::DictABM)
     delete!(agent_container(model), agent.id)
     return
 end
-
-function remove_agent_from_model!(agent::AbstractAgent, model::VecABM)
+function remove_agent_from_model!(agent::AbstractAgent, model::Union{VecABM, StructVecABM})
```
Tortar, line 43:
```suggestion
function remove_agent_from_container!(agent::AbstractAgent, model::Union{VecABM, StructVecABM})
```

for consistency this should be renamed

bergio13:
I reckon the main changes have been implemented for the single agent type and the PR is now ready to be merged

src/core/model_abstract.jl
```diff
@@ -172,11 +172,11 @@ Return a valid `id` for creating a new agent with it.
 nextid(model::ABM) = notimplemented(model)
 
 """
-    add_agent_to_model!(agent, model)
+    add_agent_to_container!(agent, model)
 Add the agent to the model's internal container, if the addition is valid
 given the agent's ID and those already in the model. Otherwise error.
 """
-add_agent_to_model!(agent, model) = notimplemented(model)
+add_agent_to_container!(agent, model) = notimplemented(model)
```
Datseris, line 179: I think this function should dispatch to `add_agent_to_container!(agent, agent_container(model))`. 

I am not convinced we need a `notimplemented` function here. The only way this function is called is anyways in the high level.

Tortar (approved):
To me the PR seems in good shape, thanks @bergio13 for the work!

bergio13:
> To me the PR seems in good shape, thanks @bergio13 for the work!

Thank you for your help and your reviews !

Tortar:
Can you bump to 6.3 @bergio13 and add to CHANGELOG.md that this new experimental feature is now available?

Tortar:
I merged so that we can consider this contribution for GSOC, I think that if there is any minor problem with this PR, we can tackle it in a small follow-up

Datseris:
> I merged so that we can consider this contribution for GSOC, I think that if there is any minor problem with this PR, we can tackle it in a small follow-up

please don't merge PRs that introduce new interface elements without an approving review from someone that has already left a review (in this case, me).

Datseris:
For now, please don't tag a release, and if you have done so retract it. I will review the code base as is now whenI can.

Tortar:
yes, I didn't want to release any new version until everything was finalized

src/core/model_accessing_API.jl
```diff
@@ -16,26 +18,29 @@ function add_agent_to_model!(agent::AbstractAgent, model::DictABM)
...
 end
 
-# This is extended for event based models (in their file)
+# This is extended for event based models
 extra_actions_after_add!(agent, model::StandardABM) = nothing
+function extra_actions_after_add!(agent, model::EventQueueABM{S,A,<:Union{AbstractDict, AbstractVector}} where {S,A})
+    getfield(model, :autogenerate_on_add) && add_event!(agent, model)
+end
+function extra_actions_after_add!(agent, model::EventQueueABM{S,A,<:StructVector} where {S,A})
+    getfield(model, :autogenerate_on_add) && add_event!(model[agent.id], model)
```
Datseris, line 36: Have we even tested that this works with Event Queue?

Datseris: Nope, there aren't any tests in the event queue file.

src/core/model_event_queue.jl
```diff
@@ -144,17 +144,25 @@ can be evolved in time.
...
-    C = construct_agent_container(container, A)
-    agents = C()
+    ) where {S<:SpaceType,P,R<:AbstractRNG}
+    if container == StructVector
+        if !(A <: SoAType)
+            @warn "The agent type passed to the model constructor is of type $A but a model with a 
+            StructVector container will have agents of type SoAType{$A}. Pass this to the constructor to 
+            remove this warning." maxlog=1
+        else
+            A = A.body.parameters[1]
```
Datseris, line 162: What is this? `A.body.parameters`? Doesn't sound very stable. Isn't there a programmatic, public-API way to obtain what you want to obtain?

src/core/model_event_queue.jl
```diff
@@ -144,17 +144,25 @@ can be evolved in time.
...
+        end
+    end
+    agents = construct_agent_container(container, A)
```
Datseris, line 165: is this one of the Interface function for the container API? If so we should put it in a subsection in the developer's docs.

Datseris: Oh, wow. This PR has introduced NO documentation changes. Yet it was merged and ready to be tagged as a new minor version...?

src/core/model_event_queue.jl
```diff
@@ -215,18 +223,12 @@ function EventQueueABM(
     )
 end
 
+function Base.getindex(model::EventQueueABM{S, A, C}, id::Int) where {S,A,C<:StructVector}
+    return AgentWrapperSoA{A}(agent_container(model), id)
+end
```
Datseris, line 228: Why do we have this function? `getindex` should be defined only once at the abstract intertface level. It should call the `agent_container` function. A new internal function `retrieve_agent(container, id)` should be defined. This way `getindex` doesn't have to be implemented twice for the different model types.

Again, we have to get the interfaces good with this PR because in the future there will be more model types beyond event que and discrete time.

src/core/model_standard.jl
```diff
@@ -152,25 +154,29 @@ function StandardABM(
         ABMObservable.
         """ maxlog=1
     end
+    if container == StructVector
+        if !(A <: SoAType)
+            @warn "The agent type passed to the model constructor is of type $A but a model with a 
+            StructVector container will have agents of type SoAType{$A}. Pass this to the constructor 
+            to remove this warning." maxlog=1
+        else
+            A = A.body.parameters[1]
```
Datseris, line 163: Same comments as above.

src/core/model_standard.jl
```diff
@@ -181,6 +187,9 @@ function remove_all_from_space!(model)
     end
 end
 
+function Base.getindex(model::StandardABM{S,A,C}, id::Int) where {S,A,C<:StructVector}
+    return AgentWrapperSoA{A}(agent_container(model), id)
```
Datseris, line 191: same comments as above.

Datseris (commented):
Merging this PR was unecessarily rushed. Besides the code changes that still need to happen, this PR didn't change any of the docs. It added the extra subsection in the docstring of the standard ABM. 

But I think we can do better. Instead, in the main docs, right after the type of the ABMs ,we should have a new subsection about the type of the containers. There we should list the 3 containers. Then the ABM docs should just say "any of the ABM containers" and cite this subsection.

src/core/model_accessing_API.jl
bergio13: I tested it on the rock-paper-scissor example if i remember correctly, but I forgot to add proper tests, I will soon make up for it !

bergio13:
> @bergio13 I've left some comments on this PR, please do consider making a second PR that addresses them.

Sure, I will soon try to adress the issues you highlighted

src/core/agent_containers.jl
```diff
@@ -0,0 +1,53 @@
...
+
+"""
+    ContainerType{A}
+
+Union type representing the possible container types for storing agents.
+"""
+ContainerType{A} = Union{AbstractDict{Int,A},AbstractVector{A}}
+
+"""
+    struct AgentWrapperSoA{C} <: AbstractAgent
```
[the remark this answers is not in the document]
bergio13: the usage of this is not documented as below we define SoAType which is a type alias for it and then we document it as follows:

The agent type should be wrapped in `SoAType` (e.g., `SoAType{MyAgent}`).

src/core/model_event_queue.jl
bergio13: unfortunately none that i know of

PR #1156, opened 2025-04-12
highlight more prominently that GraphSpace != social network

### LINKED ISSUE #1155: Inconsistencies in extracting graph object from model
Not sure whether this is an issue or a feature request, but I'm trying to extract the graph from the model to plot it with `GraphPlot` and it seems to be difficult or impossible to re-link the edge structure with the node (that is agent) properties. Ideally, I'd extract the space, defined for instance as `GraphSpace(SimpleGraph(x, y))`, and then attribute the correct agent properties to each node using the agent IDs as key.  

The problem is that, extracting the graph structure with `getfield(model, :space).graph` returns a graph object with node IDs that are not consistent with the agent ones I get from  `allagents(model)`. This means that I'm no longer able to attribute the correct properties to each node.

There may be some workaround I did not find yet, but enforcing coherence between the different elements of the model would be really useful, especially for people working with ABMs on dynamic networks.

Datseris:
Closes #1155

### DIFF
--- a/examples/sir.jl
+++ b/examples/sir.jl
@@ -5,7 +5,7 @@
 # </video>
 # ```
 #
-# This example illustrates how to use [`GraphSpace`](@ref) and how to model agents moving an graph
+# This example illustrates how to use [`GraphSpace`](@ref) and how to model agents moving on a graph
 # (network) where the transition probabilities between each node (position) is not constant.
 
 # ## SIR model
--- a/examples/sir.jl
+++ b/examples/sir.jl
@@ -1,12 +1,13 @@
-# # SIR model for the spread of COVID-19
+# # [SIR model for the spread of COVID-19](@id sir_graphspace)
 # ```@raw html
 # <video width="auto" controls autoplay loop>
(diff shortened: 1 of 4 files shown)


Datseris:
This is good to go if someone can merge and tag.

examples/sir.jl
```diff
@@ -1,12 +1,13 @@
-# # SIR model for the spread of COVID-19
+# # [SIR model for the spread of COVID-19](@id sir_graphspace)
 # ```@raw html
 # <video width="auto" controls autoplay loop>
 # <source src="../covid_evolution.mp4" type="video/mp4">
 # </video>
 # ```
 #
-# This example illustrates how to use `GraphSpace` and how to model agents on an graph
+# This example illustrates how to use [`GraphSpace`](@ref) and how to model agents moving an graph
```
Tortar, line 8:
```suggestion
# This example illustrates how to use [`GraphSpace`](@ref) and how to model agents moving on a graph
```

Tortar (approved):

PR #1159, opened 2025-04-28
Fix hover inspection for abstract spaces

hbsmith:
Currently there's a bug in Agents where trying to hover inspect a 2D continuous plot throws a DimensionMismatch error. This seems analogous to #1090 , fixed in #1092 .

You can reproduce the bug by doing something like:

```
using Agents
using GLMakie
using AgentsExampleZoo

model = AgentsExampleZoo.flocking()

fig, abmobs = abmexploration(model)

fig
```

And trying to hover over any agent. You'll get an error like:
```
nested task error: DimensionMismatch: arrays could not be broadcast to a common size
    Stacktrace:
      [1] _bcs1(a::StaticArrays.SOneTo{2}, b::StaticArrays.SOneTo{3})
        @ StaticArrays ~/.julia/packages/StaticArrays/LSPcF/src/broadcast.jl:77
      [2] _bcs(shape::Tuple{StaticArrays.SOneTo{2}}, newshape::Tuple{StaticArrays.SOneTo{3}})
        @ Base.Broadcast ./broadcast.jl:522
      [3] broadcast_shape(::Tuple{StaticArrays.SOneTo{2}}, ::Tuple{StaticArrays.SOneTo{3}})
        @ Base.Broadcast ./broadcast.jl:516
      [4] static_combine_axes
        @ ~/.julia/packages/StaticArrays/LSPcF/src/broadcast.jl:20 [inlined]
      [5] _axes(bc::Base.Broadcast.Broadcasted{StaticArraysCore.StaticArrayStyle{1}, Nothing, typeof(-), Tuple{SVector{2, Float64}, Tuple{Float64, Float64, Float64}}}, ::Nothing)
        @ StaticArrays ~/.julia/packages/StaticArrays/LSPcF/src/broadcast.jl:25
      [6] axes
        @ ./broadcast.jl:234 [inlined]
      [7] static_axes(A::Base.Broadcast.Broadcasted{StaticArraysCore.StaticArrayStyle{1}, Nothing, typeof(-), Tuple{SVector{2, Float64}, Tuple{Float64, Float64, Float64}}})
        @ StaticArrays ~/.julia/packages/StaticArrays/LSPcF/src/broadcast.jl:22
      [8] static_combine_axes(A::Base.Broadcast.Broadcasted{StaticArraysCore.StaticArrayStyle{1}, Nothing, typeof(-), Tuple{SVector{2, Float64}, Tuple{Float64, Float64, Float64}}})
[...]

PR #1162, opened 2025-05-20
3d inspection fix

hbsmith:
Currently 3D hover inspection is broken.

Minimal non working example:
```
using Agents, Plots, Random, GLMakie

@agent struct Dummy(ContinuousAgent{3,Float64}) 
     speed::Float64
end

function setup(;
    n_agents = 100,
    extent = (100, 100, 100),
)
    space3d = ContinuousSpace(extent)
    model = StandardABM(Dummy, space3d; agent_step! = agent_step,
                        container = Vector, scheduler = Schedulers.Randomly(), rng = Xoshiro(314))
    for _ in 1:n_agents
        add_agent!(model,(1.,1.,1.),1.)
    end
    return model
end

function agent_step(dummy, model)
    move_agent!(dummy, model, dummy.speed)
end

model=setup()

fig, abmobs = abmexploration(model; agent_size=3)

fig
```
Screenshot 2025-05-20 at 11 56 11

Note that there's a visible arrow but none of the default inspection information.

I don't have a great understanding of Makie, but it seems like the issue had something to do with the way the position was being translated, and thus never recognizing any Agents in the position of the mouse. The changes I made fix this as minimally as I was able to, such that it avoids Bounds errors, and displays information of multiple agents if they're in the same position.

With the fix:
Screenshot 2025-05-20 at 11 55 26

If someone thinks there's a better way to do this, be my guest, but at least it works nows.

Note that this should be merged after #1159

### DIFF
--- a/ext/AgentsVisualizations/src/spaces/abstract.jl
+++ b/ext/AgentsVisualizations/src/spaces/abstract.jl
@@ -124,35 +124,6 @@ function Makie.show_data(inspector::DataInspector,
     return true
 end
 
-# 3D space
-# function Makie.show_data(inspector::DataInspector,
-#     p::ABMP{<:Agents.AbstractSpace}, idx, source::MeshScatter)
-#     @info "show_data called for MeshScatter! idx=$idx"
-
-#     pos = Makie.position_on_plot(source, idx)
-#     @info "Position on plot: $pos"
-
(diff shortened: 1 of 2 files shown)


Tortar:
I'm wondering: isn't `idx` already `agent_idx`? we should really try to avoid `collect(allids(model))`

Tortar:
Nice work by the way :-)

PR #1165, opened 2025-06-03
Fix string allocation in model_abstract.jl

### LINKED ISSUE #1164: Big performance problem in model_abstract.jl: string allocations in setproperty
This string allocation is most of the run time of my model where I have to update a set of model property each timestep: is there a reason it's not all inside an if block? It should only be used if there is an error:

https://github.com/JuliaDynamics/Agents.jl/blob/75060017503bf307c45c8891e4e5a29824e960d2/src/core/model_abstract.jl#L128-L132

You can see the setproperty and all the string operations in the flamegraph here:

Image

Thanks

elisadonati23:
Fix #1164

### DIFF
--- a/src/core/model_abstract.jl
+++ b/src/core/model_abstract.jl
@@ -126,11 +126,10 @@ end
 function Base.setproperty!(m::ABM, s::Symbol, x)
     properties = abmproperties(m)
 
-    if properties === nothing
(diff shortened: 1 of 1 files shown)


ameligrana (approved):
Thanks, I improved the fix a bit

PR #1181, opened 2025-09-19
Add getid function to retrieve the id of the agent

Tortar:
I'd like to create a custom function which retrieves the id of a type of agent where the `id` can't be retrieved with `getproperty`. I could specialize `getproperty` but I'd like not to have branches in it, so having a specialize function to retrieve the id is better. Even if we forget in the future to use it, no big deal, it's just to facilitate some internal testing


Datseris (commented):
I assumme that for the moment this is not intended to be a public API function?

Tortar:
yes exactly

Datseris (approved):
I haven't checked the test failures but since we aren't releasing a version here it's fine.

PR #1190, opened 2025-12-11
Add chebyshev_distance utility function

jordiplam:
From a recent online discussion ([Computing distances in Agents.jl](https://discourse.julialang.org/t/computing-distances-in-agents-jl/134375)), it seems that while `:euclidean` and `:manhattan` metrics have their own distance functions, the function is missing for `metric = :chebyshev`.

This pull request introduces the `chebyshev_distance` function and tests, implemented consistently with the existing distance functions.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -168,6 +168,7 @@ get_spatial_property
 get_spatial_index
 interacting_pairs
 elastic_collision!
+chebyshev_distance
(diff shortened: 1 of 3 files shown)


Datseris:
thank you. Please note we are inbetween a big release, so it may take a while until your contribution is accessible from a new tagged version.

PR #1197, opened 2026-02-08
Fix inconsistent model state when add_agent! is called with out-of-bounds position

Vyshnavi0702:
## Summary
Fixes an inconsistency where `add_agent!` throws a `BoundsError` for
out-of-bounds coordinates but still leaves the agent inside the model
container, resulting in an invalid model state.

## Fix
Wrap the call to `add_agent_own_pos!` in a `try/catch` block and remove
the partially added agent from the container if space insertion fails,
then rethrow the error.

## Result
- Prevents corrupted model state
- Makes behavior consistent with thrown error
- Existing tests now pass for the out-of-bounds case

## Notes
One unrelated visualization test error remains due to upstream Makie
changes and is not caused by this fix.


Vyshnavi0702:
Hi! This PR fixes the inconsistent model state when `add_agent!`
is called with an out-of-bounds position.

Please let me know if any changes or additional tests are needed.
Thanks for your time and for maintaining Agents.jl!

Datseris (changes requested):
Hi, thanks for the PR! `try-catch` must be avoided in any performance-relevant code. Is also unecessary here. You can check manually whether the agent is within the allowed and if not then return the alternative clause.

Datseris:
Bump here.

Vyshnavi0702:
Hi! I have updated the implementation to avoid try/catch as suggested and resolved the merge conflicts. All tests pass locally. Please let me know if further changes are needed.

Datseris (commented):
This PR adds a lot of unecessary white spaces in various files unrelated to the PR, and as such the Runic tests fail. Please remove these.

Vyshnavi0702:
I cleaned the PR and removed unrelated file changes.
Now only `src/core/space_interaction_API.jl` is modified as intended.
The fix ensures the agent is inserted into the space before updating the container to avoid inconsistent model state.

Please let me know if any further changes are needed.

Datseris (changes requested):
This PR doesn't achieve anything. It certainly does not change any code functionality. Here you have replaced `add_agent_own_pos!` with the source code of the function. Naturally this doesn't change anything, it's exactly the same code. 

The fact that tests are missing from this PR reflects this: you need to add tests that actually test that the issue is solved, by copying the code from the original issue.

PR #1209, opened 2026-05-10
update integration example for CellListMap 0.10 interface.

lmiq:
The 0.10 version of CellListMap brought some breaking changes to its interface. This updates the integration example for the latest CellListMap version.


Datseris (commented):
We should put this version requirement to the docs/Project.toml file as a compat entry to make sure that this works as expected!

